PR #15, opened 2018-03-20
Load code block

hypest:
Note: remember to `git submodule update` to fetch the proper GB branch.

Loads the ported `Code` block from Gutenberg.

* No proper styling ported
* Input box accepts text modifications but are overridden immediately. For the time being text changes only get logged to the console. 
* Test (`'Code block is a TextInput'`) that tests the input component's type (`TextInput` atm)
* The test (`'Code block is a TextInput'`) access the child components using hardcoded array indices 😭

### DIFF
--- a/App.js
+++ b/App.js
@@ -3,16 +3,10 @@
 // @flow
 
 import React from 'react';
-import { settings as codeBlock } from './gutenberg/blocks/library/code'
 import BlockManager from './block-management/block-manager';
 
 export default class App extends React.Component<{}> {
 	render() {
-		Code = codeBlock.edit;
-        <Code
-          attributes={{content: 'Code block text'}}
-          setAttributes={ ( attrs ) => console.log( { attrs } ) }
-        />
 		return <BlockManager />;
 	}
 }
--- a/App.js
+++ b/App.js
@@ -1,15 +1,18 @@
 // @flow
 
 import React from 'react';
-import { StyleSheet, Text, View } from 'react-native';
+import { StyleSheet, View } from 'react-native';
+import { settings as codeBlock } from './gutenberg/blocks/library/code'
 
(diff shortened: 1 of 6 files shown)


maxme:
On the `gutenberg/` side, I wonder if we could factorize a bit more of the code. Especially have only one `index.mobile.js` that would be reused (imported) in dummies `index.android.js` and `index.ios.js`.

maxme:
Do you mind adding `multiline={ true }` to the `TextInput` in the plain-text component?

maxme:
Oh and `underlineColorAndroid="transparent"` as well for `android.js`

screenshot-2018-03-21_10 00 02 36

hypest:
> On the gutenberg/ side, I wonder if we could factorize a bit more of the code. Especially have only one index.mobile.js that would be reused (imported) in dummies index.android.js and index.ios.js.

We surely need to do something on that front. Duplicating code like the current code in this PR is only a cause of trouble. Can we do that in a follow up PR though so we don't postpone this one too much?

maxme:
> Can we do that in a follow up PR though so we don't postpone this one too much?

Yes sure.

hypest:
> Do you mind adding multiline={ true } to the TextInput in the plain-text component?

Defo. Was planning to gradually port more stuff from https://github.com/hypest/gbrn-experimental but `multiline` makes sense to port here in this PR. 👍

Done with 4d78fce (and 4d78fce on the GB side).

hypest:
> Oh and underlineColorAndroid="transparent" as well for android.js

Done with 21fff06.

And heh, platform differences already managed to surface 😆. Let's keep the `.android.js` and `.io.js` separate and independent for the time being to let this mature a bit. WDYT @maxme ?

maxme:
Looks good, :shipit:

PR #17, opened 2018-03-21
Use getBlockForType to get the right block edit function

maxme:


maxme:
Note: I imagine that we would move `getBlockForType` to its own file at some point. It would be similar of what we have in `gutenberg/api/index -> createBlock`

hypest:
> Note: I imagine that we would move getBlockForType to its own file at some point. It would be similar of what we have in gutenberg/api/index -> createBlock

Yes, that's my impression as well and I'll try to work to that end.

hypest:
:shipit:

PR #19, opened 2018-03-21
Register and load Code block

hypest:
Uses GB code to register and retrieve the `Code` block.

Note: Needs updating the submodule

Note 2: currently tests give out console error (`Block "core/code" is already registered.`) because the registration happens twice (2 tests in `App.test.js`). We need to figure out a better way to handle the registration lifecycle.

### DIFF
--- a/block-management/block-holder.js
+++ b/block-management/block-holder.js
@@ -34,7 +34,7 @@ export default class BlockHolder extends React.Component<PropsType, StateType> {
 
 	getBlockForType() {
 		if ( this.props.blockType === 'code' ) {
-			const blockType = getBlockType('core/code');
+			const blockType = getBlockType( 'core/code' );
 			const Code = blockType.edit;
 			// TODO: input text needs to be kept by updating the attributes
 			return (
--- a/block-management/block-holder.js
+++ b/block-management/block-holder.js
@@ -8,7 +8,7 @@ import { StyleSheet, View, Text, TouchableWithoutFeedback } from 'react-native';
 import Toolbar from './toolbar';
 
 // Gutenberg imports
-import { settings as codeBlock } from '../gutenberg/blocks/library/code';
+import { getBlockType } from '@gutenberg/blocks/api';
 
 type PropsType = {
 	index: number,
@@ -34,7 +34,8 @@ export default class BlockHolder extends React.Component<PropsType, StateType> {
 
 	getBlockForType() {
 		if ( this.props.blockType === 'code' ) {
-			const Code = codeBlock.edit;
+			const blockType = getBlockType('core/code');
+			const Code = blockType.edit;
 			// TODO: input text needs to be kept by updating the attributes
 			return (
 				<Code
--- a/block-management/block-holder.js
+++ b/block-management/block-holder.js
@@ -33,8 +33,8 @@ export default class BlockHolder extends React.Component<PropsType, StateType> {
 	}
 
 	getBlockForType() {
-		if ( this.props.blockType === 'code' ) {
-			const blockType = getBlockType( 'core/code' );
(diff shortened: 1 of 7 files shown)


maxme:
TIL about https://github.com/tleunen/babel-plugin-module-resolver

hypest:
Unfortunately, there are 3 places where module-mapping needs to be defined (babel, jest, flow). Hope there's a way to consolidate those somehow.

block-management/block-holder.js
```diff
@@ -34,7 +34,8 @@ export default class BlockHolder extends React.Component<PropsType, StateType> {
 
 	getBlockForType() {
 		if ( this.props.blockType === 'code' ) {
-			const Code = codeBlock.edit;
+			const blockType = getBlockType( 'core/code' );
```
maxme: Maybe we should just directly define 'code/code' in the sample in `block-manager.js`. I haven't checked but what kind of blockType strings will we get after the parsing step?

hypest: Yes, I think we should use the `core/code` blocktype name directly in the block list array. I can make the change in this PR (let me know!) unless we wait for the rest of the (hardcoded) blocks in the list to be loaded this way first.

> ... what kind of blockType strings will we get after the parsing step

Not sure about the parsing step but from what I see from the block definitions in the GB library, all current blocks are in the form `code/<name>`.

maxme: Ok, let's do the change in this repo and drop the `if ( this.props.blockType === 'code' ) {` here

hypest: Wait, we still need some form of the conditional since only the Code block is currently loaded that way... the rest of the blocks in the hardcoded list still need to fall back to `View`s.

maxme: Not sure to understand, can't we do this?

```diff
diff --git a/block-management/block-holder.js b/block-management/block-holder.js
index b7c11f9..1b91c49 100644
--- a/block-management/block-holder.js
+++ b/block-management/block-holder.js
@@ -33,12 +33,11 @@ export default class BlockHolder extends React.Component<PropsType, StateType> {
        }

        getBlockForType() {
-               if ( this.props.blockType === 'code' ) {
-                       const blockType = getBlockType( 'core/code' );
-                       const Code = blockType.edit;
-                       // TODO: input text needs to be kept by updating the attributes
+               const blockType = getBlockType( this.props.blockType );
+               if ( blockType ) {
+                       const Block = blockType.edit;
                        return (
-                               <Code
+                               <Block
                                        attributes={ { content: this.props.content } }
                                        setAttributes={ attrs => console.log( { attrs } ) }
                                />
```

hypest: We can, the `if ( blockType ) {` is what I was referring to as "some form of the conditional` :). That's what I'm about to do so, 👍.

hypest: Done with d55d521.

block-management/block-holder.js
```diff
@@ -33,8 +33,8 @@ export default class BlockHolder extends React.Component<PropsType, StateType> {
 	}
 
 	getBlockForType() {
-		if ( this.props.blockType === 'code' ) {
-			const blockType = getBlockType( 'core/code' );
+		const blockType = getBlockType( this.props.blockType );
+		if ( blockType ) {
 			const Code = blockType.edit;
```
maxme: Looks good, nitpicking comment: can we replace `const Code` to a more generic `const Block`?

hypest: Took the opportunity (256ab18) to generalize a bit more than just the renaming, by passing all the props to the block. Let me know what you think @maxme .

maxme:
:shipit:

PR #20, opened 2018-03-22
Introducing a Redux store

hypest:
This is an exploratory PR for adding a Redux store to hold the block list.

All current list actions (focus, movement, deletion) is performed via Redux Actions: reducers recalculate the state when each event is received. The list then gets the state change via its (React) props.

Test plan:
* Run the Expo app and start using the toolbar buttons that appear after you focus a block (code block needs tapping on its header to be "focused" currently). All buttons except the "Settings" button should work
* Try editing the content of the core/code block. Any modifications get reverted but the console should print the edited text with the modification included. The block currently misses its 'onChange' callback and the modifications get reverted by default.


hypest:
I should try and add a test or two though, to exercise the store.

store/actions/ActionTypes.js
```diff
@@ -0,0 +1,10 @@
+/** @format */
+
+export default {
+	BLOCK: {
+		FOCUS: 'BLOCK_FOCUS_ACTION',
+		MOVE_UP: 'BLOCK_MOVE_UP_ACTION',
+		MOVE_DOWN: 'BLOCK_MOVE_DOWN_ACTION',
+		DELETE_ACTION: 'BLOCK_DELETE_ACTION',
```
maxme: This should be `DELETE: 'BLOCK_DELETE_ACTION'` here not `DELETE_ACTION: 'BLOCK_DELETE_ACTION'`

hypest: Nice catch! Which makes me wonder how the tests pass :). Will look into it!

maxme: There should be a way to type check this with Flow, like in `block-management/constants.js`

maxme: > Nice catch! Which makes me wonder how the tests pass :). Will look into it!
The test passes because both are `undefined`

store/actions/test.js
```diff
@@ -0,0 +1,32 @@
...
+
+		it( 'should create an action to move block down', () => {
+			const action = actions.moveBlockDownAction( 1 );
+			expect( action.type ).toEqual( ActionTypes.BLOCK.MOVE_DOWN );
+			expect( action.rowId ).toEqual( 1 );
+		} );
+
+		it( 'should create an action to delete a block', () => {
+			const action = actions.deleteBlockAction( 1 );
+			expect( action.type ).toEqual( ActionTypes.BLOCK.DELETE );
```
maxme: We could add `expect( action.type ).toBeDefined();` here to prevent previous problem

store/actions/ActionTypes.js
hypest: OK, added some changes to make sure Actions, ActionTypes, the reduser and the store creation are included in Flow. Pretty sure more code will need the same treament. Will do in separate commits.

store/actions/test.js
hypest: Makes sense, added with `91f2180`.

maxme:
I tried something for fun and this is quite nice: 

```
diff --git a/package.json b/package.json
index a5e2cdc..3308110 100644
--- a/package.json
+++ b/package.json
@@ -38,6 +38,7 @@
     "react": "16.2.0",
     "react-native": "0.52.0",
     "react-redux": "^5.0.7",
-    "redux": "^3.7.2"
+    "redux": "^3.7.2",
+    "redux-logger": "^3.0.6"
   }
 }
diff --git a/store/index.js b/store/index.js
index 34be72e..6e7b10c 100644
--- a/store/index.js
+++ b/store/index.js
@@ -1,5 +1,6 @@
 /** @format */
-import { createStore } from 'redux';
+import { applyMiddleware, createStore } from 'redux';
+import logger from 'redux-logger'
 import { reducer } from './reducers';
 
 const initialState = {
@@ -54,6 +55,6 @@ const initialState = {
 };
 
 export function setupStore( state = initialState ) {
-	const store = createStore( reducer, state );
+	const store = createStore( reducer, state, applyMiddleware(logger));
 	return store;
 }
```

I haven't searched yet, but this would be amazing to have a tool to edit (or at least visualize) the state.

hypest:
Hmm, the Redux DevTools Chrome extension should help with visualizing/editing I think. Haven't played much with it yet though.

maxme:
> Hmm, the Redux DevTools Chrome extension should help with visualizing/editing I think. Haven't played much with it yet though.

I'll have a look. In the meantime, :shipit: !

PR #21, opened 2018-03-23
Have Travis test both platforms

hypest:
As we port GB code by utilizing the `.android.js` and `.ios.js` file variants, the code paths can differ. It will be useful to have the test suite be run for both so we can be sure that both paths are OK.

This PR:

#### Sets up a Jest config that picks up an environment variable to set the RN platform

To use that directly on your local machine: `TEST_RN_PLATFORM=android yarn test` to run with Android or or `TEST_RN_PLATFORM=ios yarn test` to run as iOS.

#### Sets up Travis to run 2 build jobs in parallel, one for each platform
Travis spawns 2 parallel build jobs.


.travis.yml
```diff
@@ -5,7 +5,18 @@ matrix:
...
     - language: node_js
       node_js: 8
-      env: LANE='node'
+      env:
+        LANE='node'
+        TEST_RN_PLATFORM='android'
+      cache:
+        yarn: true
+      script:
+        - ./.travis/travis-checks-js.sh
```
maxme: * We're going to run flow and prettier twice.
* When used locally `./.travis/travis-checks-js.sh` only run tests for default platform (Android). 

I wish we could use only one script for both travis and local checks, but we can't if we use travis specifics env. So I would remove `npm test` from `./.travis/travis-checks-js.sh`, add it to the travis config. And maybe have a second script for local checks (running `./.travis/travis-checks-js.sh` *and* `npm test` for each platform).

hypest: Good points @maxme .

I modified the Travis check script to:
* Conditionally run the flow+prettier or Jest test checks (ac7354e). Added a third parallel Travis build that runs the flow+prettier check while the other two builds will only run the Jest tests.
* If the test platform variable is not set, the script will run the test suite twice, once for each platform (6389641). That way, when used locally, the script will exercise both platforms.

Let me know what you think.

maxme:
:shipit:

PR #22, opened 2018-03-26
Wire up setAttributes

hypest:
_This PR builds upon the Redux work done in #20 and targets it_

Features in this PR:
* The `Code` block's `setAttributes` prop is wired up to the `ActionTypes.BLOCK.UPDATE_ATTRIBUTES` action
* The reducer responds to the `UPDATE_ATTRIBUTES` action causing the store to mutate and update the content of the block
* Using the block's `uid` field to identify it instead of the array index. This will bring us closer to the GB (web) data schema.

Test plan:
* Run the expo app and start editing the content in the core/code block. Editing should look just normal, with text updating as you edit it.


block-management/block-holder.js
```diff
@@ -7,14 +7,12 @@ import React from 'react';
 import { StyleSheet, View, Text, TouchableWithoutFeedback } from 'react-native';
 import Toolbar from './toolbar';
 
+import type { BlockType } from '../store/';
```
maxme: TIL

maxme:
Note: updated base branch to `master` as the redux PR was merged.

maxme:
+1 for using `uid`s instead of row ids, this will be useful when we chain actions.

block-management/block-manager.js
```diff
@@ -8,40 +8,36 @@ import { StyleSheet, Text, View, FlatList, TextInput } from 'react-native';
...
-		blockType: string,
-		content: string,
-		focused: boolean,
-	}>,
+	onChange: ( uid: string, attributes: mixed ) => void,
+	focusBlockAction: string => mixed,
+	moveBlockUpAction: string => mixed,
+	moveBlockDownAction: string => mixed,
+	deleteBlockAction: string => mixed,
+	blocks: Array<BlockType>,
```
maxme: +1 I did that as well when I started integration with the native list.

maxme:
Looks good, :shipit: 

While testing this, I realized we have an issue with the virtual keyboard and the scrolling list while editing the EditText field. I wonder if this will solved automagically by switching to the native list.

hypest:
> we have an issue with the virtual keyboard and the scrolling list while editing the EditText field

Can you expand on the issue @maxme ? I don't see something strange on my Nexus 5X.

PR #23, opened 2018-03-27
Add Redux Devtool Enhancer as a developer dependency to the project

maxme:
`devToolsEnhancer` allows to
* Visualization of state, state history and state diff, and
* Dispatch actions from Chrome Redux devtools.

screen shot 2018-03-27 at 16 40 02

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,8 @@
     "jest-expo": "25.0.0",
     "prettier": "git+https://github.com/Automattic/calypso-prettier.git#calypso-1.9",
     "react-native-scripts": "1.11.1",
-    "react-test-renderer": "16.2.0"
+    "react-test-renderer": "16.2.0",
+    "remote-redux-devtools": "^0.5.12"
   },
   "main": "./node_modules/react-native-scripts/build/bin/crna-entry.js",
   "scripts": {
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -9,7 +9,8 @@
     "jest-expo": "25.0.0",
     "prettier": "git+https://github.com/Automattic/calypso-prettier.git#calypso-1.9",
     "react-native-scripts": "1.11.1",
-    "react-test-renderer": "16.2.0"
+    "react-test-renderer": "16.2.0",
+    "remote-redux-devtools": "^0.5.12"
```
hypest: Not sure how this works but, if it's a dev only dep, will importing it into `store/index.js` (as below) become broken in production?

maxme: Good point, I'll have to add a check and enable it for dev mode only.

hypest: https://github.com/zalmoxisus/redux-devtools-extension#14-using-in-production might be relevant.

diegoreymendez: Testing, will delete.  Please ignore.

maxme: Updated in 8692a96

hypest: Verified it doesn't load in production mode.

hypest:
🛳️it

PR #24, opened 2018-03-28
Use the .native.js file extension

hypest:
The GB has been updated to use single `.native.js` files for platform specific code instead of the dual `.android.js`+`.ios.js`.


maxme:
:shipit:

PR #27, opened 2018-03-30
Auto import React, don't import it in GB code

hypest:
When imported directly in GB code, GB's eslint errors out because it
finds React as an unused import. So, leaving it out and auto-importing
it in the RN project.

To test:
* Make sure you've run `git submodule update` to get the latest of the GB branch
* run `npm run lint` inside the `gutenberg` subdir and notice no error occurs


maxme:
`cd gutenberg && npm run lint`: ✅

PR #28, opened 2018-04-02
Serialize code block

hypest:
Use the `jsx-to-string` package to serialize the block into html.

This is only a basic PR to explore how to do the serialization.


maxme:
It seems to me like a weird way to serialize the blocks. I haven't explored much, but should we use https://github.com/WordPress/gutenberg/blob/master/blocks/api/serializer.js instead?

hypest:
`serializer.js` was my starting point (`getSaveContent()` for example) but stumbled upon the code depending on `react-dom` to do the rendering-to-string. I couldn't find a practical way to work around that (RN incompatible) dep so I abandoned it. I will try again. Chances are, `jsx-to-string` will be appear again but yeah, lemme dig some more @maxme .

hypest:
👋 @maxme , this PR is now using code directly from GB to serialize to html and tests *finally* pass (bc51e55). Can we resume reviewing and perhaps merging this?

One issue known at this point is that the serialization is done but there are no html comments that define the blocks. I'd suggest we pick up that task in a separate PR. Let me know what you think.

hypest:
Turns out it wasn't too hard to use the function that serializes the GB html comments too (74729c7).

maxme:
Looking at the Web version output, I think we should add newlines between blocks to make the HTML output clearer (it's at least useful for us to debug stuff):

```diff
diff --git a/block-management/block-manager.js b/block-management/block-manager.js
index ede412d..2752796 100644
--- a/block-management/block-manager.js
+++ b/block-management/block-manager.js
@@ -81,10 +81,10 @@ export default class BlockManager extends React.Component<PropsType, StateType>
                        .map( block => {
                                const blockType = getBlockType( block.name );
                                if ( blockType ) {
-                                       return serialize( [ block ] );
+                                       return serialize( [ block ] ) + '\n\n';
                                }

-                               return '<span>' + block.attributes.content + '</span>';
+                               return '<span>' + block.attributes.content + '</span>\n\n';
                        } )
```

hypest:
👍 for easier debugging, done with 6c4f061.

block-management/block-manager.js
```diff
@@ -140,10 +170,20 @@ export default class BlockManager extends React.Component<PropsType, StateType>
...
 		flex: 1,
 		backgroundColor: '#ccc',
 	},
+	htmlView: {
+		flex: 1,
+		backgroundColor: '#fff',
+	},
+	switch: {
+		flexDirection: 'row',
+		justifyContent: 'flex-start',
```
maxme: Nitpicking, it think `flex-start` is the default option for `justifyContent`.

maxme:
Good job @hypest - :shipit:  (we don't care about https://github.com/wordpress-mobile/gutenberg-mobile/pull/28#discussion_r180727660)

block-management/block-manager.js
hypest: Correct! That's leftover code after experimenting (and failed eventually) with having the switch include the text label.

PR #32, opened 2018-04-06
New yarn clean and yarn clean-reset commands

maxme:
New commands:

* `yarn clean`: delete watchmen, delete `node_modules`, delete lock file and clean up the metro cache.
* `yarn clean-install`: run `yarn clean` + `yarn install` (pretty common).

After running `yarn clean-install`, You won't need to add the `--reset-cache` option to `yarn start`, the metro cache will also be cleaned up which is handy in most case.

Note: this is very useful when switching between branches.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,9 @@
     "test-debug": "cross-env NODE_ENV=test node --inspect-brk node_modules/jest/bin/jest.js --runInBand --verbose --config jest.config.js",
     "flow": "flow",
     "prettier": "prettier --write *.js *block-management/*.js",
-    "prettier-check": "prettier -l *.js *block-management/*.js"
+    "prettier-check": "prettier -l *.js *block-management/*.js",
+    "clean": "watchman watch-del-all; rm -rf node_modules; rm -f yarn.lock; rm -rf $TMPDIR/react-*; rm -rf $TMPDIR/metro-cache-*",
+    "clean-install": "yarn clean; yarn"
   },
   "dependencies": {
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -24,7 +24,9 @@
     "test-debug": "cross-env NODE_ENV=test node --inspect-brk node_modules/jest/bin/jest.js --runInBand --verbose --config jest.config.js",
     "flow": "flow",
     "prettier": "prettier --write *.js *block-management/*.js",
-    "prettier-check": "prettier -l *.js *block-management/*.js"
+    "prettier-check": "prettier -l *.js *block-management/*.js",
+    "clean": "watchman watch-del-all; rm -rf node_modules; rm -f yarn.lock; rm -rf $TMPDIR/react-*; rm -rf $TMPDIR/metro-cache-*",
```
hypest: How about cleaning the Jest cache as well? Command for that is `yarn test --clearCache` or `rm -Rf $TMPDIR/jest_*`.

maxme: 👍 I'll add this

maxme: Updated in 55c951c  - I added it before removing `node_modules`, but it might fail if the jest module is not installed, but won't make the full clean task fail.

hypest: Ah, I see. I don't think that anything goes wrong with the cleanup process if the node_modules dir is missing (for example, when running `yarn clean` twice one after the other) but cool, let's leave it as is 👍 .

hypest:
:shipit:

PR #36, opened 2018-04-12
Style using CSS Modules

hypest:
Start using SASS to define the styling of the views.

This solution is based on https://github.com/kristerkari/react-native-css-modules.
The idea here is to eventually be able to load the `.scss` files from the GB sourcetree, ideally with only some modifications/specializations.

* The `core/code` block is styled with setting `font-family` to `monospace`. [See here](https://github.com/wordpress-mobile/gutenberg-mobile/pull/36/files#diff-8bfcd9a8b6b235452b2a4b25c22e466aR2).
* The Jest tests don't support the CSS module loading (yet, hopefully) so, mocking the CSS modules for now
* On the Gutenberg code tree side, the styles are specialized by defining `.native.scss` versions of the SASS files


maxme:
I just noticed `font: monospace` is not working on iOS. I get the following error:

```
Unrecognized font family 'monospace'

+[RCTFont updateFont:withFamily:size:weight:style:variant:scaleMultiplier:]
    RCTFont.mm:295
-[RCTFontAttributes updateFont]
-[RCTFontAttributes setFontFamily:]
```

[Here is a list of the stock fonts by platform.](https://github.com/react-native-training/react-native-fonts) I think it should be `Courrier` or `Courrier New` for iOS.

maxme:
Should we export the style from `block-management/toolbar.js` to a `.scss` file?

block-management/block-manager.scss
```diff
@@ -0,0 +1,20 @@
+.container {
```
maxme: We should maybe start adding `/** @format */` to style files. And check/format them with prettier.

.babelrc
```diff
@@ -21,6 +21,13 @@
           ".android.js"
         ]
       }
+    ],
+    "react-native-classname-to-style",
+    [
+      "react-native-platform-specific-extensions",
+      {
+        "extensions": ["native.scss", "css", "scss", "sass"]
```
maxme: Are `ios.scss` and `android.scss` missing from this list?

hypest:
> I think it should be Courrier or Courrier New for iOS.

I changed `monospace` to `courier` as the common denominator with 05b1dd8. We can't have platform-specific files yet so, just using what's common between the platforms for now.

> Should we export the style from block-management/toolbar.js to a .scss file?

Done with 8d0821b for completeness.

Will work on the rest of the feedback.

block-management/block-manager.scss
hypest: Good idea! Done with 686c0a3.

hypest:
I think I addressed all the feedback so far @maxme , ready for another pass, thanks!

maxme:
Something weird happens on iOS:

https://bia.is/s/PQRr/style-sass-ios.mov

Note: everything looks good in the `master` branch.

hypest:
> Something weird happens on iOS:

Fixed with ccd5038 @maxme .

maxme:
:shipit: :shipit: :shipit:!

PR #37, opened 2018-04-17
Support for the more block

maxme:
There is almost no changes on this side of the project:
- Attributes type updated.
- Passed the focused state (which is not used yet in the more block).
- New more block in sample data.

Note: this one comes with a patch on the gutenberg side, but the PR is not open yet https://github.com/WordPress/gutenberg/tree/rnmobile/port-more-block


store/index.js
```diff
@@ -67,6 +67,14 @@ const initialState: StateType = {
...
 		},
 		{ ...codeBlockInstance, focused: false },
+		{
+			uid: '5',
+			name: 'core/more',
+			isValid: true,
+			attributes: { customText: undefined },
+			innerBlocks: [],
+			focused: false,
+		},
```
hypest: Can we reuse the `createBlock()` method to create the block (see `codeBlockInstance` earlier for example) to stay close to the original code generation? `createBlock` should ideally support the `core/more` block at this point.

maxme: Good point, fixed in 359eae6

hypest:
🛳️it!

PR #40, opened 2018-04-19
Update toolbar, block and html view styling

maxme:
This PR changes:
* Toolbar styling (replaced the buttons).
* Block content styling (added padding).
* HTML view styling (padding + font).
* Introduce `fonts.ios.scss`, `fonts.android.scss` that I think will be useful later to factorize some styles.


block-management/fonts.android.scss
```diff
@@ -1,5 +1,5 @@
 /** @format */
 
-.block_code {
+%monospace-font {
```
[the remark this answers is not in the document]
maxme: Updated in b402c73

package.json
```diff
@@ -3,8 +3,8 @@
   "version": "0.1.0",
   "private": true,
   "config": {
-    "jsfiles": "index.js app/*.js app/**/*.js block-management/*.js block-management/**/*.js store/*.js store/**/*.js",
-    "scssfiles": "app/*.scss app/**/*.scss block-management/*.scss block-management/**/*.scss store/*.scss store/**/*.scss"
+    "jsfiles": "index.js src/*.js src/**/*.js",
```
maxme: Moving mobile-related code to `src/` is worth it.

hypest:
I think there's an issue on the iOS app with the toolbar when the block is selected:
screen shot 2018-04-19 at 15 19 59

Does that happen on your side @maxme ?

maxme:
Ouch `TouchableNativeFeedback` is not supported on iOS... I'll have to do some changes, I'll ping you when it's ready for review.

sass-transformer.js
```diff
@@ -60,7 +60,7 @@ if ( reactNativeMinorVersion >= 52 ) {
 }
 
 // TODO: need to find a way to pass the include paths and the default asset files via some config
-const autoImportIncludePaths = [ 'gutenberg/edit-post/assets/stylesheets' ];
+const autoImportIncludePaths = [ 'src', 'gutenberg/edit-post/assets/stylesheets' ];
```
hypest: The need to add `src` seems weird and I feel it's an issue withe the `sass-transformer`. I assume you've added this to enable the `@import '../variables.scss'` statement, right?

hypest: Pushed de8629d that fixes the issue with the transformer. Follow up 0d2b501 removes the hardcoded import include path.

maxme: Good job, thanks

maxme:
> Ouch TouchableNativeFeedback is not supported on iOS... I'll have to do some changes, I'll ping you when it's ready for review.

Very disappointed about that... There are some simple solutions to have a generic button-feeling for both platform, like that: 
https://medium.com/differential/better-cross-platform-react-native-components-cb8aadeba472 - but I opted for a simpler one since we're not yet at trying to make it design-ready.

Updated in 69e6717

PR #42, opened 2018-04-27
Upgrade to RN v0.55.4

hypest:
This PR upgrades to the latest RN version (and the accompanying React version that is compatible) in preparation to start including GB-web code that depends on React 16.3.x.

~An interesting and weird situation with this upgrade is that the SCSS style files (or the transformer, not sure) causes Metro to not function correctly and the resulting bundle doesn't work. Couldn't find a clean solution to that and as a workaround I've renamed the style files so they are not parsed when Metro tries to resolve the `import` modules.~

EDIT: Turns out, because of the way Metro extends the list of extensions provided via `getSourceExts`, the same-name SCSS files were resolved first instead of the intended JS files. See https://github.com/kristerkari/react-native-sass-transformer/issues/5. Worked around the issue with 58ccfeb.


maxme:
I remember there was an issue with babel+metro+RN 0.54.x that prevented us to update. I'll check that and see if it's fixed with 0.55.x.

hypest:
Even though this upgrade works in this PR, when one tries to use it with a more recent version of GB ([this commit and up](https://github.com/WordPress/gutenberg/pull/5816/commits/d5305704ecdf1bf29b8ccf9b5d0a98f26b3e32db)) transpiling stops working properly and leads to errors like `Unexpected use of reserved work 'import' (http://localhost:8081/index.delta?platform=android&dev=true&minify=false:63160)`.

The babel configuration in `gutenberg/package.json` messes up the build. Not sure why that same configuration doesn't break it in this PR.

@maxme , is this perhaps the issue you're referring to?

hypest:
I spent more than a few days trying to figure out why the latest RN version doesn't play well with latest GB. I think I pinpointed the issue to a change in Babel (https://github.com/babel/babel/releases/tag/v7.0.0-beta.45). GB's package.json includes Babel config that gets picked up and breaks the RN build (transpilation). `beta.44` used in previous RN version works fine but `beta.45/46` doesn't.

Unfortunately, the only "simple" workaround I have found so far is to _delete_ the GB's package.json as a pre-build action. Super nasty and dirty workaround but, yeah...

BT, I tried going the `babel.config.js` route as mentioned in https://github.com/babel/babel/pull/7358 but couldn't make it work: build acts as the config is not loaded at all.

So, this means that we shouldn't merge the RN version upgrade just yet, or get ready to live with the "delete package.json" workaround until a better fix surfaces.

Any better suggestions welcome!

hypest:
Added 8bdbff1 to shortcircuit the GB Babel configuration. With this hack, the RN build completes as normal when using recent versions of GB's code.

hypest:
Upgraded to `v0.55.4` of RN with 099c524.

daniloercoli:
Tested on both platforms and worked without problems! :shipit:

PR #45, opened 2018-05-24
Use the GB parser pipeline

hypest:
This PR introduces the usage of the Gutenberg grammar + block attributes parser.

Instead of having the hardcoded blocks list be constructed by manually calling `createBlock()`, this PR uses the GB Parser to construct the blocks from html snippets, just like those will be found on a real post.


gziolo:
`src/store/post-grammar-gb.js` - why this file is required?

In Gutenberg we import it using `import { parse as grammarParse } from './post.pegjs';`. Did you try to replicate Webpack loader logic to handle it?

gziolo:
I need to setup everything locally and spend some time investigating `jsdom` implementation to give more detailed feedback. The general direction taken here makes sense. I'm wondering if we can find a better way to load grammar similar to how it is done for web.

hypest:
Thanks for having a look @gziolo !

> src/store/post-grammar-gb.js - why this file is required?
>
> In Gutenberg we import it using import { parse as grammarParse } from './post.pegjs';. Did you try to replicate Webpack loader logic to handle it?

I did give it a try utilizing babel and the transformation steps in Metro but had no success in the time frame I devoted to it. To move things forward I resorted to the "prebuilt" grammar parser. Besides, we can always try again in the future, and we should to make sure the parser stays up-to-date.

> I need to setup everything locally and spend some time investigating jsdom implementation to give more detailed feedback.

Nice! Feel free to ping me to give a hand.

> The general direction taken here makes sense.

Good to know! Let's wait for the extra pass before we merge, after you try it locally. Thanks! 🙇

gziolo:
> I did give it a try utilizing babel and the transformation steps in Metro but had no success in the time frame I devoted to it. To move things forward I resorted to the "prebuilt" grammar parser. Besides, we can always try again in the future, and we should to make sure the parser stays up-to-date.

Fair enough, it's okey as a temporary solution, but we should try to bring it closer to what web does in the near future. I hope we will tackle it somehow as part of publishing `blocks` to npm. Not sure, it seems to be a blocker for npm package and a very tricky one :)

hypest:
> Not sure, it seems to be a blocker for npm package

Oh, you mean we cannot package and publish the `src/store/post-grammar-gb.js` file atm?

gziolo:
> > Not sure, it seems to be a blocker for npm package
> 
> Oh, you mean we cannot package and publish the `src/store/post-grammar-gb.js` file atm?

Webpack does it atm, so not quite sure how this fits into the whole flow. Distribution files for packages are generated with Babel.

hypest:
Fixed conflicts with base branch with f8d875f.

gziolo:
Testing steps:

`yarn clean`
`git submodule update`
`yarn`
`yarn start`
`yarn ios` or `yarn android`

gziolo:
`src/store/post-grammar-gb.js` can be removed. We use its copy inside `blocks` module.

Otherwise, everything looks and works as expected. Let's focus on bringing a compiled version of the parser inside Gutenberg.

hypest:
> src/store/post-grammar-gb.js can be removed. We use its copy inside blocks module.

Done!

Thanks for the review @gziolo , will merge now!

PR #59, opened 2018-06-25
Build: Make app work on iOS

gziolo:
This is a set of changes I had to apply to my local copy to make it work on iOS using the simulator they provide.

### Testing
`yarn ios` - make sure it works
`yarn test` - make sure it still passes

Hopefully it should work also with:
`yarn android` :)

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -30,11 +30,7 @@ module.exports = {
 		'node',
 	],
 	moduleNameMapper: {
-		'@wordpress/blocks': '<rootDir>/gutenberg/blocks',
-		'@wordpress/editor': '<rootDir>/gutenberg/editor',
-		'@wordpress/element': '<rootDir>/gutenberg/element',
-		'@wordpress/hooks': '<rootDir>/wordpress/hooks',
-		'@wordpress/utils': '<rootDir>/gutenberg/utils',
+		'@wordpress\\/(blocks|editor|element)$': '<rootDir>/gutenberg/$1',
 		'@gutenberg': '<rootDir>/gutenberg',
 
 		// Mock the CSS modules. See https://facebook.github.io/jest/docs/en/webpack.html#handling-static-assets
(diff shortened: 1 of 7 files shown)


src/app/globals.js
```diff
@@ -1,6 +1,6 @@
 /** @format */
 
-import { createElement } from '@wordpress/element';
+import { createElement } from '@gutenberg/element';
```
gziolo: I updated it to follow the general pattern where all references in `src` folder point to `@gutenberg/*` instead of using a local copy of `@wordpress/*`. We can update it back to `@wordpress/*` as soon as we have all packages published to `npm`.

jest.config.js
```diff
@@ -30,11 +30,7 @@ module.exports = {
 		'node',
 	],
 	moduleNameMapper: {
-		'@wordpress/blocks': '<rootDir>/gutenberg/blocks',
-		'@wordpress/editor': '<rootDir>/gutenberg/editor',
-		'@wordpress/element': '<rootDir>/gutenberg/element',
-		'@wordpress/hooks': '<rootDir>/wordpress/hooks',
-		'@wordpress/utils': '<rootDir>/gutenberg/utils',
```
hypest: I wonder if we need to remove the relevant mapping in [the Flow config](https://github.com/wordpress-mobile/gutenberg-mobile/blob/update/run-ios/.flowconfig#L86) as well. WDYT?

gziolo: I wasn't aware, we don't use it so we should remove.

hypest:
Forgot to pay attention to the local `hooks.js` override in https://github.com/wordpress-mobile/gutenberg-mobile/blob/27a517de9b8bf50020ba0574086af80abf25e4fd/wordpress/hooks.js. We should probably check if that's needed anymore or perhaps conflicting with the npm package now pulled.

PR #61, opened 2018-06-28
Use latest gb master as per July 2nd, 2018

hypest:
This PR refreshes the reference to GB's master.

⚠️**Headsup**: this PR adds an new setup step: _manulally_ install the Gutenberg packages (`npm install` inside the GB directory.

Added a couple of yarn scripts (`start:reset`, `clean:runtime`, `clean:setup`) to help modularize and provide some shortcuts to frequently used CLI commands.


hypest:
The build (tests) fail because the GB package `babel-plugin-import-jsx-pragma` doesn't get resolved correctly. That package is consumed in the GB's Babel configuration directly from the `gutenberg/packages` folder and we might need to find a workaround for that until the package is published to NPM normally like the rest of the packages in that folder.

hypest:
Fixed the issue with the GB Babel configuration having a local hardcoded path to `babel-plugin-import-jsx-pragma` by employing the "bypassing" `.babelrc` file trick.

The Travis build now fails because of some issue with the shell commands around the `.babelrc` trick. Those shell commands succeed locally so this must be due to some mismatch with the shell running in the Travis VM. I've contacted Travis support to help.

Even after that shell issue gets fixed though, there's an additional issue that will make the testsuite fail: Block validation fails for the `more` block.
That happens because the `@wordpress/element` package entrypoint (`main` in its `package.json`) is pointing to the index.js file (including its file `.js` extension) which shortcircuits the Metro resolving logic, failing to pick up the native `index.native.js` of the package. That's an issue that https://github.com/WordPress/gutenberg/pull/7675 would solve but in there we opted for introducing the `react-native` entrypoint instead, as a safer choice. Under the new light where the Jest tests fail here we will need to reconsider a solution. /cc @gziolo

hypest:
I've changed (aac9681) the source from where the GB packages are consumed to point directly to the GB `packages` folder. This has these effects:
* The latest packages can be used directly while developing, plus we can quickly do/test any modifications we need in them. No need to publish the packages to NPM before using them.
* The GB packages need to be built (`npm i` in the GB directory) which takes a bit of time, especially the first time. I left this step as a **manual** one until we find an efficient way to automate it via a script.
* Modifications to the GB package sources need to be done directly in the built files (e.g. the ones in `gutenberg/packages/element/build/*`) for the changes to be picked up by the app (live-reload works too). If the ones in `src/` are modified, that will require a `npm  i` to build them.

Also, updated the GB reference to point to 2a012ff which includes the `element` package fixes (see #7675) and #7691).

With all the changes above the build is now green on Travis.

README.md
```diff
@@ -28,6 +28,14 @@ Before running the demo app, you need to download and install the project depend
 npm install
 ```
 
+The Gutenberg packages need to be installed and built. This is done via the following commands:
+
+```
+pushd gutenberg
```
gziolo: Should we put it in the shell script and run as `postinstall` script for the repository?

yarn.lock
```diff
@@ -6547,6 +6547,10 @@ remark-parse@4.0.0:
     vfile-location "^2.0.0"
     xtend "^4.0.1"
 
+rememo@^3.0.0:
```
gziolo: The only downside of the current setup is that we need to ensure that all npm packages are referenced in `package.json` in the main repository folder.

README.md
hypest: Perhaps. I refrained from automating it from the beginning so to give us time/space to see how we use it actually but, I can give it a try now and see.

daniloercoli:
Looks good! :shipit:

README.md
hypest: So, gave `postinstall`. Our patterns these days involve more than a few `yarn install` attempts and that will cause `npm i` to be called too many times, unnecessarily. Ideally, we want `npm i` to be called rarely, like, when the GB repo is cloned or when the GB branch is changed.

I think we can leave `npm i` out of automation for the following days/weeks and re-assess, depending on the patterns we notice.

hypest: And btw, thanks for the review @gziolo !

PR #62, opened 2018-06-28
Update AztecRN wrapper to the latest version and fix typing delays

daniloercoli:
This is based on the work done in https://github.com/wordpress-mobile/react-native-aztec/pull/24
and send back the `eventCount` variable together with the text.

This way we can discard superfluous upgrades on native side.

**Note** This PR uses `feature/aztec-submodules` as base branch.


daniloercoli:
Updated `react-native-aztec` hash, and ready for review.

PR #64, opened 2018-06-28
Add step to clear yarn cache of packages.

SergioEstevao:
This PR adds a extra step to the clean build option to clear the yarn cache.

This helps on cases where the folder of caches is corrupted by any reason.

Question:  I notice that one of the clean steps is `yarn test --clear cache` has a step? Do we need to test to clear the cache of the test results?


hypest:
Good catch @SergioEstevao , funny how we missed that cleaning step before. LGTM!

> Question: I notice that one of the clean steps is yarn test --clear cache has a step? Do we need to test to clear the cache of the test results?

Yes, we definitely need to. Jest tests have their own cache and not all changes get picked up when running the tests so, this cleaning step is indeed important.

That said, the `clean` script has the jest test directory deletion that is probably redundant as that's performed by the `yarn test --clearCache` step too. We could remove the deletion but doesn't exactly matter much in this case so, we might as well keep it as is.

PR #66, opened 2018-07-09
Point to the GB packages src folder

hypest:
This PR changes the way the Gutenberg packages are consumed by removing the need to compile (`npm install`) the packages first, [introduced here](https://github.com/wordpress-mobile/gutenberg-mobile/pull/61#issuecomment-402895217).

Changes:
* On the GB side, add the `react-native` package.json entry point and have it point to the `src/index` module (notice the lack of the `.js` extension, that's on purpose to let Metro do its smart resolve)
* Configure Jest to point directly to the `src/index` module as well
* Add some more transformations to be performed by Babel, to support `async/await`

To test:
* Clean the GB folder from the compiled artefacts. A handy way to do that is `git clean -fdX` which deletes the non-tracked files/folders.
* Run the RN app and it should compile and run correctly


mzorz:
Thanks for the steps provided @hypest ! I was able to make it run after going through these provided steps, copying here for completeness:

1. `git clone git@github.com:wordpress-mobile/gutenberg-mobile.git`
2. `cd gutenberg/`
3. `git checkout feature/direct-packages-src`
4. `git submodule init`
5. `git submodule update`
6. `yarn install`
7. `yarn clean:runtime && yarn start --reset-cache` <- using the new script to clear caches (`clean:runtime`) along with the usual `--reset-cache`
8. and in another terminal window: `yarn android`

.babelrc
```diff
@@ -22,6 +22,8 @@
         ]
       }
     ],
+    "transform-async-to-generator",
+    "transform-async-generator-functions",
```
mzorz: I don't understand why these two are needed, can you provide an explanation to it please? In the PR description I can read:
`Add some more transformations to be performed by Babel, to support async/await`

But still, it's not evident to me _where_  async/await is being used. Is this in preparation for an upcoming PR, maybe?

.travis/travis-checks-js.sh
```diff
@@ -19,11 +19,6 @@ if [ "$CHECK_CORRECTNESS" = true ] ; then
 fi
 
 if [ "$CHECK_TESTS" = true ] ; then
-  # build the Gutenberg packages
-  pushd gutenberg
-  npm i || pFail
-  popd
-
```
mzorz: Please remember to remove the steps from README.md as well

jest.config.js
```diff
@@ -31,7 +31,7 @@ module.exports = {
 	],
 	moduleNameMapper: {
 		'@wordpress\\/(blocks|editor)$': '<rootDir>/gutenberg/$1',
-		'@wordpress\\/(data|element|deprecated)$': '<rootDir>/gutenberg/packages/$1',
+		'@wordpress\\/(data|element|deprecated)$': '<rootDir>/gutenberg/packages/$1/src/index',
```
mzorz: I believe this is the key to this PR - could you explain how this would be benefitial, with an example, maybe? Otherwise I can only assume you're going to be using this elsewhere (it's not apparent from this PR itself, what the value it may add is)

src/store/index.js
```diff
@@ -108,7 +108,7 @@ const initialState: StateType = {
 };
 
 const devToolsEnhancer =
-	( 'development' === process.env.NODE_ENV && require( 'remote-redux-devtools' ).default ) ||
+	// ( 'development' === process.env.NODE_ENV && require( 'remote-redux-devtools' ).default ) ||
```
mzorz: If this is not needed, let's remove this line. Otherwise, if this is something you'll need for later, please let's add a `TODO` comment here, explaining what and why succinctly.

jest.config.js
mzorz: Also, this is the jest configuration, but do we have a place where it's made clear and apparent that the main configuration uses the source files?

rn-cli.config.js
```diff
@@ -7,6 +8,10 @@ enm.canvas = __dirname + '/__mocks__/nodejsMock.js';
 
 module.exports = {
 	extraNodeModules: enm,
+	getBlacklistRE: function() {
+		// blacklist the GB packages we want to consume from NPM (online) directly
+		return blacklist( [ /gutenberg\/packages\/(autop|hooks|i18n|is-shallow-equal)\/.*/ ] );
+	},
```
mzorz: Again here - I got more context from a chat elsewhere about this being denylisted as we think we won't need to modify source files for. That makes a lot of sense, but just by reading code alone (not even this PR, but the code in a general way) it's difficult to understand that we are using the source files and denylisting a subset of modules for the purpose of ease at modifying while developing. Can we add a nice comment in code that connects this part and <_wherever it's indicated in code we're using source files_>?

.babelrc
hypest: Good question! Here's what happens: by trying to use the `src` version of the files in the GB packages, we by pass the step where the GB build transpiles them to ES5. So, we need to help Babel compile those for us.

The `async/await` usage is actually in the GB packages, not the RN app. See https://github.com/WordPress/gutenberg/blob/ccbb54b4de3f0c471fa938faecb6e52c166f3991/packages/data/src/index.js#L175. For example, comment out the `"transform-async-generator-functions"` line in our Babel configuration and see the error that is emitted. It will be in code inside the GB `data` package.

Let me know if this answers your question. 👍

.travis/travis-checks-js.sh
hypest: Good catch! Will do it now.

jest.config.js
hypest: > I believe this is the key to this PR

Not quite key per se, for 2 reasons:
1. It modifies how the packages are resolved (overwritten in this case) for the _tests_ (Jest), not the app code
2. It's basically a workaround because Jest resolves the entry point of a package by reading the `main` field in the `package.json`. For example, for the `element` package the entry point is the *built* file. [See here on the GB repo](https://github.com/WordPress/gutenberg/blob/master/packages/element/package.json#L20-L22). In an ideal world, the Jest tests would use the same entrypoint as the RN app so we wouldn't need to configure Jest separately.

A question one could raise here is: *why do we even want to use the package _source_ files instead of the already transpiled ones?* The answer is that we prolly want to do that for the near future so we can iterate fast on the packages themselves. For example, with the current PR, one can modify the package `src/index.js` and have the change be picked up immediately by the RN app. With the built files, one would need to manually go through `npm install` to rebuild the packages before using the change.

I expect that in the not-so-distant future, we will revert back to using the built files after the project matures enough and we only introduce changes to the packages rarely.

> do we have a place where it's made clear and apparent that the main configuration uses the source files?

Yes, it's the `react-native` entry point, [here for example on the GB repo](https://github.com/WordPress/gutenberg/blob/master/packages/element/package.json#L22).

src/store/index.js
hypest: We don't _need_ it but the `remote-redux-devtools` offers a very nice debugger for Redux. I had to comment it out because it's not super important at the moment and I couldn't make it work in this PR in the short amount of time I was willing to devote to it. I left it in as a comment only as a reminder to try reinstate it in the future. Perhaps marking it with "TODO:" can help there.

rn-cli.config.js
hypest: Fair point. The packages we currently load from source have been chosen a bit arbitrarily, except for the `element` package in which we already have native code and so it's more likely we'll want to modify-and-try out fast. The other packages we use from source (`data` and `deprecated`) don't have native code and so, I just went ahead and made them resolved from NPM (with 8e5ebc8).

> Can we add a nice comment in code that connects this part and <wherever it's indicated in code we're using source files>?

I added a relevant comment (with 08f3626) but notice that effectively, there's nowhere something indicating that a package is loaded directly from source. What happens is that Metro automagically figures out that a module imported in code ([example](https://github.com/wordpress-mobile/gutenberg-mobile/blob/master/src/globals.js#L3)) can be satisfied by the package found inside the `gutenberg/packages/<module>` folder.

hypest:
Thanks for the review and the detailed questions @mzorz ! I tried to answer them but please, let me know if anything still doesn't make good sense.

jest.config.js
mzorz: > Not quite key per se, for 2 reasons:

Right,  I realized it was just for jest in my second comment :) - thanks for pointing to the `main`,  `module` and `react-native` keys in `package.json`, these values made sense 👍

This paragraph here is 🏅 :

> A question one could raise here is: why do we even want to use the package source files instead of the already transpiled ones? The answer is that we prolly want to do that for the near future so we can iterate fast on the packages themselves. For example, with the current PR, one can modify the package src/index.js and have the change be picked up immediately by the RN app. With the built files, one would need to manually go through npm install to rebuild the packages before using the change.

This is probably what I'd find useful the most in the PR description to understand the purpose of the PR itself - maybe a suggestion for future PRs 🙇 (thank you very much for the clarification)

.babelrc
mzorz: Thanks, I'd not find that myself other than by commenting that out as suggested

mzorz:
Thanks for your thorough answers @hypest ! Having tested and read, I've nothing to add except LGTM :shipit:

hypest:
Thanks for merging @mzorz ! I amended the PR description, adding links to the conversations we had in the comments, to help the reader understand the changes.

Among other things, I can see how the short PR description (`removing the need to compile (npm install) the packages first`) can get lost in the sea of unfamiliar notions, intricacies and workarounds. It's not always easy to figure out what's known or not so, even as part of the PR review, your questions that asked for elaborating were :gold:. Keep'em coming 😃 🙇‍♂️

Issue #71, opened 2018-07-18
Problem with resolution of babel/runtime.


diegoreymendez:
I'm getting the following error pretty often whenever I try to run the project.  This error was also reported by @SergioEstevao in our dev discussions:

<img width="526" alt="screen shot 2018-07-18 at 14 35 07" src="https://user-images.githubusercontent.com/1836005/42897899-e1343792-8a97-11e8-9dae-453cf2f46f23.png">

While it doesn't always come up, it comes up often enough that a solution to this problem is desirable.

*EDIT:* this is apparently [an external issue that will be solved shortly](https://github.com/WordPress/gutenberg/issues/7936).

PR #72, opened 2018-07-18
Solves a problem with the resolution of @babel/runtime

diegoreymendez:
### Description:

This PR solves #71.  It's not easy to reproduce the error show in #71, but I've tested this fix in a situation in which no other fix would make the error go away, and it's working consistently fine.

### Testing:

Just do a clean checkout of the project and run it for iOS to test.

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,6 +60,7 @@
     "lint:fix": "eslint $npm_package_config_jsfiles --fix"
   },
   "dependencies": {
+    "@babel/runtime": "^7.0.0-beta.54",
     "@wordpress/autop": "^1.0.6",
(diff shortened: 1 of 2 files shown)


diegoreymendez:
The weird thing on my end is that this was working perfectly fine in the morning, but I couldn't resolve the red screen in the afternoon by just cleaning the repo.  As to why that'd be the case, it's honestly beyond my knowledge at this time.

I'm fine with rolling this change back and keeping our eyes open for this issue to try and figure out what's going on, but if we prefer to get the issue out of the way entirely, this is what finally did the trick for me yesterday.

hypest:
> this is what finally did the trick for me yesterday

Since it made the situation better, I'd vote to keep it.

Issue #47, opened 2018-05-31
RichText: iOS implementation


koke:
Bring the iOS implementation to a similar point than the Android one.

Estimated: 5

diegoreymendez:
Some related progress here: https://github.com/wordpress-mobile/react-native-aztec/pull/33

hypest:
Work for this ticket has already been done, closing this.

PR #73, opened 2018-07-19
iOS: Initial Aztec-based RichText implementation.

diegoreymendez:
### Description:

This PR offers an initial Aztec-based RichtText implementation.  It integrates Aztec into `mobile-gutenberg` in a way that any developer can build and run the project.

There's a manual step in the integration (Carthage integration) that will be ironed out in an upcoming PRs.

You'll notice the Aztec view also has a yellow background, which is temporary and for testing purposes.

Moves forward #47.

### Known issues / pending work:

- Yellow background in Aztec view.
- While editing the view's heigh becomes desynchronized with the block cell space.
- Integrating the Carthage dependency requires manual steps that need to be automated.

### Before testing:

To download the Carthage Aztec dependency you need to first install [Carthage](https://github.com/Carthage/Carthage), and run the following commands from the project root directory:

```
cd react-native-aztec/ios
carthage bootstrap --platform iOS
```

### Testing:

- Make sure you've installed the Carthage dependency.
- Run the App and make sure the paragraph cell has a yellow background.
- Try typing inside of it.


hypest:
Tried with all the latest changes and can report it's working for me!

I also tried editing and while it works, it seems that it doesn't sync with the app state (Redux). I think that's OK for this PR to keep the PR small and we can work on it in a separate one.

I'd say: :shipit:

PR #76, opened 2018-07-20
Stops deleting yarn.lock on clean.

diegoreymendez:
This PR is a tiny change that will prevent yarn.lock from being deleted when doing `yarn clean` or modified when doing `yarn install`.

This is actually a positive change for development, to avoid having all PRs updating `yarn.lock` continuously, which could introduce seemingly-random and actually-unrelated issues.

It will allow us to separate updates to that file into a task of its own.

Running `yarn install` will still make sure you have the required dependencies for the branch you're in, so this should not affect our ability to stay up to date when changing branches.

### Testing:

- Do `yarn clean` and make sure the `yarn.lock` isn't deleted.
- Do `yarn install` and make sure the `yarn.lock` isn't modified.
- Run the app and make sure there are no outstanding errors.


hypest:
LGTM!

hypest:
Oops, apologies @diegoreymendez , I now noticed that I merged the PR while you had self-assigned it.

diegoreymendez:
No worries @hypest.  At least on my end I'm ok with a relaxed approach as long as we keep moving things forwards.

Thanks for taking care!

PR #80, opened 2018-07-26
Update to latest GB master

hypest:
Update to the currently latest Gutenberg master branch.

Changes:
* Using the @wordpress/blocks packages now, and directly from src
* For some reason, Flow doesn't like pointing to the root of the blocks
package (while it's not complaining for the same reason for the
`element` package). Having the flow config pointing inside the `src` dir
of the package now.
* Need to expose `window.Node` for a couple of its constants. Using
jsdom's definitions for that.
* Code block needs a css class defined (wp-block-code) in the innerHTML
for validation to succeed

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -85,6 +85,7 @@ munge_underscores=true
 
 module.name_mapper='^[./a-zA-Z0-9$_-]+\.\(bmp\|gif\|jpg\|jpeg\|png\|psd\|svg\|webp\|m4v\|mov\|mp4\|mpeg\|mpg\|webm\|aac\|aiff\|caf\|m4a\|mp3\|wav\|html\|pdf\)$' -> 'RelativeImageStub'
 module.name_mapper='@gutenberg' -> '<PROJECT_ROOT>/gutenberg'
+module.name_mapper='@wordpress/blocks' -> '<PROJECT_ROOT>/gutenberg/packages/blocks/src'
 module.name_mapper='@wordpress/data' -> '<PROJECT_ROOT>/gutenberg/packages/data'
 module.name_mapper='@wordpress/element' -> '<PROJECT_ROOT>/gutenberg/packages/element'
 module.name_mapper='@wordpress/deprecated' -> '<PROJECT_ROOT>/gutenberg/packages/deprecated'
--- a/package.json
+++ b/package.json
@@ -8,6 +8,7 @@
   },
   "devDependencies": {
     "@wordpress/babel-preset-default": "^1.1.2",
+    "@wordpress/block-serialization-spec-parser": "^1.0.0",
     "babel-eslint": "^8.2.2",
     "babel-plugin-react-native-classname-to-style": "^1.2.1",
     "babel-plugin-react-native-platform-specific-extensions": "^1.0.1",
@@ -61,6 +62,7 @@
   },
   "dependencies": {
     "@wordpress/autop": "^1.0.6",
+    "@wordpress/compose": "^1.0.1",
     "@wordpress/deprecated": "^1.0.0-alpha.2",
     "@wordpress/hooks": "^1.2.1",
     "@wordpress/i18n": "^1.1.0",
--- a/src/globals.js
+++ b/src/globals.js
@@ -2,6 +2,7 @@
 
 import { createElement } from '@wordpress/element';
 import jsdom from 'jsdom-jscore';
+import jsdomLevel1Core from 'jsdom-jscore/lib/jsdom/level1/core';
 
 global.wp = {
 	element: {
@@ -18,3 +19,7 @@ doc.implementation.createHTMLDocument = function( html ) {
 
 // `hpq` depends on `document` be available globally
 global.document = doc;
+
+if ( ! global.window.Node ) {
+	global.window.Node = jsdomLevel1Core.dom.level1.core.Node;
+}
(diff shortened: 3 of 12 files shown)


.flowconfig
```diff
@@ -85,6 +85,7 @@ munge_underscores=true
 
 module.name_mapper='^[./a-zA-Z0-9$_-]+\.\(bmp\|gif\|jpg\|jpeg\|png\|psd\|svg\|webp\|m4v\|mov\|mp4\|mpeg\|mpg\|webm\|aac\|aiff\|caf\|m4a\|mp3\|wav\|html\|pdf\)$' -> 'RelativeImageStub'
 module.name_mapper='@gutenberg' -> '<PROJECT_ROOT>/gutenberg'
+module.name_mapper='@wordpress/blocks' -> '<PROJECT_ROOT>/gutenberg/packages/blocks/src'
```
mzorz: I'm actually not sure we should name this `@wordpress` as blocks are actually `@gutenberg`'s. As a result of this observation we may also consider how other mappers are called, which I'm not sure of, but for Gutenberg blocks it sounds like it's definitely `@gutenberg`, wdyt?

Also, may I ask why do we set this here in Flow, and is not set in the module mapper plugin in `.babelrc` to indicate an alias? That's how I'd approached it, and would like to learn if that's a feasible path or not, and why this would be preferred, if the question first of all makes sense (i.e. can Flow be used for the same thing?)

package.json
```diff
@@ -8,6 +8,7 @@
   },
   "devDependencies": {
     "@wordpress/babel-preset-default": "^1.1.2",
+    "@wordpress/block-serialization-spec-parser": "^1.0.0",
```
mzorz: I couldn't realize this was needed. May I ask, even if it's a trivial question, how you come to know this needs to be set as a dependency?

package.json
```diff
@@ -61,6 +62,7 @@
   },
   "dependencies": {
     "@wordpress/autop": "^1.0.6",
+    "@wordpress/compose": "^1.0.1",
```
mzorz, line 65: why is this included as a dependency here, but then is [denylisted here](https://github.com/wordpress-mobile/gutenberg-mobile/pull/80/files#diff-41fdc10e1b5862877e4afa7e9176ee09R16)?

src/globals.js
```diff
@@ -18,3 +19,7 @@ doc.implementation.createHTMLDocument = function( html ) {
 
 // `hpq` depends on `document` be available globally
 global.document = doc;
+
+if ( ! global.window.Node ) {
```
mzorz: Ah! question, does defining this in this way fix the red screen I got saying `undefined is not an object (evaluating '_window$Node.TEXT_NODE')`?

.flowconfig
hypest: >I'm actually not sure we should name this @wordpress as blocks are actually @gutenberg's.

The package is actually called like that ([see its package definition here](https://github.com/WordPress/gutenberg/blob/master/packages/blocks/package.json#L2)) so, it makes sense to keep its original name.

> why do we set this here in Flow, and is not set in the module mapper plugin in .babelrc to indicate an alias

It's not an alias as far as the build pipeline is concerned. For Metro, `@wordpress/blocks` is a normal package and tries its best to resolve it. Metro being the smart tool it is, it actually _discovers_ the package automagically inside the `gutenberg/packages` folder.

We gradually have moved away from having to define aliases in Babel as the Gutenberg code has been npm-package-ized more and more. Previously, we had to "mock" those packages to mimic what Webpack is doing on the Gutenberg build side. But nowadays, the published Gutenberg packages can be easily picked up by Metro without alias tricks. Some packages (namely, the `editor` package` is still not published to NPM and that's why it's practically the only alias we have left in Babel.

But, `Flow` is not as smart as Metro. We need to give more direct instructions to Flow to go find the packages.

While on the subject, `Jest` (our testing tool) is also not as smart as Metro and that's why we have similar aliases in its `jest.config.js` file.

mzorz: Thanks for the detailed explanation about moving away from module resolver. I wonder why that wouldn't work then off the shelve for me (automagically as you say) when working on a similar PR in https://github.com/wordpress-mobile/gutenberg-mobile/tree/update/use-latest-gb-master

package.json
hypest: Ah, that crops up when you try to run the tests actually. If you try without the dep you'll notice Jest complaining. Haven't investigated why it happens only on the testing pipeline though.

package.json
hypest: OK, so, Metro being the smart tool it is, it actually tries to resolve the packages by looking in the local filesystem. So, when you _do_ want to use the remote package from NPM you need to denylist the local package dir.

src/globals.js
hypest: Yes! So, turns out that recent Gutenberg changes exposed the need for those constants from `window.Node` that were previously handled from `dom-react` and our `jsdom-core` dep. But yeah, with the changes we apparently need to manually add the support.

.flowconfig
mzorz: > While on the subject, Jest (our testing tool) is also not as smart as Metro and that's why we have similar aliases in its jest.config.js file

Thanks for pointing this out as well, brings useful insight

src/globals.js
mzorz: excellent, thanks for this clarification!

PR #81, opened 2018-07-26
Introduce paragraph block and RichText component for mobile

daniloercoli:
<del>This PR is branched on the top of #80 "Update to latest GB master", so don't merge this before.</del>

Pretty simple PR in this repository:
- Changed the demo app that now loads & parses valid Paragraph blocks from HTML code.
- Use an updated version of GB hash that added support for:
-- RichText component on mobile (still need to optimize it a little bit better)
-- Add a Paragraph block for mobile

Most of the work was done in a separate branch/PR in the Gutenberg repository here:
https://github.com/WordPress/gutenberg/tree/rnmobile/port-para-block-use-latest-gb-master-july

iOS: No errors, but para blocks are empty.

PR #82, opened 2018-07-26
iOS: Aztec fully wired (v1).

diegoreymendez:
This PR wraps up the initial wiring of Aztec to `gutenberg-mobile`.  It wires Aztec's content to our JavaScript state.

Closes #73.

### Details:

This PR also introduces some changes to how we store `eventCount` in our JS code.

According to https://facebook.github.io/react-native/docs/state:

```text
There are two types of data that control a component: props and state. props
are set by the parent and they are fixed throughout the lifetime of a
component. For data that is going to change, we have to use state.
```

Since `eventCount` was being changed throughout the lifetime of the Aztec TextView component, moving it to `state` made the most sense to adhere to the documentation cited above.

Storing it as a `prop` (that's what I understand we were doing before) was causing issues on iOS's side of the code.  When switching the editor to HTML mode and back, the `eventCount` was being persisted needlessly.  By removing it we were able to simplify the TextView refresh logic considerably.

### Testing:

- Make sure that you can type normally in the Aztec view.
- Switch to HTML and make sure the source code looks good.
- Switch back to visual and make sure the state wasn't lost.'

### Development tests:

✅ Tested in iOS
✅ Tested in Android (had some trouble initially to get the app running, but it was solved by resetting the cache).

### DIFF
--- a/src/block-management/block-holder.js
+++ b/src/block-management/block-holder.js
@@ -77,16 +77,18 @@ export default class BlockHolder extends React.Component<PropsType, StateType> {
 						styles[ 'aztec-editor' ],
 						{ minHeight: Math.max( _minHeight, this.state.aztecHeight ) },
 					] }
-					text={ { text: this.props.attributes.content, eventCount: this.props.attributes.eventCount } }
+					text={ { text: this.props.attributes.content, eventCount: this.state.eventCount } }
 					onContentSizeChange={ ( event ) => {
 						this.setState( { ...this.state, aztecHeight: event.nativeEvent.contentSize.height } );
 					} }
 					onChange={ ( event ) => {
 						console.log(event.nativeEvent.text),
+
+						this.setState( { ...this.state, eventCount: event.nativeEvent.eventCount } )
+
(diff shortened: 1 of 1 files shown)


diegoreymendez:
Merged the latest from master into this PR and resolved the conflicts.

daniloercoli:
Is this PR still relevant in the light of #81 being merged soon?

diegoreymendez:
@daniloercoli - Once #81 is merged I can update this one (I'd say let's try to do it in order to be fair with the earlier PRs).

But it includes changes to `block-holder.js` that I need merged.

hypest:
👋 @diegoreymendez , can you expand the PR description with some info on the `eventCount` related changes? I think I see 3 distinct changes (becoming optional, keeping it in the state vs the props and removing it from the call to `props.onChange()`. Thanks!

diegoreymendez:
@hypest - Done.

diegoreymendez:
PS: the first change is not making it optional (it already was).  What the first change does is actually "register" it as a `StateType` property.

PR #84, opened 2018-07-26
Fixes an issue in package.json.

### LINKED ISSUE #83: Issues with cross-env package when running 'yarn clean'.
### Description:

When running `yarn clean` I'm seeing this error:

```text
/bin/sh: cross-env: command not found
error Command failed with exit code 127.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

### Details:

It seems the `cross-env` package is properly defined as a dev dependency..

```bash
[diego@MonsterMac ~/Developer/gutenberg-mobile] cat package.json | grep "cross-env"
    "cross-env": "^5.1.4",
```

...properly resolved by yarn...

```bash
[diego@MonsterMac ~/Developer/gutenberg-mobile] cat yarn.lock | grep "cross-env"
cross-env@^5.1.4:
  resolved "https://registry.yarnpkg.com/cross-env/-/cross-env-5.2.0.tgz#6ecd4c015d5773e614039ee529076669b9d126f2"
```

...and properly installed in `node_modules`...

```bash
[diego@MonsterMac ~/Developer/gutenberg-mobile] cd node_modules/
[diego@MonsterMac ~/Developer/gutenberg-mobile/node_modules] ls -al | grep "cross-env"
drwxr-xr-x    8 diego  staff     256 Jul 26 12:57 cross-env
```

It would seem like the issue is that whatever alias should be created to be able to execute `cross-env` by just using its name is missing.

diegoreymendez:
### Description:

Fixes #83.

This PR resolves a problem that was causing unrelated commands to be executed when we were doing `yarn clean`.

This PR also reorganizes the script structure a bit to make it much easier to maintain in the future.

This PR also cleans the haste cache, as recommended by RN for some issues:

```
To resolve try the following:
  1. Clear watchman watches: `watchman watch-del-all`.
  2. Delete the `node_modules` folder: `rm -rf node_modules && npm install`.
  3. Reset Metro Bundler cache: `rm -rf /tmp/metro-bundler-cache-*` or `npm start -- --reset-cache`.  4. Remove haste cache: `rm -rf /tmp/haste-map-react-native-packager-*`.
```

### Details:

The culprit behind our issues was that at some point in time, our `yarn test` command was changed (probably broken too since it currently doesn't make much sense), so when running `yarn clean`, a `yarn test --clearCache` that was intended to be mapped to `jest --clearCache` wasn't properly resolved.

This PR doesn't fix our `yarn test` command, but instead maps the cleaning procedures directly to `jest`, in order to make our cleaning procedure much more resilient to the local scripts changing.

This PR also moves the `yarn clean:node` command to the last position, when doing `yarn clean`.  The reason is that some of the other cleaning commands use the binaries that exist within the `node_modules` directory (for example `yarn jest --clearCache`).

### Testing:

Run `yarn clean` and make sure that the error reported in #83 is gone.
Also make sure that the clean procedure steps are the expected ones (there shouldn't be unrelated test steps being run).

Run `yarn clean:install`, execute the App and make sure all is fine.


diegoreymendez:
@hypest, @daniloercoli - This is now ready for review.

mzorz:
Tested, works as instructed - LGTM :shipit:

PR #85, opened 2018-07-26
Add cleanup code for submodule react-native-aztec

mzorz:
As we go back and forth from testing the[ `AztecRN wrapper`](https://github.com/wordpress-mobile/react-native-aztec) and gutenberg-mobile, it's been tedious to know and make sure things are clean and can be run seamlessly.

Adding a step here to clean RN installed in `/example` and `./` folders of `react-native-aztec` subtree, and updating the subtree commit to latest master in `react-native-aztec`.

**Test steps:**

Run Aztec demo app first so RN gets installed locally there, and then run Gutenberg.
1. ```cd gutenberg-mobile/react-native-aztec/example```
2. ```yarn clean:install```
3. ```yarn start --reset-cache```
4. on another window: ```yarn android``` or ```yarn iOS```

Once you confirm the Aztec demo app works alright, proceed to try running the Gutenberg mobile demo app:

1. ```cd gutenberg-mobile/```
2. ```yarn clean:install```
3. ```yarn start --reset-cache```
4. on another window: ```yarn android``` or ```yarn iOS```

Make sure the GB mobile demo app works alright.

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,7 +55,7 @@
     "prettier": "prettier-eslint --write $npm_package_config_jsfiles $npm_package_config_scssfiles",
     "clean": "yarn clean:node; yarn clean:aztec; yarn clean:runtime;",
     "clean:runtime": "yarn cache clean; yarn test --clearCache; watchman watch-del-all; rm -rf $TMPDIR/react-*; rm -rf $TMPDIR/metro-cache-*; rm -rf $TMPDIR/jest_*",
-    "clean:aztec": "pushd react-native-aztec && (yarn clean; pushd example && yarn clean; popd; popd;)",
+    "clean:aztec": "pushd react-native-aztec && (yarn clean; pushd example && (yarn clean; popd); popd)",
     "clean:node": "rm -rf node_modules",
     "clean:install": "yarn clean; yarn",
     "lint": "eslint $npm_package_config_jsfiles",
--- a/package.json
+++ b/package.json
@@ -53,9 +53,10 @@
     "test:debug": "yarn pre-build && cross-env NODE_ENV=test node --inspect-brk node_modules/jest/bin/jest.js --runInBand --verbose --config jest.config.js",
     "flow": "flow",
     "prettier": "prettier-eslint --write $npm_package_config_jsfiles $npm_package_config_scssfiles",
-    "clean": "yarn clean:setup; yarn clean:runtime",
+    "clean": "yarn clean:node; yarn clean:aztec; yarn clean:runtime;",
     "clean:runtime": "yarn cache clean; yarn test --clearCache; watchman watch-del-all; rm -rf $TMPDIR/react-*; rm -rf $TMPDIR/metro-cache-*; rm -rf $TMPDIR/jest_*",
-    "clean:setup": "rm -rf node_modules",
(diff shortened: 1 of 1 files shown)


diegoreymendez:
@mzorz I'm not sure we should be cleaning a dependency manually.  My understanding is that the dependencies we consume take care of that on their own through `npm` / `yarn`.

While I have found scenarios myself in which it seems like there's trouble with the cache, I've recently started to realize that the main reasons behind red screens is badly resolved dependencies (this was the case for the `@babel/runtime` red screen and for the `@wordpress/blocks` red screen).

Is there a specific scenario that's solved by doing this cleanup manually?

Pinging @hypest for an opinion on this.

PS: There are two earlier PRs that update `react-nativez-aztec` to the latest from master, and that update the cleanup scripts.  If we decided to go forward with this modification let's try to merge them in order, to avoid other issues.

mzorz:
> @mzorz I'm not sure we should be cleaning a dependency manually. My understanding is that the dependencies we consume take care of that on their own through npm / yarn.

The problem here is that the example folders in the submodules have their own scripts in `package.json` that we have on purpose there in order to run the example apps. Following instructions to run example apps there makes yarn install RN and dependencies there. So installing RN there and then going up to `gutenberg-mobile` and running `yarn clean:install` makes for duplicates. May I suggest trying the test steps without this PR, there you'll see the errors that this PR fixes.

We've been having to do this manually each time we go back and forth from one project to the other, often not knowing what was clean and what not - I don't see a strong reason not to have this change merged.

> PS: There are two earlier PRs that update react-nativez-aztec to the latest from master, and that update the cleanup scripts. If we decided to go forward with this modification let's try to merge them in order, to avoid other issues.

Ok - we can wait for that 👍

diegoreymendez:
For reference the earlier PRs are #82 and #84.  Hoping we can get those in soon.

hypest:
I consider the `react-native-aztec` example app as a source of noise and possibly errors atm. We shouldn't use the submodule to run the example app at all. Besides, the whole point of the submodule is to stop using its example app and rather, test the Aztec wrapper code directly via the main app.

That said, I don't mind having this extra cleaning step to cater for the cases where someone does use the wrapper example app in the submodule.

package.json
```diff
@@ -53,9 +53,10 @@
     "test:debug": "yarn pre-build && cross-env NODE_ENV=test node --inspect-brk node_modules/jest/bin/jest.js --runInBand --verbose --config jest.config.js",
     "flow": "flow",
     "prettier": "prettier-eslint --write $npm_package_config_jsfiles $npm_package_config_scssfiles",
-    "clean": "yarn clean:setup; yarn clean:runtime",
+    "clean": "yarn clean:node; yarn clean:aztec; yarn clean:runtime;",
     "clean:runtime": "yarn cache clean; yarn test --clearCache; watchman watch-del-all; rm -rf $TMPDIR/react-*; rm -rf $TMPDIR/metro-cache-*; rm -rf $TMPDIR/jest_*",
-    "clean:setup": "rm -rf node_modules",
+    "clean:aztec": "pushd react-native-aztec && (yarn clean; pushd example && yarn clean; popd; popd;)",
```
diegoreymendez: This is mostly fine, but if you want to make it slightly safer, this would be:

`pushd react-native-aztec && (yarn clean; pushd example && (yarn clean; popd); popd)`

Issue #58, opened 2018-06-12
Inserter


koke:
Add an insert button that shows a list of registered blocks, and adds the selected block after the currently focused block.

![image](https://user-images.githubusercontent.com/6597771/43523855-14593f5a-9574-11e8-8269-f93e0a7a1caa.png)

Requirements for alpha:
- The inserter button will be in the toolbar as the design shows (depends on #55).
- The block picker should show the right icon for each block. (done in #129)
- After the block is inserted, switch focus to the newly inserted block.

## Tasks

Integration with toolbar (2 weeks)
- [x] add new `+` adder control placeholder (to be connected to the toolbar)
- [x] extract and define adder <-> toolbar interface
- [x] actually connect with toolbar

Other (1 week)

- [ ] UI pass (make it not only work but look as per the designs) -> #349 (Beta)

mzorz:
Adding rough design draft for v1 here:
![image](https://user-images.githubusercontent.com/6597771/43523855-14593f5a-9574-11e8-8269-f93e0a7a1caa.png)


This is how I plan to approach this:

- [ ] add basic support for model addition
- [ ] model addition: list item id generator / evaluate usages of IDs for list items in block manager
- [ ] make direct access inserter (insert anywhere, not just append to list)
- [ ] add new `+` adder control placeholder (to be connected to the toolbar)
- [ ] extract and define adder <-> toolbar interface
- [ ] add block types selection mechanism (obtain list of registered types, get icons for each, show list, implement bottom sheet, attach empty model type, maybe extend empty model reference object for each type as needed)
- [ ] UI pass (make it not only work but look as per the designs)

hypest:
Headsup @mzorz , I went ahead and struck out iteration No4 in https://github.com/wordpress-mobile/gutenberg-mobile/issues/58#issuecomment-409571383 as I think it should not block the feature completion and it can be performed at a future point in time.

koke:
I updated the description for this one. @mzorz do you think I missed anything in the alpha requirements? Can you take a look at estimations since it looks like you already have a good idea of what needs to be done?

mzorz:
#### Focus handling
While not directly related to the Inserter, it's of basic need that the focus state and UI representation gets handled right when inserting / moving blocks / switching focus.
- [x] check focus handling in redux (state management) works as expected
- [ ] Tapping into RichText doesn't handle focus right #135
- [ ] Inserting a new block doesn't make cursor immediately available on new block #113 
[...]

PR #88, opened 2018-08-01
Inserter - basic support for block addition

mzorz:
This PR has the minimum needed in order to make new blocks appear in the list.

1. tap on a non-editable bock to gain focus and make the item toolbar appear
2. tap on the `+` icon
3. scroll down the list and observe a new block has been added.
4. repeat as many times as you wish, and observe you can also delete blocks.

For now, we're producing only one type of block `paragraph` (even when the [function is parameterized](https://github.com/wordpress-mobile/gutenberg-mobile/compare/feature/inserter...try/inserter-take1#diff-cc5612fd38596495ffa481dcd29bb336R16)) and only _appending_ to the list (I have plans to make direct insertion). Also, we make use of the current item toolbar, as I plan to make small steps / PRs as we go to make small additions (like a standalone + that checks the currently focused item and appends a new block right below, and so on).

The list of steps [is available here](https://github.com/wordpress-mobile/gutenberg-mobile/issues/58#issuecomment-409571383) which I'll use as a master issue and link this and upcoming PRs there as well.

I plan to make all PRs against the `feature/inserter`, and merge that one to `master` once all done.

~**Note (and question at the same time) for the reviewers**: in order to make this work, I'm creating an empty block object in two places: [one for the list](https://github.com/wordpress-mobile/gutenberg-mobile/compare/feature/inserter...try/inserter-take1#diff-2d50daf729e068ecebcce113d1ce7f9aR88) as in the UI element, and another [one in the reducer](https://github.com/wordpress-mobile/gutenberg-mobile/compare/feature/inserter...try/inserter-take1#diff-3a96ec2ab60725b086c581badfe4e7cdR118) as in the Redux model. It bugs me that I'm creating 2 different objects, so I'm most probably not doing something right. While I'm exploring it, please don't hesitate to provide any insights as to whether this should be fine or not, and if not, feel free to provide any alternatives you can think of to make this cleaner.~

~**Known issue**: the item toolbar does not appear when the new items are focused (tapped on). Working on that for a bit, but I don't think is something I should spend much time on, as the toolbar will be handled differently anyway.~

Known issue: https://github.com/wordpress-mobile/gutenberg-mobile/pull/88#issuecomment-411477900


mzorz:
> Known issue: the item toolbar does not appear when the new items are focused (tapped on). Working on that for a bit, but I don't think is something I should spend much time on, as the toolbar will be handled differently anyway.

Actually looking closer I think this is quite tight related to the other observation (having 2 different objects is no good). Idea: I think we should rather create in one place (the redux store) and then request a re-render as the list has changed, but not change the in-memory list (it should change itself by following the cycle)

mzorz:
Fixed the duplicate object / focus problems by correctly implementing `createBlock` in 2beaebd

mzorz:
This PR has a problem keeping up with the `focused` state on newly created blocks, which is addressed in #95 (targeting this branch). That PR should be merged first 👍

src/block-management/block-manager.js
```diff
@@ -8,20 +8,18 @@ import { Platform, Switch, Text, View, FlatList } from 'react-native';
...
-import { getBlockType, serialize } from '@wordpress/blocks';
+import { getBlockType, serialize, createBlock } from '@wordpress/blocks';
 
 export type BlockListType = {
 	onChange: ( clientId: string, attributes: mixed ) => void,
 	focusBlockAction: string => mixed,
 	moveBlockUpAction: string => mixed,
 	moveBlockDownAction: string => mixed,
 	deleteBlockAction: string => mixed,
+	createBlockAction: ( string, BlockType ) => mixed,
```
hypest: Hmm, apparently we're passing an already instantiated block to the action so, I wonder if the naming ("createBlock...") is a bit off here. I'd kinda expect to pass some identifier and/or params in order for the reducer to instantiate the block. Not sure which way is the web side of Gutenberg's doing this.

src/store/actions/test.js
```diff
@@ -32,5 +35,13 @@ describe( 'Store', () => {
 			expect( action.type ).toEqual( ActionTypes.BLOCK.DELETE );
 			expect( action.clientId ).toEqual( '1' );
 		} );
+
+		it( 'should create an action to create a block', () => {
+			registerCoreBlocks();
```
hypest: This feels like a big thing to do while in a section of an otherwise simple test where we just test the fields of a newly created action object.

Can we perhaps mock the block to pass it to the action creation function?

On the other hand, judging from the convo we had over Slack, this will probably change when the block creation itself might move into the reducer so, perhaps we can leave this PR as is.

src/store/actions/test.js
```diff
@@ -32,5 +35,13 @@ describe( 'Store', () => {
...
+			const newBlock = createBlock( 'core/code', { content: 'new test text for a core/code block' } );
+			const action = actions.createBlockAction( '1', newBlock );
+			expect( action.type ).toEqual( ActionTypes.BLOCK.CREATE );
+		} );
```
hypest: Any particular reason for not testing the `clientId` value too as in the rest if this test case?

src/block-management/block-manager.js
mzorz: Rightly spotted @hypest  - as discussed on Slack, this is something I had plans for in a later PR, to take care of bringing this closer to GB 👍 
Trackerd in my item list here https://github.com/wordpress-mobile/gutenberg-mobile/issues/58#issuecomment-409571383

src/store/actions/test.js
mzorz: That's a good observation - also this test for now tests for creation action type only but I think should check for each of the registered blocks if that makes sense (once the function actually creates a block instead of just carrying a block to be added/inserted).  
Added to my item list https://github.com/wordpress-mobile/gutenberg-mobile/issues/58#issuecomment-409571383

src/store/actions/test.js
mzorz: addressed in 6ad3813 - and for completeness added check for the block in e556378

src/store/actions/test.js
hypest: I'm still not comfortable with `registerCodeBlocks` being inside the testcase since it's not the subject of the test itself. Let's move it to the `beforeAll` section so it's clear that it's part of the setup, not the test. WDYT?

mzorz: 👍

src/store/actions/test.js
```diff
@@ -2,6 +2,9 @@
 
 import * as actions from './';
 import ActionTypes from './ActionTypes';
+// Gutenberg imports
+import { getBlockType, serialize, createBlock } from '@wordpress/blocks';
```
hypest: `getBlockType` and `serialize` don't seem to be used anymore, can we remove them?

src/store/actions/test.js
mzorz: addressed in a04c5e9

src/store/actions/test.js
mzorz: nice catch! addressed in 54220f8

hypest: Thanks for the change @mzorz ! Apparently, the eslint configuration is actually faulty at the moment since this one should be caught during `yarn lint` (which is already part of the Travis tests).

I will produce a separate PR with the eslint fix (and also fix there a few other violations) so to not pollute this PR, cool?

mzorz: Thanks for explaining and describing where things are / need to be and follow up actions🙇 thank you! Also think the decision to make a separate PR for eslint fixes is in the right direction 👍

PR #89, opened 2018-08-02
Inserter - use GB createBlock to generate new blocks

mzorz:
While looking to replace and find a function in GB that would generate IDs, realized we should completely use GB's `createBlock` function instead of the mocked up `buildEmptyBlock` I was using.

This PR removes the use of the mocked up method and uses `import { createBlock } from '@wordpress/blocks';` to actually create a new block when the + icon is pressed on any of the items toolbar.


daniloercoli:
Looks good!

PR #96, opened 2018-08-08
Inserter - added direct insertion to block manager

mzorz:
This PR builds on top of #95, and adds direct insertion at a specific index in the list state. For the sake of the exercise in the GB mobile application for now we're inserting right below the tapped block.

Before, it would only _append_ new blocks to the list.

#### Android
insertertake1direct

#### iOS
insertertake1direct_ios


hypest:
With #95 closed, I think it makes sense to target this one against `feature/inserter` directly? WDYT @mzorz ?

mzorz:
Closing in favor of #110

Issue #57, opened 2018-06-12
Heading block


koke:
Implement a heading block with its toolbar for heading level selection

Estimated: 5

PR #98, opened 2018-08-09
Implement heading block

koke:
This ports the heading block (without toolbar or settings) to gutenberg-mobile.
The bulk of the code is in TBD, this PR just adds a heading block to the initial list.

See #57


gziolo:
We no longer use `uid` on the web - it is named `clientId` to better explain its intent.

Issue #99, opened 2018-08-09
Editable HTML view master issue.


etoledom:
This issue is meant to track progress on the `Editable HTML view` feature:

- [x] Make html view editable. #94 
- [x] Parse existing Gutenberg blocks from html to blocks. #100 
- Fix html `InputView` UX on Android. (#118)
  - It is shown scrolled to the bottom.
  - On focus, it will always scroll to the bottom.
  - When is empty, the cursor is at the middle of the screen.

etoledom:
Moved the InputView UX issues to its own issue #118

etoledom:
It's ready!
ref: #119

PR #100, opened 2018-08-10
Parse html into known blocks

etoledom:
Part of #99

This PR adds the ability of parsing the html string in the html view back into blocks.

For simplicity, the current implementation will just parse the known blocks, and anything that is outside of those blocks will be ignored.

An example on iOS:

parsing_ios

On Android, this currently won't work, but this issue was taken care of by #95. I tested a fast merge and it works:

parse_html_android

----

There are many details I wasn't sure about, I chose the simplest solutions for all of them, but I hope this PR is a good place to discuss them. The most important ones are:

* The only way to modify blocks is with a redux action.

The current reducer seems to me that is meant for actions over a single block, but the `parse` action will regenerate the whole list of blocks.

I was tempted to create another reducer specifically for parser, but it immediately made this implementation (and PR) grow on complexity, so I opted to add an extra action to the current reducer.

* HTML InputView content as part of the component state.

An interesting question I had was where to keep the state of the text that the user is editing.
My first test was to store it as part of the component state (as the [example](https://facebook.github.io/react-native/docs/textinput.html) shows), but for each time the state changes, a new instance of the view is created. This makes the whole UX very bad when editing on a scrolled position.

I also though about the redux store, but I wasn't sure about it and it was more complex than the next option.

I opted to simply store the edited text in a variable.


src/store/reducers/index.js
```diff
@@ -108,6 +108,10 @@ export const reducer = (
 			blocks.splice( index, 1 );
 			return { blocks: blocks, refresh: ! state.refresh };
 		}
+		case ActionTypes.BLOCK.PARSE: {
+			const parsed = action.parse(action.payload)
+			return { blocks: parsed, refresh: state.refresh };
```
gziolo: Why not import `parse` in this file instead of passing it down every time you want to use it?

gziolo: The whole idea of using Redux store is to take advantage of immutability. when you update reference to the `blocks`, you can treat it as an indicator if something has changed. This makes usage of `refresh` boolean flag obsolete as it tracks the state which can be derived from `blocks`.

src/store/actions/index.js
```diff
@@ -37,3 +37,10 @@ export const deleteBlockAction: BlockActionType = uid => ( {
 	type: ActionTypes.BLOCK.DELETE,
 	uid: uid,
 } );
+
+export const parseBlocksAction: BlockActionType = (payload, parser) => ( {
+	type: ActionTypes.BLOCK.PARSE,
+	uid: '',
```
gziolo: `uid` can be removed, it is never used. When parsing blocks from HTML you can get multiple blocks and as far as I see, it replaces all content.

src/block-management/block-manager.js
```diff
@@ -178,8 +207,9 @@ export default class BlockManager extends React.Component<PropsType, StateType>
 				<TextInput
 					multiline
 					numberOfLines={ 0 }
-					style={ styles.htmlView }>
-					{ this.serializeToHtml() }
+					style={ styles.htmlView }
+					onChangeText={ ( html ) => this.props.html = html }>
```
gziolo: You should never mutate `props`, you should track changes to `html` in `state`. In this case, you rather should pass down callback to this component which will update `html` outside.

src/block-management/block-manager.js
```diff
@@ -44,6 +51,10 @@ export default class BlockManager extends React.Component<PropsType, StateType>
 		};
 	}
 
+	componentDidMount() {
+		this.serializeToHtml();
```
gziolo: This function returns string, so it should be assigned to a variable.

src/store/reducers/index.js
etoledom: The idea was to avoid a hard coded dependency.
But for simplicity I'd agree that it would be better to import it directly. 👍

src/block-management/block-manager.js
etoledom: This was an experiment and I forgot to remove it 😞 
Thanks for catching it!

src/block-management/block-manager.js
etoledom: My original idea was to keep track of the `html` updates in a simple property `this.html`, but I kept getting a `flow` warning about it not being part of the `BlockManager` type.

I tried to add it as part of the component `state` but on every update, the InputView is reloaded and it feels weird from a UX point of view (but it works).

What would be the advantage of lifting the `html` state up vs keeping it in the component's state?

src/store/actions/index.js
etoledom: If I remove `uid`, `flow` gets angry saying that `'uid' is missing in object literal`, since `BlockActionType` expects a `uid` property.
That's one reason why I thought that a new reducer operating in the same store might be a good idea.

Instead, I'm inclined to define a `ParseActionType` to keep it simple, remove `uid` and keep `flow` happy.

src/store/reducers/index.js
etoledom: So, we should remove `refresh` from the store, right?
In that case I'd prefer to do that on a separate PR.

@hypest or @mzorz might know why it was necessary to add it?

src/block-management/block-manager.js
etoledom: I also tried what is described [here (lifting state up)](https://reactjs.org/docs/lifting-state-up.html#lifting-state-up), but the jumpiness remains.

mzorz:
Tests are failing but they're passing for me locally when running `yarn test` 🤔

hypest:
> Tests are failing but they're passing for me locally when running yarn test 🤔
FWIW, the tests fail for me locally with the same errors seen on Travis. cc @mzorz

etoledom:
I got the same tests error locally when I updated the branch from master, it was solved when I run `git submodule update`.

src/store/reducers/test.js
```diff
@@ -4,6 +4,9 @@
 
 import { reducer } from './';
 import * as actions from '../actions/';
+import { registerCoreBlocks } from '@gutenberg/core-blocks';
+
+registerCoreBlocks()
```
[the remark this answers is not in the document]
etoledom: I was following the example from `block-parser-code.test.js` and the others, but if it makes more sense to put them in the test case itself, no problem!

What it feels weird to me is that we are testing parsing two times, one from the `parse` function directly and other from the redux action indirectly. Maybe we can find a way to do it just once.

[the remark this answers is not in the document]
mzorz: hi! - I just moved a similar case to `beforeAll()` as per https://github.com/wordpress-mobile/gutenberg-mobile/pull/88/files/e55637838c3f16fefc5ef8de028855ebeaac209f#r209576735  in commit https://github.com/wordpress-mobile/gutenberg-mobile/commit/a04c5e90a381dee1f1fcdffb9f34ca537494749a  - wdyt?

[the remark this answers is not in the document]
etoledom: Sounds good! I'll go for it. Thank you

etoledom:
@mzorz I pushed a couple more commits. I hope those handle your comments :)

About the tests, I think I merged wrongly the reference to the `gutenberg` project and that's why the tests keep failing. I'm looking into it

mzorz:
Very nice @etoledom ! 🙇  been testing it quite a bit today

> On Android, this currently won't work, but this issue was taken care of by #95. I tested a fast merge and it works:

Given #95 has been closed without merging, due to a better solution to our specific case having been found in #108, I think we still need to address this here. Knowing this might be a bit too much to ask for to do in this PR alone, I went ahead and tried something on my own in #116 (I just didn't feel suggesting something without making sure it would work first was good enough, and so I ended up with a PR 😄).

Hope that's alright! Please feel free to comment / indicate otherwise. Ccing @hypest

PR #103, opened 2018-08-10
Reinstate clean:runtime script so start:reset works

hypest:
Reinstates the `clean:runtime` script that runs all the cleaning procedures needed after package installations and before Metro starts. This also fixes the broken `start:reset` script.

So, ideally, if the dependencies list hasn't changed (no new deps or upgraded deps or removed deps) than one just needs to `yarn start:reset` to clean-start Metro.

To test:
1. Run `yarn start:reset` to clean-start Metro and notice that it start correctly


hypest:
Thanks for the review @SergioEstevao ! Will go ahead and merge.

PR #104, opened 2018-08-13
Fix paragraph splits on the Paragraph block

SergioEstevao:
This PR fixes a regression where inserting multiple lines on a paragraph block was crashing the example app.

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
     "pre-build": "yarn check-gb-unused-babelrc && yarn bypass-gb-babel",
     "test:pre-build": "yarn check-gb-unused-babelrc && yarn test:bypass-gb-babel",
     "start": "yarn pre-build && yarn react-native start",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -45,7 +45,7 @@
     "pre-build": "yarn check-gb-unused-babelrc && yarn bypass-gb-babel",
     "test:pre-build": "yarn check-gb-unused-babelrc && yarn test:bypass-gb-babel",
     "start": "yarn pre-build && yarn react-native start",
-    "start:reset": "yarn clean:runtime && yarn start --reset-cache",
+    "start:reset": "yarn clean:metro && yarn start --reset-cache",
```
koke: There's #103 which looks like it's also taking care of this. Can you maybe review that one and get it merged? Otherwise, I think there will be a conflict

SergioEstevao:
@koke thanks for the heads up! I just reviewed the other PR and reverted the changes on this PR.

PR #105, opened 2018-08-13
Inserter - show Picker with registered blocks to choose from

mzorz:
This completes a very draft, initial 1st iteration on implementing the inserter flow. With this, the flow goes roughly along the lines of the expected actions like this:

1. focusing on an item shows an item toolbar which for now currently includes the `+` button to insert a new item (this `+` trigger will be moved to the toolbar as per the designs https://github.com/wordpress-mobile/gutenberg-mobile/issues/58#issuecomment-409571383)
2. tapping on `+ ` makes a RN `Picker` show up
3. tap on the picker to choose the type of block you'd like to insert
4. a newly created block is inserted right below the block that was being focused on.

inserter-picker

Known issues:
- the [RN `Picker`](https://facebook.github.io/react-native/docs/picker) component, though easy to use, as far as I could check only reacts when an item other than the currently selected value is chosen, and has had variable behavior in the recent past:
https://github.com/facebook/react-native/issues/15556
https://github.com/facebook/react-native/issues/15672
https://github.com/facebook/react-native/issues/13204
https://stackoverflow.com/questions/48465288/detect-selecting-current-selectedvalue-with-react-native-picker
Kind of makes sense with the prop name `onValueChange`, (i.e. I think from the name itself it derive that if the value didn't change, it won't be triggered) although it's not exactly what can be expected of the control (even the docs describe the prop function as `Callback for when an item is selected.`)

For this I'm working on another implementation with a modal dialog and a flatlist, but wanted to continue to put this PR up and complete a first "working" flow at least.

### DIFF
--- a/src/block-management/block-manager.js
+++ b/src/block-management/block-manager.js
@@ -4,14 +4,14 @@
  */
 
 import React from 'react';
-import { Platform, Switch, Text, View, FlatList } from 'react-native';
+import { Platform, Switch, Text, View, FlatList, Picker } from 'react-native';
 import RecyclerViewList, { DataSource } from 'react-native-recyclerview-list';
 import BlockHolder from './block-holder';
 import { ToolbarButton } from './constants';
 import type { BlockType } from '../store/';
 import styles from './block-manager.scss';
 // Gutenberg imports
-import { getBlockType, serialize, createBlock } from '@wordpress/blocks';
+import { getBlockType, getBlockTypes, serialize, createBlock } from '@wordpress/blocks';
 
 export type BlockListType = {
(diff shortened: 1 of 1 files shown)


koke:
The UX makes sense on Android but it's quite odd on iOS: it's showing a picker and inserting the block when the value changes.

inserter-ios

This is unusual on iOS where you'd require an extra step to validate the selection. The equivalent to the android UI would be a `UIAlertController` with the `actionSheet` style.

This might be good enough for a temporary design but I wanted to note it's a very odd experience on iOS.

mzorz:
Tracked issue in #113  - closing this PR in favor of #112

PR #106, opened 2018-08-13
Unsupported Block

### LINKED ISSUE #50: Unsupported: Implement unsupported block code/UI (render, serialize back)
Create a new “unsupported” block. If it’s easy, let’s try to get close to Thomas’ design for it, otherwise a simple “unsupported” label could do the trick.

Estimated: 2

diegoreymendez:
### Description:

This PR implements the unsupported block for blocks we don't offer support for.

Fixes #50.

### Details:

This PR introduces a new block type to display all the blocks we don't support.  Adding support for a block happens through regular registration of the block type in Gutenberg.

Next steps: introduce this new block type as the native version of the `freeform` block.

### Known issues / limitations:

After the changes submitted by @koke in 
https://github.com/wordpress-mobile/gutenberg-mobile/pull/127, I could no longer retrieve the original block name to display in visual mode.

This should be fine though, as unsupported blocks could be changing once this is merged: https://github.com/WordPress/gutenberg/pull/8274

### Testing:

- Load the standard post content.
- Switch to HTML mode and make sure the block contents are properly maintained.


etoledom:
I tested an alternative strategy to show the unsupported block UI, based on the conversation we had in slack. It's [described here](https://github.com/wordpress-mobile/gutenberg-mobile/issues/109).

src/block-types/unsupported-block/edit.js
```diff
@@ -0,0 +1,45 @@
...
+import { getBlockType } from '@wordpress/blocks';
+import { getUnknownTypeHandlerName } from '@wordpress/blocks';
+
+export default class UnsupportedBlockEdit extends React.Component<PropsType, StateType> {
+	constructor( props: PropsType ) {
+		super( props );
+	}
+
+	render() {
+		let blockName = this.props.name.charAt(0).toUpperCase() + this.props.name.slice(1);
```
koke: Wouldn’t this always be “Gmobile/unsupported” now?

diegoreymendez: Yeppers.  I can just remove it if it feels redundant, although I left it in cos otherwise the block feels a bit empty at this time.

Let me know which way you prefer it.

koke: Since the other line already says unsupported, I think we can remove it for now

koke:
I've fixed the conflicts with master and the tests to get this ready for today's demo. It also turned out that with the latest changes, the block manager didn't need any special case handling for the unsupported block, so that was simplified as well when resolving the conflict: 1a2bfc0f31f447b8976a731998142248aeee2d69

Merging when green

PR #110, opened 2018-08-14
Inserter - added direct insertion to block manager (take2)

mzorz:
_Closing #96 in favor of this one, as the base branch of #95 is not needed anymore, and avoid messing with commit history_

Adds direct insertion at a specific index in the list state. For the sake of the exercise in the GB mobile application for now we're inserting right below the tapped block.

Before, it would only _append_ new blocks to the list.

#### Android
insertertake1direct

#### iOS
insertertake1direct_ios

### DIFF
--- a/src/block-management/block-manager.js
+++ b/src/block-management/block-manager.js
@@ -19,7 +19,7 @@ export type BlockListType = {
 	moveBlockUpAction: string => mixed,
 	moveBlockDownAction: string => mixed,
 	deleteBlockAction: string => mixed,
-	createBlockAction: ( string, BlockType ) => mixed,
+	createBlockAction: ( string, BlockType, string ) => mixed,
 	blocks: Array<BlockType>,
 	aztechtml: string,
 	refresh: boolean,
@@ -72,16 +72,11 @@ export default class BlockManager extends React.Component<PropsType, StateType> 
 				this.props.deleteBlockAction( clientId );
 				break;
 			case ToolbarButton.PLUS:
-				// TODO: direct access insertion: it would be nice to pass the dataSourceBlockIndex here,
-				// so in this way we know the new block should be inserted right after this one
-				// instead of being appended to the end.
-				// this.props.createBlockAction( uid, dataSourceBlockIndex );
-
 				// TODO: block type picker here instead of hardcoding a core/code block
 				const newBlock = createBlock( 'core/paragraph', { content: 'new test text for a core/paragraph block' } );
 				const newBlockWithFocusedState = { ...newBlock, focused: false };
-				this.state.dataSource.push( newBlockWithFocusedState );
-				this.props.createBlockAction( newBlockWithFocusedState.clientId, newBlockWithFocusedState );
+				this.state.dataSource.splice( dataSourceBlockIndex + 1, 0, newBlockWithFocusedState );
+				this.props.createBlockAction( newBlockWithFocusedState.clientId, newBlockWithFocusedState, clientId );
 				break;
 			case ToolbarButton.SETTINGS:
 				// TODO: implement settings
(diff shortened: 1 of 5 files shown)


src/block-management/block-manager.js
```diff
@@ -72,16 +72,11 @@ export default class BlockManager extends React.Component<PropsType, StateType>
...
 				break;
 			case ToolbarButton.PLUS:
-				// TODO: direct access insertion: it would be nice to pass the dataSourceBlockIndex here,
-				// so in this way we know the new block should be inserted right after this one
-				// instead of being appended to the end.
-				// this.props.createBlockAction( uid, dataSourceBlockIndex );
-
 				// TODO: block type picker here instead of hardcoding a core/code block
 				const newBlock = createBlock( 'core/paragraph', { content: 'new test text for a core/paragraph block' } );
 				const newBlockWithFocusedState = { ...newBlock, focused: false };
```
hypest: I know it's not part of the PR's diff but, wdyt about doing a small optimization here by just appending to the object a new field instead of creating a full new object?

We can just do:
```
newBlock.focused = false;
```
and use the `newBlock` instead of construct the `newBlockWithFocusedState`.

src/store/actions/test.js
```diff
@@ -42,7 +42,7 @@ describe( 'Store', () => {
 
 		it( 'should create an action to create a block', () => {
 			const newBlock = createBlock( 'core/code', { content: 'new test text for a core/code block' } );
-			const action = actions.createBlockAction( '1', newBlock );
+			const action = actions.createBlockAction( '1', newBlock, '0' );
```
hypest: Let's also test the `clientIdAbove` while at it, wdyt?

src/store/reducers/index.js
```diff
@@ -112,6 +127,7 @@ export const reducer = (
 			// TODO we need to set focused: true and search for the currently focused block and
 			// set that one to `focused: false`.
 			blocks.push(action.block);
+			insertBlock(blocks, action.block, action.clientIdAbove);
```
hypest: Hmm, should this `insertBlock` replace the `blocks.push` in the previous line?

src/store/reducers/index.js
```diff
@@ -21,6 +21,21 @@ function findBlockIndex( blocks, clientId: string ) {
...
 }
 
+/*
+ * insert block into blocks[], below / after block having clientIdAbove
+*/
+function insertBlock( blocks, block, clientIdAbove ) {
+	// TODO we need to set focused: true and search for the currently focused block and
+	// set that one to `focused: false`.
+	const insertionIndex = findBlockIndex( blocks, clientIdAbove );
+	if (insertionIndex === blocks.length - 1) {
```
hypest: Do we really need the check? If it's just for guarding against the end of the array, I think `splice` takes care of that already so we don't need to. [See here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/splice).

hypest:
Nice one @mzorz ! I left a few comments, have a look and let me know if something is not clear! Thanks!

src/block-management/block-manager.js
mzorz: good call! decided to address that in 39f6500 in #112 given this part of the code gets moved around in there 👍

src/store/actions/test.js
mzorz: addressed in 8814dc0

src/store/reducers/index.js
mzorz: ooops thank Lord you catched that, addressed c1b51b9

mzorz:
flow has given its OK now with the new `CreateActionType` @hypest - ready for a third round :)

PR #111, opened 2018-08-14
Inserter - have the RecyclerView update and re-draw item in onChange

mzorz:
After the change introduced in https://github.com/wordpress-mobile/gutenberg-mobile/pull/108, the first time the RecyclerView renders its items these will all have the same height, rendering some of the blocks "cut".

**BEFORE**
screen shot 2018-08-14 at 11 21 29

**AFTER**
screen shot 2018-08-14 at 11 22 09

This PR addresses this by making sure to set these items dirty, [signaling the RecyclerView to re-draw them](https://github.com/wordpress-mobile/react-native-recyclerview-list/blob/master/src/RecyclerViewList.js#L106) as `onChange` is received.

### DIFF
--- a/src/block-management/block-manager.js
+++ b/src/block-management/block-manager.js
@@ -117,6 +117,7 @@ export default class BlockManager extends React.Component<PropsType, StateType> 
 		const dataSource = this.state.dataSource;
 		const block = dataSource.get( index );
 		block.attributes = attributes;
+		dataSource.set( index, block );
 		// Update Redux store
 		this.props.onChange( clientId, attributes );
 	}


hypest:
The `flow` error in this PR seems to be present in `feature/inserter` anyway so, to keep this PR the small one it is, let's fix the error in one of the follow up Inserter PRs @mzorz , wdyt?

mzorz:
> The flow error in this PR seems to be present in feature/inserter anyway so, to keep this PR the small one it is, let's fix the error in one of the follow up Inserter PRs @mzorz , wdyt?

Agreed! Will try and address shortly 🙇

PR #112, opened 2018-08-14
Inserter - show Picker with registered blocks to choose from

mzorz:
_Supersedes already reviewed/approved #105 to clean commit history_

This completes a very draft, initial 1st iteration on implementing the inserter flow. With this, the flow goes roughly along the lines of the expected actions like this:

1. focusing on an item shows an item toolbar which for now currently includes the `+` button to insert a new item (this `+` trigger will be moved to the toolbar as per the designs https://github.com/wordpress-mobile/gutenberg-mobile/issues/58#issuecomment-409571383)
2. tapping on `+ ` makes a RN `Picker` show up
3. tap on the picker to choose the type of block you'd like to insert
4. a newly created block is inserted right below the block that was being focused on.

inserter-picker

Known issues:
- the [RN `Picker`](https://facebook.github.io/react-native/docs/picker) component, though easy to use, as far as I could check only reacts when an item other than the currently selected value is chosen, and has had variable behavior in the recent past:
https://github.com/facebook/react-native/issues/15556
https://github.com/facebook/react-native/issues/15672
https://github.com/facebook/react-native/issues/13204
https://stackoverflow.com/questions/48465288/detect-selecting-current-selectedvalue-with-react-native-picker
Kind of makes sense with the prop name `onValueChange`, (i.e. I think from the name itself it derive that if the value didn't change, it won't be triggered) although it's not exactly what can be expected of the control (even the docs describe the prop function as `Callback for when an item is selected.`)

For this I'm working on another implementation with a modal dialog and a flatlist, but wanted to continue to put this PR up and complete a first "working" flow at least.

### DIFF
--- a/src/block-management/block-manager.js
+++ b/src/block-management/block-manager.js
@@ -86,14 +86,14 @@ export default class BlockManager extends React.Component<PropsType, StateType> 
 
 		// create an empty block of the selected type
 		const newBlock = createBlock( itemValue, { content: 'new test text for a ' + itemValue + ' block' } );
-		const newBlockWithFocusedState = { ...newBlock, focused: false };
+		newBlock.focused = false;
 
 		// set it into the datasource, and use the same object instance to send it to props/redux
-		this.state.dataSource.splice( focusedItemIndex + 1, 0, newBlockWithFocusedState );
-		this.props.createBlockAction( newBlockWithFocusedState.clientId, newBlockWithFocusedState, clientIdFocused );
+		this.state.dataSource.splice( focusedItemIndex + 1, 0, newBlock );
+		this.props.createBlockAction( newBlock.clientId, newBlock, clientIdFocused );
 
 		// now set the focus
-		this.props.focusBlockAction( newBlockWithFocusedState.clientId );
+		this.props.focusBlockAction( newBlock.clientId );
 	}
 
 	onToolbarButtonPressed( button: number, clientId: string ) {
--- a/src/block-management/block-manager.js
+++ b/src/block-management/block-manager.js
@@ -4,14 +4,14 @@
  */
 
 import React from 'react';
-import { Platform, Switch, Text, View, FlatList } from 'react-native';
+import { Platform, Switch, Text, View, FlatList, Picker } from 'react-native';
 import RecyclerViewList, { DataSource } from 'react-native-recyclerview-list';
 import BlockHolder from './block-holder';
 import { ToolbarButton } from './constants';
 import type { BlockType } from '../store/';
 import styles from './block-manager.scss';
 // Gutenberg imports
-import { getBlockType, serialize, createBlock } from '@wordpress/blocks';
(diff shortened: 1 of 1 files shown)


mzorz:
@koke This code has already been reviewed / approved in #105  except for this particular line here https://github.com/wordpress-mobile/gutenberg-mobile/pull/112/files#diff-2d50daf729e068ecebcce113d1ce7f9aR92  which has been added after the fix in #108. Would you like to give it a quick look and re-approve, or otherwise make any comments as you see fit? Thanks in advance 🙇

src/block-management/block-manager.js
```diff
@@ -56,6 +61,41 @@ export default class BlockManager extends React.Component<PropsType, StateType>
...
+		// find currently focused block
+		const focusedItemIndex = this.findDataSourceIndexForFocusedItem();
+		const clientIdFocused = this.state.dataSource.get( focusedItemIndex ).clientId;
+
+		// create an empty block of the selected type
+		const newBlock = createBlock( itemValue, { content: 'new test text for a ' + itemValue + ' block' } );
+		newBlock.focused = false;
+
+		// set it into the datasource, and use the same object instance to send it to props/redux
+		this.state.dataSource.splice( focusedItemIndex + 1, 0, newBlock );
```
koke: Aren't you supposed to modify `state` only by calling `setState`? I still have a lot to learn about React and I'm not sure which side effects this might have.

mzorz: I know, right? 😉  That's exactly my understanding from my learnup that ended in the elaborate description of #95. But, one of the troublesome things that appear here is that the list will redraw itself on each state change as per React's mandate, which is something we still have trouble digesting (feels super counterintuitive after having learnt to deal with recycled views and making good use of list updates for years).  Also this kills the native animations when moving blocks, which was something that was tested early on when picking the List component, as ultimately we want to have a best UX on Gutenberg when moving blocks around. 

Long story short, we are keeping the same state on both sides (RecyclerView and redux) to make sure we can still use the fine grained RecyclerView API to move items around and get native animations to work, without incurring in heavy lifting. More info in #108.

koke: Got it. I'm going to go with 👍 for now since I don't want to be blocking on this, but it seems to me as if the recyclerlist library wasn't implemented in a way that's consistent with React's declarative style. If I understood the code correctly, the list isn't implementing `shouldComponentUpdate`, and the README seems to imply that you should update the contents by manipulating the data source directly.

What I would expect is that you'd call setState and the list would deal with calculating what changed from the previous one and what needs to be inserted/deleted/updated. But this is what we have for now 🤷‍♂️

mzorz: On the same boat here 😄.  We could set some time apart to further investigate and see if using a `FlatList` and non-native animations makes the cut for both platforms and eventually go with it, making good use of the React pattern 👍

PR #115, opened 2018-08-14
Fix jumpy html text input on iOS

etoledom:
This PR fixes an UX issue with the html text input on iOS, where the text scrolls up each time the text change.
jumpy_html_ios

This happens because, with the usual `setState()` setup, RN recreates the element when the state changes.

To fix it I decided to let the component itself manage its state, and later on retrieve the text from its reference, similar to how is usually done on native iOS. It would be great if there's a better way.

**To test:**
- Run the project on iOS.
- Switch to html editor.
- Edit the text.
- Check that the cursor won't jump to the top.
- Check that the html changes are taken into account when switching back to blocks as described in #100 (working just with registered Gutenberg blocks for now)


mzorz:
FYI restarted a failed Travis job and it's green now 👍

PR #116, opened 2018-08-14
HTML parser - restart datasource after full parse of html

mzorz:
This PR performs a `dataSource` re-init (by recreating the object in the `state`) when a full `parse` has taken place. While we've avoided re-creating the list's `dataSource` in other cases, I think it makes sense to do so after an html-to-blocks conversion has taken place. 

This makes sure the state of `dataSource`, and `props` are aligned, and fixes the detail described in #100 here:

> On Android, this currently won't work, but this issue was taken care of by #95. [...]

To test (on Android):
1. start the demo app
2. edit one of the known blocks such as paragraph or heading
3. wait a bit so Aztec makes the change signaling happen (at least 500 miliseconds ;) ) _this is something we'll probably need to take a look into at some point_
3. switch to HTML view
4. observe the html contents show the entered text in the edited blocks
5. edit such blocks again (without breaking the delimiters)
6. switch back from html
7. observe the new text is in place, in the visual mode.

Idea to make this better at some point in the future: only pass the `fullparse: true` value when no edits have been in place (maybe at least check the length?)

### DIFF
--- a/src/block-management/block-manager.js
+++ b/src/block-management/block-manager.js
@@ -61,6 +61,17 @@ export default class BlockManager extends React.Component<PropsType, StateType> 
 		return -1;
 	}
 
+	static getDerivedStateFromProps( props: PropsType, state: StateType ) {
+		if ( props.fullparse === true ) {
+			return {
+				...state,
+				dataSource: new DataSource( props.blocks, ( item: BlockType ) => item.clientId ),
+			};
+		}
+		// no state change necessary
+		return null;
+	}
+
 	onToolbarButtonPressed( button: number, clientId: string ) {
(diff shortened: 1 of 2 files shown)


mzorz:
> My only concern is adding more flags to the redux store. I don't really know if it's a common pattern or not but we could look into it in the future too.

TBH I don't know either 😄 I think it's a valid concern and I wonder that too 🤔 -  if you agree we can merge this one and if @hypest (or anyone else for the case) thinks there's a better solution we can always revert this one, as it's just one commit (586526b75a09a5ca74fc86710b29abc2d15b22b2)

mzorz:
Merging, as agreed over DM, with possibility to revert if necessary :shipit:

PR #126, opened 2018-08-21
Add VSCode configuration for React Native extension

koke:
I've been using the [React Native extension](https://marketplace.visualstudio.com/items?itemName=vsmobile.vscode-react-native) for Visual Studio Code for a couple weeks and it's a really nice addition.

The integration with the debugger is not as solid as we might be used to with Xcode or Android Studio, but the improvements over running everything from a terminal, and being able to set breakpoints and debug from the IDE are worth it.

This also adds the React Native extension to the recommended list.


koke:
Oh good call, I'll add a note to the README 👍

koke:
Added a VScode section to the README, ready for another look

koke:
I'm merging this since it only has README changes, but please take a look and let me know if you have any feedback, and I'll amend the README in another PR

PR #129, opened 2018-08-24
Inserter - block icons

mzorz:
This PR makes the block icons visible in the `BlockPicker`:

screen shot 2018-08-23 at 21 48 42

To test:
1. update the branch
2. focus on a block and tap the `+` icon
3. observe the picker now shows the block icons

Relies on the following Gutenberg PR https://github.com/WordPress/gutenberg/pull/9294


gziolo:
It's a quite simplified version comparing to web :)
Next step would be to use [BlockIcon](https://github.com/WordPress/gutenberg/tree/master/packages/editor/src/components/block-icon) components instead which uses `div` thus requires some refactor in RN context.
Yet another step would be to reuse [BlockTypesList](https://github.com/WordPress/gutenberg/blob/master/packages/editor/src/components/block-types-list/index.js) which has even more DOM elements in use...

mzorz:
Thank you @SergioEstevao ! 🙇

PR #130, opened 2018-08-24
Update Aztec RN version

SergioEstevao:
This PR updates the version being used of the aztec-rn library.

Because of that it updates the following:
- build tools version for Android
- Update event interface for contentSizeChange
- Small improvements on display of list to avoid the keyboard.


daniloercoli:
:shipit:

PR #131, opened 2018-08-24
Nicer UI from the demo

koke:
I tweaked some of the styles to make things nicer for the demo and hid the block titles by default, adding a switch for that.

screenshot_1535013662
simulator screen shot - iphone 6 - 2018-08-23 at 10 51 13


koke:
Hmm, those screenshots are old. I played with adding a border to the blocks but I left it out in the end.

PR #133, opened 2018-09-03
Heading toolbar

SergioEstevao:
This PR adds support for the toolbar specific items to change the type of Header (h2, h3, h4).

screen shot 2018-09-03 at 12 44 37

### DIFF
--- a/src/block-management/block-manager.js
+++ b/src/block-management/block-manager.js
@@ -170,7 +170,7 @@ export default class BlockManager extends React.Component<PropsType, StateType> 
 		// Change the data source
 		const index = this.getDataSourceIndexFromClientId( clientId );
 		const dataSource = this.state.dataSource;
-		const block = { ...( dataSource.get( index ) ) };
+		const block = dataSource.get( index );
 		block.attributes = attributes;
 		dataSource.set( index, block );
 	}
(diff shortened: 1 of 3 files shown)


src/block-management/block-manager.js
```diff
@@ -164,14 +164,15 @@ export default class BlockManager extends React.Component<PropsType, StateType>
 	}
 
 	onChange( clientId: string, attributes: mixed ) {
-		// Update datasource UI
+		// Update Redux store
+		this.props.onChange( clientId, attributes );
```
SergioEstevao: I pulled the onChange/redux change to before  to make act and mark the refresh variable before we go and change the block ourselfs bellow.

src/block-management/block-manager.js
```diff
@@ -164,14 +164,15 @@ export default class BlockManager extends React.Component<PropsType, StateType>
...
+
+		// Change the data source
 		const index = this.getDataSourceIndexFromClientId( clientId );
 		const dataSource = this.state.dataSource;
-		const block = dataSource.get( index );
+		const block = { ...( dataSource.get( index ) ) };
```
SergioEstevao: Ah sorry, I was trying to do another change here to see if I could just do a copy of the block and avoid conflicts with the reducer. Let me change that back.

hypest:
Perhaps minor, I'm not seeing a toolbar button for the `h1` case as mentioned in the PR description @SergioEstevao. Should there be one?

SergioEstevao:
It looks it's not part of the options that Gutenberg uses, I will update the ticket description.

PR #137, opened 2018-09-10
Update React Native Aztec version.

SergioEstevao:
This PR updates the version of Aztec being used in the project.

This new version fixes an issue where changes of style/format where done on the content of a block and where not being propagated to the HTML.

How to test:
 - Start the demo app
 - Select text on a paragraph block
 - Tap on the Bold button on the style toolbar
 - Switch to HTML
 - Check the the `strong` tag was applied.

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -32,7 +32,7 @@ module.exports = {
 		'node',
 	],
 	moduleNameMapper: {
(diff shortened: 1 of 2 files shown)


diegoreymendez:
@SergioEstevao - Since I can't see the changes to `react-native-aztec` in this PR (and there's no PR for that apparently), could you submit a PR for those changes?

Thanks!

SergioEstevao:
There is no need of changes we are just updating the Aztec version referes. The changes are only a different SHA.

PR #139, opened 2018-09-11
Ability to run the app and tests from inside Gutenberg

hypest:
This PR introduces an addition set of RN and Jest run configurations to enable the app and tests to run when the code is a subfolder inside the main Gutenberg code filetree. This is necessary as a first step towards running the mobile testsuite via Gutenberg's test pipeline.

In particular, this PR:

1. Adds a test for the `heading` block to test the text prop passed to is to match the string literal defined in the initial app data
2. Adds `rn-cli-inside-gb.config.js` as the run configuration to launch the RN app and use the Gutenberg code in the parent directory. Adds the yarn scripts to use it as well.
3. Adds `jest_gb.config.js` as the Jest configuration of the tests, using the Gutenberg code in the parent directory. Adds the yarn scripts to use it as well.
4. All Gutenberg `packages/*` packages are added as root source folders in the "inside-gb" case. Apparently this makes the relevant `package.json` deps redundant, although it's awkward. I hope we can make this situation better.
5. There is some code duplication in the sass-transformer.js files and the jest.config.js files but, that can be fixed either in a later PR
6. We need to run `npm install` in the parent Gutenberg folder to compile the packages since the compiled versions are used by the mobile testsuite. I couldn't find a quick way to have that use Babel on the fly and use the `src` versions directly. Hope we can make this better too.
7. The Travis config doesn't include the "inside-gb" scenario yet

### To test A

Normal `yarn start`, `yarn android|ios`  and `yarn test` should work as normal and successfully.

### To test B

This is a different than usual setup and involves switching the roles of the repos (Gutenberg becomes the parent and the mobile repo becomes the child):
1. Clone the main Gutenberg repo in a new root folder
2. Change into the Gutenberg folder
3. `npm install` to install and compile packages
4. While still inside Gutenberg, clone or copy the whole gutenberg-mobile code into a subfolder and change into that one. Follow the normal setup instructions to complete the mobile code setup from its README.md file.
5. Run `yarn test:inside-gb` to run the testsuite and verify it's running green
6. Run `yarn start:inside-gb` to start Metro using the Gutenberg code from the parent directory
7. In a separate terminal window, run `yarn android` or `yarn ios` to run the native app and verify that it works fine

You may want to delete the nested `gutenberg` folder (the submodule) to verify that the code in use is the one in the parent folder but, this is not mandatory. The setup accounts for the nested folder and "blacklists" it.


koke:
I think this is a very interesting proof of concept, but I'm a bit uncomfortable about merging all the extra commands and files. I think I'd rather make the build scripts detect where gutenberg is (e.g. 85b3f2b711943b2a181121b19aa1a19a5d8a3428) or focus directly on integrating gutenberg-mobile into the Gutenberg repo

hypest:
Thanks for having a look @koke !

> I think I'd rather make the build scripts detect where gutenberg is or focus directly on integrating gutenberg-mobile into the Gutenberg repo

I consider this PR just a first step to build upon. Even the presence of the two parallel run configurations is a bit too much but, it's a temporary solution to avoid disrupting the work of others when/if this PR gets merged to master. I expect us to have only one run configuration, not multiple, when we finalize how to "merge" with Gutenberg. In short, I betting more on the "merging repos" idea.

In any case, can we defer optimizations to follow up PRs to help with focus?

package.json
```diff
@@ -41,20 +41,19 @@
...
-    "test:bypass-gb-babel": "echo '{\"presets\": [\"@wordpress/default\"]}' > gutenberg/.babelrc",
-    "pre-build": "yarn check-gb-unused-babelrc && yarn bypass-gb-babel",
-    "test:pre-build": "yarn check-gb-unused-babelrc && yarn test:bypass-gb-babel",
-    "start": "yarn pre-build && yarn react-native start",
+    "start": "yarn react-native start",
+    "start:inside-gb": "yarn react-native start --config `pwd`/rn-cli-inside-gb.config.js",
     "start:reset": "yarn clean:runtime && yarn start --reset-cache",
-    "start:debug": "yarn pre-build && node --inspect-brk node_modules/.bin/react-native start",
+    "start:inside-gb:reset": "yarn clean:runtime && yarn start:inside-gb --reset-cache",
+    "start:debug": "node --inspect-brk node_modules/.bin/react-native start",
```
SergioEstevao: This doens't look right, shouldn't this be kept has before? At the moment this command looks exactly the same has the one bellow.

PR #144, opened 2018-09-14
First steps for splitting para blocks on enter.key

daniloercoli:
This PR introduces a proposal on how to handle the enter.key event to split the content of a para block in 2 blocks. (does't matter if you're at the middle or the end of the para).

It adds a way to receive back the HTML content available in the editor with the cursor tag in it. Once the HTML+cursor tag is received JS side, we can write the logic that splits the para in 2 blocks in one place for both platforms.

In details:
1. The user taps on Enter. The event is detected by the native platform and a new event is emitted
2. RichText `OnEnter` method is called
3. RichText then calls a new command over the editor `returnHTMLWithCursor`
4. The native wrapper catches the request, get the content + cursor and emits a new event
5. RichText `onHTMLContentWithCursor` is finally called 

Further improvements will provide Para and heading with the proper logic on how to react to this new event.

Companion PRs are here:
- https://github.com/wordpress-mobile/react-native-aztec/pull/55
- https://github.com/WordPress/gutenberg/pull/9907

### DIFF
--- a/.travis/travis-checks-js.sh
+++ b/.travis/travis-checks-js.sh
@@ -18,12 +18,18 @@ if [ "$CHECK_CORRECTNESS" = true ] ; then
   npm run lint || pFail
 fi
 
+if [ "$GUTENBERG_AS_PARENT" = true ] ; then
+  TEST_SCRIPT_NAME='test:inside-gb'
+else
+  TEST_SCRIPT_NAME='test'
+fi
+
 if [ "$CHECK_TESTS" = true ] ; then
   # we'll run the tests twich (once for each platform) if the platform env var is not set
   if [[ -z "${TEST_RN_PLATFORM}" ]] ; then
-    TEST_RN_PLATFORM=android npm test || pFail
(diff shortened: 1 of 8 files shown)


daniloercoli:
With this logic in place, the JS side will receive an event that contains the `text` and the `cursorPosition` when the enter.key is pressed. 
It's up to the JS side to use those values and decide how to react/modify the text once received.

Note that if you press enter in the middle of a random paragraph, say in the middle of a bold word:
`<b>Hello-CURSOR HERE- There!</b> and other text follow`
the event does just return the info - without closing any tags, or anything else. In this particular case it will return `<p><b>Hello There!</b> and other text follow</p>` and `11`. The JS side should be able to close relevant tags and open some other.
Not sure if this is feasible or we want to rely on Aztec-native specific logic for that, and does return 2 para contents in the returning event. cc @SergioEstevao @diegoreymendez

daniloercoli:
GB hash was already updated in a previous/another PR. This PR updates ReactNativeAztec hash only.

Should I merge this since change to RNAztec was already reviewed?

PR #146, opened 2018-09-17
Update GB mobile and fix run-time red error screen with GB master

daniloercoli:
There was an update on GB that broke the native side.  Ref: https://github.com/WordPress/gutenberg/pull/9507

In this PR i've blacklisted `redux-routine` and made it working over NPM online.

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -32,7 +32,7 @@ module.exports = {
 		'node',
 	],
 	moduleNameMapper: {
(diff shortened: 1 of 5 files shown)


hypest:
Headsup: even though this PR seems to fix the issue (I presume it was the `Can't find variable: Symbol` error?) the issue will resurface when the latest `@wordpress/redux-routine` package gets published to NPM. It looks fixed atm because the NPM version is not using the `rungen` package which seems incompatible with the current Gutenberg mobile Android app.

PR #148, opened 2018-09-18
Fix Android html text input

etoledom:
This PR fixes the html text input that stoped responding to changes on Android.

**On iOS:**
If the html string is stored on the state and it changes, the whole TextInput is recreated, making jump on certain circumstances. (ref: #115)

**On Android:**
If the html string is **not** stored on the component's state, the state of the TextInput won't change even when the user tries. In other words, #115 broke Android's html TextInput.

This solution implements a different way to hold and get the html string for iOS and Android.

* I'm inclined to create an` HTMLTextInput` component that handles this problem internally, but I prefer to ask for suggestions before implementing it.

### DIFF
--- a/src/block-management/block-manager.js
+++ b/src/block-management/block-manager.js
@@ -221,7 +221,7 @@ export default class BlockManager extends React.Component<PropsType, StateType> 
 				} }
 				onValueSelected={ ( itemValue ) => {
 					this.onBlockTypeSelected( itemValue );
-				} } 
+				} }
 			/>
 		);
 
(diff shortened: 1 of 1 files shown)


hypest:
> I'm inclined to create anHTMLTextInput component that handles this problem internally, but I prefer to ask for suggestions before implementing it.

Yes, that would actually be useful, especially now that it seems the different mobile platforms demand different implementations.

You think you can work on that in this PR? If not, happy to have it in a separate PR.

etoledom:
Thank you @hypest !

I'll merge this to fix the issue on master.
Right after that I'll send a PR with that improvement.

Issue #118, opened 2018-08-15
HTML TextInput UX issues


etoledom:
This is a follow up of many UX issues regarding the HTML editor.

As an alternative of using `InputView`, we could try to use Aztec component on HTML edit mode. It might save us some troubles (or add more?)

**Both platforms:**
- [ ] When there is more than one screen of text, and the user scrolls the content, it will always start editing showing the keyboard.

**Android:**
- [ ] When it first appear, It is shown scrolled to the bottom.
- [ ] On focus, it will always scroll to the bottom and show the cursor at the end.
- [ ] When is empty, the cursor is at the middle of the screen. #114

**iOS**
- [ ] When there's more than one screen of text, adding text will automatically jump the cursor to the top of the screen. #115

daniloercoli:
> On focus, it will always scroll to the bottom and show the cursor at the end.

This is an ugly problem IMO, especially when you switch to HTML to be able to edit a long post (with maybe unsupported blocks). You scroll the view until you find the part of the text you want to modify, and then tap on the screen....It took a while to figure out where to put back the caret.

etoledom:
I like the idea of using the HTML Editor used on Aztec. Hopefully that will solve these problems 🙏

SergioEstevao:
@hypest should we try to fix the issues above or move to the solution where we use the native HTML editors?

hypest:
I'd favor using a native-based html editor (Aztec's can be an option), thinking that since html mode is a long string of rich-text content (think: syntax highlighted) it might not be very performant to rely on RN/JS for the editing UX.

PR #150, opened 2018-09-19
Adding HTMLTextInput component

etoledom:
This PR adds a new component `HTMLTextInput` to hide the different InputView implementation between iOS and Android.

It should continue working as before. That means that the remaining Android UX issues specified in #118 are not yet fixed, but this gives a better place to fix them in the future.

By creating a separated component, we can now use the component's `shouldComponentUpdate` method to give a better solution to the Android/iOS update issue .

To Test:
- Run the app on both iOS and Android.
- Switch to HTML Output.
- Edit the text.
- Check that on iOS it works correctly.
- Check that on Android it can be modified.
- Go back to Blocks view.
- Check that the changes on html are represented on the blocks view.

### DIFF
--- a/src/block-management/block-manager.js
+++ b/src/block-management/block-manager.js
@@ -11,6 +11,7 @@ import { ToolbarButton } from './constants';
 import type { BlockType } from '../store/';
 import styles from './block-manager.scss';
 import BlockPicker from './block-picker';
+import HTMLTextInput from '../components/html-text-input'
 
 // Gutenberg imports
 import {
@@ -29,6 +30,8 @@ export type BlockListType = {
 	blocks: Array<BlockType>,
 	aztechtml: string,
 	refresh: boolean,
+	showBlocks: void => void,
+	showHtml: void => void,
 };
 
 type PropsType = BlockListType;
@@ -42,7 +45,6 @@ type StateType = {
 };
 
 export default class BlockManager extends React.Component<PropsType, StateType> {
-	_htmlTextInput: TextInput = null;
 
 	constructor( props: PropsType ) {
 		super( props );
@@ -52,7 +54,7 @@ export default class BlockManager extends React.Component<PropsType, StateType> 
 			inspectBlocks: false,
 			blockTypePickerVisible: false,
 			selectedBlockType: 'core/paragraph', // just any valid type to start from
-			html: '', // This is used to hold the Android html text input state.
+			html: '',
 		};
 	}
 
@@ -154,9 +156,9 @@ export default class BlockManager extends React.Component<PropsType, StateType> 
 			}, '' );
 	}
 
-	parseHTML( html: string ) {
+	parseHTML() {
 		const { parseBlocksAction } = this.props;
-		parseBlocksAction( html );
+		parseBlocksAction( this.state.html );
 	}
 
 	componentDidUpdate() {
@@ -251,12 +253,17 @@ export default class BlockManager extends React.Component<PropsType, StateType> 
 	}
 
 	handleSwitchEditor = ( showHtml: boolean ) => {
-		if ( ! showHtml ) {
-			const html = this._htmlTextInput._lastNativeText;
-			this.parseHTML( html );
-		}
+		showHtml ? this.showHtml() : this.showBlocks()
+	}
+
+	showBlocks() {
+		this.parseHTML();
+		this.setState( { showHtml: false } );
+	}
+
+	showHtml() {
 		const html = this.serializeToHtml();
-		this.setState( { showHtml, html } );
+		this.setState( { showHtml: true, html } );
 	}
 
 	handleInspectBlocksChanged = ( inspectBlocks: boolean ) => {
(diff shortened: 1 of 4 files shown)


src/block-management/block-manager.js
```diff
@@ -29,6 +30,8 @@ export type BlockListType = {
 	blocks: Array<BlockType>,
 	aztechtml: string,
 	refresh: boolean,
+	showBlocks: void => void,
+	showHTML: void => void,
```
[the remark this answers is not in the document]
etoledom: Right! My mistake.
I believe I saw a `flow` error when I defined the functions, but I probably got confused. Those were meant to be internal `BlockManager` functions, so you are right - they shouldn't be defined in `BlockListType`. 👍

src/block-management/block-manager.js
```diff
@@ -52,7 +53,7 @@ export default class BlockManager extends React.Component<PropsType, StateType>
 			inspectBlocks: false,
 			blockTypePickerVisible: false,
 			selectedBlockType: 'core/paragraph', // just any valid type to start from
-			html: '', // This is used to hold the Android html text input state.
+			html: '',
```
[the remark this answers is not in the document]
hypest: > My original idea was to let HTMLInputView handle just text

It was a perfect start, nothing wrong there. It just turns out that html related code/logic is found in many places in `block-manager.js` while the block manager ideally has nothing to do with the raw html. So, it now makes sense to try to abstract all html handling away.

> even though it would create a coupling (probably not an issue right now?)

Hmm, not sure which coupling you have in mind. Coupling to the a block's list? The block list is currently the master source of info and is stored in Redux so, ideally we could even have the HTMLTextInput be completely separated from the BlockManager, to get rendered by the `MainScreen` if the `showHtml` prop/state is true. In a sense, I think I see a path to completely decouple the html editing from the block list itself that way. Does that make sense?

All in all, it's not a high priority to go down that route now so, not going to block this PR on this one.

> We would have to generate the block list on every key stroke and give it back via props function

Ah, I see what you mean. Sounds like a case where we would only want to update the Redux state via a debouncer or when HTMLTextInput gets blurred. I wonder how GB-web does it with its full post html editor.

> We hold a reference to the component and call a function on it to get the new set of blocks (or the html text itself). It will probably work but I believe it's not too react-like.

Yeah, woudn't like us to end up there if we can avoid it.

All in all, I'd say let's timebox this one. Can't think of great benefit optimizing the code separation too much right now.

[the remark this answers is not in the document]
etoledom: > Ideally we could even have the HTMLTextInput be completely separated from the BlockManager, to get rendered by the MainScreen.

I see. That sounds like a great plan!
If it was the plan all along, I'm sorry I missed it 😆

I see that my approach wouldn't work, but you gave so many tips that I will try again with this new approach. 

I'll keep you posted. 😁

etoledom:
@hypest - Thanks for the advices!

I believe this is ready for a second look.
Following to out conversation, the `HTMLTextInput` component handles all html related parsing and serializing, and it updates the redux store.

Serializing happens on **mount** and parsing happens on **unmount**, so no extra work for each keystroke.

`HTMLTextInput` is still part of the `BlockManager` hierarchy, but it should be ready to work independently.

src/components/html-text-input.js
```diff
@@ -0,0 +1,76 @@
...
+	blocks: Array<Object>,
+	parseBlocksAction: string => mixed,
+};
+
+type StateType = {
+	html: string,
+};
+
+export default class HTMLInputView extends React.Component<PropsType, StateType> {
+	state = {
```
hypest: Should we initialize the state using a constructor? If I interpret the doc properly, that's the recommended way as per the [React doc](https://reactjs.org/docs/react-component.html#constructor)?

> Constructor is the only place where you should assign this.state directly. In all other methods, you need to use this.setState() instead.

src/components/html-text-input.js
```diff
@@ -0,0 +1,76 @@
...
+	componentWillUnmount() {
+		//TODO: Blocking main thread
+		this.props.parseBlocksAction( this.state.html );
+	}
+
+	shouldComponentUpdate() {
+		const isIOS = Platform.OS === 'ios';
+		if ( isIOS ) {
+			// iOS TextInput gets jumpy if it's updated on every key stroke.
+			return this.state.html === '';
```
hypest: Not a blocker for the PR but, I wonder if it's intentional that we're checking the _current_ state for empty html instead of checking the _next_ state.

In case it was not intentional, should we check the _next_ State? `shouldComponentUpdate` accepts the next Props and State as parameters we can check. See https://reactjs.org/docs/react-component.html#shouldcomponentupdate.

etoledom: Yes! It is intentional. The first render of the component will always be empty. And since on iOS we were blocking updates, it was never showing content. 
So the logic is: If the previous state was empty, we let update the component to show the next state.

Since I had to explain it here, probably I should explain it better in code too. 😆

src/components/html-text-input.js
etoledom: I took it from [this example](https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html#initializing-state).
I believe I had read it too, that if the constructor is only initializing the state, it is valid to remove the constructor and init the state it in this way. Not sure if it's just a convenience.

hypest: Perfect, let's keep it as it currently is in the PR! 👍

PR #154, opened 2018-09-24
Fix broken jest platform resolution

### LINKED ISSUE #153: Jest test picks up iOS files while running for Android
The Jest tests setup is mistakenly picking up iOS platform files (`.ios.js` modules) while running in "android" platform mode.

### Steps to reproduce:
* Make a platform specific version of the `src/app/App.js` module as `src/app/App.ios.js`
* Edit the iOS variant and put a `console.log( 'This is an iOS specific module!' );` early on top of the module
* Run `yarn test`. Notice that without any command line parameters like that, the tests are set up to run the Android codepaths (`Setting RN platform to: default (android)` printed in the console).
* Notice the `This is an iOS specific module!` being printed while it shouldn't

hypest:
## Fixes #153 

This PR alters the way packages are resolved during the Jest run. Up to now, we were relying on the `moduleNameMapper` Jest config to map various `@wordpres/...` packages directly to the `src/index` module and call it a day.

Relevant PR on the Gutenberg repo: https://github.com/WordPress/gutenberg/pull/10131

### Background

Unfortunately, the Jest platform resolution setup was actually broken for some time now. See #153. Rootcause for #153 is the extensions defined in `moduleFileExtensions`. Unfortunately those were effectively bypassing the proper platform resolution mechanism in Jest/Haste.

By removing the excess extension definitions (leaving `moduleFileExtensions` with its defaults by deleting the whole field anyway) surfaced another problem, where the module mapper in Jest was breaking how Haste resolves packages: For example, Haste was having the `@wordpress/a11n` module in its cache but, because the mapping (apparently) happens after the cache is built the mapped module (`<fullpath>/gutenberg/packages/a11n/src/index`) cannot be found in the cache and fails to resolve. Removing the mapping fixes the issue but, Jest can no longer find the packages at all.

That's where the folder with the symlinks come into play. `symlinked-packages` is a folder that has a `@wordpress` subfolder in which there are now a set of symlinks, each named after the package we're interested in loading from the Gutenberg code. The symlink is pointing directly to the `src/` folder of the respective package. `symlinked-packages` is added as an extra node modules directory in Jest's config. When Jest tries to resolve, say, `@wordpress/a11n` it can find it inside `symlinked-packages`. Similar situation with `symlinked-packages-in-parent` but for the inside-gb setup.

In a slightly unrelated way, it happens that at the time of writing, the `@wordpress/element` package on NPM doesn't match the local one and the local one has proper serialization code for the blocks that `save` using React's `Provider`s. Using the package from NPM leads to all block get serialized to empty string, failing all the `isValid()` block checks. It's clear that we should either use *all* the packages from NPM or *all* the packages from local. This PR changes the dependencies to only use the local packages.

### "Can't find variable: Symbol" error

It also happens that the `@wordpress/redux-routine` package (local version at the time of writing) introduces the use of the `rungen` package which happens to be incompatible with the version of the JavascriptCore used in the Android RN app. This PR follows https://github.com/react-community/jsc-android-buildscripts#how-to-use-it-with-my-react-native-app to include an updated version of JavascriptCore but bumps the Android `minSdkVersion` to 21 at the same time. That makes the app requiring Android 5.0 and above.

## To Test

**Test A:**
No new functionality is added by this PR and in general, the app and tests should work as normal.

**Test B:**
Steps to test the #153 fix:
* Make a platform specific version of the `src/app/App.js` module as `src/app/App.ios.js`
* Edit the iOS variant and put a `console.log( 'This is an iOS specific module!' );` early on top of the module
* `yarn test src/app/App.test.js` to run a single test module so we can more easily observe the console logs. Notice that without any command line parameters like that, the tests are set up to run the Android codepaths (`Setting RN platform to: default (android)` printed in the console).
* Notice that the `This is an iOS specific module!` log is *not* printed.
* Now, use `TEST_RN_PLATFORM=ios yarn test src/app/App.test.js` to run the same test module but setting the platform to `ios`. Notice the `This is an iOS specific module!` is now printed.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,4 +1,5 @@
 ; ignore the submodules
 gutenberg
+symlinked-packages-in-parent
 react-native-aztec
 
--- a/.flowconfig
+++ b/.flowconfig
@@ -74,6 +74,8 @@ libdefs.js
 emoji=true
 
 module.system=haste
+module.system.node.resolve_dirname=node_modules
+module.system.node.resolve_dirname=symlinked-packages
 
 module.file_ext=.js
 module.file_ext=.jsx
@@ -84,10 +86,6 @@ module.file_ext=.scss
 munge_underscores=true
 
 module.name_mapper='^[./a-zA-Z0-9$_-]+\.\(bmp\|gif\|jpg\|jpeg\|png\|psd\|svg\|webp\|m4v\|mov\|mp4\|mpeg\|mpg\|webm\|aac\|aiff\|caf\|m4a\|mp3\|wav\|html\|pdf\)$' -> 'RelativeImageStub'
-module.name_mapper='@wordpress/block-library' -> '<PROJECT_ROOT>/gutenberg/packages/block-library/src'
-module.name_mapper='@wordpress/blocks' -> '<PROJECT_ROOT>/gutenberg/packages/blocks/src'
-module.name_mapper='@wordpress/element' -> '<PROJECT_ROOT>/gutenberg/packages/element/src'
-module.name_mapper='@wordpress/components' -> '<PROJECT_ROOT>/gutenberg/packages/components/src'
 
 ; mock/ignore style files
 module.name_mapper='.*\(.scss\)' -> 'empty/object'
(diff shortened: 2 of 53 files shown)


daniloercoli:
Looks good - Nice work!

PR #155, opened 2018-09-25
Update React Native to version 0.57

SergioEstevao:
This PR updates Rect Native to version 0.57.1

This causes a lot of other dependencies to be update to, mainly the following:

- React to version 16.5.0
- Babel to version 7
- Metro to version 0.45.5 
- Jest to 0.23.6

I started the upgrade job by following the steps [here](https://github.com/react-native-community/react-native-releases/blob/master/CHANGELOG.md#updating-to-this-version).

Used this [Metro page](https://facebook.github.io/metro/docs/en/configuration) to see how the new config file worked.

Babel 7 also brought a lot of new things so I used this [page](https://babeljs.io/docs/en/next/v7-migration) for reference. The main changes here was to use the new scoped pages and the `babel.config.js` instead of `.babelrc`.

For Jest I started with this [page](https://jestjs.io/docs/en/tutorial-react-native), but this was still giving me all kinds of problems so I searched the internet for similar problems and the best solution was provided [here](https://github.com/facebook/react-native/issues/20876).

This still needed some extra wrangle to make work, but we finally have a working repo for version 0.57 of React Native.


babel.config.js
```diff
@@ -0,0 +1,32 @@
+module.exports = function( api ) {
+	api.cache( true );
```
hypest: Since there's apparently a cache being used here, can we make sure we cover the cleaning of it with the current clean:... scripts or perhaps add a new one for it?

SergioEstevao: I added a new step to the scripts `clean:babel-cache` that was added to the `start:reset` script.

package.json
```diff
@@ -7,15 +7,18 @@
     "scssfiles": "src/*.scss src/**/*.scss"
   },
   "devDependencies": {
+    "@babel/core": "^7.1.0",
+    "@babel/plugin-proposal-async-generator-functions": "^7.1.0",
+	"@babel/plugin-transform-async-to-generator": "^7.1.0",
```
hypest: Nit: a few lines in `package.json` seem to have a different indentation style. Let's align those with the rest (4 spaces I think?)

package.json
```diff
@@ -67,12 +71,13 @@
...
     "clean:node": "rm -rf node_modules",
-    "clean:watchman": "watchman watch-del-all",
+	"clean:watchman": "watchman watch-del-all",
+	"clean:babel-cache": "rm -rf ./node_modules/.cache/babel-loader/*",
     "lint": "eslint . --ext .js",
     "lint:fix": "yarn lint --fix"
   },
-  "dependencies": {
-    "babel-preset-es2015": "^6.24.1",
+  "dependencies": {
```
hypest: Nit: trailing space

rn-cli-inside-gb.config.js
```diff
@@ -1,6 +1,7 @@
 /** @format */
-const blacklist = require( 'metro' ).createBlacklist;
 const path = require( 'path' );
+const blacklist = require( 'metro-config/src/defaults/blacklist' );
+const blacklistElements = blacklist( [ new RegExp( path.basename( __dirname ) + '/gutenberg/.*' ) ] );
```
hypest: It seems that while moving this exclusion here, we lost the inline comment that was there before. Let's reinstate the comment, or just perhaps just move the line back as it was, without defining the const here. WDYT @SergioEstevao ?

(Same for the `rn-cli.config.js` module too)

hypest:
Nice work @SergioEstevao 🎉 

LGTM! Feel free to merge when Travis gives the 💚

PR #158, opened 2018-10-08
Update GB hash to latest master

daniloercoli:
Title has it. 

Better to update to GB `master` asap, and test it well since new richText format already landed in GB. (Ref: https://github.com/WordPress/gutenberg/pull/7890)


hypest:
Confirmed Android, iOS app working both with `yarn start` and `yarn start:inside-gb` (when from inside Gutenberg), and tests are green from inside Gutenberg too. :shipit: !

PR #159, opened 2018-10-12
Add a new empty para block on Enter.KEY

daniloercoli:
This PR updates GB hash, and adds the code to insert a new para block after the current focused when enter.key is pressed.

There are few problems at the moment, not strictly related to this PR, explained below:
- `this.props.focusBlockAction( newBlock.clientId )` in `block-manager.js` is not working fine, the new block is not getting the focus, both if you tap `+` on the bottom toolbar and both if you press Enter.key in a para/heading block. cc @mzorz 
- There are problems getting the focus on blocks on Android, and consequence of this is that the bottom block management toolbar is not visible, and the index of the current block unknown. consequence of this is that I needed to add a hack to insert the block in the middle of the list atm.

GB side PR is available here: https://github.com/WordPress/gutenberg/pull/10553

### DIFF
--- a/src/block-management/block-holder.js
+++ b/src/block-management/block-holder.js
@@ -69,6 +69,7 @@ export default class BlockHolder extends React.Component<PropsType, StateType> {
 				setAttributes={ ( attrs ) =>
 					this.props.onChange( this.props.clientId, { ...this.props.attributes, ...attrs } )
 				}
+				insertBlocksAfter={ this.props.insertBlocksAfter }
 				isSelected={ this.props.focused }
 				style={ style }
 			/>
--- a/src/block-management/block-holder.js
+++ b/src/block-management/block-holder.js
@@ -50,14 +50,6 @@ export default class BlockHolder extends React.Component<PropsType, StateType> {
 		return <View />;
 	}
 
-	_insertBlocksAfter( blocks: Array<Object> ) {
-		if ( ! this.props.insertBlocksAfter ) {
-			return;
-		}
-		const { insertBlocksAfter } = this.props;
-		insertBlocksAfter( this.props.clientId, blocks );
-	}
-
 	getBlockForType() {
 		// Since unsupported blocks are handled in block-manager.js, at this point the block should definitely
 		// be supported.
@@ -78,7 +70,7 @@ export default class BlockHolder extends React.Component<PropsType, StateType> {
 				setAttributes={ ( attrs ) =>
 					this.props.onChange( this.props.clientId, { ...this.props.attributes, ...attrs } )
 				}
-				insertBlocksAfter={ this._insertBlocksAfter.bind( this ) }
+				insertBlocksAfter={ this.props.insertBlocksAfter }
 				isSelected={ this.props.focused }
 				style={ style }
 			/>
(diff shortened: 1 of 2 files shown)


src/block-management/block-manager.js
```diff
@@ -139,6 +139,26 @@ export default class BlockManager extends React.Component<PropsType, StateType>
 		this.state.dataSource.setDirty();
 	}
 
+	insertBlocksAfter( blocks: Array<Object> ) {
+		// find currently focused block
+		let focusedItemIndex = this.findDataSourceIndexForFocusedItem();
+		if ( focusedItemIndex == -1 ) {
+			// no focus yet? This is a well knwon bug on Android...
```
mzorz: let's add a ToDo comment here to fix it instead of leaving the magic number 3 with no other explanation

daniloercoli:
Removed empty lines and added a TODO - Thanks for your review @mzorz !

src/block-management/block-manager.js
```diff
@@ -139,6 +139,27 @@ export default class BlockManager extends React.Component<PropsType, StateType>
 		this.state.dataSource.setDirty();
 	}
 
+	insertBlocksAfter( blocks: Array<Object> ) {
```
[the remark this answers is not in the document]
daniloercoli: Interesting...will try that!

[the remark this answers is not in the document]
hypest: Awesome! Just to expand a bit, we can make block-manager's `insertBlocksAfter` accept the `clientID` but we can still pass a wrapped versions down to the block-holder that only needs the `blocks` parameter.

src/block-management/block-holder.js
```diff
@@ -49,6 +50,14 @@ export default class BlockHolder extends React.Component<PropsType, StateType> {
 		return <View />;
 	}
 
+	_insertBlocksAfter( blocks: Array<Object> ) {
```
hypest: In https://github.com/wordpress-mobile/gutenberg-mobile/pull/159/files#diff-2d50daf729e068ecebcce113d1ce7f9aR270, we pass the `insertBlocksAfter` method to all the rendered items so, I don't think we need to check whether the prop exists, right?

Then again, I think we don't even need this `_insertBlocksAfter` here as we can construct a wrapped method with just the `blocks` parameter right there in https://github.com/wordpress-mobile/gutenberg-mobile/pull/159/files#diff-2d50daf729e068ecebcce113d1ce7f9aR270.

For example we can have this in block-manager.js:
```
insertBlocksAfter={ ( blocks ) =>
	this.insertBlocksAfter.bind( this )( blocks, value.item.clientId )
}
```
WDYT?

src/block-management/block-holder.js
```diff
@@ -19,6 +19,7 @@ type PropsType = BlockType & {
 	onChange: ( clientId: string, attributes: mixed ) => void,
 	onToolbarButtonPressed: ( button: number, clientId: string ) => void,
 	onBlockHolderPressed: ( clientId: string ) => void,
+	insertBlocksAfter: ( clientId: string, blocks: Array<Object> ) => void,
```
hypest: I think we need to remove the `clientId` parameter from the definition here. Essentially, what we will be passing down to the block-holder is a function with the `clientId` string already "embedded" in it so, we will just call it by passing the new blocks.

daniloercoli:
Thanks @hypest and @mzorz for your review!

PR #162, opened 2018-10-15
Fix: html editor

etoledom:
This was a regression from (I think) updating `react-native`.
Some internal on how the TextView works on iOS broke our (hacky) HTML editor implementation.

The original problem look like this:
editor

Now if we set a value directly to the TextInput component, and we don't let it re-render, the component will keep that state not allowing editions:
editor_2

This PR solves that problem by setting the TextView text value directly using a reference.

@marecar3 Could you please take a look at this?

To test:
- Run the project on iOS
- Check that the HTML Editor behaves normally.

### DIFF
--- a/src/components/html-text-input.js
+++ b/src/components/html-text-input.js
@@ -29,8 +29,8 @@ export default class HTMLInputView extends React.Component<PropsType, StateType>
 	componentDidMount() {
 		const html = this.serializeBlocksToHtml();
 		this.setState( { html } );
-		if (this.isIOS) {
-			this.textInput.setNativeProps({text: html});
+		if ( this.isIOS ) {
+			this.textInput.setNativeProps( { text: html } );
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


src/components/html-text-input.js
```diff
@@ -23,10 +23,15 @@ export default class HTMLInputView extends React.Component<PropsType, StateType>
 	state = {
 		html: '',
 	}
+	isIOS: boolean = Platform.OS === 'ios';
```
marecar3: Maybe we can use this import `import { isAppleOS } from './platform';`
from gutenberg (file : [platform](https://github.com/WordPress/gutenberg/blob/master/packages/keycodes/src/platform.native.js)) so if something changes in the future we will have better control of that part of the code.

etoledom: Good catch!
It looks like that specific implementation is an internal of `@wordpress/keycodes`, but still it's a good idea to abstract it in the same way in our codebase.

Let's do that and replace all the instances of `Platform.OS === 'ios'` on a separate PR.

marecar3: I agree @etoledom

Issue #167, opened 2018-10-17
Block split and merge


koke:
When a user presses return the block should split in two at the cursor. Similarly, backspace at the beginning of a block should merge it with the previous block (or _forward delete_ at the end of a block)

There are some extra rules for this so double check what Gutenberg does and follow the same patterns. I don't think we have to worry about many of those yet since we only support a few block types, but one that I've identified so far is:

- When you press return, the newly created block will be a paragraph, unless it's a multiline block (I think we only have `code` in the alpha).

- [x] Merge the content (or return false to let Aztec native to handle the backspace)
- [x] Merge 2 blocks of different types
- [x] Resize after block splitting
- [x] Fix problem with content lost on merging
- [x] Fix delayed updated from native side (not completely related to this feature, but we need it)
- [ ] Focus is moved to the wrong block after merging/splitting (We may need to wait until focus problems are resolved)
- [ ] ~On blocks merging, should we add a white space character before the content of the 2nd block? (if necessary). Otherwise there is a situation where the last word of the first block, and the first word of the 2nd block, are merged together firing up the spell checker inline suggestions.~
- [x] ~Test on RTL languages~ #380
- [x] Block Splitting:  If `before` content is omitted, treat as intent to delete block and call `onReplace( [] );`
- [x] Check `Code` block in regard to enter and backspace

- [x] Using backspace while the first word is highlighted should not merge the 2 blocks, only remove the currently selected word. It's working as expected on Android, but iOS does merge the 2 blocks. #325

- [ ] Add Enter.key handling/detection to PlainText and PlainText powered blocks #324

diegoreymendez:
Issues remaining (at least in iOS):

1. @danilo and @hypest [reported a problem](
https://github.com/wordpress-mobile/react-native-aztec/pull/62#issuecomment-431324688) with the text at the right position of the caret not being removed from the original block.
2. Blocks are not resizing correctly after being split.
3. The iOS native caret position should be compatible with unicode characters.  This is especially important for languages such as Japanese and Chinese.

koke:
@daniloercoli @diegoreymendez since you're already working on this, can you make a task list of what needs to be done and add some estimations by Wednesday?

> The iOS native caret position should be compatible with unicode characters. This is especially important for languages such as Japanese and Chinese

Is this not working now? How is it behaving?

koke:
Also, I initially put this in #168 but take into account as well updating the Heading (and any other ported) block so it works correctly

[...]

PR #172, opened 2018-10-17
Update ReactNativeAztec hash - Include cursor selection on onHTMLContentWithCursor 

daniloercoli:
This PR updates ReactNativeAztec hash and update the event `onHTMLContentWithCursor`, that now includes selectionStart, selectionEnd, together with the HTML content.

Previously only the cursor position was returned.


diegoreymendez:
Related to #167

PR #173, opened 2018-10-17
Improve dependency checking for scripts

### LINKED ISSUE #157: Build on iOS fails silently if Carthage is not installed
I ran `yarn run iOS` on a new machine and I got an unrelated error. It took me a while to scroll back and see the `carthage: command not found`.

We should fail early if any of the required tools are missing

koke:
The (pre)ios command should now fail early if carthage is not installed
The clean:watchman command should now fail if watchman is not installed
The clean command will still perform the other clean steps

Includes https://github.com/wordpress-mobile/react-native-aztec/pull/60

@daniloercoli when updating the react-native-aztec submodule to master, it picked up some other changes, can you confirm those are good to go?

Fixes #157


daniloercoli:
Yes, those changes were good....anyway we've already merged an updated copy of `react-native-aztec` into master.

Feel free to update this PR and merge it!

:shipit:

PR #174, opened 2018-10-17
Updated README file

marecar3:
Updated readme file.

1. We have added a command which installs the Carthage as it is needed when Mac OS is first time used.

2. Removed deprecated link. I tried to find some other which suits most but the newest version of the file relies on EXPO [NEW README](https://github.com/react-community/create-react-native-app) which doesn't suit our needs.

If you have something else to add on this README, please leave a commend on this PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,13 +40,22 @@ Before running the demo app, you need to download and install the project depend
 yarn install
 ```
 
-## Pre-Setup iOS
+## Prerequisites
 
-Carthage is also needed for fetching the Aztec dependency and you can get it running following command:
+Carthage is also needed for fetching the Aztec dependency.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -40,9 +40,15 @@ Before running the demo app, you need to download and install the project depend
 yarn install
 ```
 
-## Run
+## Pre-Setup iOS
+
+Carthage is also needed for fetching the Aztec dependency and you can get it running following command:
```
koke: I don't think `preios` will *install* Carthage. I think we just need to add Carthage to the "Prerequisites" section

marecar3: Thanks @koke , you are right. It confused me for some reason.
I have added the "Prerequisites" section. Please, take a look.

PR #175, opened 2018-10-18
Updated yarn to 1.10.1

koke:
When I installed yarn in my new laptop I got the latest version (1.10) which
adds an integrity field to each entry in yarn.lock

I ran `yarn install` and by suggestion of @diegoreymendez added the specific
yarn version to the `engines` section in `package.json`, so yarn will prompt you
to upgrade if you use an older version

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,9 +6,6 @@
     "jsfiles": "./*.js src/*.js src/**/*.js src/**/**/*.js",
     "scssfiles": "src/*.scss src/**/*.scss"
   },
-  "engines": {
-    "yarn": "^1.10.1"
-  },
   "devDependencies": {
     "@babel/core": "^7.1.0",
     "@babel/plugin-proposal-async-generator-functions": "^7.1.0",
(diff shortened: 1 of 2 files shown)


hypest:
Quick question: have we tested this yet with the "inside-gb" setup? Like, what happens when the mobile project is run from within the Gutenberg repo. In that setup, `yarn` is a normal package dependency and installed via `npm install`.

diegoreymendez:
@hypest (cc @koke) - Regarding the compatibility question: Is there a way in which running the test from the web repo can pick up `gutenberg-mobile`s package.json?

In order to stay on point in this PR, I was thinking to just test this regularly, since changes to the mobile `package.json` file should not affect Gutenberg (Web).  If there's more we want to test we can always do it as part of a specific task.

hypest:
> Is there a way in which executing something from the web repo can pick up gutenberg-mobiles package.json

Yes. What actually happens is that the web's `npm test` script runs the mobile's `yarn test:inside-gb` script which of course depends on the mobile's package.json. So, I wonder/guess that could fail if the yarn version (installed as a _web's_ dependency [here](https://github.com/WordPress/gutenberg/blob/master/package.json#L118)) is not matching the one set in mobile's package.json?

koke:
OK, maybe not a good idea to enforce the upgrade yet

koke:
BTW, this is ready for another look

PR #176, opened 2018-10-18
Block splitting - Handle the case when the split occurs at the trailing or leading edge

daniloercoli:
This PR adds handles the case when the split occurs at the trailing or leading edge of the field.

Testing: 
1) Put the cursor at the end of a para block and hit enter. It should add an empty block after the current.
2) Put the cursor at the end of a para block and hit Enter. It should add a block after the current with all the content. Current block will be blank after.
3) Put the cursor in the middle of a block, and hit Enter.

Note: there are some known issues with Heading blocks and validation. Will address that in another PR.

Companion PR:
- GB side: https://github.com/wordpress-mobile/react-native-aztec/pull/61
- AztecReactNative: https://github.com/WordPress/gutenberg/pull/10737


diegoreymendez:
Related to #167

PR #178, opened 2018-10-19
Fix validation after block splitting

daniloercoli:
This PR fixes the validation problem on Heading blocks after block splitting, by making sure to remove the correct root tags returned by AztecWrapper.

It was a known issue, and we've fixed it for all blocks, even if the block splitting logic is in place for `Heading` and `Paragraph` block only.

GB side PR available here: https://github.com/WordPress/gutenberg/pull/10797


koke:
Shouldn't we merge the PR on Gutenberg first?

daniloercoli:
Yes, we should merge the GB side PR first -  Note there were problems with GH yesterday, and the other PR given us problem on merging.

PR #184, opened 2018-10-23
Use createElement from @wordpress/element

gziolo:
This PR tries to refactor code to use `createElement` through `@wordpres/element` rather than directly to stay closer to how Gutenberg web works.

It's old PR which worked with Babel 6, we probably would have to rework it to make it up to date. 

It also is part of my exploration of overriding `createElement` to allow usage of HTML tags as component names.

Requires https://github.com/WordPress/gutenberg/pull/11129

/cc @Tug

### DIFF
--- a/src/app/App.js
+++ b/src/app/App.js
@@ -3,8 +3,6 @@
 
 import '../globals';
 
-import React from 'react';
-
 // Gutenberg imports
 import { registerCoreBlocks } from '@wordpress/block-library';
 import { registerBlockType, setUnregisteredTypeHandlerName } from '@wordpress/blocks';
--- a/src/app/AppContainer.js
+++ b/src/app/AppContainer.js
@@ -1,7 +1,7 @@
 /** @flow
  * @format */
 
-import React from 'react';
+import { createElement, Component } from '@wordpress/element';
 import { parse, serialize } from '@wordpress/blocks';
 import { withDispatch, withSelect } from '@wordpress/data';
 import { compose } from '@wordpress/compose';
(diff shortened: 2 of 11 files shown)


gziolo:
I added Gutenberg part as a reference to how I tried to override `createElement` in https://github.com/WordPress/gutenberg/pull/10934.

In a nutshell, I tried the following:

```js
/**
 * External dependencies
 */
import { get, isString } from 'lodash';
import { createElement as createElementDefault } from 'react';
import {
	Circle,
	Defs,
	Ellipse,
	G,
	Line,
	Path,
	Polygon,
	TextPath,
	TSpan,
	Svg,
} from 'react-native-svg';

const htmlTagToNativeComponent = {
	circle: Circle,
	defs: Defs,
	ellipse: Ellipse,
	g: G,
	line: Line,
	path: Path,
	polygon: Polygon,
	svg: Svg,
	textPath: TextPath,
	tspan: TSpan,
	/*Polyline,
	Text,
	Rect,
	Use,
	Image,
	Symbol,
	LinearGradient,
	RadialGradient,
	Stop,
	ClipPath*/
};

const mapHTMLTagToNativeComponent = ( type ) => {
	if ( ! isString( type ) ) {
		return type;
	}

	return get( htmlTagToNativeComponent, [ type ], type );
};

/**
 * Returns a new element of given type. Type can be either a string tag name or
 * another function which itself returns an element.
 *
 * @param {?(string|Function)} type     Tag name or element creator.
 * @param {Object}             props    Element properties, either attribute
 *                                       set to apply to node or values to
 *                                       pass through to element creator.
 * @param {...WPElement}       children Descendant elements.
 *
 * @return {WPElement} Element.
 */
const createElement = ( type, props, ...children ) => {
	return createElementDefault(
		mapHTMLTagToNativeComponent( type ),
		props,
		...children
	);
};

export default createElement;
```

gziolo:
You made Flow and Jest unhappy, it probably needs more works for the configuration. However, this is exactly what should happen 👍

gziolo:
I don't think it is still valid. We can always reopen :)

PR #185, opened 2018-10-23
Fix error on iOS when adding a new block from paragraph block.

etoledom:
This PR is a proposal to fix an issue on iOS when selecting a `paragraph` block and creating a new block from it:

screen_shot_2018-10-23_at_16 31 47

We use `findDataSourceIndexForFocusedItem` to get the index of the focused block and insert the new block below it.

`findDataSourceIndexForFocusedItem` use `this.state.dataSource` to get the block state, and I noticed that the `focused` state of the focused block was `false`. But the same block had `focused` state `true` on `this.props.blocks`.

Using `this.props.blocks` instead of `this.state.dataSource` on iOS solved the problem.

This still feels like a workaround other than finding and fixing the real problem. Hopefully @tug 's work on Gutenberg internal state will fix this more permanently 😁

To test:
- Run the iOS example app.
- Select a paragraph block.
- Tap on the + button to add a new block.
- Try to add a new block.
- Check that the block is added successfully.

### DIFF
--- a/src/block-management/block-manager.js
+++ b/src/block-management/block-manager.js
@@ -66,12 +66,18 @@ export default class BlockManager extends React.Component<PropsType, StateType> 
 
 	findDataSourceIndexForFocusedItem() {
 		for ( let i = 0; i < this.state.dataSource.size(); ++i ) {
-			const block = this.state.dataSource.get( i );
+			const block = this.findBlockAtIndex( i );
 			if ( block.focused === true ) {
 				return i;
 			}
 		}
 		return -1;
+	}
+
+	findBlockAtIndex( index: number ): Object {
+		return Platform.OS === 'ios' ?
+			this.props.blocks[ index ] :
+			this.state.dataSource.get( index );
 	}
 
 	// TODO: in the near future this will likely be changed to onShowBlockTypePicker and bound to this.props
--- a/src/block-management/block-manager.js
+++ b/src/block-management/block-manager.js
@@ -66,18 +66,12 @@ export default class BlockManager extends React.Component<PropsType, StateType> 
 
 	findDataSourceIndexForFocusedItem() {
 		for ( let i = 0; i < this.state.dataSource.size(); ++i ) {
-			const block = this.findBlockAtIndex( i );
+			const block = this.state.dataSource.get( i );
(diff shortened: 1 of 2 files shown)


src/block-management/block-manager.js
```diff
@@ -66,14 +66,20 @@ export default class BlockManager extends React.Component<PropsType, StateType>
...
+			const block = this.findBlockAtIndex( i );
 			if ( block.focused === true ) {
 				return i;
 			}
 		}
 		return -1;
 	}
 
+	findBlockAtIndex( index: number ): Object {
+		return Platform.OS === 'ios' ?
```
hypest: I think that the root cause of the issue at hand is that the two lists (the one in the props and the other that's kept in the dataSource) get out of sync. And I think they get out of sync [here](https://github.com/wordpress-mobile/gutenberg-mobile/blob/master/src/store/reducers/index.js#L74-L77), where a block is copied to a new instance and replaced in the list. The `dataSource` is still having the old version of the block in that index and then the focusing action fails to "select" both instances.

That syncing problem can be solved if we mutate the block instead of recreating it, by only setting its attrbibutes:

in `/src/store/reducers/index.js`:
```diff
-                       const index = findBlockIndex( blocks, action.clientId );
-                       blocks[ index ] = {
-                               ...block,
-                               attributes: nextAttributes,
-                       };
+                       block.attributes = nextAttributes;
```

Wdyt?

etoledom: This is great! I completely missed that.
I tested it and works perfectly. Pushing a commit with the change.

Thank you!

diegoreymendez:
I noticed that our focus actions [seems to be off](https://github.com/wordpress-mobile/gutenberg-mobile/blob/master/src/store/reducers/index.js#L91), at least for paragraphs and headings.

It is my understanding that tapping again on a paragraph or heading should not unfocus it.

Here's an example of what I'm seeing (keep an eye on the bottom toolbar, which I believe is what we call the inserter?).

problemswithdoublefocus

diegoreymendez:
I think we can safely switch this line to always set to `true`.  I tried that and it works fine (unless there's a scenario I'm missing).

Issue #166, opened 2018-10-17
Tap to write


koke:
When the user taps into a text block, they should start editing right away with the cursor in the position they tapped, even if it's not the currently selected block.

koke:
@diegoreymendez @daniloercoli can you update this one by Wednesday with what we need to do to ensure this and how much it should take?

koke:
Let's also consider #135 as part of this

daniloercoli:
There are a couple of problems with focus handling on the Android side. 
I know @mzorz was working on a fix for them, but I'm not quite sure the current state of the fix. IIRC the problem was with our `react-native-recyclerview-list` implementation, that was acting badly under some circumstances.
Once those "focus" problems are solved, I think we need to modify Aztec + Aztec Wrapper to correctly move the cursor to the position tapped by the user.

diegoreymendez:
It's a bit complicated because I'm not sure what's needed for this.  But assuming:

- Discovery work to learn what's necessary. (1 day)
- Initial implementation. (2 days)
- Testing / fixing issues (1 or 2 days)

I'd say 5 days would be my initial estimation.

Does this work for you @koke?

diegoreymendez:
Actually @koke, I assumed work was necessary in iOS, but it seems like it works out of the box.

Can you review and confirm if this behavior is fine for the purpose of this issue?

![taptoeditios](https://user-images.githubusercontent.com/1836005/47372655-9401a200-d6c0-11e8-93ff-62de60e1a5a5.gif)

koke:
@diegoreymendez yeah, that looks like it works as expected. Great news 👏

koke:
After chatting with @daniloercoli, this also seems to be working on Android and should be ok once we fix #135

koke:
After chatting with @mzorz we are estimating 3 days for #135, although he might come up with a better estimate after more research today.

diegoreymendez:
I removed myself from the assignment since there's nothing else I can do for this task right now.

PR #191, opened 2018-10-24
Tapping twice no longer removes focus.

### LINKED ISSUE #190: Problem with focus
I noticed that our focus action seems to be off, at least for paragraphs and headings.

It is my understanding that tapping again on a paragraph or heading should not unfocus it.

Here's an example of what I'm seeing (keep an eye on the bottom toolbar, which I believe is what we call the inserter?).

problemswithdoublefocus

I think we can safely switch [this line](
https://github.com/wordpress-mobile/gutenberg-mobile/blob/master/src/store/reducers/index.js#L91) to always set to `true`.  I tried that and it works fine (unless there's a scenario I'm missing).

diegoreymendez:
### Description:

Fixes #190 

Tapping twice on the same block no longer removes focus from it.

### Testing:

Just tap multiple times in the same text block and make sure the bottom toolbar is never removed.

### DIFF
--- a/src/store/reducers/index.js
+++ b/src/store/reducers/index.js
@@ -83,8 +83,7 @@ export const reducer = (
 				block.focused = false;
 			}
 
-			// Select or deselect pressed block
-			destBlock.focused = ! destBlockState;
+			destBlock.focused = true;
 			return { blocks: blocks, refresh: ! state.refresh };
 		}
 		case ActionTypes.BLOCK.MOVE_UP: {
--- a/src/store/reducers/index.js
+++ b/src/store/reducers/index.js
@@ -76,7 +76,6 @@ export const reducer = (
 		}
 		case ActionTypes.BLOCK.FOCUS: {
 			const destBlock = findBlock( blocks, action.clientId );
-			const destBlockState = destBlock.focused;
 
 			// Deselect all blocks
 			for ( const block of blocks ) {


diegoreymendez:
> The travis error seems to be just a lint issue

Yes.  I'll change the code to fix this.

> There is not current way to deselect all blocks other than switching to html view and back. But I didn't find a way to deselect all from the mobile web version either. 🤔

My opinion is that it may not be a requirement at all to be able to "unfocus" all blocks.

If you think of Aztec and most mobile editors there's no way to remove the caret / focus / responder status from them.

iamthomasbishop:
> My opinion is that it may not be a requirement at all to be able to "unfocus" all blocks.

Tapping outside of a focused block should in theory unfocus all blocks. For example: if you're focused on a block, you have to tap once outside the block to unfocus it. That means at this time, no blocks are focused. You then can tap on another block to focus that one. 

The nature of the block editor means a slight shift in approach vs. a typical single-stream editor, but if this adds too much friction in practice, I'm all for adjusting.

etoledom:
The problem I see with the `tapping outside to unfocus` approach, is that on mobile, unless you are at the very top or very bottom of the post, any space outside the current focused block, will be another block, and taping on it will focus that other block instead.

They use all the width of the screen and they are vertically pasted each other.

As a user, I'd like to give a fast preview of what I'm writing without the tool bars. 
But again, not important at all for Alpha, and we will have a better insight later on. 👍

diegoreymendez:
> Tapping outside of a focused block should in theory unfocus all blocks

That doesn't match [this requirement](https://github.com/wordpress-mobile/gutenberg-mobile/issues/166), though.

Are you sure we want to have to tap twice to move the focus from block to block?

iamthomasbishop:
> Are you sure we want to have to tap twice to move the focus from block to block?

There are a lot of things at play with this interaction that are hard to articulate, but I'm not 100% tied to this if it adds too much friction – especially in terms of getting us to the initial version of Mobile GB. The most important thing to me is that we ensure that switching across blocks isn't jarring, whether via single or multiple taps.

I agree with @etoledom that an intermediate writing state where we remove all focus is desired. But as @koke pointed out, this could very well be achieved by adding the `dismiss keyboard` button on the toolbar (the original designs didn't have that, so it's now a consideration and maybe much less of a concern).

PR #194, opened 2018-10-24
Updates react-native-aztec to support backspace interception in iOS.

diegoreymendez:
### Description:

Brings forward #167.

Implements backspace interception for iOS.

### Additional info:

`react-native-aztec` PR here: https://github.com/wordpress-mobile/react-native-aztec/pull/65

### Testing:

To test this, set a console.log [here](https://github.com/WordPress/gutenberg/pull/10997/files#diff-ed195425f515f40db41c23d586760ef9R189) and make sure the event is called when the caret is at the beginning of the block, and backspace is pressed.


daniloercoli:
Looks good! :shipit:

PR #200, opened 2018-10-26
Blocks Merging - Introduce the JS callback

daniloercoli:
This is a simple PR that introduces the routine `mergeBlocks` in `BlockManager`, that does hold initial part of the the blocks merging logic, and passes it down to blocks.

Companion PR is here: https://github.com/WordPress/gutenberg/pull/11112

**To test**
- Place the caret at the beginning of a Para block that has another Para block on the top of it
- Tap backspace and see the block merge together

The same works if you use Heading blocks in the testing steps.

Still not yet addressed:
- Resize of the content after block merging
- Merging 2 block of different types does't work at the moment. There is an if condition that blocks the execution https://github.com/wordpress-mobile/gutenberg-mobile/pull/200/files#diff-2d50daf729e068ecebcce113d1ce7f9aR190
- If merging is blocked on a block we should move the focus on the block above


daniloercoli:
Sorry @diegoreymendez, I've added few other commits right before your approval. Will ping @hypest for another set of eyes.

diegoreymendez:
No worries, I thought it was ready since it was assigned, but I'll also review the complete PR again.

Issue #182, opened 2018-10-22
Integrate Gutenberg UI into the app editors


koke:
Once we bring Gutenberg in the apps with #163, we'll need to connect the UI pieces so that the editor:
- Shows a top bar with the same options as Aztec (preview, document options)
- Can load a post from the app
- Can save changes and upload a post

Task list:

iOS
- [x] Working setup of a `Native Module` for communication between RN and Native.
- [x] Send initial html content (Native -> RN)
~- Close button (RN -> Native [?])~ (Handled in the Native side)
- [x] Retrieve current content (Native -> RN -> Native) (Maybe not needed for Alpha)
- [x] Save button implementation in WPiOS side.
- [x] Create new (empty) post. 1d
- [x] Implement content change detection, similar to Aztec
More button options:
- [x] Implement Save as draft 1d
- [x] Implement Preview 1d
- [x] Implement Post settings 1d
- [x] Switch to html 1d
- [x] Add History option(if it is not a big deal) 1d

Android

- [x] Shows a top bar with the same options as Aztec (preview, document options) (3days including all project scaffolding)
- [x] Can load a post from the app (1day)
- [x] Can save changes and upload a post (3days)
- [x] Implement content change detection, similar to Aztec (1day)
- [x] Create new (empty) post (1d)
- [x] Handle device rotation (2days)
- [x] Fix reverting to a revision from History (if it is not a big deal) 1d
- [x] Cleanup the GutenbergEditorFragment class (remove unused lines, unused menus) (1day)

etoledom:
After playing around with this tasks, I believe that this is more about overall communication between RN and Native, namely `Native Modules`.

- [ ] Working setup of a `Native Module` for communication between RN and Native.

It would be good to identify all the communication points needed to satisfy the Alpha milestone. So far I can think of the following:

- [ ] Send initial html content (Native -> RN)
- [ ] Save button (RN -> Native [?])
- [ ] Close button (RN -> Native [?])
- [ ] Retrieve current content (Native -> RN -> Native) (Maybe not needed for Alpha)
- [ ] Show media picker (config info) (RN -> Native)
- [ ] Image selected (Native -> RN)
- [ ] Image networking state [uploading, finished, error) (Native -> RN)

The list could grow (or shrink). Some of them are super simple (few hours or less) and others might take longer (few days). Many of them are already working on the example branch (iOS) but none is yet merged.

[...]

PR #201, opened 2018-10-27
Full circle communication with parent app

hypest:
Addresses #182 , #163 .

## In this PR:

Note: Only the Android side is fully working/tested. There's a test failing on iOS, possibly because we need to add some proper implementation of the bridge. @etoledom , let me know if the iOS side here makes sense and we can talk. @mzorz , feel free to give this a go, at least to make sure that it works on Android on your side too. Thanks!

* Introducing `react-native-gutenberg-bridge`, a local npm module that holds a "Bridge" NativeModule project to establish a communication channel between a native parent app and the Gutenberg-mobile editor
* The parent native app sends events and the RN app responds by calling a method back to the parent app. Flow of events/actions:
1. Parent app sends the "requestGetHtml" event down to JS via `RCTDeviceEventEmitter` (and possibly blocks waiting for a method call in the opposite direction)
2. RN side (`MainScreen` component) listens to that event and issues a Redux action (`ActionTypes.BLOCK.SERIALIZE_ALL`)
3. Redux middleware (`gutenbergBridgeMiddleware`) responds to that action by serializing all blocks to html and calling the `provideToNative_Html(html)` method on the bridge
4. Parent app unblocks by the method call in 3. and receives the html content from the method parameter

Accompanying PRs:
* https://github.com/wordpress-mobile/react-native-aztec/pull/66
* https://github.com/wordpress-mobile/react-native-recyclerview-list/pull/9

Main ideas:
* a subproject (3 platforms: JS, Android, iOS) that implements both the native sides and the JS side of the bridge
* handle the request from the parent as just another Store action
* Redux middleware to separate the concern of responding back to the parent

### To test on Android:

1. Use the `feature/integrate-gutenberg` branch from the WP Android app https://github.com/wordpress-mobile/WordPress-Android/tree/feature/integrate-gutenberg
2. Update the submodule (`git submodule update --init`)
3. cd into `gutenberg-mobile` and do the usual (`git submodule update --init`, `yarn install`, `yarn start:reset`) to start Metro
4. Run the wpandroid app by issuing `yarn wpandroid`. This will compile and launch the parent WP Android app.
5. Open a GB post. GB-mobile should come up after bootstrapping from Metro and it should show the post.
6. Do some minor text editing
7. Go Back. That will trigger a back and forth from parent to RN to fetch the post content html. The post will be marked as having "Local changes" and the changes will be visible in the Preview.

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -1,7 +1,7 @@
 /** @flow
  * @format */
 
-const defaultPlatform = 'ios';
+const defaultPlatform = 'android';
 const rnPlatform = process.env.TEST_RN_PLATFORM || defaultPlatform;
 if ( process.env.TEST_RN_PLATFORM ) {
 	// eslint-disable-next-line no-console


etoledom:
@hypest - I've added a commit with a simple implementation of the iOS side of the bridge, so it doesn't error.

**To test:**
- run `yarn iOS`
- Check that it runs the iOS app as before.

----

The bridge implementation looks super good to me! I believe it will be quite strait forward to use this approach on my previous bride work. It might even solve a couple of workarounds in place. 👍

etoledom:
Travis:
It keep failing on:
`Setting RN platform to: ios`
`Invariant Violation: Native module cannot be null.`

That's the error I just fixed, I wonder why it is still failing on Travis.
I'll double check.

etoledom:
> this error results from creating a NativeEventEmitter on a module that hasn't been mocked
From [this thread.](https://github.com/facebook/jest/issues/2208#issuecomment-264733133)

Added a mock of `react-native-gutenberg-bridge`. That should fix the Travis error.
I'm not sure if it's a temporal workaround or a real fix.

hypest:
> I'm not sure if it's a temporal workaround or a real fix.

I opened https://github.com/wordpress-mobile/gutenberg-mobile/issues/210 just to have it as a note to check it out.

mzorz:
This is so cooooooollll!! ❤️

PR #205, opened 2018-10-29
Feature/port nextpage block

pinarol:
This is a simple PR that adds the [react-native-hr](https://github.com/Riglerr/react-native-hr) library to GB mobile. It will be used by the nextpage block.

### DIFF
--- a/package.json
+++ b/package.json
@@ -95,6 +95,7 @@
     "react-native-modal": "^6.5.0",
     "react-native-recyclerview-list": "git+https://github.com/wordpress-mobile/react-native-recyclerview-list.git#c073c0335ebe0398c85648ffeb11cbed4ee79229",
     "react-native-svg": "^6.5.2",
+    "react-native-hr": "git+https://github.com/Riglerr/react-native-hr.git#2d01a5cf77212d100e8b99e0310cce5234f977b3",
     "react-redux": "^5.0.7",
     "redux": "^3.7.2",
     "rememo": "^3.0.0",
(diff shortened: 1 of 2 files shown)


hypest:
👋 @pinarol , is there perhaps a GB-web side branch or PR to link with in this PR? I don't see GB-web ref being updated or a PR referenced so, not sure how to test this one? Thanks!

hypest:
This PR is a followup from https://github.com/wordpress-mobile/gutenberg-mobile/pull/209 so, let's wait for #209 before moving on with this one. Is that OK @pinarol ? Thanks!

pinarol:
right let's wait for #209 firstly

pinarol:
this is ready to review @hypest I updated the GB-web ref to include the example usage

etoledom:
Hey @pinarol - This is looking great! 🎉

I couldn't find the `Gutenberg` side PR, though.
Can we have a review on that side before merging this one?

Thank you!

hypest:
I notice that the Gutenberg side PR only touches `.native` files and as of https://github.com/WordPress/gutenberg/pull/11318 Travis will happily say everything's fines. So, the only way to test if the GB side PR will break mobile master is to try it manually (try using that GB branch against mobile master that is).

Now, we do actually know that the import on the GB side will break against current mobile master so, let's review and give the green light here and then quickly merge the GB side, update this PR's ref and merge it too. The total disruption to mobile master will be small. WDYT?

pinarol:
@hypest I agree, it looks like it won't be a problem as soon as I'm quick between PR merges. So once PR reviews are done I can just merge GB PR and then update this PR's ref and merge it too 👍

PR #209, opened 2018-10-29
Support for the nextpage block

pinarol:
Add support for nextpage block. Most of the changes reside [in this PR](https://github.com/WordPress/gutenberg/pull/11192).

## How has this been tested?

- [x]  Make sure the (web) GB build and webapp work normally
- [x]  Make sure the GB mobile app works normally and nextpage block is available for adding.
- [x]  Make sure the GB mobile demo page has a next page block added initially


etoledom:
Not sure if this is already figured out, but it looks like the travis errors are not related to this PR.
I followed these steps:

- Checkout `gutenberg-mobile` `master`
  - Tests pass ✅ 
- `cd gutenberg`
- `git checkout master` (update Gutenberg reference to master)
  - Test fail with this same error ❌

We will probably need to update the `gutenberg` reference in our master and solve these tests issue there first.

cc @hypest

hypest:
I tried the PR locally and can confirm that the tests get fixed if we create a new symlink, for the `notices` package. The change seems rather small so, I'd suggest we do it in this PR instead of a new one. WDYT @pinarol ?

pinarol:
thanks @etoledom for your suggestion, since the fix is very simple I just did it and got it out of our way as @hypest suggested :)

symlinked-packages/@wordpress/notices
```diff
@@ -0,0 +1 @@
+../../gutenberg/packages/notices/src
```
hypest: Since we're on it, let's also do the related symlink in the `symlinked-packages-in-parent` too, to be used in the web side setup, where the mobile code is a submodule to the parent Gutenberg code, cool?

pinarol: I just pushed a new commit for this and tested with "npm run test-mobile" on a separate Gutenberg repo after checking out this branch on its gutenberg-mobile submodule and tests ran OK. 👍

PR #212, opened 2018-10-30
[Android Aztec] - Fixes refresh issues on merging blocks 

daniloercoli:
### Description:

Fixes refresh issues in Android when merging blocks.

Moves #167 forward.

### Related PRs:

Aztec RN Wrapper PR: https://github.com/wordpress-mobile/react-native-aztec/pull/67 [Merged]

### Testing:

1. Make sure you have 2 blocks of the same type (paragraph x 2 or heading x2).
2. Place the caret at the beginning of the bottom block.
3. Press backspace.
4. Make sure the resulting block is as expected and refreshed properly.


hypest:
Hmm, doesn't seem to work for me when following the steps @daniloercoli . The second block completely disappears. I think it did work only once though thought my tests.

daniloercoli:
> The second block completely disappears

Not quite sure I've got the problem you're experiencing.
When merging 2 blocks the second block should disappear, and its content moved into the first (the one above).

hypest:
If it's any help, here's a video from my Pixel 2 XL where the second blocks just disappears on "backspace", along with its contents: https://cloudup.com/cSFnyuS3jpF

daniloercoli:
I think the problems with the caret that jumps to another block, and the other problem with the content of the 2nd block not being copied into the first, live already in `master`.  I've just tested the iOS side, and happens to have the same problems.   https://cloudup.com/ccxN4XcmrGp

Will open issues for those problems.

This PR just adds a delay on the `onContentSizeChanged` event on Android, but should not alter the content in any way.

hypest:
I think I have more info. Try to make changes to both blocks before merging them @daniloercoli .

hypest:
Verified that the merging behavior is not working OK for me on `master` too @daniloercoli , so, it's not a problem of this PR. Think is, I don't know how to then test the change in this PR. Any different steps or something different to pay attention to to verify these changes?

daniloercoli:
It should work without making changes to blocks.
Put the caret at the beginning of the 1st para block, and try to merge it with the Heading block above.

PR #214, opened 2018-10-31
Port Quote block step 1

daniloercoli:
Add a very basic initial support for quote blocks. 

GB side PR: 

## How has this been tested?

- [ ]  Make sure the (web) GB build and webapp work normally
- [ ]  Make sure the GB mobile demo page has a quote block added initially

There are still things need to be addressed, like adding the ability to create a block from toolbar.
Also the ability to add/remove the `cite` part is missing, as well as basic interactions "within" the block: move the cursor up/down between those 2 editors. 

Gutenberg Side PR: https://github.com/WordPress/gutenberg/pull/11311


koke:

daniloercoli:
This PR is very old, and we've recently agreed to re-use much of the web code as possible when porting new blocks to mobile. Experimenting this new way in the List block here: https://github.com/WordPress/gutenberg/pull/14249

Also the GB side of this PR was already closed.

PR #215, opened 2018-11-01
Experiment with reusing the gutenberg store in the block manager

Tug:
WIP


hypest:
Had a sneak peak in this PR and it seems that good things are coming, thanks @Tug !

Got to say though that it will be a bit hard to review it from our small mobile side experience so far so, please make an effort to add a PR description when this is ready for review that explains what components are added, short mention of their role and use and so on. Thanks!

PR #216, opened 2018-11-01
Added turbo-combine-reducers dependency to fix failed builds

marecar3:
With latest [PR](https://github.com/WordPress/gutenberg/pull/11255/files) on gutenberg, we have experienced failing builds on our end. This quick fix will catch up us with gutenberg and our builds will pass with success.


hypest:
Verified build fixed and app working as expected, nice work @marecar3!

Issue #192, opened 2018-10-24
Extend communication between RN and Host apps.


etoledom:
Following #182 , there are more instances where we need to communicate between RN and the Native side:

- [ ] Show media picker (RN -> Native)
- [ ] Image selected (Native -> RN)
- [ ] Image networking state [uploading, finished, error) (Native -> RN)

koke:
I had this split into two task in the backlog (picker, upload with progress). Do you think it makes sense to split in two, or are they too interrelated?

koke:
Moved the selection and upload parts to #206

PR #220, opened 2018-11-05
Image set via parent media library picker

hypest:
Addresses #192 for the Android side of the Bridge.

This PR adds a function to the bridge interface to trigger the media library image picker on the parent native app. The function provides a callback for the Bridge to invoke to send the selected image URL back into RN.

Depends on the GB repo PR (needs merge first): https://github.com/WordPress/gutenberg/pull/11511

To test (Android):

1. Use the `feature/integrate-gutenberg-image-set` branch from https://github.com/wordpress-mobile/WordPress-Android. [PR here](https://github.com/wordpress-mobile/WordPress-Android/pull/8553).
2. update all submodules at every level
3. Run `yarn wpandroid` at the wpandroid root to compile the native app, run Metro, install and run the mobile app in a single command
4. Open a GB post in wpandroid and add a new image block
5. Tap on the "MEDIA LIBRARY" button and select an image from the *media library*
6. App should update the block and show the image


hypest:
cc @etoledom as you might want to have a peak at the Bridge change proposal here. Check out https://github.com/wordpress-mobile/gutenberg-mobile/pull/220/files#diff-0ae1e2c19a7a43836c4694c61985e963R6

etoledom:
Using the JS callback looks like a good approach! 

I checked out the Android side to compare with iOS and it seems like the Media Pickers work in a similar way: providing the selected image via "delegate" callback. 

So I believe that the WPiOS side implementation will be quite similar too. ✅

PR #223, opened 2018-11-06
Fix blocks onMerging and onChange red screen errors

daniloercoli:
This PR makes sure the block is still available when calling an action over it from the native side.

If you type into a block, and than delete a block very fast via the toolbar, the block is not there anymore when the `onChange` call is received from the native side, and the JS code throws the error screen, since there weren't any checks in place to ensure block validity

Same happens `onMerge`, if you type and the fast merge a block.

Companion PR on GB is here https://github.com/WordPress/gutenberg/pull/11537
that removes the delay on `onChange` and makes sure the correct content is used on change, merging, or splitting.

**To test**
- Change the content of a block
- Delete the block immediately
- You should not see the Red Screen

Another test is with merging:
- Try to add/remove/merge block(s)
- Even if the content of the merged block (s) is not correct, the app should not crash


daniloercoli:
GB side was merged! Ready for review @hypest !

src/block-management/block-manager.js
```diff
@@ -159,6 +159,11 @@ export default class BlockManager extends React.Component<PropsType, StateType>
 	mergeBlocks( clientId: string, forward: boolean ) {
 		// find currently focused block
 		const focusedItemIndex = this.getDataSourceIndexFromClientId( clientId );
+		if ( focusedItemIndex === -1 ) {
+			// do nothing if it's not found.
```
hypest: Let's expand the comment a bit, to include how can this happen (update form native arriving late and block already been deleted). Same with the other comments below in the diff.

We'll need to pay attention to this while we will be moving to the new stores (GB ones) so, a descriptive comment here will help us there. Thanks!

daniloercoli: I've updated those comments. 👍

PR #224, opened 2018-11-07
Fix blocks merging UI refresh problem

daniloercoli:
Fix the problem where merging 2 blocks together doesn't refresh the native UI properly.

To test this:
- Start the demo app
- modify the content of a para block
- merge it with another para or heading block

The new resulting block should contain the correct text.

GB side PR: https://github.com/WordPress/gutenberg/pull/11576

Merge this after #223


marecar3:
Hey @daniloercoli ,  probably we also want to update yarn.lock because there is a change in package.json?

PR #225, opened 2018-11-07
Detect content changed

hypest:
Addresses part of #182 

Fit the Bridge interface with a extra boolean to denote if the returned html is changed compared to the original html the editor was started with. Parent app can then optimize for not the case where the content didn't change (e.g. not marking the post as changed, save it and so on).

### Changes in this PR

* State adopted the `initialHtmlHash` member. Reducer code adapted to support it.
* Html change detection is using `md5` hashes to compare the html versions
* `BlockType` and `StateType` moved to own file. Similar with a couple of state util functions.
* Bridge middleware compares html hashes on BLOCK.SERIALIZE_ALL, to report to parent app

### To test A:

App should compile and run as usual. No visible feature change.

### To test B (Android only):

* Use the `gb/detect-content-change` branch from wpandroid. Usual `submodule update`, `yarn install`, `yarn start:reset` dance to launch Metro.
* `yarn wpandroid` to launch wpandroid
* Open a Gutenberg post that has no local changes. Don't make any changes and go "Back".
* Notice the post retaining the status it had before, without marking the post as having local changes.

cc @etoledom , @pinarol as this includes Bridge changes to pass the "changed" flag.

### DIFF
--- a/src/store/reducers/index.js
+++ b/src/store/reducers/index.js
@@ -5,10 +5,11 @@
 
 import { find, findIndex, reduce } from 'lodash';
 
+import { html2State } from '../';
+
 import ActionTypes from '../actions/ActionTypes';
-import type { StateType } from '../';
+import type { StateType } from '../types';
 import type { BlockActionType } from '../actions';
-import { parse } from '@wordpress/blocks';
 
 function findBlock( blocks, clientId: string ) {
 	return find( blocks, ( obj ) => {
@@ -32,7 +33,7 @@ function insertBlock( blocks, block, clientIdAbove ) {
 }
 
 export const reducer = (
-	state: StateType = { blocks: [], refresh: false },
+	state: StateType = { blocks: [], initialHtmlHash: '', refresh: false, fullparse: false },
 	action: BlockActionType
 ) => {
 	const blocks = [ ...state.blocks ];
@@ -72,7 +73,7 @@ export const reducer = (
 			// Otherwise merge attributes into state
 			block.attributes = nextAttributes;
 
-			return { blocks: blocks, refresh: ! state.refresh };
+			return { blocks: blocks, initialHtmlHash: state.initialHtmlHash, refresh: ! state.refresh };
 		}
 		case ActionTypes.BLOCK.FOCUS: {
 			const destBlock = findBlock( blocks, action.clientId );
@@ -83,7 +84,7 @@ export const reducer = (
 			}
 
 			destBlock.focused = true;
-			return { blocks: blocks, refresh: ! state.refresh };
+			return { blocks: blocks, initialHtmlHash: state.initialHtmlHash, refresh: ! state.refresh };
 		}
 		case ActionTypes.BLOCK.MOVE_UP: {
 			if ( blocks[ 0 ].clientId === action.clientId ) {
@@ -94,7 +95,7 @@ export const reducer = (
 			const tmp = blocks[ index ];
 			blocks[ index ] = blocks[ index - 1 ];
 			blocks[ index - 1 ] = tmp;
-			return { blocks: blocks, refresh: ! state.refresh };
+			return { blocks: blocks, initialHtmlHash: state.initialHtmlHash, refresh: ! state.refresh };
 		}
 		case ActionTypes.BLOCK.MOVE_DOWN: {
 			if ( blocks[ blocks.length - 1 ].clientId === action.clientId ) {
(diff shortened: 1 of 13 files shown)


koke:
I wonder why not use the same approach that the [editor store uses](https://github.com/WordPress/gutenberg/blob/16a718a4bf359c53f0fb9c3626b08e2434a6fd7d/packages/editor/src/store/reducer.js#L242) with [withChangeDetection](https://github.com/WordPress/gutenberg/tree/master/packages/editor/src/utils/with-change-detection).

I know we've used md5 hashes in the past because it was convenient and we didn't have a good way to keep track of changes, but having redux in place and a util that already does this, it seems like a better choice and less wasteful.

hypest:
> I wonder why not use the same approach that the editor store uses with withChangeDetection.

I guess no particular reason to _avoid_ `withChangeDetection` other than it seems to depend on assumptions on the Reducer and the State schema itself that seem not true for the mobile project atm. We currently have a couple of state fields (`refresh`, `fullparse` that we use for _signaling_ instead of state per se. I think those will trigger the `isDirty` flag without the html content actually changing.

Then again, I can be wrong but, `withChangeDetection` seems to detect changes only between two consecutive states (one action). So, even if after a few actions the final html is the same as the original, `isDirty` will have been triggered. Can you confirm?

koke:
It looks like you can configure which actions to ignore. I think it would stay dirty if you end up with the same content after a few edits but that’s ok with me

hypest:
I'd be careful to avoid changing the interface with the main app at this point. We have content-based change detection in Aztec-Android atm and I'd lean on keeping that until we re-attack this problem when we port the GB-web's store, soon.

That said, @diegoreymendez let me ask, what's the current pattern on iOS and Aztec to report if the post has changes or not? Can you share some info on that? Thanks!

koke:
After some chat, I'm fine with the approach for the time being, although I'd like to revisit this in the future when we have better integration with Gutenberg in the store layer and some more time to spare.

src/store/reducers/index.js
```diff
@@ -72,7 +73,7 @@ export const reducer = (
 			// Otherwise merge attributes into state
 			block.attributes = nextAttributes;
 
-			return { blocks: blocks, refresh: ! state.refresh };
+			return { blocks: blocks, initialHtmlHash: state.initialHtmlHash, refresh: ! state.refresh };
```
koke: Nitpicky, but for all these updates we can use the spread operator and property shorthand:

```suggestion
			return { ...state, blocks, refresh: ! state.refresh };
```

hypest:
> I'd like to revisit this in the future when we have better integration with Gutenberg in the store layer and some more time to spare.

Similar to other cases, I opened a ticket to keep it in our radar: https://github.com/wordpress-mobile/gutenberg-mobile/issues/227

src/store/reducers/index.js
koke: I think it's easier to read and understand when you only make explicit what is changing

hypest: Done with 5180b0c.

src/store/utils.js
```diff
@@ -0,0 +1,35 @@
...
+import md5 from 'md5';
+import { Store } from 'redux';
+
+import type { BlockType, StateType } from './';
+
+export function html2State( html: string ) {
+	const blocksFromHtml = parse( html );
+	const reserialized = blocks2html( blocksFromHtml );
+	const hash = md5( reserialized );
+	const state: StateType = {
```
pinarol: just an idea: we can make use of ES6 classes for State to make use of constructors, getter&setters, for example every time the html is set we can update blocks and initialHtmlHash inside the class. or we can define a constructor that is taking html as input and update blocks and initialHtmlHash in that constructor. again this is just an idea current version is also ok for me.

hypest:
Thanks for the review @pinarol !

> Test 2:
>
>    Open a blog post and change its title

Good idea of a test! You might also want to try editing something inside the post content since, as it currently stands, the title itself is a different input field, not rendered by the RN app.

src/store/utils.js
hypest: Let me see if I understand the proposal: Will that class update its internal vars?

If yes, I'm not sure that's what we want for Redux. My understanding is that the reducer is supposed to return a new "state" object and avoid mutating the old one, isn't it? In that case, I'm not sure a State class would be too useful since we would essentially only use its constructor, and that's pretty much what `html2State` performs.

Let me know if there is something in the proposal that I'm missing, thanks!

hypest: Oh, and btw, we might don't want to spend too much time iterating on this since we will be using GB's stores soon #226 anyway.

pinarol:
> Thanks for the review @pinarol !
> 
> > Test 2:
> > Open a blog post and change its title
> 
> Good idea of a test! You might also want to try editing something inside the post content since, as it currently stands, the title itself is a different input field, not rendered by the RN app.

Right! totally missed that! Thanks for pointing out.

hypest:
Thanks for the thorough testing @pinarol ! Will go ahead and merge now 🙇‍♂️!

src/store/utils.js
pinarol: You are right about mutating. we could figure it out though maybe there's an easy way to clone a class object. I'll investigate this further on a more free time and If I discover a useful practice I will post a p2 about it.

PR #228, opened 2018-11-08
Workaround to make yarn ios work with Xcode 10

koke:
It seems the issue is that Xcode is not automatically downloading required dependencies, so this adds a preios step to download them before trying to build the app.

See https://github.com/facebook/react-native/issues/20774

I don't have Xcode 9 installed, but I believe it should not affect building with 9.


hypest:
FWIW, I tried this PR with Xcode 9 and worked fine.

> $ xcodebuild -version
> Xcode 9.4.1
> Build version 9F2000

PR #232, opened 2018-11-13
Inserter button in toolbar

hypest:
Addresses part of #58 

Adds a bottom toolbar button to trigger the Inserter. This is the minimum needed in order to be able to add a block when no block is selected or there are no blocks at all (e.g. when a new post, which is part of #182)

Changes in this PR:
1. Allow for empty string as initial data from the parent app
2. Inserter button in the bottom toolbar
3. Removal of inserter button from the inline toolbar

Scope not handled by this PR:
* Styling of the Inserter button to match https://github.com/wordpress-mobile/gutenberg-mobile/issues/58

screenshot-1542108851184

To test 1:
* Run the app. No blocks are focused yet.
* Tap on the bottom toolbar Inserter button
* Select a block from the Inserter bottom sheet
* Scroll at the beginning of the content and notice the new block is there

To test 1:
* Run the app. No blocks are focused yet.
* Tap on the Image block to focus it
* Tap on the bottom toolbar Inserter button
* Select a block from the Inserter bottom sheet
* Notice the new block is inserted right after the Image block

### DIFF
--- a/src/block-management/block-toolbar.js
+++ b/src/block-management/block-toolbar.js
@@ -1,11 +1,28 @@
+/**
+ * @format
+ * @flow
+ */
+
 import React, { Component } from 'react';
 import { View } from 'react-native';
+import { IconButton } from '@wordpress/components';
 import { BlockFormatControls, BlockControls } from '@wordpress/editor';
+import { __ } from '@wordpress/i18n';
 
-export default class BlockToolbar extends Component {
+type PropsType = {
+	onInsertClick: void => void,
+};
+
+export default class BlockToolbar extends Component<PropsType> {
 	render() {
 		return (
 			<View style={ { height: 50, backgroundColor: '#DCDCDC', flexDirection: 'row' } }>
+				<IconButton
+					className="block-list-appender__toggle"
+					label={ __( 'Add block' ) }
+					icon="insert"
+					onClick={ this.props.onInsertClick }
+				/>
 				<BlockControls.Slot />
 				<BlockFormatControls.Slot />
 			</View>
--- a/src/block-management/block-toolbar.js
+++ b/src/block-management/block-toolbar.js
@@ -18,7 +18,6 @@ export default class BlockToolbar extends Component<PropsType> {
 		return (
 			<View style={ { height: 50, backgroundColor: '#DCDCDC', flexDirection: 'row' } }>
 				<IconButton
-					className="block-list-appender__toggle"
 					label={ __( 'Add block' ) }
 					icon="insert"
 					onClick={ this.props.onInsertClick }
--- a/src/app/App.js
+++ b/src/app/App.js
@@ -28,7 +28,7 @@ export default class AppProvider extends React.Component<PropsType, StateType> {
 			store:
 				typeof props.initialData === 'object' ?
 					props.initialData :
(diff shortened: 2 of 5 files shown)


src/block-management/block-toolbar.js
```diff
@@ -1,11 +1,28 @@
...
 
-export default class BlockToolbar extends Component {
+export default class BlockToolbar extends Component<PropsType> {
 	render() {
 		return (
 			<View style={ { height: 50, backgroundColor: '#DCDCDC', flexDirection: 'row' } }>
+				<IconButton
+					className="block-list-appender__toggle"
+					label={ __( 'Add block' ) }
+					icon="insert"
```
marecar3: The only concern here would be that we didn't share some constant with gutenberg project, so in the future, if someone changes those strings, it will break gutenberg-mobile.

src/block-management/block-toolbar.js
```diff
@@ -1,11 +1,28 @@
...
+	onInsertClick: void => void,
+};
 
-export default class BlockToolbar extends Component {
+export default class BlockToolbar extends Component<PropsType> {
 	render() {
 		return (
 			<View style={ { height: 50, backgroundColor: '#DCDCDC', flexDirection: 'row' } }>
+				<IconButton
+					className="block-list-appender__toggle"
```
marecar3: Not sure if we need className and what value it has for us?

etoledom:
One detail to notice:

In Gutenberg-Web, when I open an existing post with content, and add a new block with the `+` button at the top right corner, the new block is appended at the end of the document.

Should we try to copy that behavior in this PR or it is not important for now?
cc @hypest .

src/block-management/block-toolbar.js
hypest: Ah, I left that in to tweak the button's styling later but, I shall remove it from this PR since the styling is not handled here.

hypest:
> In Gutenberg-Web, when I open an existing post with content, and add a new block with the + button at the top right corner, the new block is appended at the end of the document.
> 
> Should we try to copy that behavior in this PR or it is not important for now?

Sounds reasonable, I'll add the logic, thanks @etoledom 👍

hypest:
Changed the code to add the new block at the end of the list with fb4dfda.
While at it, added a scroll animation when on Android. See d86dc09 and its commit comment.

src/block-management/block-toolbar.js
hypest: Done with e8df805.

hypest:
Addressed review comments, ready for another pass @marecar3 , @etoledom , thanks!

src/block-management/block-toolbar.js
marecar3: Sounds like a good plan :)

etoledom:
Looks great to me ✅ 

I envy the scroll to the new block on Android. Let's add a ticket to implement that on iOS :)

marecar3:
Looks good to me ✅

hypest:
> I envy the scroll to the new block on Android. 

Heh, that's a happy accident that it's quick to do on the DataSource. I don't think that the current way of doing it will survive for long. We need to figure out the proper way to do UI adjustments like this (and anims).

PR #239, opened 2018-11-15
Fix Android back button not dismissing the inserter's block-type picker

mzorz:
This PR adds the behavior to dismiss the block-type picker modal when the user taps `back` on Android without needing to touch outside the modal or swipe it down.

First reported by @daniloercoli .

To test:
1. open the demo app on Android
2. tap on the `+` inserter button
3. tap `back` 
4. observe the modal is dismissed.


daniloercoli:
LGTM!

Issue #240, opened 2018-11-15
Expose focus() and blur() API in order to get working keyboard dismiss functionality


marecar3:
Currently, we can't use RN `Keyboard.dismiss()` API as we are missing `focus()` and `blur()` API on the Aztec level.

koke:
Added this to the alpha milestone, but not a blocker. If it doesn't make it in time, we'll disable the button to hide the keyboard until this is ready

koke:
This one seemed to be working fine, should we close it?

PR #241, opened 2018-11-15
DO NOT MERGE - Implemented keyboard hide button

marecar3:
Added keyboard hide button in Toolbar. 

**Note**: Currently we can't dismiss keyboard as we are missing some logic on the Aztec level which will expose `focus()` and `blur()` methods. That should be done in [separate ticket ](https://github.com/wordpress-mobile/gutenberg-mobile/issues/240)and probably @mzorz can look at it.

keyboard_hide_button

### DIFF
--- a/src/block-management/block-manager.js
+++ b/src/block-management/block-manager.js
@@ -6,7 +6,7 @@
 import React from 'react';
 import { xorBy, isEqual } from 'lodash';
 
-import { Platform, Switch, Text, View, FlatList, KeyboardAvoidingView } from 'react-native';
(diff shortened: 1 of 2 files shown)


marecar3:
@daniloercoli has a good point regarding this one. As it is working with block `code` we can merge it into `feature/toolbar_component_ui`, and in the future, if we don't have `focus()` and `blur()` implemented on alpha, we can just hide the button and wait for the implementation. @mzorz

marecar3:
Thanks @marecar3

koke:
I just tested this one on an iPhone 8 and I'm not seeing the toolbar when the keyboard is up. I noticed this is against `feature/toolbar_component_ui` so you might want to target `master` instead after the toolbar changes are merged.

PR #242, opened 2018-11-15
Send `changed` to the native bridge if html changed since last time

Tug:
https://github.com/wordpress-mobile/gutenberg-mobile/pull/226 caused a regression in the js to native bridge. We're not sending the `changed` parameter anymore when calling `provideToNative_Html`. This PR fixes it.

> Note: it's currently causing a problem in https://github.com/wordpress-mobile/WordPress-iOS/pull/10461 when you tap "Publish" on parent app it crashes.

### DIFF
--- a/package.json
+++ b/package.json
@@ -91,7 +91,6 @@
     "jsc-android": "224109.x.x",
     "jsdom-jscore": "git+https://github.com/iamcco/jsdom-jscore-rn.git#6eac88dd5d5e7c21ce6382abde7dbc376d7f7f59",
     "jsx-to-string": "^1.3.1",
-    "md5": "^2.2.1",
     "memize": "^1.0.5",
     "moment": "^2.22.1",
     "moment-timezone": "^0.5.16",
--- a/src/app/AppContainer.js
+++ b/src/app/AppContainer.js
@@ -1,12 +1,13 @@
 /** @flow
  * @format */
 
-import MainApp from './MainApp';
 import React from 'react';
 import { parse, serialize } from '@wordpress/blocks';
 import { withDispatch, withSelect } from '@wordpress/data';
 import { compose } from '@wordpress/compose';
 import RNReactNativeGutenbergBridge from 'react-native-gutenberg-bridge';
+
+import MainApp from './MainApp';
 import type { BlockType } from '../store/types';
 
 type PropsType = {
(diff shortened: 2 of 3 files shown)


pinarol:
cc @hypest

hypest:
If I understand the code correctly, it detects the change compared to the last time it was _sent_ up to the parent. I think that's not the intended logic and it's not what it was in place before.

Because it seems that parsing and then serializing the blocks again does not produce the original html, we need to check if the html we pass back to the parent is the same as the html originally passed down to the RN app. The previous solution was keeping the md5 hash of the original html so to compare it with the current and report "changed" or not.

I tested with the wpandroid integration and indeed, the post is reported as "changed" with the current PR (`master` by now), I'm guessing because `lastHtmlSent` starts undefined anyway? Let's create a new PR to fix this by comparing to the original html passed down. WDYT?

pinarol:
That's right @hypest I missed this because this flag is not utilized in iOS side while closing the editor. There's an internal logic to decide if content has changed so it is not using the Javascript flag while closing. I am just setting the last version of html and it is taking care of the rest(to give more detail, it creates a revision on opening and compares the new content with that revision), if there's no change it just closes the editor without extra prompts. 

The only use case of such a flag on iOS side is enabling/disabling the Publish button. Currently the publish button is always active because we don't have the capability yet on the Gutenberg side that informs us the every editing action unlike Aztec. So it turns out currently I am not actually utilizing that flag. 🤔 (I mean I am using it but it is ignored.)

hypest:
> There's an internal logic to decide if content has changed so it is not using the Javascript flag while closing

Where does that logic live? Is it the parent app, the bridge or the RN layer?

Since what we're dealing with here are involuntary changes by the editing component itself, the logic needs to live in the RN side and the initial "snapshot" to compare against needs to take into account those self-inflicted changed. That's why the original implementation needed to serialize the blocks again before it computed the "initial hash". See here in the original PR: https://github.com/wordpress-mobile/gutenberg-mobile/pull/225/files#diff-969b5637d99e6ae15745bc85e7f3519fR16

cc @Tug

pinarol:
> Where does that logic live? Is it the parent app, the bridge or the RN layer?

@hypest  In the parent iOS app. If you want to dig more into it you can check out the class AbstractPost, there's a hasLocalChanges method in it. It simply compares the latest content with its original content and returns the result, this has always been this way. Whenever user makes a change on aztec editor the lates content in the post is updated, "hasLocalChanges" is computed where necessary. 

So what happens is:
- We open the Gutenberg editor with the original content
- Tap close button
- Request html from gutenberg side
- Gutenberg html arrives with a callback, we update the latest content in the AbstractPost
- Proceed with close operation
- Close operation inside checks the ”hasLocalChanges” flag in the post and and does what’s necessary, if it is true for example it prompts the user if they want to keep/discard changes.

There are huge differences between iOS and Android apps both in terms of user experience and implementation. So I am having hard time saying both sides need the exact same flag. I think we can move this conversation to the #mobile-gutenberg

PR #243, opened 2018-11-15
Reestablish possibility to pass initial data from a parent app

etoledom:
This PR reestablish possibility to pass initial data from a parent app.

As a bonus, on iOS it's now possible to send an empty string as initial data and we will get an empty post instead of the example post. (Not tested on Android)

**To test:**
- Check that the example app displays as always.

**On WPiOS:**
- Checkout https://github.com/wordpress-mobile/WordPress-iOS/pull/10467
- Check that posts from the post list open the proper post data.
- Check that a new post opens an empty post.

### DIFF
--- a/src/app/App.js
+++ b/src/app/App.js
@@ -23,12 +23,12 @@ type PropsType = {
 };
 
 const AppProvider = ( { initialData }: PropsType ) => {
-	if (initialData === undefined) {
+	if ( initialData === undefined ) {
 		initialData = initialHtml;
 	}
 	return (
 		<AppContainer initialHtml={ initialData } />
 	);
-}
+};
 
 export default AppProvider;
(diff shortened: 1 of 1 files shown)


hypest:
> (Not tested on Android)

🎉 Tested those on Android and they work:
✅ Check that posts from the post list open the proper post data
✅ Check that a new post opens an empty post

Tug:
We could probably update AppContainer to handle initialHtml changes, curently it won’t react to changes

etoledom:
Thank you all ! ✨ 

@Tug - I just check and it is possible to reset initial props (I thought it wasn't possible).
Anyway it seems to just work if we pass a new value different to the one we gave it initially.

Since it's already working for us on Android and iOS, I'd prefer if we check this special case when (and if) we need it in a future point.

I hope that's fine! :)

PR #244, opened 2018-11-15
Recover insert block at the end, if no block is selected.

etoledom:
This PR recovers the functionality that insert a new block at the end, if no block is selected.

create_bottom_demo

To test:
- Remove all blocks from the post.
- Add a new block.
- [x] Check that it's created without problems.
- Switch between HTML and Visual mode to deselect the previously created block.
- Create a new block. 
- [x] Check that it's inserted at the end.
- Select the block at the top.
- Create a new block.
- [x] Check that it's inserted after the first block

### DIFF
--- a/src/app/AppContainer.js
+++ b/src/app/AppContainer.js
@@ -58,7 +58,9 @@ class AppContainer extends React.Component<PropsType> {
 	};
 
 	createBlockAction = ( clientId, block ) => {
-		this.props.onInsertBlock( block, this.props.selectedBlockIndex + 1, this.props.rootClientId );
+		const indexAfterSelected = this.props.selectedBlockIndex + 1;
+		const insertionIndex = indexAfterSelected || this.props.blocks.length;
+		this.props.onInsertBlock( block, insertionIndex, this.props.rootClientId );
 	};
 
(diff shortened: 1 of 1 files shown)


mzorz:
> Remove all blocks from the post

1. start the demo
2. remove each block by hand (by selecting one block and then tapping on the trash icon each time)
3. when you try to remove the last item, a red screen appears:

screen shot 2018-11-16 at 13 33 25

mzorz:
ok as seen the problem reported here is not related to this PR and is being addressed in https://github.com/WordPress/gutenberg/pull/11988 👍

etoledom:
Thank you @mzorz !

hypest:
👋 @mchowning , noticed this merged PR was added to the v1.10 milestone. Not sure, maybe you were meaning to move a different one into the milestone?

mchowning:
> noticed this merged PR was added to the v1.10 milestone. Not sure, maybe you were meaning to move a different one into the milestone?

I certainly hope so! Thanks for catching that. I removed the 1.10 milestone label. I can't tell what, if any, label was applied before though, so it is currently unlabeled.

PR #247, opened 2018-11-16
Add yarn scripts to build the release bundle

koke:
As part of some research about the integration, I added these scripts to build release bundles.

- `yarn bundle`
- `yarn bundle:ios`
- `yarn bundle:android`

These will build the JS bundle and source map in `build/`, which is ignored by git


koke:
No, they're not used yet, I didn't get anywhere with that, but I thought those would be useful for experimenting

PR #248, opened 2018-11-16
Fix Picker touch handling, avoiding to have to touch twice if unfocused

mzorz:
This fixes an issue with the inserter's Block-Type picker that would not respond to touch at first touch when the focus was taken by a PlainText-backed block such as `Code` or `More` (implemented with InputText in React Native).

Before:
before
(it's difficult to see, but believe me I had to tap twice :D - you'll be able to reproduce following the steps below).

After:
after

**To test:**
1. open the demo app
2. tap into a `Read More`  or `Code` block
3. observe the cursor appears and keyboard goes up
4. now tap on the inserter icon `+` 
5. see the block type picker appear
6. tap on any icon
7. observe it responds to touch at first touch.

### DIFF
--- a/src/block-management/block-picker.js
+++ b/src/block-management/block-picker.js
@@ -52,7 +52,7 @@ export default class BlockPicker extends Component<PropsType, StateType> {
 					</View>
 					<View style={ styles.lineStyle } />
 					<FlatList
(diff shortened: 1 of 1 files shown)


daniloercoli:
LGTM!

PR #250, opened 2018-11-16
isSelected handling on PlainText-backed blocks

### LINKED ISSUE #113: Inserter - cursor not following along focus when inserting a new block
Coming from https://github.com/wordpress-mobile/gutenberg-mobile/pull/105#pullrequestreview-145980349

> If I'm editing a text block and I add another, the "focus" switches to the newly inserted block, but the cursor stays in the previous block. I'm not sure if we're worrying about that at this point, but leaving a note here.

~**ToDo:**~
- [x] ~#237 handle focus on RichText-based blocks~
- [x] ~#250 handle focus on other blocks~

**EDIT: 
ToDo:**
Recreating the ToDo list here after [migration to FlatList](https://github.com/wordpress-mobile/gutenberg-mobile/pull/256#pullrequestreview-177652478) 
- [ ] handle focus on RichText-based blocks
- [x] #250 handle focus on PlainText-based blocks

mzorz:
Fix #113 

This updates the GB hash to use the `isSelected` prop as an indication for the PlainText-backed blocks (currently the `More` and `Code` blocks) to gain focus when such a prop is `true`.

Related Gutenberg PR: https://github.com/WordPress/gutenberg/pull/11989

Before, the focus would not be gotten when a new "PlainText"-backed block was inserted:
cursor_after_insert_okno

Now, the cursor appears blinking after insertion:
cursor_after_insert_ok

To test:
1. make sure to select a block (tap on a block and see the up/down arrows appear, etc)
2. now tap on the `+` inserter icon
3. choose one of Code or More blocks (the only ones that currently are backed by `PlainText `implementation)
4. observe the new block is inserted, the block is shown as selected (it has the up/down arrows and such) and the cursor starts blinking there (that is, it got the focus).


daniloercoli:
LGTM once the companion Gutenberg PR is merged.

PR #253, opened 2018-11-17
Revert "isSelected handling on PlainText-backed blocks"

mzorz:
Reverts wordpress-mobile/gutenberg-mobile#250


mzorz:
Tested that this branch works alright  (git clone and git submodule init, git submodule update and then running it works as expected). The merge broke master as GB hash wasn't properly updated.

Reverting :shipit: 🔙

PR #254, opened 2018-11-17
Update GB hash and fix matching package version for simple-html-tokenizer

mzorz:
Get #250 back into `master` after the revert  in #253.

This PR fixes the GB hash and the package.json file.

### DIFF
--- a/package.json
+++ b/package.json
@@ -113,7 +113,7 @@
     "refx": "^3.0.0",
     "rememo": "^3.0.0",
     "shallowequal": "^1.0.2",
-    "simple-html-tokenizer": "^0.5.1",
+    "simple-html-tokenizer": "^0.4.1",
     "tinycolor2": "^1.4.1",
     "turbo-combine-reducers": "^1.0.2",
     "tannin": "^1.0.1"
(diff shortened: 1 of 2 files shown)


mzorz:
In order to bring the latest Gutenberg commit into our repo, in this PR I changed the following in `package.json`:
```
"simple-html-tokenizer": "^0.5.1",
```
for 
```
"simple-html-tokenizer": "^0.4.1",
```
as that's the version as declared in Gutenberg's `package.json`.

While this makes it run and is at least not a blocker for development, I noticed later on that this breaks tests with the following error:

```
({"Object.<anonymous>":function(module,exports,require,__dirname,__filename,global,jest){import EventedTokenizer from './evented-tokenizer';
                                                                                             ^^^^^^
    
    SyntaxError: Unexpected token import

```

I've been looking into the latest Gutenberg PRs to see if anything around that changed, so I added (this change locally) the following key as observed in [this other PR](https://github.com/WordPress/gutenberg/pull/11771/files#diff-6b0584709544530e1ccbee1bf679be9aR17): 
```
transformIgnorePatterns: [ '<rootDir>/node_modules/(?!(simple-html-tokenizer)/)' ],
```
to `jest.config.js`. That makes the error go away and turn into this other error:


```
gutenberg-mobile/node_modules/react-native/jest/mockComponent.js:20
        static displayName = 'Component';
                           ^
    
    SyntaxError: Unexpected token =
      
```

I tried then to understand what was going on... with not much luck.

cc @hypest for help ⛑ - wdyt? can we merge and continue try fix tests in parallel in different PR, as this contains code back from #250 that I need to continue the work on focus?

mzorz:
Closing in favor of #257 

With the update to Gutenberg's `master` there ☝️ , the function introduced in #250 will automatically start working in mobile-gutenberg

PR #255, opened 2018-11-19
Create new podspec so WPiOS can use Gutenberg directly from git

koke:
I moved the Podspec to the top level, renamed it to Gutenberg, and add some tweaks so it picks up the right versions from `package.json`.

The main difference is that now the built JS is included in the repo inside `bundle` so the apps don't have to worry about building JS (and requiring the whole toolchain). This means we'll have to start doing release management. To make this easier, every time you run `yarn version` to bump the package version, it will automatically rebuild the bundle and include it in the version bump commit.

WPiOS PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/10487


koke:
This has been tested and approved as part of https://github.com/wordpress-mobile/WordPress-iOS/pull/10487

PR #258, opened 2018-11-20
Allow node v10 engine in package.json

Tug:
This add v10 and above to the list of allowed node engines to work with this project.

We haven't yet defined a fixed version of node to use, so in the meantime let's be permissive and just exclude the engines we know don't work (v9).


daniloercoli:
Tested on both Android and iOS. LGTM!

PR #260, opened 2018-11-20
Wire onReplace on blocks splitting

daniloercoli:
This PR adds the ability to replace a block when necessary on blocks splitting logic.

To properly test this PR is necessary to follow the instructions below:
- Add the following code `before = null;` right above
https://github.com/WordPress/gutenberg/pull/12125/files#diff-43430773cad44b3406827c3db7bc9d7fR75
to make the current `before` value null
- Then start the demo app
- Put the caret in the middle of a para block and tap enter
- The block should be splitted and the content before the caret should be gone


src/block-management/block-manager.js
```diff
@@ -366,6 +377,9 @@ export default class BlockManager extends React.Component<PropsType, StateType>
 						this.insertBlocksAfter.bind( this )( value.item.clientId, blocks )
 					}
 					mergeBlocks={ this.mergeBlocks }
+					onReplace={ ( blocks ) =>
+						this.onReplace.bind( this )( value.item.clientId, blocks )
```
Tug: You shouldn't need to bind to `this` here, just `this.onReplace( value.item.clientId, blocks ) should work

PR #263, opened 2018-11-21
Adding doc for making new releases

koke:
Document the changes in #247 and #255 to generate new releases with `yarn version`

### DIFF
--- a/docs/Releasing.md
+++ b/docs/Releasing.md
@@ -1,0 +1,30 @@
+# Making a release
+
+The `bundle` directory contains the production version of the project's Javascript. This is what the WordPress apps use to avoid having to build Gutenberg.
+
+You can rebuild those files at any time by running
+
+```
+yarn bundle
+```
+
+This is useful in case you want to test some changes in the apps before pushing a new version.
+
+The recommended method is that you bump the version when you do that. To publish a new version of gutenberg-mobile you just need to run:
+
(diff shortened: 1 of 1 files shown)


docs/Releasing.md
```diff
@@ -0,0 +1,30 @@
...
+
+The `bundle` directory contains the production version of the project's Javascript. This is what the WordPress apps use to avoid having to build Gutenberg.
+
+You can rebuild those files at any time by running
+
+```
+yarn bundle
+```
+
+This is useful in case you want to test some changes in the apps before pushing a new version.
```
hypest: I'm assuming that testing some changes cannot be done just by producing the bundle so, can we make it slightly more clear here? Something like `...in case you want to take the bundle and use it in the apps before pushing...` or similar. Thanks!

koke: Ah, good catch, that's really not clear enough. How about...

> This is useful in case you want to use an unreleased version of the bundle in the apps. For instance, on a PR that's a work in progress, you might want to include to a specific `gutenberg-mobile` branch in the apps with an updated bundle so reviewers can see the latest changes before approving them (and releasing a new version)

hypest: That version makes sense to me @koke 👍

koke:
Ready for another look

PR #265, opened 2018-11-21
Updated react-native-recyclerview-list hash to include selection fix

### LINKED ISSUE #135: Tapping into RichText doesn't move the cursor properly on Android
It seems that there are problems with the way the tap event is intercepted by the main list / block bottom toolbar / RichText component, that does result in weird behavior when you try to move the cursor to another part of the text (into the same RichText item) or when you try to cut&paste content.

- I found out that long tapping into the field doesn't' show the Cut&Paste menu.
- Also double tapping over a word doesn't select it.
- If you tap to move the cursor to a different position, the cursor is not moved but the bottom toolbar just toggle its state.

https://cloudup.com/cT2GShQAyY8

Above is a video that show how the Cut&Paste acts on the current version on a Nexus 5X.

Edit: I've just noticed that the video does't show the taps. 🤕  So it's hard to follow my interactions.

### LINKED ISSUE #267: Tapping into PlainText doesn't allow to select text on Android
Creating issue here to keep track of scenarios tested and make a specific case for the solution built in PR #265.

 Trying Gutenberg as is, realized that it was difficult (or impossible) to select text on PlainText-backed blocks (that is currently Code and More blocks).

The caret would appear upon touching, but not the context menu to select / copy / paste.

gutenberg_plain_text_selection_not_ok

Leaving information about the experiments done here:

1. I noted that just opening the demo app, rotating the device to horizontal and then rotating it back to vertical will make the issue go away - try doing it and then selecting text on any `TextInput` becomes possible.

gutenberg_as_is_rotate

2. so as a workaround, it seems the component needs to re-render itself, for that I tried a hack as described here https://github.com/facebook/react-native/issues/9958#issuecomment-365624780  that simply sets the width of the InputText to 99% and then after having the component attached to the window we reset the state to force a redraw. This effectively makes the field's content selectable again.

3. Back to the context of using `TextInput` in `RecyclerViewList`, I found out the workaround does *not work* for the elements that are *not actually drawn on the screen when the list is shown* (that is, the elements in the RecyclerViewList that are off the screen on startup).  For example, try showing the keyboard and letting it up. Now reload the app, and try scrolling down the list and selecting  one of the items that was “behind” the keyboard. You’ll see the text cannot be selected there.

One possible solution is to listen into the the adapter’s `onViewAttachedToWindow(viewholder)` callback and try to refresh the EditText there.

mzorz:
Fixes #267  for `PlainText` blocks on Android.

[EDITED]
~was: "fixes #135  for both `RichText` and `PlainText` blocks on Android", updated to reflect as per https://github.com/wordpress-mobile/gutenberg-mobile/pull/265#issuecomment-440796735.~

See https://github.com/wordpress-mobile/react-native-recyclerview-list/pull/11  for more information.

To test:
Before:
1. open the demo app
2. try to select text on ~RichText and~ PlainText-backed blocks (~Header, Paragraph are rich-text~, while Code and Read More are based on PlainText)
3. observe you can't select text (contextual menu doesn't appear, can't copy, etc.)

gutenberg_selection_works_not_ok

After:
1. open the demo app
2. try to select text on ~RichText and~ PlainText-backed blocks (~Header, Paragraph are rich-text~, while Code and Read More are based on PlainText)
3. observe you can select, copy, paste and such.

gutenberg_offscreen_works_ok

### DIFF
--- a/package.json
+++ b/package.json
@@ -110,7 +110,7 @@
     "react-native": "0.57.5",
     "react-native-hr": "git+https://github.com/Riglerr/react-native-hr.git#2d01a5cf77212d100e8b99e0310cce5234f977b3",
     "react-native-modal": "^6.5.0",
-    "react-native-recyclerview-list": "git+https://github.com/wordpress-mobile/react-native-recyclerview-list.git#v1.0.0",
+    "react-native-recyclerview-list": "git+https://github.com/wordpress-mobile/react-native-recyclerview-list.git#4a4b11ceb6244837c08b8def1c0356909b6a8c52",
     "react-native-svg": "6.5.2",
     "react-redux": "^5.0.7",
     "redux": "^3.7.2",
--- a/yarn.lock
+++ b/yarn.lock
@@ -5,21 +5,18 @@
 "@babel/code-frame@7.0.0-beta.44":
   version "7.0.0-beta.44"
   resolved "https://registry.yarnpkg.com/@babel/code-frame/-/code-frame-7.0.0-beta.44.tgz#2a02643368de80916162be70865c97774f3adbd9"
-  integrity sha512-cuAuTTIQ9RqcFRJ/Y8PvTh+paepNcaGxwQwjIDRWPXmzzyAeCO4KqS9ikMvq0MCbRk6GlYKwfzStrcP3/jSL8g==
   dependencies:
     "@babel/highlight" "7.0.0-beta.44"
 
 "@babel/code-frame@^7.0.0", "@babel/code-frame@^7.0.0-beta.35":
   version "7.0.0"
   resolved "https://registry.yarnpkg.com/@babel/code-frame/-/code-frame-7.0.0.tgz#06e2ab19bdb535385559aabb5ba59729482800f8"
-  integrity sha512-OfC2uemaknXr87bdLUkWog7nYuliM9Ij5HUcajsVcMCpQrcLmtxRbVFTIqmcSkSeYRBFBRxs2FiUqFJDLdiebA==
   dependencies:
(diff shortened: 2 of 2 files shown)


daniloercoli:
Tested on Android 8.0.0 - Samsung S9, and the behavior is identical to what reported above by @hypest.

mzorz:
Thanks for the reports here @daniloercoli and @hypest ! I actually did not test on the real device 😬  - will get back to you after checking myself and trying some few bullets I may have 🙇

mzorz:
Just a note here that tested the RecyclerListView [demo project PR](https://github.com/wordpress-mobile/react-native-recyclerview-list/pull/11) on a Pixel 2 and it seems to be working alright for me. Will continue looking on the Gutenberg demo now 👍

mzorz:
> The RichText blocks seem to not acquire focus/caret even though the block itself does get selected (I can see the inline toolbar appearing on each). I wonder if you see similar behavior on any of your devices.

Confirmed RichText is not fixed by this, but PlainText is 🤦‍♂️ . In fact that's what I have been trying all along, and only tried RichText on the emulator. I'll create another ticket for the proper handling of PlainText selection and associate this PR to it as this PR addresses such situation, and continue to work on #135 (RichText) on another PR.

mzorz:
PR description has been updated to reflect it fixes the problem for the RN PlainText component only (#267)

mzorz:
Closing this PR for now, as we're moving towards use of `FlatList` and removing `RecyclerViewList` as indicated in https://github.com/wordpress-mobile/gutenberg-mobile/pull/256#pullrequestreview-177652478

PR #275, opened 2018-11-23
Use format-library for the formatting bar

koke:
This PR is here to help test the changes in gutenberg: https://github.com/WordPress/gutenberg/pull/12249

Requires changes in Aztec: https://github.com/wordpress-mobile/AztecEditor-Android/pull/777

~~Requires RN Aztec changes https://github.com/wordpress-mobile/react-native-aztec/pull/105~~ 

### Testing Instructions (for Android)
- Apply this patch
- Apply the following patch: https://gist.github.com/Tug/50f0a51b7833a94d5f05f4fa7020a289
You can use the following command for this:
```
curl "https://gist.githubusercontent.com/Tug/50f0a51b7833a94d5f05f4fa7020a289/raw/3317f9ab4dd096cb2a358c44bb5e1085a8a15550/debug-with-aztec.patch" | git apply -v
```
- Have https://github.com/wordpress-mobile/AztecEditor-Android in the same directory as `gutenberg-mobile` and checkout [this patch](https://github.com/wordpress-mobile/AztecEditor-Android/pull/777)
- Run `yarn android` or launch with Android Studio

### Testing Instructions (for iOS)
- Apply this patch
- Run `yarn ios` or launch with XCode


hypest:
I tried this on Android and (besides the fact that I had to temporarily comment out the [moduleDidMount](https://github.com/wordpress-mobile/gutenberg-mobile/blob/8061eb742e829415308fcf2b333653afc80b8d98/src/app/App.js#L33) call which is not available on Android) it seems that the selected word is not picked up as a whole. See the following gif: (the `r` is missing from the selected word)

link-word-length-android-problem

Tug:
@hypest Thanks for reviewing. Strangely, I'm unable to reproduce this issue. Could it be possible that you tested while I was rebasing the gutenberg PR about an hour ago?

hypest:
I can try again now just to be sure and will report back 👍

hypest:
Still happening for me on 08e345a. Using a Pixel 2 XL, Android 9.

koke:
Those buttons look really odd there, is that the final design?

Tug:
@hypest Thanks! I thought I had fixed it with https://github.com/WordPress/gutenberg/pull/12249/commits/f7513a822b420f903d4dbb9012aa1d1962e4c487 but it seems it only works in Aztec iOS... I just added [a hack](https://github.com/WordPress/gutenberg/pull/12249/commits/6c2cc0774f05622fecbb5b5286770e3fd09eb308) to fix the problem. I think it's good enough for our need for now, but we might want to revisit later.

> Those buttons look really odd there, is that the final design?

I think they should be good enough for this first version, I used `Button` from `react-native` so they get the platform's styling. I explored using `Button` from `@wordpress/components` but it seems the styles were hardcoded there so they look like icon buttton (aspectRation: 1...). It will require a whole different set of changes so they work for both the Block toolbar and for those modal buttons.

koke:
@iamthomasbishop what do you think? If there's no way we can make those look more like the designs, I think it might be better to at least drop the colors and keep them both white

hypest:
I can confirm that latest fixes have removed the [missing character issue](https://github.com/wordpress-mobile/gutenberg-mobile/pull/275#issuecomment-448558823) @Tug 👍.

hypest:
Can you also update this PR from master @Tug ? There's an annoying issue where on Android the app red-screens on start because it tries to contact the bridge, which is already fixed in master.

Tug:
@hypest Sure, it's done now :)

hypest:
@tug, can you verify that Bold/Italic/Strikethrough are working for you in the paragraph block (possibly in other too) when just toggling the format without anything selected? When I try it on Android I don't see the formats "stick".

Steps:

1. Focus a paragraph block with some text already in it
2. Tap on a word but don't select any part of it
3. Tap on the "Bold" format button on the toolbar
4. Notice the "Undo" button becoming active but no other visual change happening, including no "pressed" state for the Bold button
5. Type in some characters. Notice that the new characters are not in bold 💥

Tug:
@koke @iamthomasbishop Updated the Buttons to use a custom version instead (will need to be refactored in `@wordpress/components` later) using TouchableOpacity, View and Text. It seems to work pretty well's indeed an improvement on android imo:
image

@hypest Haven't tested Android much tbh. Looking now 👍

koke:
Tested on iPhone 8, example app:

I selected a word, tapped the link button, got a red screen

```
Invariant Violation: Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: undefined. You likely forgot to export your component from the file it's defined in, or you might have mixed up default and named imports.

Check the render method of `InlineLinkUI`.

This error is located at:
    in InlineLinkUI (at with-spoken-messages/index.js:38)
    in WithSpokenMessages(InlineLinkUI) (at link/index.js:109)
    in LinkEdit (at with-spoken-messages/index.js:38)
    in WithSpokenMessages(LinkEdit) (at format-edit.js:21)
    in FormatEdit (at with-select/index.js:144)
    in ComponentWithSelect (at with-select/index.js:151)
    in WithSelect(FormatEdit) (at rich-text/index.native.js:305)
    in RCTView (at View.js:45)
    in RichText (at with-instance-id/index.js:30)
    in WithInstanceId(RichText) (at edit.native.js:87)
    in RCTView (at View.js:45)
    in ParagraphEdit (at edit.native.js:18)
    in Edit (at with-filters/index.js:64)
    in WithFilters(Edit) (at block-edit/index.js:46)
    in BlockEdit (at block-holder.js:44)
    in RCTView (at View.js:45)
    in RCTView (at View.js:45)
    in TouchableWithoutFeedback (at block-holder.js:69)
    in BlockHolder (at block-manager.js:283)
    in RCTView (at View.js:45)
    in RCTView (at View.js:45)
    in CellRenderer (at VirtualizedList.js:687)
    in RCTScrollContentView (at ScrollView.js:852)
    in RCTScrollView (at ScrollView.js:977)
    in ScrollView (at VirtualizedList.js:1062)
    in VirtualizedList (at FlatList.js:663)
    in FlatList (at block-manager.js:202)
    in RCTView (at View.js:45)
    in KeyboardAvoidingView (at keyboard-avoiding-view.ios.js:45)
    in KeyboardAvoidingView (at block-manager.js:211)
    in RCTView (at View.js:45)
    in BlockManager (at MainApp.js:33)
    in SlotFillProvider (at MainApp.js:32)
    in MainScreen (at AppContainer.js:82)
    in AppContainer (at with-dispatch/index.js:66)
    in ComponentWithDispatch (at with-dispatch/index.js:73)
    in WithDispatch(AppContainer) (at with-select/index.js:144)
    in ComponentWithSelect (at with-select/index.js:151)
    in WithSelect(WithDispatch(AppContainer)) (at App.js:31)
    in AppProvider (at renderApplication.js:34)
    in RCTView (at View.js:45)
    in RCTView (at View.js:45)
    in AppContainer (at renderApplication.js:33)

createFiberFromTypeAndProps
    ReactNativeRenderer-dev.js:5672:10
createFiberFromElement
    ReactNativeRenderer-dev.js:5700:4
reconcileSingleElement
    ReactNativeRenderer-dev.js:10547:8
reconcileChildFibers
    ReactNativeRenderer-dev.js:10631:14
reconcileChildren
    ReactNativeRenderer-dev.js:11101:6
updateFragment
    ReactNativeRenderer-dev.js:11290:4
performUnitOfWork
    ReactNativeRenderer-dev.js:15852:21
workLoop
    ReactNativeRenderer-dev.js:15892:41
renderRoot
    ReactNativeRenderer-dev.js:15996:15
performWorkOnRoot
    ReactNativeRenderer-dev.js:16976:17
performWork
    ReactNativeRenderer-dev.js:16877:24
performSyncWork
    ReactNativeRenderer-dev.js:16838:14
batchedUpdates$1
    ReactNativeRenderer-dev.js:17079:21
batchedUpdates
    ReactNativeRenderer-dev.js:2614:31
_receiveRootNodeIDEvent
    ReactNativeRenderer-dev.js:2709:17
receiveTouches
    ReactNativeRenderer-dev.js:2785:28
__callFunction
    MessageQueue.js:349:47
<unknown>
    MessageQueue.js:106:26
__guard
    MessageQueue.js:297:10
callFunctionReturnFlushedQueue
    MessageQueue.js:105:17
```

koke:
I removed my previous comment since I hadn't actually pulled the latest changes correctly 🤦‍♂️

koke:
Tested on iPhone 8, and the UI is not visible somehow

gb-links

iamthomasbishop:
Sorry for the delay! This is looking pretty dang solid for the quick first iteration! Well done. A couple immediate thoughts to get this shipped and in test hands:

* The border-radii of bottom-left and bottom-right should be 0.
* Can we round off the left/right sides of the drag handle to be fully rounded?
* I believe the margin between the top of the sheet and the drag handle should be 5pt
* What size are the table cell labels/values? I think they should probably be 16 to be safe
* What is the plan for `Link to Existing Content` function, as that was recently added and is a very highly requested feature – I'd hate to lose that

For reference, here's kinda where the designs were at for links between platforms, but I'm not necessarily expecting us to get there immediately:

image

Thanks!

koke:

Tug:
Did some important update to the gutenberg PR so we can write in a specific format (even in multiple ones). So @hypest issue is fixed now.

I also did some minor styling improvement to follow @iamthomasbishop recommendations.

- The border-radii of bottom-left and bottom-right should be 0. ✅ 
- Can we round off the left/right sides of the drag handle to be fully rounded? ✅ 
- I believe the margin between the top of the sheet and the drag handle should be 5pt ✅ (I reduced it to 6px as react native doesn't support pt)
- What size are the table cell labels/values? I think they should probably be 16 to be safe
The text is 14px, I think it looks about the same size as on the mockups, I can always increase it if necessary. There's a 10px padding (top and bottom) around each label/values now, and 5px for the Switch which is bigger
- What is the plan for Link to Existing Content function, as that was recently added and is a very highly requested feature – I'd hate to lose that
We'll certainly want to add it but it's outside the scope of the PR imo. We need to be able to connect the app to the API so it's no simple task and this PR is already large enough to be painful to review.

Also, I'm not sure where the Done button is for Android in the mockup?

hypest:
Confirmed that the [reported issue](https://github.com/wordpress-mobile/gutenberg-mobile/pull/275#issuecomment-448646641) is fixed now 🎉.

> Also added an android version to the react-native-aztec changes but it seems to have some problems as onSelectionChange is emitted in a loop just after render.

Maybe @daniloercoli or @mzorz can lend a hand on this one ^?

daniloercoli:
While testing this PR I noticed some other problems when formatting the text. 
- I tried to apply "Italic" to new text in a para block, by tapping on the `I`  toolbar button. The button was highlighted, and then reverted after a while. 
- Navigating thru text with the keyboard does not keep the toolbar properly updated.
- Other weird behavior where the "bold" style was selected in toolbar, but the text was plain.

Se the animated GIF below for details:

selection-android

hypest:
> Since there is currently a bug in the react-native-aztec PR for android

Is the bug still there @Tug ? Not sure which one is it though. Is there a comment or ticket somewhere about it?

Tug:
> Is the bug still there @Tug ?

Yes it's there. I mentioned it quickly in a comment:

> Also added an android version to the react-native-aztec changes but it seems to have some problems as onSelectionChange is emitted in a loop just after render.

But we mainly discussed it on slack.

Here is how it looks like:

android

Tug:
So there are at least 2 errors left (tested on android):
- It doesn't look like aztec android supports writing using the next character formats. You can try and apply this patch and see the the result: https://gist.github.com/Tug/e9e64412dd2d87e205fecbbf8db98540
- Aztec and the RichText component may desync since they both use their own formatting lib and might generate different HTML output. You can try writing in multiple formats and then editing inside the formatted text, switching formats and jumping to new parts of the formatted text until you see you're not writing in the selected formats and formatted text around your cursor may switch. In this case you will see an error being logged in the console due to this "try catch" here:  https://github.com/WordPress/gutenberg/pull/12249/files#diff-ed195425f515f40db41c23d586760ef9R156

@hypest Could you try to reproduce those 2 issues first before we discuss a "solution"?

hypest:
> It doesn't look like aztec android supports writing using the next character formats. 

I think I see the problem happening on both iOS and Android. In particular, while debugging the Android side, I notice that Aztec correctly identifies the styles after the caret (even at the start of text) but the JS side doesn't and send down an empty formats list.

To test:
1. with this html post `<p><strong>Hello</strong> World!</p>`, place the caret at the beginning of text.
2. Notice that the "Bold" button is not highlighted
3. Start writing and notice that the text is not in bold. Happens both on iOS and Android.

> Aztec and the RichText component may desync

Hmm, haven't been able to reproduce that yet 🤔

hypest:
At this stage, I wonder if the issues so far can be worked on a separate/follow up PR. The main feature of this PR is the links-UI and it seems to work fine so far.

diegoreymendez:
Created a WPiOS PR for the latest commit number from this PR here: https://github.com/wordpress-mobile/WordPress-iOS/pull/10958

At least from my tests this feature seems to be working fine without any changes on the iOS side.

daniloercoli:
As per chat convo and video call, this PR (and linked PRs) are working fine and LGTM from my side, even though there are still some issues with the format bar buttons that could not be in par with the text.

Problem 1 - Format bar buttons act weirdly. You turn style off but they go back to on after a while.
On Android:
- Add a new para block 
- Tap on Bold and start typing
- Tap on Italic and type some other text
- Remove both Bold and Italic
- Boom, after a short while the tool is back with Bold and Italic both ON
- If you keep typing the styles are still there applied to the new text

Problem 2 - In other cases format bar buttons are not in par with the styling of text in the block. 
Try to write some styled text and move caret around. 

Problem 3 - Rendering loop
- Put the cursor in the middle of a bold word and tap Enter.Key
- The block is split correctly in 2 blocks
- But there is a loop on the first block going...
https://cloudup.com/c5NT20tZOrJ

Tug:
@danielebogo Right I did notice those problems as well at different points but didn't manage to find the cause except for the fact that, due to a sequence of events, **Aztec and gutenberg-mobile gets de-synchronized** and that one editor does not convert to the the same html as the other.

@hypest @koke Imo it's still not ready to merge, I'd like to spend some more time on it and see if we can mitigate those issues and at least fix the rendering loop problem.

koke:
😢 The rendering loop sounds like a merge blocker, I think the other two we could solve incrementally on a separate PR

Tug:
Another thing: it seems AztecEditor-iOS returns `<strike>` tags instead of `<del>` in some cases. It causes inconsistencies in the selected formats on iOS.

koke:
I found a couple issues:

* On Android, toggling the format for a selection works, but it unselects the text. This is a minor annoyance and we can deal with this later.
* On iOS (iPhone 8), tapping on the link button seems to present the bottom sheet but it's not visible, the screen just darkens:
ios-links 2019-02-15 11_38_28

hypest:
Just mentioning it here for tracking it: there's a known issue: "b" tags are not recognised by the format-library as "bold" so, the toolbar doesn't highlight the "B" button when the cursor is on a `<b></b>` section. That behavior is same on GB-web.

koke:
Another thing I noticed on iOS is that when the link settings come up, the block list behind it scrolls up and the block is no longer visible:

link-scroll

iamthomasbishop:
A note on the styling of this sheet as we work towards shipping it – the styling of the Sheet component has been changed/consolidated, so we'll need to update the styling slightly. I understand we're getting close to shipping so I'm hoping this doesn't add much work – it'll just feel odd relative to the other updated sheets (Inserter in-progress, Image Settings done, Media stuff done).

Here's what the updated "neutral" Sheet looks like:

image
(iOS left, Android right)

The notable differences:

* Sheet Title is removed `*`
* Top Bar actions are removed – `remove link` is added as a button in the Tableview, and user can dismiss by swiping down or tapping the scrim.
* We are using a new "neutral" Table style. The only diffs between platforms are in font-family (SF Pro Text on iOS, Roboto on Android), and row indentation (w/ icons 56pt on iOS, 72dp on Android).

`*` – The plan is to do some usability testing to see if we need to show an explicit title for clarity, but this aligns with the majority of examples we referenced during discovery.

Might be worth coordinating with @etoledom and @marecar3 to figure out how they've been styling/re-using the component.

daniloercoli:
> I think the other two we could solve incrementally on a separate PR

Just ended another round of testing and the rendering loop issue is gone. Although I've found the problem with toolbar/styling pretty annoying. Once it happens you cannot go back to plain text. Even if you switch to another block , and then back to the "problematic" one, the issue is still there.
At the end of the video you will see me trying to remove styling without luck.

https://cloudup.com/cXNXc52YxKK

I've also tried by jumping to another block and back, with no luck.

hypest:
Here's a quick list of the issues that still seem outstanding (will create tickets later):

1. [Android] Word gets "bold": Start with two words text like "two words". Place the cursor between the space and the "w". Tap on "B" to turn on bold. Type a few chars and press "space". Notice the "word" becoming *strong* while it shouldn't.
2. [iOS] Style doesn't apply: Start with "two words". Place the cursor between the space and the "w". Tap on "B" to turn on bold. Type a few chars and notice that only the first new char is in bold. The rest are not styled.
3. [Android] Selection is dropped. Select a word and style it (example: to bold) by tapping on a toolbar button. Word gets stlyed but selection is removed.
4. Haven't replicated this but as per Tug: it seems AztecEditor-iOS returns `<strike>` tags instead of `<del>` in some cases. It causes inconsistencies in the selected formats on iOS.
5. [Android] Known issue: "b" tags are not recognized by the format-library as "bold" so, the toolbar doesn't highlight the "B" button when the cursor is on a `<b></b>` section. That behavior is same on GB-web but I think it's counter-intuitive.
5b. [iOS] on iOS, the editor mutates the markup, replacing `<b>` with `<strong>`s. Issue it that this doesn't align with what Gutenberg-web does.
6. [Both platforms] Styling issues as [reported by Thomas above](https://github.com/wordpress-mobile/gutenberg-mobile/pull/275#issuecomment-464078628).
7. [Android] Style button gets highlighted again. See video reported by @daniloercoli : https://cloudup.com/cXNXc52YxKK
8. [Android] Note for the next pass/fixes: _setting_ the last content from the "onSelectionChange" event feels wrong and we should remove it. Makes the logic more complex and it may introduce Aztec/RN sync bugs.

Will now update the PRs and start the merge domino since we haven't dimmed any of the issues as blockers for merging this and iterating on follow up PRs. cc @koke

koke:
> [iOS] Style doesn't apply: Start with "two words". Place the cursor between the space and the "w". Tap on "B" to turn on bold. Type a few chars and notice that only the first new char is in bold. The rest are not styled.

Tested on 12.1 and it seems to work fine 🎉

Issue #251, opened 2018-11-16
Tapping on PlainText powered blocks does not make the block-holder selected


mzorz:
What I want now is to be able to tap into a PlainText block, see the cursor appears, and make the Block automatically get selected (bubble the event up to get `isSelected` to be set to `true`).
Currently, what happens is you have to tap on the borders of the block to make it be selected.

This happens for PlainText-based blocks as well, so it should be easier to isolate the problem by not using RecyclerView nor Aztec to try and understand what's going on.

Ideas:
- make the touch event be grabbed first by the touchable area in the block-holder? (I mean, investigate around that)
- if that doesn't work, then using `onFocus` in PlainText to bubble up a `Selected` event, that makes `isSelected` turn to true for the block holder representing such a block.
- furthermore, there are plans to use InputTextState-like state, but that's on a different level (keyboard handling will need this to be able to hide/show the keyboard not only on InputText but also on Aztec RN wrapper).

mzorz:
Possiby related issue https://github.com/facebook/react-native/issues/14958#issuecomment-324237317

mzorz:
> furthermore, there are plans to use InputTextState-like state, but that's on a different level (keyboard handling will need this to be able to hide/show the keyboard not only on InputText but also on Aztec RN wrapper).

Plans referred to up here are exposed in #273.

daniloercoli:
It seems that the problem is not only related to PlainText, since i'm experiencing it on Rich Text fields on a real device.

I've just put some logs in the code, and moving the caret from a Para block to a Code blocks doesn't update the props properly.
See the name of the component that holds the focus after the tapping, during the rendering phase:
```
2018-11-23 19:12:26.696 28060-28100/? I/ReactNativeJS: 1542996746696
2018-11-23 19:12:26.696 28060-28100/? I/ReactNativeJS: core/image
2018-11-23 19:12:26.696 28060-28100/? I/ReactNativeJS: NOT FOCUSED
2018-11-23 19:12:26.733 28060-28100/? I/ReactNativeJS: 1542996746733
2018-11-23 19:12:26.733 28060-28100/? I/ReactNativeJS: core/image
2018-11-23 19:12:26.734 28060-28100/? I/ReactNativeJS: NOT FOCUSED
2018-11-23 19:12:26.750 28060-28100/? I/ReactNativeJS: 1542996746750
[...]

PR #277, opened 2018-11-26
Fix block toolbar that sometime doesn't appear when the block get the focus

daniloercoli:
This PR fixes an issue where tapping on a `PlainText` or `AztecWrapper` block does not make the bottom bar following the focus, and be visible attached to the correct block. #251

It seems that the problem was related to the `onFocus` method not properly propagated up in the software layers, and our data model was lagging one step behind the focus. 

Companion GB PR is here: https://github.com/WordPress/gutenberg/pull/12316

To test:
- Start the demo app
- Tap on a `Para` block and check that the toolbar is visible
- Tap on a `Code` block and check that the toolbar is visible
- Tap on a `More` block and check that the toolbar is visible

### DIFF
--- a/src/block-management/block-holder.js
+++ b/src/block-management/block-holder.js
@@ -48,7 +48,7 @@ export default class BlockHolder extends React.Component<PropsType> {
 				setAttributes={ ( attrs ) =>
 					this.props.onChange( this.props.clientId, { ...this.props.attributes, ...attrs } )
 				}
-				onFocus={this.props.onBlockHolderPressed.bind( this, this.props.clientId )}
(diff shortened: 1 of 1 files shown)


daniloercoli:
Thanks @mzorz for your review! I've address your comments and updated the ref. Travis is working, hope it will turn green soon! 😅 

Once this is merged will try to apply the same changes to `onBlur`, since i've noticed edge cases if you tap on a block that doesn't get the focus. (Unsupported block).

mzorz:
> Start the demo app
Tap on a Para block and check that the toolbar is visible
Tap on a Code block and check that the toolbar is visible
Tap on a More block and check that the toolbar is visible

Added step:
- Tap on a Heading block and check that the toolbar is visible

Issue #279, opened 2018-11-26
Toolbar isn't visible in WPiOS app


marecar3:
For some reason, Toolbar isn't visible in WPiOS application.

<img width="352" alt="screen shot 2018-11-23 at 20 01 19" src="https://user-images.githubusercontent.com/28219092/49047499-e5c5ac80-f1d7-11e8-82bb-6f668d6707e0.png">

PR #280, opened 2018-11-26
Fix/toolbar position in wp

marecar3:
Fixed [problem with Toolbar](https://github.com/wordpress-mobile/gutenberg-mobile/issues/279) in WP app. 
(The toolbar wasn't visible in WP iOS app.)

**iPhone 8 WP app**
simulator screen shot - iphone 8 - 2018-11-26 at 23 43 38

**iPhone XR Gutenberg app**
simulator screen shot - iphone xr - 2018-11-26 at 23 47 08


src/block-management/block-manager.js
```diff
@@ -23,6 +22,8 @@ import EventEmitter from 'events';
 const keyboardDidShow = 'keyboardDidShow';
 const keyboardDidHide = 'keyboardDidHide';
 
+const { height: fullHeight } = Dimensions.get( 'window' );
```
koke: From [Dimensions.get](https://facebook.github.io/react-native/docs/dimensions.html#get):

> Although dimensions are available immediately, they may change (e.g due to device rotation) so any rendering logic or styles that depend on these constants should try to call this function on every render, rather than caching the value

marecar3: Good point @koke.
Fixed.

PR #281, opened 2018-11-27
[iOS] Adding aztec attachment delegate to Gutenberg bridge.

etoledom:
This PR introduces the changes necessary to add an interface to the `Gutenberg` bridge for clients to easily assign an `Aztec.TextViewAttachmentDelegate` to the internal Aztec components of Aztec.

The current behavior (before this PR) is to crash the app if a paragraph block contains an image.

Now it will be able to load and display the image appropriately.
- The current implementation in the example app is to display a dummy "Aztec" image instead of loading the image from the given URL.

Since `Aztec.TextViewAttachmentDelegate` is a swift protocol, I decided to translate our Gutenberg bridge to swift. It also give us a nicer "swift-y" interface and all the goodies of the language for future implementations.

**To test:**

- Open the example app on Xcode and clean build folder.
- Run `yarn start`
- Run the example app with `yarn ios` (or from Xcode).
- Check that everything compiles and run correctly.
- Check that the previous bridge methods continue to work.
- Switch to html view and replace everything with this:
```html
<!-- wp:paragraph -->
<p>Hi there</p>
<img src="https://etoledomatomicsite01.blog/wp-content/uploads/2018/04/amazing-beautiful-breathtaking-clouds-1024x684.jpg" alt="" class="wp-image-43">
<!-- /wp:paragraph -->
```
- Switch back to blocks view.
- If you see a red screen saying: `console.error: "Block validation: Block validation failed ..."` that is expected, since paragraph blocks are not supposed to have img tags.
- Press `escape` and check that the paragraph block displays the Aztec logo.
- Success!

screen shot 2018-11-27 at 11 07 06 am


etoledom:
I just noticed a couple of issues. I'll mark this as `Not ready for review` while I fix them,

etoledom:
I believe it's ready now 👍

react-native-gutenberg-bridge/ios/RNReactNativeGutenbergBridge.xcodeproj/project.pbxproj
```diff
@@ -201,7 +222,7 @@
 				GCC_WARN_UNINITIALIZED_AUTOS = YES_AGGRESSIVE;
 				GCC_WARN_UNUSED_FUNCTION = YES;
 				GCC_WARN_UNUSED_VARIABLE = YES;
-				IPHONEOS_DEPLOYMENT_TARGET = 8.0;
+				IPHONEOS_DEPLOYMENT_TARGET = 9.0;
```
pinarol: we can even set it to 10 in the future since WPiOS uses 10.

PR #286, opened 2018-11-28
Make block holder and toolbar look nicer

koke:
**Note:** This needs https://github.com/WordPress/gutenberg/pull/12403 to be merged first

Some changes to make the block holder look better:

- Uses dashicons instead of Emoji
- Update layout so buttons have the expected size (44x44)
- Remove unused settings button
- Push the delete button to the right
- Disable up/down when you can't move a block (it's the first or last)
- Added borders to the block holder when focused
- A bit of a refactor and better folder structure for InlineToolbar

@iamthomasbishop I know your mockups use Gridicons, but we still don't have those available, and dashicons are already quite an improvement over emoji 😉. Also, I used `$light-gray-500` for the block borders, but I'm not sure if it's the right one, I think your mockups looked more blue than this

inline-toolbar


marecar3:
Adding @iamthomasbishop.

koke:
This needs https://github.com/WordPress/gutenberg/pull/12403 to be merged first

koke:
Ready for another look:

screen shot 2018-11-28 at 23 25 58

PR #290, opened 2018-11-28
Travis generated bundles

hypest:
This PR adds an extra job in Travis which produces the JS bundle and uploads them to Amazon S3.

The final URLs are:
`https://s3-us-west-1.amazonaws.com/gutenberg-mobile-js-bundle/wordpress-mobile/gutenberg-mobile/<commit hash>/android/App.js`

`https://s3-us-west-1.amazonaws.com/gutenberg-mobile-js-bundle/wordpress-mobile/gutenberg-mobile/<commit hash>/android/App.js.map`

`https://s3-us-west-1.amazonaws.com/gutenberg-mobile-js-bundle/wordpress-mobile/gutenberg-mobile/<commit hash>/ios/App.js`

`https://s3-us-west-1.amazonaws.com/gutenberg-mobile-js-bundle/wordpress-mobile/gutenberg-mobile/<commit hash>/ios/App.js.map`

PR #294, opened 2018-11-28
Bridge: Android bundle in bridge assets

hypest:
This PR adds support for automatically fetch the Android JS bundle from S3 (see #290) and including it in the assets. WPAndroidGlueCode.java then uses it from there.

Changes in this PR:

1. Downloads the Android JS bundle from S3. Detects the commit hash and uses it for the URL.
2. Defines a filename in gradle for the JS bundle and use it in the Java code too (via BuildConfig)
3. Places the JS bundle in the assets folder to be reachable in runtime via the typical Android `assets://` URL scheme.
4. Only download the file when building the library standalone or via jitpack.

### To test A (building the library)

1. Go into the `react-native-gutenberg-bridge-android` folder
2. Run `./gradlew assemble` to start building the bridge library
3. Notice that at some point, gradle runs `downloadJSBundle` and prints "Downloading JS bundle from ...." and that the URL includes the commit hash in this PR (at the time of writing it's
ec5649e)
4. The build finishes OK
5. Feel free to examine the `build/outputs/aar/react-native-gutenberg-bridge-debug.aar` file (it's a zip file essentially) to verify that the JS bundle lies inside the assets folder.

### To test B (using the library in wpandroid)

1. Use https://github.com/wordpress-mobile/WordPress-Android/pull/8664 (hash b356abc and beyond) to run wpandroid against this library
2. Avoid or remove the `wp.BUILD_GUTENBERG_FROM_SOURCE` flag from the root `gradle.properties` file so the build will depend on binaries only
3. Running wpandroid as normal and open a Gutenberg post. It should open the block editor just fine.

### DIFF
--- a/react-native-gutenberg-bridge/android/build.gradle
+++ b/react-native-gutenberg-bridge/android/build.gradle
@@ -82,3 +82,32 @@ dependencies {
     }
     implementation 'com.facebook.react:react-native:+'
 }
+
+def getGitHeadHash() {
+    def stdout = new ByteArrayOutputStream()
+    exec {
+        workingDir project.projectDir
+        commandLine 'git', 'rev-parse', 'HEAD'
+        standardOutput = stdout
+    }
+    return stdout.toString().trim()
+}
+
+task downloadJSBundle << {
+    def assetsFolderName = "${project.projectDir}/src/main/assets"
+    File assetsFolder = new File(assetsFolderName)
+    if (! assetsFolder.exists()){
+        assetsFolder.mkdirs();
+    }
+
+    def filename = 'gutenberg-mobile-bundle.js'
+    def targetFile = new File("${assetsFolderName}/${filename}")
+    def hash = getGitHeadHash()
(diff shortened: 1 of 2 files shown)


hypest:
Closing this without merging in favor of https://github.com/wordpress-mobile/WordPress-Android/pull/8696. Even though this PR makes things look "cleaner" by having the JS bundle be included in the bridge jar instead of the wpandroid, it introduces a sequence of build dependencies which makes the build fail easily and having to manually orchestrate them afterwards.

For context, the sequence of events it introduces is:
1. wpandroid Travis waits on gutenberg-mobile Jitpack to build the bridge jar
2. gutenberg-mobile Jitpack waits on gutenberg-mobile Travis to build the bundles
3. gutenberg-mobile Travis builds the bundle
4. 3. takes some time to copmlete
5. 2. times out waiting on 3.
6. 1. fails because 2 is not available
7. One needs to manually wait for 3. to complete and restart 2.
8. One needs to wait for 2. to restart 1.

PR #298, opened 2018-11-29
Focus on RichText when selection changes

koke:
This combines https://github.com/WordPress/gutenberg/pull/12422 and https://github.com/wordpress-mobile/react-native-aztec/pull/85

When a RichText becomes selected, it will call `focus` on the aztec view to ensure it gets the right focus. The focus was working OK when tapping, but this fixes the cases where the block focus changes programatically, like inserting a new paragraph or splitting a block.

To test:

- Insert a new paragraph, verify it has focus
- Type some words, press return, verify the new block has focus

### DIFF
--- a/src/block-management/block-manager.js
+++ b/src/block-management/block-manager.js
@@ -85,7 +85,7 @@ export default class BlockManager extends React.Component<PropsType, StateType> 
 		this.setState( { ...this.state, selectedBlockType: itemValue, blockTypePickerVisible: false } );
 
 		// create an empty block of the selected type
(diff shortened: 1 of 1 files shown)


koke:
@mzorz since you probably have a better idea of the current issues, feel free to add any this PR fixes to the description

mzorz:
Nothing much to add 😄 the code is quite simple to read and you got things right.
This https://github.com/WordPress/gutenberg/pull/12422 also fixes a similar problem I saw on Android (breaking focus when `enter` was pressed and a new block was inserted) 👍

PR #299, opened 2018-11-29
Show default block appender component when post is empty

Tug:
Requires https://github.com/WordPress/gutenberg/pull/12434

This PR adds a default text in the content area when no blocks are available. Clicking on it will replace it with an empty paragraph block by default.
This is close to what's available in the web version with with `BlockListAppender` even though it's simpler at this stage.

You can test this without using the parent app if you manually set `initialHtml` to `""` (empty string).
For instance in `App.js`: 
```
<AppContainer initialHtml="" />
```

### DIFF
--- a/src/app/AppContainer.js
+++ b/src/app/AppContainer.js
@@ -41,7 +41,7 @@ class AppContainer extends React.Component<PropsType> {
 		const post = props.post || {
 			id: 1,
 			content: {
-				raw: props.initialHtml,
+				raw: '',// props.initialHtml,
 			},
 			type: 'draft',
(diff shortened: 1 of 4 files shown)


src/app/AppContainer.js
```diff
@@ -41,7 +41,7 @@ class AppContainer extends React.Component<PropsType> {
 		const post = props.post || {
 			id: 1,
 			content: {
-				raw: props.initialHtml,
+				raw: '', // props.initialHtml,
```
Tug: TODO: revert this before merge

PR #310, opened 2018-12-03
Fix block validation errors

Tug:
Requires https://github.com/WordPress/gutenberg/pull/12552

Fix "Block validation failed for `core/paragraph`... " errors when trying the demo post content from gutenberg.

For this we remove  the babel transform `react-native-classname-to-style` which would cause unexpected result when saving blocks and thus validation errors.

This PR also removes some unused dependencies and others that are already provided by `@wordpress/babel-preset-default`.

Lastly, it updates the `initial-html.js` and appends the whole content of the gutenberg demo post which can be found here https://wordpress.org/gutenberg/


.flowconfig
```diff
@@ -39,6 +39,9 @@
 <PROJECT_ROOT>/node_modules/fbjs/.*
 <PROJECT_ROOT>/node_modules/graphql/.*
 <PROJECT_ROOT>/node_modules/prettier/.*
+<PROJECT_ROOT>/node_modules/jsx-to-string/.*
```
Tug: Seems like those are necessary sometimes, let's just add those

PR #313, opened 2018-12-03
Update html content through the bridge

pinarol:
This change adds ability to update the editor content through the bridge. This is for being able to return back to a previous revision that parent app will provide(a.k.a. History feature).

updatecontent1

**To Test:**

Copy following content:

```
<!-- wp:paragraph -->
<p><b>Hello</b> World!</p>
<!-- /wp:paragraph -->

```
Test 1 - Update content on Visual mode

- Tap (...) on right top corner, select Update HTML
- Paste the above content
- Verify that visual editor content updated with the given html
- Tap 'Save' and check the XCode logs to verify the latest html is passed
- Switch to HTML mode and verify contents are matching

Test 2 - Update content on HTML mode

- Switch to HTML editing mode
- Tap (...) on right top corner, select Update HTML
- Paste the above content
- Verify that html editor content updated with the given html
- Tap 'Save' and check the XCode logs to verify the latest html is passed
- Switch back to Visual mode and verify contents are matching

WPiOS side can be tested also with instructions here https://github.com/wordpress-mobile/WordPress-iOS/pull/10593


pinarol:
cc @hypest this is about the History option.

ios/gutenberg/GutenbergViewController.swift
```diff
@@ -79,20 +79,48 @@ extension GutenbergViewController {
...
+    }
+    
+    func alertWithTextInput(using handler: ((String?) -> Void)?) -> UIAlertController {
+        let alert = UIAlertController(title: "Enter HTML", message: nil, preferredStyle: .alert)
+        alert.addTextField()
+        let submitAction = UIAlertAction(title: "Submit", style: .default) { [unowned alert] (action) in
+            handler?(alert.textFields?.first?.text)
+        }
+        alert.addAction(submitAction)
+        alert.addAction(UIAlertAction(title: "Cancel", style: .cancel))
```
marecar3: Also here, maybe it would be good to put those strings in some localization file.

react-native-gutenberg-bridge/ios/Gutenberg.swift
```diff
@@ -51,6 +51,10 @@ public class Gutenberg: NSObject {
     public func toggleHTMLMode() {
         bridgeModule.sendEvent(withName: EventName.toggleHTMLMode, body: nil)
     }
+    
+    public func updateHtml(_ html: String) {
+        bridgeModule.sendEvent(withName: EventName.updateHtml, body: [ "html": html ])
+    }
 }
 
 extension Gutenberg: RCTBridgeDelegate {
```
marecar3: I know it's not a part of this PR, but file name `Gutenberg.swift` doesn't tell anything about the purpose of this class, maybe we can consider re-naming in the future?

ios/gutenberg/GutenbergViewController.swift
```diff
@@ -79,20 +79,48 @@ extension GutenbergViewController {
...
+        alert.addAction(updateHtmlAction)
         alert.addAction(cancelAction)
 
         present(alert, animated: true)
     }
-
+    
+    var toggleHTMLModeAction: UIAlertAction {
+        return UIAlertAction(
+            title: htmlMode ? "Switch To Visual" : "Switch to HTML",
```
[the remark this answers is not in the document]
pinarol: This is just the example app for us to make development tests, not sure if we'll have any benefit localizing the texts here.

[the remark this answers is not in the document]
marecar3: @pinarol thanks for the update. Definitely no benefit to localize that :)

react-native-gutenberg-bridge/ios/Gutenberg.swift
pinarol: That seemed a little odd to me too at the beginning but then I started to think that it is a fine name. Gutenberg class provides an interface for all the capabilities of the ReactNative Gutenberg bridge, if you look at that way it is fine to call it Gutenberg because it is THE actual Gutenberg for the parent app. What do you think @etoledom ?

hypest: FWIW, on the Android side of things, I've introduced the `WPAndroidGlueCode` class that manages the ReactRootView, the RN dependencies and the bridge instance. The name is still poor IMHO and I'll revise it at some point but, it was a deliberate choice to clear from a too ambiguous "Gutenberg". I most probably introduce something like "WPGutenbergGlue" in the near future.

pinarol:
@etoledom I reverted one line of code in order to fix that cursor issue that happens on iOS and then the componentDidMount() method became necessary. I think it is ready for you to take one more look

pinarol:
@etoledom yes and now I've fixed that revertign back what I did last. It turns out we have no option as using "value={ this.isIOS ? null : this.props.editedPostContent }". Because if we use that then the content will not be updated when we try to set HTML from outside so we really need to remove that. 

I think the problem of cursor position when you do a fast typing is about the react native component, because it doesn't happen with a simpler content as below:

```
<!-- wp:paragraph -->
<p><b>Hello</b> World!</p>
<!-- /wp:paragraph -->

<!-- wp:nextpage -->
<!--nextpage-->
<!-- /wp:nextpage -->
```
And assuming also people won't be able to type very fast in the real device I think it is an edge case, so we can use our opportunity to not focus on this a lot currently.

Unless there's a better solution of course.

I have updated the code to make it more similar to web side also.

Issue #317, opened 2018-12-04
Some toolbar buttons are hidden when in a heading block


hypest:
When a heading block is focused, the Toolbar gets populated with enough buttons to cause an "overflow" and some buttons are pushed outside the view. The toolbar is not scrollable so, those overflown buttons are not reachable.

![screenshot-1543937884364](https://user-images.githubusercontent.com/1032913/49453454-d48a2a80-f7eb-11e8-8a4b-e7c094873f59.jpg)

marecar3:
Hey @hypest ,
Should I solve this one with as high priority task?

Maybe adding horizontal scroll view would be a good and quick solution for the alpha release? @koke @hypest @iamthomasbishop 

Thanks.

hypest:
A horizontal scroll could be a stop gap solution for the Alpha but I'll let @iamthomasbishop chime in as well.

iamthomasbishop:
Absolutely, the user should be able to scroll the toolbar horizontally.

PR #326, opened 2018-12-05
Make toolbar horizontally scrollable

pinarol:
Fixes https://github.com/wordpress-mobile/gutenberg-mobile/issues/317

ezgif com-video-to-gif

The block toolbar is made scrollable horizontally. If the scrollbar content is smaller than the scrollbar itself then it won't scroll, thus, it won't scroll for blocks other than Heading right now, but in the future if other blocks have also lots of toolbar buttons then it will scroll also for them with no extra effort.

**To Test:**

- Add a Heading block and type something in it
- Verify that scrolling horizontally is possible
- Verify that you can tap toolbar buttons with no problem

### DIFF
--- a/src/block-management/block-toolbar.js
+++ b/src/block-management/block-toolbar.js
@@ -4,7 +4,7 @@
  */
 
 import React, { Component } from 'react';
-import { View } from 'react-native';
+import { View, ScrollView } from 'react-native';
 import { withSelect, withDispatch } from '@wordpress/data';
 import { compose } from '@wordpress/compose';
 import { Toolbar, ToolbarButton } from '@wordpress/components';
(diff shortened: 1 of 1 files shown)


src/block-management/block-toolbar.js
```diff
@@ -37,34 +37,36 @@ export class BlockToolbar extends Component<PropsType> {
...
-				{ showKeyboardHideButton && ( <Toolbar>
-					<ToolbarButton
-						label={ __( 'Keyboard hide' ) }
-						icon="arrow-down"
-						onClick={ onKeyboardHide }
-					/>
-				</Toolbar> ) }
-				<BlockControls.Slot />
-				<BlockFormatControls.Slot />
+				<ScrollView horizontal={ true } showsHorizontalScrollIndicator={ false } keyboardShouldPersistTaps={ 'always' } alwaysBounceHorizontal={ false } >
```
marecar3: Hey @pinarol , is there a chance that we break this line into multiple lines?
I think it would be much easier for review :)

pinarol: sure

PR #327, opened 2018-12-05
added replaceBlockAction handling

### LINKED ISSUE #315: Replace empty block when inserting another
When focused in an empty block and then go to insert another block via the Inserter, the new block should replace the empty one.

mzorz:
Fixes #315 

#### What this PR does
If the currently selected block is empty and the BlockPicker is actioned, the currently selected block is now replaced with the new block type as selected in the BlockPicker.

~There are some cases that I'll want to make an exception for (will check Gutenberg on the web for that), as for example an image block may not have an image set but still we want to keep it?. Will look and make a final description on this PR.~

Only these blocks `core/heading
core/paragraph
core/code
` are taken into account for the condition described above (these are basically blocks where the user is expected to enter some text in, and where it makes sense to check for emptiness). All other blocks will not be replaced.

### ToDo
- [x] replace title `ADD BLOCK` for `REPLACE BLOCK` as corresponds 0a210a9
- [x] check which cases should not be replaced (`core/image` blocks? `core/more` block? etc) d87a8f7  - whitelist so only these blocks `core/heading
core/paragraph
core/code
` are taken into account.

replaceorinsert

#### To test:
1. load the demo app
2. tap on the + inserter icon to bring the BLockType picker up
3. choose Paragraph type (first on the left)
4. observe a new, empty Paragraph block is appended to the list and the editor scrolls there (the Paragraph editor gets the focus and the block is selected)
5. leaving the focus/selection there, tap on the + inserter icon again
6. now select T title block
7. observe the block is now a Title block type, as opposed to appending a new empty block once again.

Repeat the steps 5-7 again with different blocks: when the currently selected block is either of Paragraph, Title, or Code any other selected block in the picker will take its place, whereas if the currently  selected block is any other then whatever pick you make on the Picker will introduce a new block and insert it below.


src/block-management/block-picker.js
```diff
@@ -32,6 +33,8 @@ export default class BlockPicker extends Component<PropsType, StateType> {
 	}
 
 	render() {
+		const titleForAdd = 'ADD BLOCK';
+		const titleForReplace = 'REPLACE BLOCK';
```
Tug: I'm not convinced by this "title" feature, user might pick a block that won't replace the currently selected block, because the block is not replaceable, in that case the block will be added even though it shows "REPLACE BLOCK".

Moreover I think it's really easy to replace a block unintentionally, especially since we don't yet have the `BlockListAppender` component at the end of the list and the (+) button is global to the app currently.

That said, I don't mind much fixing those usability issues later since we're still aiming for alpha.

mzorz: > I'm not convinced by this "title" feature, user might pick a block that won't replace the currently selected block, because the block is not replaceable, in that case the block will be added even though it shows "REPLACE BLOCK".

This wouldn't happen - it should say "REPLACE BLOCK" _only if the block is really going to be replaced_ :) - for that, this check is getting run: https://github.com/wordpress-mobile/gutenberg-mobile/pull/327/files#diff-2d50daf729e068ecebcce113d1ce7f9aR280

So this way there shouldn't be a risk in using the wrong title. On the contrary, leaving the "add block" title for all cases when we're introducing "replacement" as an action for empty editable blocks (`Paragraph`, `Title`, and `Code` for now) seemed misleading as "adding" is not what ends up happening with the introduction of this behavior.

Tug: Nevermind about replacing a block unintentionally, we check that the block is empty first 👍

Tug: > This wouldn't happen - it should say "REPLACE BLOCK" only if the block is really going to be replaced :)
Oh you're right, again we check that the block is empty, missed that 🙇

Tug:
LGTM 👍 

I did some update in how replaceBlock is given as props to BlockManager. We did port the store to gutenberg by doing minimal changes in AppContainer, but it's good practice to connect our components to the store whenever it's possible to minimize the amount of props which are passed down and also avoid calling the whole render chain from the top component when just one props changes.

In another PR we'll want to refactor BlockManager by
- Moving most of the props from AppContainer to BlockManager
- Removing state.blocks and the `focused` property, replacing it with isBlockSelected in other component whenever it's possible.

src/block-management/block-manager.js
```diff
@@ -86,13 +91,19 @@ export default class BlockManager extends React.Component<PropsType, StateType>
 	}
 
 	onBlockTypeSelected( itemValue: string ) {
-		this.setState( { ...this.state, selectedBlockType: itemValue, blockTypePickerVisible: false } );
```
Tug: Note: It's not necessary to return the whole state here, react will merge with the current state already

mzorz:
> I did some update in how replaceBlock is given as props to BlockManager. We did port the store to gutenberg by doing minimal changes in AppContainer, but it's good practice to connect our components to the store whenever it's possible to minimize the amount of props which are passed down and also avoid calling the whole render chain from the top component when just one props changes.

Thank you for introducing these changes and explaining the whys here @Tug ! 🙇

src/block-management/block-manager.js
```diff
@@ -260,6 +269,23 @@ export default class BlockManager extends React.Component<PropsType, StateType>
...
+
+	isCandidateForReplaceBlock( block: BlockType ) {
+		return ( block.name === 'core/paragraph' || block.name === 'core/heading' || block.name === 'core/code' );
+	}
+
+	isReplaceable( block: ?BlockType ) {
+		if ( ! block ) {
+			return false;
+		}
+		return this.isEmptyBlock( block ) && this.isCandidateForReplaceBlock( block );
```
koke: Late to the party, but we should reuse the same logic that Gutenberg for this: `isUnmodifiedDefaultBlock` from `@wordpress/blocks`: https://github.com/WordPress/gutenberg/blob/40d1282a992702e1933118ef626f0666692baf2b/packages/editor/src/components/inserter/menu.js#L389

Tug: Good catch! I miss that 👍

mzorz: Good catch indeed! opening a PR for this valuable change

PR #335, opened 2018-12-07
Refactor BlockManager and BlockHolder to use selectors and actions from the store

Tug:
This PR is a janitorial PR that mainly updates BlockManager and BlockHolder to use information from the store (instead of function passed as props from AppContainer) and remove multiple anti-patterns.

Changes introduced here:
#### 1. `src/app/AppContainer.js`
- **[React anti-pattern]** Remove all props that are not directly used by this component
- Update PropsType
- Deduplicate `parseBlocksAction` and `updateHtmlAction`, keep only `updateHtmlAction`
- **[React anti-pattern]** Switch from passing `blocks` to `getBlocks` as props to avoid rerender `AppContainer` on `blocks` change

#### 2. `src/app/MainApp.js`
- Remove props that were just passed from parent to children

#### 3. `src/block-management/block-manager.js`
- Reconnect all the selectors and actions that passed from `AppContainer`
- Remove `block.focused` attribute and thus the need for `state.blocks`
- Remove unused property in state such as `selectedBlockType` and `refresh`
- Move `onInlineToolbarButtonPressed`, `onFocus`, `insertBlocksAfter` and `mergeBlocks ` to `BlockHolder` as those make more sense in the scope of a block. Also try to mirror the gutenberg version of those components.
- **[React anti-pattern]** Remove `Function.bind` call in `renderItem`

#### 4. `src/block-management/block-holder.js`
- Add selectors and actions necessary for `onInlineToolbarButtonPressed`, `onFocus`, `insertBlocksAfter` and `mergeBlocks ` to work properly
- Slightly reimplement those to use data from the store instead of looking into the `blocks` array
- **[React anti-pattern]** Feed block.name and block.attributes from withSelect not from BlockManager
- **[React anti-pattern]** Switch from passing `order` to `getBlockIndex ` as props to avoid rerender `BlockHolder` on order change, only if it's the first or the last

#### 5. `src/block-management/block-picker.js`
- Remove unused state properties
- **[React anti-pattern]** Remove unnecessary `Function.bind` calls in `render`
- Update onValueSelected propType

#### 6. `src/block-management/block-toolbar.js`
- Move `onKeyboardHide` from `BlockManager` to `BlockToolbar`

#### 7. `src/block-types/unsupported-block/edit.js`
- Remove unused StateType definition

### Anti-pattern details
**Remove all props that are not directly used by this component**: The first goal is to modularize our code so we keep it as short as possible and focus on its own logic, not the one of its children. Moreover, having too many props, especially ones that change often will cause our component to re-render on each props change. This is bad for performance. In general we should avoid passing props which modifications are not used to render our component.

Thus the reason for **Switch from passing `blocks` to `getBlocks`**. Here it's only for performance,  we don't want AppContainer to re-render on each `blocks` change. 

**Calling `Function.bind` on a prop inside `render`** should be avoided as it will create a new function on each render call. Meaning the props of the children element will change and thus it will re-render. This can cause a re-render chain in the app in cases where there is nothing to update.

Binding a function to its class is only ever needed when the function is called from another context than the class (an event listener on the DOM, a native interface, another component).
Gutenberg's approach to this is to bind the functions in the constructor as class attributes, but Flow doesn't really like the syntax (we have a workaround though: https://github.com/wordpress-mobile/gutenberg-mobile/blob/master/src/block-management/inline-toolbar/index.js#L26). The alternative is to define our function as an attribute of the class (instead of a method) and use a fat-arrow function as those are bound locally by default.

**Feed block.name and block.attributes from withSelect not from BlockManager**: Minor but again the less props we have the better. Here BlockManager should not decide if BlockHolder has to rerender on `name` or `attributes` change.

### Testing Instructions
- Try on both iOS and android
- Boot the app
- Try switching from visual to html and back
- Try moving blocks up and down
- Try adding a new block of every type
- Try splitting an existing block
- Try adding a new block while an empty/replaceable block is selected
- Try removing blocks
- Try editing in blocks
- Try undo/redo changes


hypest:
Hey @Tug , nice work!

While i'm still reviewing the code changes, I tried out this branch in wpandroid and seems to red-screen with `Invariant Violation: Event cannot be both direct and bubbling: topSelectionChange`. Can you try replicating and looking into it? Thanks!

mzorz:
Re: this
> Invariant Violation: Event cannot be both direct and bubbling: topSelectionChange

Could this have something to do? iOS has a contradicting declaration. https://github.com/wordpress-mobile/react-native-aztec/pull/94#issuecomment-446303751

koke:

Tug:
@koke @hypest Updated with last changes from develop

hypest:
> Updated with last changes from develop

I think I see the `code` and `more` blocks available in the Inserter and those are disabled in develop, currently. Perhaps some extra updating from develop (and or Gutenberg master perhaps) is needed?

src/app/AppContainer.js
```diff
@@ -12,25 +12,14 @@ import type { BlockType } from '../store/types';
...
-	onResetBlocks: Array<BlockType> => mixed,
-	onSelect: string => mixed,
-	clearSelectedBlock: void => void,
-	onAttributesUpdate: ( string, mixed ) => mixed,
 	initialHtml: string,
+	resetBlocks: Array<BlockType> => mixed,
 	setupEditor: ( mixed, ?mixed ) => mixed,
-	clientId: string,
+	toggleBlockMode: ?string => mixed,
+	getBlocks: () => Array<BlockType>,
```
hypest: WDYT about adding the `post` property to the PropsType definition? We are optionally expecting it in line 31 even though Flow doesn't complain (I guess because of the `||`ing).

src/block-management/block-holder.js
```diff
@@ -45,36 +54,109 @@ export class BlockHolder extends React.Component<PropsType, StateType> {
...
+				TextInputState.blurTextInput( currentlyFocusedTextInput );
+			}
 		}
+		this.props.onSelect( this.props.clientId );
+	};
+
+	onInlineToolbarButtonPressed = ( button: number, clientId: string ) => {
+		switch ( button ) {
+			case InlineToolbarActions.UP:
+				this.props.moveBlocksUp( clientId );
```
hypest: I'm a bit bugged that we're accepting and passing `clientId` here even though the clientId is a constant for each block. Is it perhaps an anti-pattern to modify the `withDispatch` to return a curried version of `moveBlocksUp` that doesn't require the parameter? Same for numerous other functions passed as props from `withDispatch` that require `clientId` but don't exactly need it locally. WDYT?

Tug: Agreed 👍

Tug:
@hypest Thanks for your feedback, I updated and fixed the issues you mentioned and cleaned up a bit the proptypes as well. This should be good for another go!

src/block-management/block-holder.js
```diff
@@ -20,16 +28,25 @@ import { BlockEdit } from '@wordpress/editor';
...
 	isSelected: boolean,
+	isFirstBlock: boolean,
+	isLastBlock: boolean,
 	showTitle: boolean,
-	onChange: ( clientId: string, attributes: mixed ) => void,
+	getBlockIndex: ( clientId: string, rootClientId: string ) => number,
+	getPreviousBlockClientId: ( clientId: string ) => string,
+	getNextBlockClientId: ( clientId: string ) => string,
+	onChange: ( attributes: mixed ) => void,
 	onReplace: ( blocks: Array<Object> ) => void,
```
hypest: I think `onReplace()` is not passed in props anymore so, let's remove it from the propTypes as well.

src/block-management/block-holder.js
```diff
@@ -20,16 +28,25 @@ import { BlockEdit } from '@wordpress/editor';
...
-	onInlineToolbarButtonPressed: ( button: number, clientId: string ) => void,
-	onSelect: void => void,
-	insertBlocksAfter: ( blocks: Array<Object> ) => void,
-	mergeBlocks: ( forward: boolean ) => void,
-	canMoveUp: boolean,
-	canMoveDown: boolean,
+	onInsertBlocks: ( blocks: Array<Object>, index: number ) => void,
+	onInlineToolbarButtonPressed: ( button: number ) => void,
```
hypest: Similarly, `onInlineToolbarButtonPressed` is not passed via props anymore, right?

src/block-management/block-holder.js
```diff
@@ -45,36 +62,109 @@ export class BlockHolder extends React.Component<PropsType, StateType> {
...
-					this.props.onChange( this.props.clientId, { ...this.props.attributes, ...attrs } )
-				}
-				onFocus={ this.props.onSelect }
-				onReplace={ this.props.onReplace }
-				insertBlocksAfter={ this.props.insertBlocksAfter }
-				mergeBlocks={ this.props.mergeBlocks }
+				attributes={ this.props.attributes }
+				setAttributes={ this.props.onChange }
+				onFocus={ this.onFocus }
+				onReplace={ this.props.replaceBlock }
```
hypest: Hmm, it seems `replaceBlock()` has a very different signature than `onReplace()` that was passed before. Are we sure `BlockEdit` is OK with the `replaceBlock()`?

src/block-management/block-holder.js
```diff
@@ -45,36 +62,109 @@ export class BlockHolder extends React.Component<PropsType, StateType> {
...
+				this.props.moveBlockDown();
+				break;
+			case InlineToolbarActions.DELETE:
+				this.props.removeBlock();
+				break;
+		}
+	};
+
+	insertBlocksAfter = ( blocks: Array<Object> ) => {
+		const order = this.props.getBlockIndex( this.props.clientId, this.props.rootClientId );
```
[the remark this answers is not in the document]
hypest: Scratch that, I see this falls under the "Switch from passing order to getBlockIndex as props to avoid rerender BlockHolder on order change" that you mention in the PR description.

Regarding the `getBlockIndex()` method though, can we do the same treatment to it and remove the parameters since those are constants for each BlockHolder?

src/block-management/block-holder.js
hypest: Can we do the same for `getPreviousBlockClientId()` and `getNextBlockClientId()`?

src/block-management/block-holder.js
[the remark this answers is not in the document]
Tug: We could but it's not the best practice since `getBlockIndex()` is a selector, and thus injected in our props with `withSelect`. If we do that:
```
withSelect( ( select, { clientId, rootClientId } ) => {
	const { getBlockIndex } = select( 'core/editor' );
	return {
		...
		getBlockIndex: () => getBlockIndex( clientId, rootClientId ),
	};
} )
```
Then our `getBlockIndex` props will be re-created on each props change, which isn't optimal if let's say, `getBlockIndex` is passed as a props to another child component as it will force a refresh.

src/block-management/block-holder.js
Tug: Indeed, I must have messed up here, fixing that 👍

Tug:
@hypest I have addressed your concerns, do you think we can have this merged and iterate later if needed?

src/block-management/block-holder.js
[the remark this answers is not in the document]
Tug: Some precision about my last comment which can be confusing:
1 - Passing down props to a child is not a good practice in general, except in very few cases such as HOCs, or some very generic components.
2 - The main reason we don't want to wrap `getBlockIndex` inside a simpler version without parameters is that defining any function inside `withSelect` is bad practice as a new function will be created each time and given as new props. And `BlockHolder` could skip rerendering if no props have actually changed.

src/block-management/block-holder.js
hypest: > Can we do the same for getPreviousBlockClientId() and getNextBlockClientId()?

Answering my own question, those fall under the same reasoning as https://github.com/wordpress-mobile/gutenberg-mobile/pull/335#discussion_r245754229 so, leaving them as is 👍.

PR #340, opened 2018-12-10
First step on fixing keyboard state

### LINKED ISSUE #308: Tapping on a Page Break block does make the block-holder selected but keep the keyboard up, and the caret to the Para block
This is related to https://github.com/wordpress-mobile/gutenberg-mobile/issues/306 - This time the selection of Page Break block happens by the "user" with the cursor/tapping, instead of programatically like in #306, but the end result is the same.

Steps to repro:
- Tap on a para block
- Tap on the page break block

daniloercoli:
This PR fixes #308 by making sure that the field that gets the focus on tapping is a field registered in TextInputState - That has handles focus on all TextInput fields.

React Native Aztec PR is here: https://github.com/wordpress-mobile/react-native-aztec/pull/92

Steps to repro:

- Tap on a para block
- Tap on the page break block
- The block holder should be visible on the page break block, and the keyboard dismissed
- Repeat the steps above by tapping on any of the TextInput "block", and then tap on another kind of block --> the keyboard should be visible only when the focus is on a TextInput field.

Note: The caret is still visible (and blinking) in the TextInput "powered" block that lost the focus. We will fix this in another PR.

### DIFF
--- a/src/block-management/block-holder.js
+++ b/src/block-management/block-holder.js
@@ -111,8 +111,8 @@ export default compose( [
 			onSelect: ( event ) => {
 				if ( event ) {
 					// This fix an issue with the way we are handling focus on TextIput fields and handling the keyboard state.
-					// When moving from a TextInput field to another kind of field the call that hides the keyboard was not invoked
-					// properly resulting in the keyboard up when it should not be there.
+					// When moving from a TextInput field to another kind of field the call that hides the keyboard was not invoked properly,
+					// resulting in keyboard up when it should not be there.
 					const currentlyFocusedTextInput = TextInputState.currentlyFocusedField();
 					if ( event.nativeEvent.target !== currentlyFocusedTextInput && ! TextInputState.isTextInput( event.nativeEvent.target ) ) {
 						// This check that the current TextInputState "focused" field is the one that has the focus.
--- a/src/block-management/block-holder.js
+++ b/src/block-management/block-holder.js
@@ -17,6 +17,8 @@ import styles from './block-holder.scss';
 // Gutenberg imports
 import { BlockEdit } from '@wordpress/editor';
 
+import TextInputState from 'react-native/lib/TextInputState';
+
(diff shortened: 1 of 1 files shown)


src/block-management/block-holder.js
```diff
@@ -106,7 +108,17 @@ export default compose( [
...
+				if ( event ) {
+					// This fix an issue with the way we are handling focus on TextIput fields and handling the keyboard state.
+					// When moving from a TextInput field to another kind of field the call that hides the keyboard was not invoked properly,
+					// resulting in keyboard up when it should not be there.
+					const currentlyFocusedTextInput = TextInputState.currentlyFocusedField();
+					if ( event.nativeEvent.target !== currentlyFocusedTextInput && ! TextInputState.isTextInput( event.nativeEvent.target ) ) {
+						// This check that the current TextInputState "focused" field is the one that has the focus.
+						TextInputState.blurTextInput( currentlyFocusedTextInput );
+					}
+				}
```
koke: I'm not sure this logic belongs here. The opposite (focusing on select) is handled on `RichText.componentDidUpdate`, and I think we could add the inverse there (if it was selected and now it's not, then blur). What do you think?

mzorz: I've tested your suggested approach locally @koke and it seems to work well 👍 thank you for bringing this clarity with that

Branch for Mobile Gutenberg here @daniloercoli  https://github.com/wordpress-mobile/gutenberg-mobile/compare/issue/308-call-blur-when-not-selected?expand=1

React Native Aztec:
https://github.com/wordpress-mobile/react-native-aztec/compare/add/blur-method?expand=1

And Gutenberg:
https://github.com/WordPress/gutenberg/compare/mobile...add/blur-signal-when-rich-text-deselected?expand=1

daniloercoli: Yes, I did exactly the same and was preparing a new PR for it.
Please open a PR then, since my code is the same of your.

PR #346, opened 2018-12-11
Feature/update html

marecar3:
This PR adds the ability to update the editor content through the bridge. This is for being able to return back to a previous revision that parent app will provide(a.k.a. History feature).

To test:

It can be tested from the WPAndroid app based on [this PR](https://github.com/wordpress-mobile/WordPress-Android/pull/8768).

### DIFF
--- a/react-native-gutenberg-bridge/android/src/main/java/org/wordpress/mobile/ReactNativeGutenbergBridge/RNReactNativeGutenbergBridgeModule.java
+++ b/react-native-gutenberg-bridge/android/src/main/java/org/wordpress/mobile/ReactNativeGutenbergBridge/RNReactNativeGutenbergBridgeModule.java
@@ -16,6 +16,11 @@ public class RNReactNativeGutenbergBridgeModule extends ReactContextBaseJavaModu
     private final ReactApplicationContext mReactContext;
     private final GutenbergBridgeJS2Parent mGutenbergBridgeJS2Parent;
 
+    private static final String EVENT_NAME_REQUEST_GET_HTML = "requestGetHtml";
+    private static final String EVENT_NAME_UPDATE_HTML = "updateHtml";
+
+    private static final String MAP_KEY_UPDATE_HTML = "html";
+
     public RNReactNativeGutenbergBridgeModule(ReactApplicationContext reactContext,
             GutenbergBridgeJS2Parent gutenbergBridgeJS2Parent) {
         super(reactContext);
@@ -33,13 +38,13 @@ public class RNReactNativeGutenbergBridgeModule extends ReactContextBaseJavaModu
     }
 
     public void getHtmlFromJS() {
-        emitToJS("requestGetHtml", null);
+        emitToJS(EVENT_NAME_REQUEST_GET_HTML, null);
     }
 
     public void updateHtml(String html) {
         WritableMap writableMap = new WritableNativeMap();
-        writableMap.putString("html", html);
-        emitToJS("updateHtml", writableMap);
+        writableMap.putString(MAP_KEY_UPDATE_HTML, html);
+        emitToJS(EVENT_NAME_UPDATE_HTML, writableMap);
     }
 
     @ReactMethod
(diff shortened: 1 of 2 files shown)


react-native-gutenberg-bridge/android/src/main/java/org/wordpress/mobile/WPAndroidGlue/WPAndroidGlueCode.java
```diff
@@ -160,14 +161,29 @@ public void setContent(String postContent) {
             return;
         }
 
+        if (mShouldUpdateContent) {
```
hypest: I think there's some non obvious reasoning behind the logic here so, let's add a comment about what happens here. WDYT?

marecar3: Good one, will add it, thanks :)

react-native-gutenberg-bridge/android/src/main/java/org/wordpress/mobile/ReactNativeGutenbergBridge/RNReactNativeGutenbergBridgeModule.java
```diff
@@ -32,7 +38,13 @@ private void emitToJS(String eventName, @Nullable WritableMap data) {
     }
 
     public void getHtmlFromJS() {
-        emitToJS("requestGetHtml", null);
+        emitToJS(EVENT_NAME_REQUEST_GET_HTML, null);
+    }
+
+    public void updateHtml(String html) {
```
hypest: I feel the `updateHtml` method name can be a little confusing in the context of the bridge since, it's not clear which way the update is going to happen. I think in this case the update is from the parent app to the RN side so, how about naming this "setHtmlInJS()` or similarly to make it more explicit?

marecar3: Yeah, let's do it. Fixed.

mzorz: > Fixed.

When a request / suggestion is addressed, it's far easier to read when the comment is answered to with the commit hash. Could you then also put a reference to the commit where this is fixed here (valid for future changes as well ;) )?

mzorz: agh - hit `send` too quickly: thanks in advance! 👍

react-native-gutenberg-bridge/android/src/main/java/org/wordpress/mobile/WPAndroidGlue/WPAndroidGlueCode.java
marecar3: Added :)

mzorz: Confirmed - addressed in fa8eaabd72f4cf15b4926dc1d7b3acdf50eb96fa 👍

hypest: Thanks for adding the comment @marecar3 .

> Content can be set directly to RootView only once (per RootView instance)

I guess that it's a convention at this point not to push the content again via the app props, right? I mean, it's already done twice, one in `onCreateView` via `startReactApplication()` and once here in `setContent` via `setAppProperties()`. My impression is that we don't to do it again via props because gutenberg-mobile's state management would be bootstrapped again and we don't want that. Right? If so, let's revise the comment a bit to make this more clear. Sorry if i'm nitpicking about a comment, it's just that there is this important detail we need to convey to our future selves when we read this comment in the future :)

marecar3: Hey, @hypest I agree that we need to be clear and detail with the comment.
I wrote this comment based on how things are working when you test the implementation. 

> // Content can be set directly to RootView only once (per RootView instance)
// otherwise it should be done through module interface

So if you try to pass different conent multiple times thrugh props, with current implementation, nothing will happen, content will stay the same. 

But be free to leave comment that you think it's more suitable for this case and I will update it in this PR.

hypest: Yes, let's add the "why/because" to this `can be set directly to RootView only once`. Thanks!

marecar3: @hypest done.

PR #348, opened 2018-12-11
Upgrade react-native-svg to v8.0.9, use the forked lib

hypest:
This PR:

- Upgrades to v8.0.9 of `react-native-svg`, needed to make the android build work (otherwise gradle complains with `Could not find com.android.tools.build:gradle:2.2.3`
- Start using the forked library, now at https://github.com/wordpress-mobile/react-native-svg, and especially the `feature/enable-android-build-of-standalone-package` branch. Background: we need to maintain the fork for now to have support for building `react-native-svg` as a standalone library via JitPack, used in the binary build pipeline of wpandroid ([See here](https://github.com/wordpress-mobile/gutenberg-mobile/blob/master/react-native-gutenberg-bridge/android/build.gradle#L82)).

To test:

* Run the demo app and notice that the various icons are working as before


marecar3:
@hypest tested and it seems that nothing is broken! 
Approved :)

PR #353, opened 2018-12-12
[iOS] Adds support for forward merging of blocks when pressing DELETE.

diegoreymendez:
### Description:

Pressing FN + Delete now properly merges blocks forward when appropriate.

### Related PRs:

`gutenberg`: https://github.com/WordPress/gutenberg/pull/12809
`react-native-aztec`: https://github.com/wordpress-mobile/react-native-aztec/pull/96

### Details:

TBD


koke:

PR #356, opened 2018-12-12
Adding SafeAreaView to block-toolbar component.

### LINKED ISSUE #354: [iOS]: Make toolbar follow Safe Areas.
The current implementation doesn't follow Safe Areas, this makes the toolbar to be lower than users might expect on newer devices (X family):

screen shot 2018-12-12 at 10 35 13 am

etoledom:
Fixes #354 

This PR adds `SafeAreaView` to the toolbar to make it aware of iOS Safe Areas.

screen shot 2018-12-12 at 11 24 24 am

safe_area

To Test:

- Build and run the iOS example project.
- Check that the tool bar is over the bottom safe area (as shown in the image).
- Tap on a block to show the keyboard.
- Check that the tool bar is still properly visible.

This should not affect Android.
- Check that Android toolbar is unchanged.

### DIFF
--- a/src/block-management/block-toolbar.js
+++ b/src/block-management/block-toolbar.js
@@ -4,7 +4,7 @@
  */
 
 import React, { Component } from 'react';
-import { View, ScrollView } from 'react-native';
+import { View, ScrollView, SafeAreaView } from 'react-native';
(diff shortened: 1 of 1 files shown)


etoledom:
Thank you @marecar3 !

PR #382, opened 2018-12-14
Keeps reference to mReactInstanceManager to avoid re-creating on rotation

mzorz:
The title has it: re-creating everything on rotation meant a lot of time spent on reloading the content, so the WPAndroid bridge now keeps the reference value of mReactInstanceManager to avoid this.

This is to be tested in conjunction with https://github.com/wordpress-mobile/WordPress-Android/pull/8790
(see test steps there)


mzorz:
closing in favor of https://github.com/wordpress-mobile/WordPress-Android/pull/8800

Issue #197, opened 2018-10-26
Keyboard management


koke:
There are some issues related to keyboard, keeping track of them here:

- [ ] When a block is selected, ensure the cursor is visible, scrolling the list if necessary.
- [x] #308 When a non textual block (e.g. image, more) is selected, hide the keyboard.
- [x] #291 when keyboard is up and user taps on another RichText component, keyboard is dismissed
- [ ] #381 keyboard gets dismissed when a new block is created tapping `Enter`. (keyboard should stay open to let the user continue typing)

This is related to #196 and the keyboard tasks in #55 

~I think 2 weeks should be enough but it needs further analysis.~

koke:
I took a first stab at scrolling to a newly created block in e663cbf928214ee019a6cc8cbdfd9fa89e62ce11, but that will try to scroll every time the selection changes. It seems like we'd need to keep track of which items are visible via `onViewableItemsChanged` and only scroll if the desired item isn't visible.

PR #387, opened 2018-12-17
Fix scroll position on iOS

pinarol:
Fixes https://github.com/wordpress-mobile/gutenberg-mobile/issues/197 partially

This change fixes inconsistent scroll behavior with input texts on iOS. Android behavior didn't change since it wasn't problematic.

iOS Before:
ezgif com-resize-2

After:

ezgif com-video-to-gif-2

**Testing prerequisites**

For WPiOS

- Checkout the PRs branch to any arbitrary folder and cd .. to it
- run `yarn install`, `yarn start:reset`
- Open XCode WPiOS on the latest develop
- Clean build folder on Xcode, and then run the app

For Android

- open grade.properties at WordPress-Android folder
- add `wp.BUILD_GUTENBERG_FROM_SOURCE = true` to grade.properties
- checkout the PRs branch in the subrepo of WordPress-Android repo
- cd to WordPress-Android/libs/gutenberg-mobile
- run `yarn install`, `yarn start:reset`
- `yarn wpandroid` on a separate terminal in the same directory

**To Test**

Test 1 - Empty post on WPiOS with return button

- Open an empty post via WPiOS app
- Start typing and press return button to create a new block
- Repeat this until you add about 10 paragraph blocks
- Verify that typed text input is viewable
- Close the keyboard
- Verify that scroll position goes back to normal

Test 2 - Add block with (+) button
- Open some profiled post via WPiOS app
- Add new text blocks with (+) button
- Verify that recently added text input is viewable

Test 3 - Verify inner block buttons are working

- Use up&down&trash buttons when keyboard is open or closed
- Verify that they are working

Test 4 - Repeat tests for Android and make sure we didn't break anything, Android behavior shouldn't change with this PR.


pinarol:
cc @daniloercoli

pinarol:
I will work on travis failures but feel free to take a look @koke

pinarol:
we'll need to merge https://github.com/WordPress/gutenberg/pull/12946 first

src/components/keyboard-aware-flat-list.ios.js
```diff
@@ -0,0 +1,65 @@
...
+	return (
+		<KeyboardAwareScrollView
+			style={ { flex: 1 } }
+			keyboardDismissMode={ 'none' }
+			enableResetScrollToCoords={ false }
+			keyboardShouldPersistTaps={ 'handled' }
+			extraScrollHeight={ extraScrollHeight }
+			extraHeight={ 0 }
+			innerRef={ ( ref ) => {
+				( this: any ).flatList = ref;
```
etoledom: I believe that all these `( this: any )` notation in this file are not necessary. I tried changing them to `this.` and it works fine and flow is still happy.

src/components/keyboard-aware-flat-list.ios.js
```diff
@@ -0,0 +1,65 @@
+import { KeyboardAwareScrollView } from 'react-native-keyboard-aware-scroll-view';
+/**
+* @format
```
etoledom: Small nitpicking:
Can we have the `import` bellow those comments?

src/components/keyboard-aware-flat-list.ios.js
etoledom: Also, this reference seems to be to the `scrollView` and not to the `flatList`? I might be a bit lost 😅

src/components/keyboard-aware-flat-list.ios.js
pinarol: I put this here as a solution to flow error "Cannot resolve module `react-native-keyboard-aware-scroll-view`"
I think that happens because I added `react-native-keyboard-aware-scroll-view` to flowconfig ignore list because flow was trying to validate library source code as well. if there's another solution to this I can apply that happily I just couldn't find any.

src/components/keyboard-aware-flat-list.ios.js
pinarol: > I believe that all these ( this: any ) notation in this file are not necessary. I tried changing them to this. and it works fine and flow is still happy.

flow isn't happy to me :( that was the reason I used `( this: any )` in the first place, maybe you can have error if you run travis check in your local.

screen shot 2018-12-18 at 12 26 02


> Also, this reference seems to be to the scrollView and not to the flatList? I might be a bit lost 

right, I was using KeyboardAwareFlatList before and then changed it to KeyboardAwareScrollList but it looks like I forgot updating the name `flatList`. Will push a new commit for this 👍

pinarol: pushed https://github.com/wordpress-mobile/gutenberg-mobile/pull/387/commits/401d69875ec9938f30d20aa1b24df3b6d8188e50

src/components/keyboard-aware-flat-list.ios.js
etoledom: I see, that makes sense. In that case can we put a small comment explaining that? :)

src/components/keyboard-aware-flat-list.ios.js
etoledom: > flow isn't happy to me :( that was the reason I used ( this: any ) in the first place, maybe you can have error if you run travis check in your local.

May that be because we are declaring variables "on the fly" in a stateless function component  instead of "properly" declaring them in a class before accessing them?

etoledom: But again, I'm not asking to making that change, just trying to understand the issue 👍

src/components/keyboard-aware-flat-list.ios.js
pinarol: @hypest offered a new solution and I've pushed a new commit for it https://github.com/wordpress-mobile/gutenberg-mobile/pull/387/commits/42540721435b273b7d5ee6afcea582325f59ffe6 It is so much better this way 🎉

etoledom: Great solution! 😁

src/components/keyboard-aware-flat-list.ios.js
pinarol: yes right.. but we don't have a class here, KeyboardAwareFlatList is a function and we want these variable to be shared among these sub functions of KeyboardAwareFlatList. 
so by using `this` we just insert the variables in the current context, not the ideal solution I agree. 
I tried declaring KeyboardAwareFlatList as a React.PureCompoent/React.Compoent also but it causes unnecessary re-renders because extraData prop of FlatList changes and that change makes KeyboardAwareScrollView re-render. 
Another solution would be declaring these variables in BlockManager and updating them via functions declared in the KeyboardAwareFlatList props, but I think that wouldn't be good since that variables are only needed here in iOS code :(

etoledom: I see now. Sounds good to me! Let's ✅

pinarol:
@etoledom  I've opened a [bug](https://github.com/wordpress-mobile/gutenberg-mobile/issues/395) for that issue, I don't think it is easily fixable, it is related with the inner mechanisms of KeyboardAwareScrollView.

mzorz:
> Test 4 - Repeat tests for Android and make sure we didn't break anything, Android behavior shouldn't change with this PR.

I've been testing this for a while (smoke tests mostly), it seems to be OK as per my tests.

pinarol:
@etoledom @mzorz about showing the inner toolbar on Android: as far as I know KeyboardAvoidingView can't do it, I had tried KeyboardAwareScrollView on Android side also to achieve that but it behaves different, it first scrolls to the cursor and after some little time it scrolls again, this time for showing the inner toolbar. This is because Android already provides scrolling to the cursor position in the OS level(I guess) and the library just handles the extra scroll, this makes it look scrolling happen in 2 parts and doesn't look very good.

src/components/keyboard-aware-flat-list.android.js
```diff
@@ -0,0 +1,23 @@
+/**
+* @format
+* @flow
+*/
+import { FlatList, KeyboardAvoidingView } from 'react-native';
```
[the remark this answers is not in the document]
pinarol: thanks for the heads up @marecar3 addressing that in https://github.com/wordpress-mobile/gutenberg-mobile/pull/397

PR #391, opened 2018-12-17
Fix iOS heading behavior

### LINKED ISSUE #355: [iOS] Heading block glitch
Adding a new heading block and trying to change its size seems to be surfacing a glitch.

Steps to reproduce:
1. Run the demo app and add a *new* heading block
2. Input some text in the heading block
3. Change the size by tapping any of the H3/H4 toolbar buttons
4. 💥. The size is not actually changing and instead, an empty line appears above the text. See the attached gif.

heading-glitch-ios

etoledom:
Fixes #355

This PR implements the changes made to `react-native-aztec-view` and `gutenberg` to fix the wrong Heading block behavior, where the style shown doesn't correspond to the Heading level selected. 

Those changes force the typing attributes to be corresponding with the Heading level, when Aztec is used to display Heading blocks.

### Refertences:
`react-native-aztec` PR: https://github.com/wordpress-mobile/react-native-aztec/pull/98
`gutenberg` PR: https://github.com/WordPress/gutenberg/pull/12869
`Aztec-iOS` PR https://github.com/wordpress-mobile/AztecEditor-iOS/pull/1108

heading

To Test:

**Setup:**
- Run `git submodule update` to checkout submodules commits.
- Run `yarn preios` to force update the Aztec version.
- Run `yarn start:reset`
- Open the iOS example project on Xcode `open ios/gutenberg.xcodeproj`.
- Clean build folder.
- Build and run (from Xcode or from `yarn ios`)

**Testing:**
- Create a Heading block.
- Check that the heading block style is the correct.
- Type and delete text, change Heading levels, etc...
- Check that everything looks fine.
- Format a word of the title with Italic and change Heading levels.
- Make sure that the Italic format is not removed.
  - Bold has an issue but is already solved on master.

Thank you!


etoledom:
Thank you for the review!

I'm checking the submodules and it looks like `rn-aztec` is already pointing to `master` HEAD, and `gutenberg` is pointing to `mobile` HEAD (that's the good one, right?).

I tested the behavior you described, and I don't see the issue mentioned.
I did notice something odd, is that the block is not deleted when pressing backspace on an empty Heading block, but I could also reproduce that on `master`. So it seems that this PR is not causing it.

heading_delete

PR #396, opened 2018-12-18
Sending `editorDidLayout` message to Native Bridge side.

etoledom:
This PR is meant to solve a problem on iOS, where the toolbar is wrongly positioned when is automatically presented on load (WPiOS)

https://github.com/wordpress-mobile/gutenberg-mobile/pull/386 Added a `SafeAreaView` to `block-manager`. This affects the layout calculations but it's normally not a problem.

The problem happens when we try to present the toolbar before that calculation is resolved (before onLayout is called).

The idea of `editorDidLayout` is to let the Native side know when this happens, so it can present the keyboard at an appropriate time, so the toolbar displays correctly.

Check https://github.com/wordpress-mobile/WordPress-iOS/pull/10682 for more info and how to test it.

### DIFF
--- a/react-native-gutenberg-bridge/index.js
+++ b/react-native-gutenberg-bridge/index.js
@@ -12,7 +12,7 @@ const gutenbergBridgeEvents = new NativeEventEmitter( RNReactNativeGutenbergBrid
 export function sendNativeEditorDidLayout() {
 	// For now, this is only needed on iOS to solve layout issues with the toolbar.
 	// If this become necessary on Android in the future, we can try to build a registration API from Native
-	// to register messages it wants to receive, similar to the Natice -> JS messages listener system.
+	// to register messages it wants to receive, similar to the Native -> JS messages listener system.
 	if ( isIOS ) {
 		RNReactNativeGutenbergBridge.editorDidLayout();
 	}
--- a/react-native-gutenberg-bridge/index.js
+++ b/react-native-gutenberg-bridge/index.js
@@ -3,7 +3,7 @@
 import { NativeModules, NativeEventEmitter, Platform } from 'react-native';
 
 const { RNReactNativeGutenbergBridge } = NativeModules;
-const isIOS: boolean = Platform.OS === 'ios';
+const isIOS = Platform.OS === 'ios';
 
 const gutenbergBridgeEvents = new NativeEventEmitter( RNReactNativeGutenbergBridge );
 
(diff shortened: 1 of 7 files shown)


react-native-gutenberg-bridge/ios/RNReactNativeGutenbergBridge.swift
```diff
@@ -42,6 +43,15 @@ extension RNReactNativeGutenbergBridge {
     public override static func requiresMainQueueSetup() -> Bool {
         return true
     }
+
+    public override func batchDidComplete() {
+        if isJSLoading {
+            isJSLoading = false
+            DispatchQueue.main.async {
+                self.delegate?.gutenbergDidLoad()
```
etoledom: We are not using `gutenbergDidLoad` anymore, but I thought on keeping it for a while in case is needed in the future.

src/app/App.js
```diff
@@ -8,7 +8,6 @@ import React from 'react';
 // Gutenberg imports
```
etoledom: I mistakenly pushed changes to this file directly to master 😱
This is reverting those changes.

src/block-management/block-toolbar.js
```diff
@@ -36,45 +36,43 @@ export class BlockToolbar extends Component<PropsType> {
 		} = this.props;
 
 		return (
-			<SafeAreaView>
```
etoledom: SaveAreaView is not needed in `block-toolbar` anymore. The one inside `block-manager` takes care of the toolbar.

react-native-gutenberg-bridge/index.js
```diff
@@ -1,11 +1,25 @@
...
 
+// Send messages
+
+export function sendNativeEditorDidLayout() {
+	// For now, this is only needed on iOS to solve layout issues with the toolbar.
+	// If this become necessary on Android in the future, we can try to build a registration API from Native
+	// to register messages it wants to receive, similar to the Native -> JS messages listener system.
+	if ( isIOS ) {
+		RNReactNativeGutenbergBridge.editorDidLayout();
+	}
```
etoledom: Avoid breaking Android example app.
This is the approach we decided to take with @hypest

etoledom:
Added a validation to avoid breaking the Android example app.
@pinarol I believe this is ready for review.

Thank you!

PR #399, opened 2018-12-18
Remove block top/bottom padding

### LINKED ISSUE #393: Paragraph  / Heading blocks have extra bottom margin above block controls
First reported by @iamthomasbishop - there's an extra space in between the end of the content area for Paragraph and Heading blocks, and their corresponding container (with the up/down arrows and trash controls)

iOS:

image from ios 12

Android:
screenshot_20181217-202721

mzorz:
Fixes #393 

Removes the top/bottom padding on `block-container` as it was adding an unneeded extra space.

Before:
screen shot 2018-12-18 at 11 57 55

After:
screen shot 2018-12-18 at 11 56 39

Before (WPAndroid):
screen shot 2018-12-18 at 12 02 18

After (WPAndroid):
screen shot 2018-12-18 at 12 03 29

Before (demo app on iOS):
screen shot 2018-12-18 at 12 13 50

After (demo app on iOS):

screen shot 2018-12-18 at 12 19 03

### DIFF
--- a/src/block-management/block-holder.js
+++ b/src/block-management/block-holder.js
@@ -80,7 +80,7 @@ export class BlockHolder extends React.Component<PropsType> {
 			<TouchableWithoutFeedback onPress={ this.props.onSelect } >
 				<View style={ [ styles.blockHolder, focused && styles.blockHolderFocused ] }>
 					{ this.props.showTitle && this.renderBlockTitle() }
-					<View style={ styles.blockContainer }>{ this.getBlockForType() }</View>
+					<View style={ [ !focused && styles.blockContainer, focused && styles.blockContainerFocused ] }>{ this.getBlockForType() }</View>
 					{ this.renderToolbarIfBlockFocused() }
 				</View>
 			</TouchableWithoutFeedback>
--- a/src/block-management/block-holder.js
+++ b/src/block-management/block-holder.js
@@ -80,7 +80,7 @@ export class BlockHolder extends React.Component<PropsType> {
 			<TouchableWithoutFeedback onPress={ this.props.onSelect } >
 				<View style={ [ styles.blockHolder, focused && styles.blockHolderFocused ] }>
 					{ this.props.showTitle && this.renderBlockTitle() }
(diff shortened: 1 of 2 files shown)


iamthomasbishop:
Thanks for pinging for review. Unsupported blocks should still maintain that top/bottom padding, so how it was in the "before" examples was proper. Example:

image

Note: each block, regardless of type, has the same padding: 12pt top/bottom, 16pt left/right.

I think we should also think about implementing the styling shown here, but that's for a separate issue.

mzorz:
Updated in 9f21f87dbe6699ea2e8e089d4300483222c1c426, where it's been updated to match that 12dp value for top/bottom for `unsupported` blocks _only_:

#### Before
screen shot 2018-12-18 at 12 58 09


#### After
screen shot 2018-12-18 at 12 58 30


does it look good now @iamthomasbishop ?

mzorz:
> Note: each block, regardless of type, has the same padding: 12pt top/bottom, 16pt left/right.

oops! this is fixed then in a1bdd09

results:

screen shot 2018-12-18 at 13 09 35

mzorz:
ok - ready for another pass @iamthomasbishop , implemented the specs described in  https://github.com/wordpress-mobile/gutenberg-mobile/issues/393#issuecomment-448302105   in 81d9704, this is how it looks like (before/after):

screen shot 2018-12-18 at 15 03 48

mzorz:
Merging as agreed

iamthomasbishop:
As discussed in dm, the block inner body padding/margin stuff looks good here, so feel free to 🚢! 

We can handle this in a separate issue, but the spacings on the inline toolbar (up/down/trash) needs to be slightly tweaked to include a slight (2pt) margin on both left/right edges of the toolbar. This helps to visually align the block content a little better.

image

mzorz:
no worries, opened issue here https://github.com/wordpress-mobile/gutenberg-mobile/issues/409 - let's follow discussion there or on upcoming PR

PR #404, opened 2018-12-19
[Alpha] Hide the keyboard when tapping on non-inputField block

### LINKED ISSUE #308: Tapping on a Page Break block does make the block-holder selected but keep the keyboard up, and the caret to the Para block

daniloercoli:
This PR fixes #308 for the Alpha, by dismissing the keyboard (calling blur on the last TextInput field) when the field that now gets the focus is a non-textual field.

Please test extensively on iOS, and Android.

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -72,6 +72,7 @@ node_modules/react-native/Libraries/react-native/react-native-interface.js
 node_modules/react-native/flow/
 node_modules/expo/flow/
 libdefs.js
+flow-typed
 
 [options]
 emoji=true
--- a/flow-typed/TextInputState.js
+++ b/flow-typed/TextInputState.js
@@ -1,0 +1,4 @@
+// @flow
+declare module 'react-native/lib/TextInputState' {
+    declare module.exports: any;
+}
--- a/src/block-management/block-holder.js
+++ b/src/block-management/block-holder.js
@@ -110,12 +110,12 @@ export default compose( [
 		return {
 			onSelect: (event) => {
 				if ( event ) {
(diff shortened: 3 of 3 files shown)


daniloercoli:
I'd like @hypest review the last commit https://github.com/wordpress-mobile/gutenberg-mobile/pull/404/commits/d99a268b1892ff531407136309b947a353570b70 , since introduced a new module to make flow happy.

koke:
I think you could add the new definitions to the existing `libdefs.js` file, but this looks OK to me.

koke:
I'm merging this to make sure it gets there before the merge, feel free to do a follow up PR if you want to change the flow config

daniloercoli:
Thanks @koke, your suggestion was right. will open a new PR for flow only.

hypest:
> I'd like @hypest review the last commit d99a268 , since introduced a new module to make flow happy.

👋 @daniloercoli , late to the party but, has this been worked on already perhaps? If not, can't we directly import `TextInputState` from `react-native`? I think we won't need to add extra definitions in `libdefs.js` that way.

daniloercoli:
> directly import TextInputState from react-native

We tried that but was not working, since TextInputState is not exported. Looking at docs on Internet, we implemented the suggested way of importing it.

Issue #398, opened 2018-12-18
Disable code and more blocks for the alpha


koke:
Since we've had some issues with `PlainText`, we'll disable the `code` and `more` blocks for a bit, until we can polish some of the issues to avoid extra feedback about those.

After the alpha, we'll want to evaluate if it makes more sense to have an Aztec-powered PlainText instead of using TextInput

pinarol:
Is this for the problems we have while focusing/scrolling to code/more blocks? if so, they are fixed with https://github.com/wordpress-mobile/gutenberg-mobile/pull/387 on iOS side. And Android doesn't seem to have problem on focusing to code/more blocks.

koke:
Fixed in #406

PR #406, opened 2018-12-19
Disable Code and More blocks for release

Tug:
Addresses https://github.com/wordpress-mobile/gutenberg-mobile/issues/398

This PR unregisters `Code` and `More` for the prod build.

### Testing Instructions
- Run locally using `yarn start` and `yarn android` or `yarn ios` and notice the Code and More blocks are visible
- Run a Release build with `node_modules/.bin/react-native run-ios --configuration Release` and notice the Code and More blocks appear as unsupported blocks.

### DIFF
--- a/src/app/App.js
+++ b/src/app/App.js
@@ -7,7 +7,7 @@ import React from 'react';
 
 // Gutenberg imports
 import { registerCoreBlocks } from '@wordpress/block-library';
-import { registerBlockType, setUnregisteredTypeHandlerName } from '@wordpress/blocks';
+import { registerBlockType, setUnregisteredTypeHandlerName, unregisterBlockType } from '@wordpress/blocks';
 
 import AppContainer from './AppContainer';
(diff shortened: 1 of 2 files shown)


src/app/App.js
```diff
@@ -18,6 +18,12 @@ registerCoreBlocks();
 registerBlockType( UnsupportedBlock.name, UnsupportedBlock.settings );
 setUnregisteredTypeHandlerName( UnsupportedBlock.name );
 
+// disable Code and More blocks for release
+if ( ! __DEV__ ) {
```
koke: How does `__DEV__` get set?

Tug: I'm surprised by the fact that the react-native documentation only has a very succinct [mention](https://facebook.github.io/react-native/docs/javascript-environment) of this global variable.
Still though, it's provided by react-native to differentiate a development build from a release one.

Issue #402, opened 2018-12-18
iOS - iPhone X Landscape/iPad devices - Missing borders on the left/right side of the editor


pinarol:
Right now left/right border’s don’t exist so the block doesn’t feel “contained”. Those should be added as below:

<img width="1234" alt="screen shot 2018-12-19 at 8 41 51 am" src="https://user-images.githubusercontent.com/5032900/50227616-2ff51480-03b7-11e9-90f0-c789cbf410e8.png">

cc @iamthomasbishop

koke:
I think this is by design #386, although the title should respect the safe area as well

pinarol:
we are fixing the title with https://github.com/wordpress-mobile/WordPress-iOS/pull/10692

PR #407, opened 2018-12-19
Add left/right borders to focused block

pinarol:
Fixes https://github.com/wordpress-mobile/gutenberg-mobile/issues/402

Left&Right borders are currently viewable on iPhone X generation devices & landscape modes but other types of devices and portrait mode should also be tested to make sure it doesn't have unwanted consequences.

Left&Right borders should appear on the focused block as below:
screen shot 2018-12-19 at 18 59 21

Portait mode shouldn't change.

**Testing**

- The borders on the focused block:

- should be visible in iPhone X generation devices in Landscape mode.

- should not be visible in Portrait mode, although they still do exist on Portrait mode we should check that they are hardly noticeable.

- Android side also should be checked to verify right&left borders are hardly noticeable in Portrait mode.

### DIFF
--- a/src/block-management/block-holder.scss
+++ b/src/block-management/block-holder.scss
@@ -11,6 +11,8 @@
 	border-color: $gray-lighten-30;
 	border-top-width: 1px;
 	border-bottom-width: 1px;
+	border-left-width: 1px;
+	border-right-width: 1px;
 }
 
 .blockContainer {


pinarol:
@mzorz could you check this doesn't have unwanted consequences on Android?

pinarol:
cc @koke

PR #408, opened 2018-12-19
Remove flow extra config file

daniloercoli:
Remove flow extra config file, and move the new flow declaration for TextInputState to `libdefs.js`.


koke:
It seems like travis is happy, but `.flowconfig` is still referencing `flow-typed`

PR #411, opened 2018-12-20
Only show toolbar when a block is focused

### LINKED ISSUE #405: [Android] Toolbar is presented and enabled when focus is on the Title
1. Open WPAndroid app
2. Enable Gutenberg
3. Open the existing Post 
4. Click on any block (Toolbar is presented)
4. Click on the Title

Result : Toolbar is presented 
Expected: Toolbar shouldn't be presented

mzorz:
Fixes #405 

Given we only need the toolbar when an item _in the list_ is focused, but not when an item outside it is focused (for example, the Title field in the WPAndroid app), this PR implements a workaround that hides the toolbar when there is no focused item, and renders it otherwise.
While this is not the required functionality, in conjunction with WPAndroid this should effectively workaround the issue.

NOTE: it's not working and I can't figure out why (it shows the toolbar right after loading the demo app regardless of no item being focused), therefore I ask a couple more eyes to check it out

To test (what should happen):
1. load the demo app
2. without selecting anything, the toolbar should not be there.
3. tap on any block, the toolbar should appear.

### DIFF
--- a/src/block-management/block-manager.js
+++ b/src/block-management/block-manager.js
@@ -226,7 +226,16 @@ export class BlockManager extends React.Component<PropsType, StateType> {
 		return item.clientId;
 	}
 
+	isAnyBlockSelected() {
+		const focusedItemIndex = this.state.blocks.findIndex( ( block ) => block.focused );
+		if ( focusedItemIndex === -1 ) {
+			return false;
+		}
+		return true;
+	}
+
 	renderList() {
+
 		// TODO: we won't need this. This just a temporary solution until we implement the RecyclerViewList native code for iOS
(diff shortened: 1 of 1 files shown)


koke:
Yes, the issue in #405 is that because the title is in the apps, when you focus on the title, the selected block stays selected and keeps putting its formatting controls inside the toolbar

mzorz:
> is that because the title is in the apps, when you focus on the title, the selected block stays selected and keeps putting its formatting controls inside the toolbar

I got the wrong approach here, as this (if working)  will only hide the toolbar when no block is selected, so it won't fix the actual problem described in #405. Closing this PR

PR #413, opened 2018-12-20
Adding KeyboardHideButton component.

etoledom:
This PR creates a `KeyboardHideButton` component that renders the proper hide-button icon using the RNSVG library.

This is a temporary change since we should ultimately use `dash-icons` with the `ToolbarButton` component.

**Notes:**
The original .svg file for the icon seems to be too complex for RNSVG, so I redrew it with simple shapes.

hide-keyboard

To test:
- Run the example app.
- Tap on a textual block to show the keyboard.
- Check that the new icon shows correctly.
- Tap on the hide-keyboard button.
- Check that the keyboard hides.

cc @koke


etoledom:
Thank you!

PR #414, opened 2018-12-20
iOS only - Fix bottom padding of focused block to the keyboard

pinarol:
This change fixes the bottom padding of the focused block to the keyboard. There was a small extra padding left after this PR https://github.com/wordpress-mobile/gutenberg-mobile/pull/399 so we are fixing that by importing the new value.

Before:
screen shot 2018-12-20 at 11 36 18

After:
screen shot 2018-12-20 at 13 50 28

**To Test**

- Run yarn start
- Open WPiOS app in debug mode and open a post
- Add more than 5-6 paragraph blocks so that the content size exceeds the screen size
- Tap one of the block at the bottom of the page
- Verify that the scrolling position doesn't have unnecessary padding between the keyboard and the bottom of the focused block.

### DIFF
--- a/__mocks__/styleMock.js
+++ b/__mocks__/styleMock.js
@@ -13,4 +13,7 @@ module.exports = {
 	toolbar: {
 		height: 44,
 	},
+	blockContainerFocused: {
+		paddingBottom: 0,
+	},
 };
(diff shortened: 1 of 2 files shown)


__mocks__/styleMock.js
```diff
@@ -13,4 +13,7 @@ module.exports = {
 	toolbar: {
 		height: 44,
 	},
+	blockContainerFocused: {
```
mzorz: nice spot!

PR #417, opened 2018-12-20
updated gutenberg hash to hide link formatting toolbar option

mzorz:
Title has it, check related GB PR in https://github.com/WordPress/gutenberg/pull/13036


hypest:
This seems to point to the Gutenberg _branch_ hash. Let's make it point to the hash in Gutenberg `master` instead, wdyt @mzorz ?

mzorz:
>  wdyt @mzorz ?

Nothing to think about 😅 

Done in 2496c1b

PR #420, opened 2018-12-20
Fix tapping on non-textual block may trigger an error

### LINKED ISSUE #415: [wp-android-integration] Tapping on a picture block does trigger an error

Recap:
`

daniloercoli:
This PR fixes #415 by making sure there is at least one Textual block in the list, before applying the hack about the keyboard hiding in the alpha release.

Steps to repro:
- Start the demo app
- Tap on a para block
- Tap on the page break block

**Important test**
- Start the demo app with empty content
- Tap on '+' symbol and add a page break
- Tap on the page break block and see there is no error

Repeat the steps above by selecting other kind of blocks


daniloercoli:
Closing in favor of #423 due to the reasons expressed in #423 (comment)

PR #423, opened 2018-12-20
Resign focus on richtext when unmounted

### LINKED ISSUE #415: [wp-android-integration] Tapping on a picture block does trigger an error

koke:
Fixes an issue where TextInputState would still believe a specific RichText
instance was focused after it was removed from view

Fixes #415 
An alternative to #420 and #421


mzorz:
Tested on emulators:

- Nexus 4 emulator with Android 5.1.1
- Nexus 5 emulator with Android 7.0

Real device:
- Pixel with Android 8.1

the only thing that we'll need to tackle separately and is a known issue atm is that when there's nothing focused Android grabs the focus on some of the other Aztec editables, so for example in the demo app:
1. tap on the + icon
2. choose paragraph block (scroll all the way down if necessary so you can see it)
3. once now tap + again and choose image block
4. observe the focus (blinking caret)  goes to the next one or two input fields up, which is the one that reads "Dig in deeper" or "Gutenberg is available..."  with the demo content.

Other than that, this achieves the same purpose as intended in #421 and #420 with the more meaningful and clarity. I'd title this `if focused and going to unmount: release focus`.

Thanks for the brevity and taking a look into how `TextInput` works (https://github.com/facebook/react-native/blob/master/Libraries/Components/TextInput/TextInput.js#L875). ❤️

mzorz:
Just for completeness, a note regarding this:

> the only thing that we'll need to tackle separately and is a known issue atm is that when there's nothing focused Android grabs the focus on some of the other Aztec editables, so for example in the demo app:

> 1. tap on the + icon
> 2. choose paragraph block (scroll all the way down if necessary so you can see it)
> 3. once now tap + again and choose image block
> 4. observe the focus (blinking caret) goes to the next one or two input fields up, which is the one > that 5. reads "Dig in deeper" or "Gutenberg is available..." with the demo content.

I confirm this one is fixed on the WPAndroid side, it seems the change introduced in https://github.com/wordpress-mobile/WordPress-Android/pull/8826 suffices already, specifically this change:

```
 android:focusable="false"
     android:focusableInTouchMode="true">
```

It will still exist on the demo Gutenberg app.

To test:
1. start a new draft
2. insert a paragraph, type something
3. insert a Heading, type something
4. now insert an Image block
5. observe the caret is gone away, instead of re-focusing on the Heading or paragraph blocks

PR #431, opened 2018-12-21
[Android bridge] Start the RN app only after we got the real post content

### LINKED ISSUE #390: Opening the demo content on mobile Gutenberg shows an empty post
### Expected behavior
Opening the demo content on the integrated WPAndroid - mobile Gutenberg editor should open normally and certainly not make the app crash / restart.

### Actual behavior
The app ~crashes / restarts itself when opening the demo content, and/or~ just doesn't load the content (sits there showing an empty body)

1. copy the demo content in https://github.com/wordpress-mobile/gutenberg-mobile/blob/master/src/app/initial-html.js into a new Gutenberg post in a self-hosted WordPress site using Gutenberg. 
2. verify your app has Gutenberg enabled (go to Me->App settings -> scroll down and enable Gutenberg).
3. refresh your posts list 
4. Then, trying to open the post from the Posts list
5. observe ~the app being restarted / or~ content not loaded.

hypest:
It seems that RN doesn't like it when we start early and then re-set the
app props very soon after. So, instead, starting the RN app later, as
soon as we get the post content via `setContent()`.

This PR probably fixes #390 too.

To test:
1. Use the web to create a post that uses the demo content from https://github.com/wordpress-mobile/gutenberg-mobile/blob/develop/src/app/initial-html.js
2. Use wpandroid `zalphaRelease` (remember to enable Gutenberg from App settings)
3. Open the post and notice the content loading normally
4. Try going back and opening the post more than a few times. It should always open normally.

### DIFF
--- a/react-native-gutenberg-bridge/android/src/main/java/org/wordpress/mobile/WPAndroidGlue/WPAndroidGlueCode.java
+++ b/react-native-gutenberg-bridge/android/src/main/java/org/wordpress/mobile/WPAndroidGlue/WPAndroidGlueCode.java
@@ -118,7 +118,7 @@ public class WPAndroidGlueCode {
 
         // The string here (e.g. "MyReactNativeApp") has to match
         // the string in AppRegistry.registerComponent() in index.js
-        mReactRootView.startReactApplication(mReactInstanceManager, "gutenberg", initialProps);
+        mReactRootView.setAppProperties(initialProps);
     }
 
     public void onPause(Activity activity) {
@@ -181,7 +181,7 @@ public class WPAndroidGlueCode {
             appProps = new Bundle();
         }
         appProps.putString(PROP_NAME_INITIAL_DATA, content);
-        mReactRootView.setAppProperties(appProps);
+        mReactRootView.startReactApplication(mReactInstanceManager, "gutenberg", appProps);
     }
 
     private void updateContent(String content) {


react-native-gutenberg-bridge/android/src/main/java/org/wordpress/mobile/WPAndroidGlue/WPAndroidGlueCode.java
```diff
@@ -118,7 +118,7 @@ public void onReactContextInitialized(ReactContext context) {
 
         // The string here (e.g. "MyReactNativeApp") has to match
         // the string in AppRegistry.registerComponent() in index.js
-        mReactRootView.startReactApplication(mReactInstanceManager, "gutenberg", initialProps);
+        mReactRootView.setAppProperties(initialProps);
```
koke: I don't know a lot about the bridge but this seems odd. If the app hasn't started yet, why are we setting the properties?

hypest: The properties are kept even though the App may not have been started yet. Looking at the RN source seems to indicate that this is fine.

I can do some more tests in the meantime.

hypest: Did some runtime debugging and indeed, `setAppProperties()` won't try to run the app since this condition fails: https://github.com/facebook/react-native/blob/0.57-stable/ReactAndroid/src/main/java/com/facebook/react/ReactRootView.java#L463

koke: I was thinking less about it being problematic than it being useless, since we'll just set those on initContent, right?

hypest: Oh, not useless. It passes the `htmlModeEnabled` flag that is currently relayed via `onCreateView`. It is then preserved in `initContent()` because of the `getAppProperties()` call there. I could try to refactor this all but I preferred to keep the PR to the minimum of change possible.

koke: I see, that's the bit of context I was missing 👍

PR #124, opened 2018-08-20
Feature/inserter modal

mzorz:
**UPDATE:** this PR is ready for review now.

This PR adds a new `BlockPicker` component that shows the list of available (registered) blocks on a `FlatList` using the modal library [`react-native-modal`](https://github.com/react-native-community/react-native-modal).

- First blocks and action/props connectivity is in place
- worked a bit on styles, to make a better match (though not final) to what the designs show in #58 
- a new "ADD BLOCK HERE" label is shown when the picker is shown, to hint the user about where the block insertion is going to happen

screenshot_1534964690 copy

**ToDo**
- To be tackled separately: the rendering of the icon for each `BlockType` from `dashicons` will be done once https://github.com/WordPress/gutenberg/pull/9012 and https://github.com/wordpress-mobile/gutenberg-mobile/pull/117 get merged

**Updated gif**
insertermodal3

PR #445, opened 2018-12-24
only show Add Block Here indicator when block is not replaceable

### LINKED ISSUE #439: Misleading "Add block here"
When there is just one empty (default) block, and you want to add a new one, you see the text "ADD BLOCK HERE" indicating the position below the default block. But when this new block is inserted, it actually replaces the existing one.

add_block_here

In Gutenberg Web, the Add new block is shown over the existing default block. It's not very clear since the default block is completely white there.

mzorz:
Fix #439 

This PR checks whether the selected block is replaceable and only shows the ADD BLOCK HERE when it is _not_ going to be replaced.

Also, we're using the i18n API now for the `ADD BLOCK HERE` string.

addblockhere

### DIFF
--- a/src/block-management/block-manager.js
+++ b/src/block-management/block-manager.js
@@ -3,6 +3,11 @@
  * @flow
  */
 
+ /**
+ * WordPress dependencies
+ */
+import { __ } from '@wordpress/i18n';
+
 import React from 'react';
 import { isEqual } from 'lodash';
 
@@ -306,10 +311,11 @@ export class BlockManager extends React.Component<PropsType, StateType> {
 	}
 
 	renderItem( value: { item: BlockType, index: number } ) {
+		const titleForAddPlaceIndicator = __( 'ADD BLOCK HERE' );
 		const insertHere = (
 			<View style={ styles.containerStyleAddHere } >
 				<View style={ styles.lineStyleAddHere }></View>
-				<Text style={ styles.labelStyleAddHere } >ADD BLOCK HERE</Text>
+				<Text style={ styles.labelStyleAddHere } > {titleForAddPlaceIndicator } </Text>
 				<View style={ styles.lineStyleAddHere }></View>
 			</View>
 		);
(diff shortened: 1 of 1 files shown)


iamthomasbishop:
I wouldn't do this – The inserter trigger should always act the same (insert a new block) regardless of what block type is focused. We should probably avoid block replace/transforms in general until we have the transform flow built into the toolbar.

koke:
The thing here is that we haven't planned to do block transformations for the beta. I understand how it might seem confusing in some scenarios, but the lack of it would also be confusing in others without block transformations:

- Write a paragraph
- You want to add an image so instinctively tap enter to break away from the paragraph
- Insert an image

Without this logic, the image would be added after the new empty paragraph, and you would have to remove it manually.

BTW, right now the only thing that is replaced is an empty paragraph, which is the default block. I see the web version also replaces empty headings, but I think that's actually done by block transformations.

mzorz:
I think this PR just makes things consistent with the introduction of "add block here" label in #124 by following the original inserter designs here:

gutenberg - inserter v1 - android

Also it's important to note that the replacement behavior was introduced in https://github.com/wordpress-mobile/gutenberg-mobile/pull/327 and slightly modified in https://github.com/wordpress-mobile/gutenberg-mobile/pull/338 with the reasoning behind it being to match the web's behavior at that moment. This PR in particular is a fix to make things consistent with all these 3 things.

Given it's totally valid we either change our minds and/or Gutenberg changes behavior as well, I think we should merge this particular PR and bring the discussion of transformations and their utility within the scope of "inserting when on an empty block" on another issue with designs that can be followed up on different PRs. Wdyt @koke @iamthomasbishop ?

mzorz:
Thank you for the review @etoledom ! 🙇 ❤️

iamthomasbishop:
I still think the behavior is slightly confusing and think we should be intentional about keeping the two things (insert, replace/transform) separate, but for now, how about we just remove the sheet title so the user doesn't have to worry about the distinction here?

I think eventually, we need to show the block type icon in the toolbar to indicate what type of block the user is on, but this should ease the concern for now.

koke:
> another ticket to introduce the proper default block placeholder.

That discussion is going on in #449

mzorz:
Merging this one and letting the conversation continue until a decision is made

mzorz:
> think it should be OK to merge it and also open a new ticket to lose the "REPLACE BLOCK" title case and another ticket to introduce the proper default block placeholder.

opened new ticket in https://github.com/wordpress-mobile/gutenberg-mobile/issues/452 and https://github.com/wordpress-mobile/gutenberg-mobile/issues/453 as per  this ☝️

PR #447, opened 2018-12-25
Fix image cropped vertically

### LINKED ISSUE #444: Image block cropps image vertically.
When adding an image to a post, the image is cropped at a shortened height. 
The image should be the full-size image (scaled to max-width of 100% of the container)

image_block_height

etoledom:
Fixes #444 

Gutenberg side: https://github.com/WordPress/gutenberg/pull/13096

The image size implementations tries to follow closely the web implementation. I tried to share some code from `ImageSize` but most of it is platform specific. I made a small `utils` to share some math calculations.

On a future PR I plan to implement some kind of loading indicator.
cc @iamthomasbishop 

image_height

To test:
- Run the example project.
- Scroll down to the last block.
- Check that the image displays correctly as shown in the attached image.


iamthomasbishop:
Awesome! Thanks for fixing this @etoledom!!

etoledom:
Hey @iamthomasbishop - glad you like it :) 

I'm wandering what do you think about an activity indicator when the image is loading?
This is how it looks now:

image_loading

(That is pretty much the behavior on web, but on mobile I would expect some kind of activity indicator)

iamthomasbishop:
I would also expect a loading indicator. Is there any chance we can utilize the same approach that we took on Aztec uploads, where you see a loader over the top of the media as it loads?

etoledom:
I created a new issue for the activity indicator task.
Let's move that conversation to: #451

etoledom:
Hey @Tug !
I updated the gutenberg reference to latest master. Everything is working in my local tests.
Please take a look before the final merge.
Thank you!

etoledom:
The `gutenberg` reference already was updated in develop so this PR is not needed anymore.

Issue #453, opened 2018-12-27
Introduce the proper "default block placeholder"


mzorz:
Comes from https://github.com/wordpress-mobile/gutenberg-mobile/pull/445#issuecomment-450102887

> the Inserter we have on mobile, is matching the _contextual_ inserter button from the web, not the global one. The contextual one always "replaces" the empty block placeholder instead of appending a new block. It's title though still is "Add block". The default block placeholder should not be considered a block though... it's only a fancy/elaborate indicator.

> Long story short, I think we should work towards the proper behavior anyway, which I think is:
> * Introduce the proper "default block placeholder", which is not the paragraph block

cc @hypest to elaborate? I think we might need more information here to make the issue's description "complete"

SergioEstevao:
@iamthomasbishop is this something that still needs to be done?

iamthomasbishop:
@SergioEstevao I wouldn’t consider it a blocker by any means, but from what I gather we still need to implement behavior that moves selection to the title block when the only block on the canvas is removed or contents cleared. I think other parts of this have been solved in various issues (particularly the part about not showing the movers on the inline toolbar — bc those were removed a while back).

SergioEstevao:
@iamthomasbishop I don't see that behavior on the web side after you remove all the blocks the selection stays on the placeholder block. 
Do we want to do something different for mobile?

iamthomasbishop:
I don't have strong feelings about the behavior, but it seems like the expected behavior is what I described above. I was referring to what we should try to do on mobile specifically, if possible.

hypest:
Since the ticket's original intention was to try and reuse the placeholder as it's found on the web (along with its recently used blocks quick icons) and most of its functionality is already in place per the discussion above, I will close this at this point.

@iamthomasbishop , can you please open a ticket about the "move focus to title block" behavior you are describing? Best to track it separately I think. Thanks!

PR #454, opened 2018-12-27
Block Picker: remove 'replace block' title, only keep 'add block' title

### LINKED ISSUE #452: Remove "REPLACE BLOCK" case indicator and sheet title from Block Picker
As discussed, the behavior is changing and so we should reflect that in Gutenberg mobile.
See related discussion in https://github.com/wordpress-mobile/gutenberg-mobile/pull/445#issuecomment-450102887

mzorz:
Fix #452 

This PR drops the ` isReplacement`  prop from the BlockPicker and always shows:

- block picker title "ADD BLOCK" , dropping "REPLACE BLOCK" 

**IMPORTANT** This PR assumes we will also work on #453

### DIFF
--- a/src/block-management/block-manager.js
+++ b/src/block-management/block-manager.js
@@ -353,7 +353,7 @@ export class BlockManager extends React.Component<PropsType, StateType> {
 					onReplace={ ( block ) => this.onReplace( value.item.clientId, block ) }
 					{ ...value.item }
 				/>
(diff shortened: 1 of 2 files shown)


mzorz:
> May I also suggest we expand the PR description with something like "This PR assumes we will also work on #453" or similar,

Done - thanks!

PR #457, opened 2018-12-27
[Android] Fix for Aztec wrapper that is a single line field

daniloercoli:
This PR updates the Aztec Wrapper hash and makes sure AztecWrapper on Android is a multi-line field with blocks that start capitalized by default and after a period.

Testing step in the companion PR here: 
https://github.com/wordpress-mobile/react-native-aztec/pull/108


hypest:
The react-native-aztec side PR is already merged. I will update the hash to point to the commit in master and will merge the PR since @daniloercoli is AFK.

Issue #436, opened 2018-12-21
iOS - Scrolling behavior scrolls to bottom when inserting new block


etoledom:
When there is one or two blocks, and we add a new block, the current scroll behavior will scroll the new block to the bottom of the screen, when it's not necessary.

It comes back to normal after one tap, as the following gif shows.

![scroll](https://user-images.githubusercontent.com/9772967/50349480-a5422000-0544-11e9-91f0-79b850d5ed3b.gif)

Issue #432, opened 2018-12-21
iOS - Scrolling behavior hides the caret when tapped at the beginning of a very long text


pinarol:
Scrolling behavior hides the caret when user taps at the beginning of a very long text. This happens because KeyboardAwareScrollView always focuses at the end of the selected react element. Let's dig in some more to make it scroll to the cursor instead.

![ezgif com-video-to-gif-4](https://user-images.githubusercontent.com/5032900/50338682-2938df80-0526-11e9-8a4a-cd038922ec85.gif)

koke:
Related to #427

SergioEstevao:
I'm no longer able to reproduce this on iOS so I'm closing this ticket.

PR #461, opened 2019-01-02
Scroll to caret position on focused field

pinarol:
**Fixes** https://github.com/wordpress-mobile/gutenberg-mobile/issues/436 https://github.com/wordpress-mobile/gutenberg-mobile/issues/432 

This PR solves some scrolling issues when keyboard opens. With this PR we are starting to make use of caret position to arrange the scroll position. 

Description of the new mechanism can be found in the [child PR](https://github.com/wordpress-mobile/react-native-keyboard-aware-scroll-view/pull/2).

Preview
ezgif com-video-to-gif-6

**Implementation Details**
The position of the currently focused input in the window had been measured with a native call `UIManager.measureInWindow` previously. I've added a similar call `UIManager.measureSelectionInWindow` but this time it also fetched the caret position.
Scrolling to caret is only available for paragraph and heading blocks currently because it requires a TextView, currently Aztec's text input supports this. React native's text input doesn't provide us the caret position and hides the inner TextView from us. React native text inputs will continue to scroll to the bottom of the component. In the future we can replace those with Aztec's inputs to achieve same thing.

**Testing Instructions**

**Testing Prerequisites**

- Run `yarn install` and `yarn start`

To test with iOS example app 
- Run `yarn ios`

To test with WPiOS
- Follow instructions at [WPiOS PR](https://github.com/wordpress-mobile/WordPress-iOS/pull/10730)
https://github.com/wordpress-mobile/gutenberg-mobile/issues/436 needs to be tested with WPiOS

Test 1 - Test with long text
This tests https://github.com/wordpress-mobile/gutenberg-mobile/issues/432
- Add a paragraph block and add a text that is longer than the viewable area when keyboard is open
- Close keyboard if open
- Align top of the paragraph block to top of the screen
- Tap at the first line
- Verify that no scrolling occurs because caret is already in the viewable area
- Change orientation and verify caret is still visible

Test 2 - Test with the last line
- Use a paragraph block with some multiple line of input
- Close keyboard if open
- Align the block to the bottom of the screen
- Tap to the last line
- Verify that scroll position is adjusted and inner toolbar is also visible
- Change orientation and verify caret is still visible

Test 3 - Test with a line other than the last line
- Use a paragraph block with some multiple line of input
- Close keyboard if open
- Align the block to the bottom of the screen
- Tap to some line other than the last line
- Verify that scroll position is adjusted to make caret visible only
- Change orientation and verify caret is still visible

Test devices should include:
- iPhone X variety device
- iPad 
- Either one of iPhone 6-7-8

Test 4 - Android
- Android side wasn't affected so a simple test would be enough
Open Android example app and verify it opens successfully


iamthomasbishop:
This looks so great in the preview – nice work! 👏

pinarol:
Great catches! @etoledom I've opened issues for those https://github.com/wordpress-mobile/gutenberg-mobile/issues/463 https://github.com/wordpress-mobile/gutenberg-mobile/issues/464

464 will be a whole separate issue, currently we only have a mechanism to adjust the scroll on keyboard open/close. We will need a new solution for that.

I'll look into 463 separately, I am guessing it is a natural result of closing and opening keyboard between two blocks(it happens so fast so it is not noticeable). When keyboard closes the scroll position tries to go back to normal. Maybe I can have a solution for it, it'd be nice to get rid of that glitch.

PR #462, opened 2019-01-03
README: Update Link & Typo

torres126:
Noticed a link wasn't working, so this PR proposes updating it to one which _does_ work, but I'm not sure if that's the place it's supposed to redirect to? Also fixes a typo where an extra "the" is present.

New link: https://github.com/react-community/create-react-native-app/blob/fc1ef3c74626191ad4c610cc57cc8ebe97104021/react-native-scripts/template/App.test.js

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,7 +107,7 @@ Then, open `chrome://inspect` in Chrome to attach the debugger (look into the "R
 
 ## Writing and Running Tests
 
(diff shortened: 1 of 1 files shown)


hypest:
Thanks for opening a PR on this @torres126 !

Apparently, the `create-react-native-app` tool doesn't live in the original repo anymore and I think that pointing to the now legacy template file is not super useful. Instead, may I suggest changing the link to just point to an example test file in this very repo directly? For example, we have our own App.test.js inside the `src/app` folder we could link to. What do you think?

torres126:
Seems good to me! Do you know if I could just put `src/app/App.test.js` instead or does that not really matter? I tried it but it seems to link to the one in my fork for some reason, which isn't ideal. :/

torres126:
Sure, any ideas on what to? How about `See an example test here.`?

torres126:
Added! :+1:

torres126:
Thanks @hypest! :)

PR #465, opened 2019-01-03
Fold aztec wrapper in main repo

### LINKED ISSUE #460: Fold react-native-aztec into main gutenberg-mobile repo
Currently, the Aztec wrapper project lives in [its own repo](https://github.com/wordpress-mobile/react-native-aztec). Although the separation of concerns is generally useful, the wrapper is highly coupled with the main project and it ends up being a constant pain to do development with it. Main problem is the need to always open at least 2 PRs (one in the wrapper and one in the main project) to develop and test a feature.

On the other hand, it being a separate repo helps with having JitPack compile a library out of the wrapper.

It would be better though if we find a way to stop having the wrapper as a separate repo to help us with development turnaround time.

hypest:
### Fixes #460 

This PR replaces the `react-native-aztec` Aztec wrapper project submodule, putting in its place a `git subtree`'d version of its files and git history.

Followed these steps to remove the submodule: https://gist.github.com/myusuf3/7f645819ded92bda6677
And used this command to introduce the subtree: `g subtree add --prefix=react-native-aztec git@github.com:wordpress-mobile/react-native-aztec.git master`.

Also made changes to the Android side of the bridge and react-native-aztec projects to have them compile successfully on JitPack as before, but as a single (multi-module) project instead of separate repos.

With this PR, development of the react-native-aztec subproject can happen directly in the gutenberg-mobile repo, without the need to open a PR at https://github.com/wordpress-mobile/react-native-aztec. We can deprecate that repo soon if we want to.

Note: this PR has lots of commits, but all the ones dated in 2018 are actually carbon copies of the `react-native-aztec` commits from the separate repo. Only the ones dated on Jan 3rd (today at the time of writing) are new and are related to the work to make JitPack export 2 different Android modules: "gutenberg-mobile" which is the bridge project and `react-native-aztec` which is the Aztec wrapper.

### Test 1:
Run the gutenberg-mobile demo app as normal. It should work without needing to update a react-native-aztec submodule (there's no such git submodule anymore)

### Test 2 (Android):
Use wpandroid and point the gutenberg-mobile submodule to this PR. The build should work as normal.

### Test 3 (iOS): (not ready yet, needs iOS dev)
Build the wpios app. It should work as normal.


etoledom:
@hypest @diegoreymendez ,
I made a small change to expose `RNTAztecView.podspec.json` and load `RNTAztecView` from within this repo.

If there is a version update to `RNTAztecView`, we will need to run `./bin/generate-podspecs.sh` from `gutenberg-mobile` root folder to update that `.json` file.

I also pushed a commit implementing this change in WPiOS `try/integrate-gutenberg-without-aztec-wrapper-submodule-with-subtree`

One thing that I don't like is that our `RNTAztecView.podspec.json` is under the `third-party-podspecs` directly being it ours, but it plays nicely with the current pods implementation.

react-native-aztec/RNTAztecView.podspec
```diff
@@ -0,0 +1,22 @@
...
+package = JSON.parse(File.read(File.join(__dir__, 'package.json')))
+
+Pod::Spec.new do |s|
+  s.name             = 'RNTAztecView'
+  s.version          = package['version']
+  s.summary          = 'Aztec editor for React Native'
+  s.license          = package['license']
+  s.homepage         = 'https://github.com/wordpress-mobile/gutenberg-mobile'
+  s.authors          = 'Automattic'
+  s.source           = { :git => 'https://github.com/wordpress-mobile/gutenberg-mobile.git', :branch => 'feature/fold-aztec-wrapper-in-main-repo'}
```
etoledom: The `:branch` specification here is to make WPiOS branch `try/integrate-gutenberg-without-aztec-wrapper-submodule-with-subtree` work. It should be removed before merging this PR and update the `podspect.json` using `$./bin/generate-podspecs.sh`

koke:
A bit late to the conversation but would cocoapods subspecs help here? Instead of having a separate podspec in a subdirectory (which is slightly problematic for versioning), we can add have `Gutenberg/Bridge` and `Gutenberg/Aztec` in `Gutenberg.podspec`.

I'm not sure if it's worth the effort, but wanted to share the idea.

mzorz:
Nice work @hypest ! Gave it a quick look, found out the **Test 1** (`yarn run android`) gives this when trying on a clean checkout:

```
* Where:
Build file '/Users/mariozorz/proyectos/tests/gutenberg-mobile/react-native-aztec/android/build.gradle' line: 88

* What went wrong:
A problem occurred evaluating project ':react-native-aztec'.
> Cannot get property 'buildGutenbergFromSource' on extra properties extension as it does not exist
```
Tried a couple of things without luck though, so thought of at least reporting here but afraid I can't offer any solution as a follow up 😞

hypest:
Thanks for the report @mzorz ! I pushed aa88979 with a fix. The issue was that the root project (in `yarn android`'s case it's the demo app) needs to set the flag to some value. Copied the logic from the wpandroid for this one so, the flag can be set via the gradle.properties file too.

hypest:
Updated from develop to resolve conflicts.

The update commit is 0de87da and then fcfa370 is for updating from react-native-aztec too.

That update was rather painful in general since the submodule is removed from this branch and I had to figure out how to wrangle git to convince it to keep the commit history adopted from the subtree.

For the record, here are the steps I think I performed to do the update:
```
git merge develop

# Resolve conficts. Chosen the Gutenberg hash from develop but the submodule deletion from local
git mergetool 

# Ignore the Aztec wrapper changes that came from develop
git rm --cached react-native-aztec

# finish the merge
git commit

# get the latest commits from the react-native-aztec master
git subtree pull --prefix=react-native-aztec git@github.com:wordpress-mobile/react-native-aztec.git master
```

hypest:
👋 @etoledom , @diegoreymendez , @koke . Are you all content with the iOS side of changes in this PR so we can merge it? Ideally, is the solution in this PR workable down the road and hopefully not a deadend? Also, will it still be an improvement over how we develop the Aztec wrapper so far (the separate repo)?

etoledom:
I made a test of @koke's idea about using subspec and looks promising (
https://github.com/wordpress-mobile/gutenberg-mobile/pull/465#issuecomment-454358981)

I could send an update on early Monday, but it can be a later improvement too (future PR).

etoledom:
This was my test:
https://github.com/wordpress-mobile/gutenberg-mobile/compare/feature/fold-aztec-wrapper-in-main-repo-updated-from-develop...feature/aztec-wrapper-subspec

hypest:
> but it can be a later improvement too 

I'd personally favor a future PR so we avoid having to update this PR from develop again. It's not exactly fun :)

etoledom:
I found a problem while testing the subspec approach.
The `import RNTAztecView` from `/react-native-gutenberg-bridge/ios/Gutenberg.swift` fails on WPiOS (dependencies installed with cocoapods), and it is solved by deleting that line.

But that import is required to build the example project.

I'm not sure how to solve that problem. Maybe we could use cocoapods in the example project?

Anyway, I think this PR is good to go, and as mentioned, we can improve the iOS side later.

hypest:
Thanks for the update @etoledom!

I think we should go ahead and merge this PR. Also, regarding https://github.com/wordpress-mobile/gutenberg-mobile/pull/465#discussion_r245275007, can you do that change before we merge?

etoledom:
Sure! I'll make a PR for that as soon as this one is merged 👍

etoledom:
Should be ready to 🛳

PR #478, opened 2019-01-17
Upload media file in iOS

SergioEstevao:


SergioEstevao:
> ```swift
> if progress.fractionCompleted > 0.5 {
>       gutenberg.mediaUploadUpdate(id: mediaID, state: .failed, progress: Float(progress.fractionCompleted), url: nil)
>       timer.invalidate()
> }

@etoledom I was thinking of implementing logic on the code to make it alternate between success and fail on each upload, what do you think?

etoledom:
> I was thinking of implementing logic on the code to make it alternate between success and fail on each upload.

I don't think that it is really necessary, and it might be confusing for other people not aware of that detail. 
Maybe it is better to figure out a way to test those extra cases on unit tests? Even though the bridge itself is quite simple.
If we decide this, I'd say that we could implement them on a separate PR since we might need to change a couple of things outside the bounds of this PR.

PR #481, opened 2019-01-21
Make Aztec send height information to JS after pasting text

### LINKED ISSUE #438: Pasting issues on Paragraph and Heading blocks
### Paragraph block
When pasting a big block of text, the paragraph block won't refresh its size to contain the new amount of text. It is fixed after tapping on the block (as shown in the gif).

paste

### Heading block
In this case, the block does refresh its size but: 
- Pasting using the normal `Paste` button, the heading style is lost.
- Pasting using the `Paste without formatting` seems to work fine.

paste_heading

etoledom:
Fixes #438 

This is a small PR that forces layout when the user paste text. I noticed that `layoutSubviews` is called every time `textViewDidChange` is called, except for pasted text. Adding an explicit call to `setNeedsLayout()` solves the issue.

I tested that `layoutSubviews` is not called more than once for every text change, and now it's properly called after pasting text.

pasting

**To test:**
- Test pasting a long text in a paragraph block.
- Check that its heights grows accordingly.

### DIFF
--- a/react-native-aztec/ios/RNTAztecView/RCTAztecView.swift
+++ b/react-native-aztec/ios/RNTAztecView/RCTAztecView.swift
@@ -302,6 +302,8 @@ extension RCTAztecView: UITextViewDelegate {
         forceTypingAttributesIfNeeded()
         propagateFormatChanges()
         propagateContentChanges()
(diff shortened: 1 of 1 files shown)


etoledom:
Thank you!

PR #483, opened 2019-01-21
Ability to attach ReactRootView to different Activity

hypest:
This is useful for retaining the RN view/context while rotating the device (and letting the Activity get recreated) or even when interested in attaching the view to a completely different Activity (say, preloading RN before the editor Activity appears).

RN is still able to adapt to the new Activity's layout, which is super useful for portrait/landscape/etc.

This PR is not opened for merging yet. Rather, it's purpose is to be able to try it easily and have a discussion.

### Implementation Details

The solution here is based on `MutableContextWrapper` with its ability to change the Context after the ReactRootView has already been created. RN will still receive lifecycle events before and affter the change since those are propagated.

After the Context gets changed, we need to remove the View from its previous parent (if any, happens on rotation) and add it to a new parent.

### To test

Use this PR on wpandroid https://github.com/wordpress-mobile/WordPress-Android/pull/9030.


react-native-gutenberg-bridge/android/src/main/java/org/wordpress/mobile/WPAndroidGlue/WPAndroidGlueCode.java
```diff
@@ -94,33 +103,35 @@ public void responseHtml(String title, String html, boolean changed) {
...
                 mPendingMediaUploadCallback = mediaUploadCallback;
-                onMediaLibraryButtonListener.onCapturePhotoButtonClicked();
+                mOnMediaLibraryButtonListener.onCapturePhotoButtonClicked();
             }
 
             @Override
             public void mediaUploadSync(MediaUploadCallback mediaUploadCallback) {
                 mPendingMediaUploadCallback = mediaUploadCallback;
-                onReattachQueryListener.onQueryCurrentProgressForUploadingMedia();
+                mOnReattachQueryListener.onQueryCurrentProgressForUploadingMedia();
```
daniloercoli: Wondering if there could be a case where this listener can be called before the view is attached to a container. In that case we need to check that are not null(s)?

react-native-gutenberg-bridge/android/src/main/java/org/wordpress/mobile/WPAndroidGlue/WPAndroidGlueCode.java
```diff
@@ -167,15 +176,48 @@ public void onReactContextInitialized(ReactContext context) {
...
+        viewGroup.addView(mReactRootView, 0,
+                new LayoutParams(ViewGroup.LayoutParams.MATCH_PARENT, ViewGroup.LayoutParams.MATCH_PARENT));
+
+        refocus();
+    }
+
+    private void refocus() {
+        if (mLastFocusedView != null) {
+            // schedule a request for focus
+            new Handler(Looper.getMainLooper()).post(new Runnable() {
```
daniloercoli: As mentioned in chat convo this could introduce some lag, but in my opinion it's unlikely to happen on "normal" workflow. Even it that happens, the focus/caret is restored properly with just a bit of delay.

daniloercoli:
I wasn't fully aware of what were the benefits of `MutableContextWrapper` till today.
I googled it up a bit and found out the following warning:

`Definitely not safe. There are actions on Context that require counter-actions in order to free up resources, e.g. Context.registerComponentCallabacks, Context.registerReceiver, and if you change the Context in the meantime, that means the counter-actions will not be called for the initial context, resulting in resource leaks. `

Not quite sure if this can affect us in some way, since we could set the Title and the Content, and then start the RN application. Later we can attach it to a new view, and a different context.

hypest:
I think we will need to keep an eye on this solution anyway. It's hacky from its start. I mean, trying to retain _a view_ while its Activity is rotating is quite against the Android best practices.

It's only a necessity because we want to not touch the Aztec Activity (EditPostActivity over at the wpandroid repo) and continue to allow it to recreate on rotation. We'll want to revise that preference (to not touch Aztec-standalone at all) down the road as Gutenberg-mobile gets more stable.

PR #486, opened 2019-01-22
Re-introduce the react-native-aztec submodule to make JitPack green

hypest:
This PR re-introduces the git submodule [we removed recently](https://github.com/wordpress-mobile/gutenberg-mobile/pull/465), ~but putting it in an obscure place~.

We need to do *no* work to update this submodule at all. Its code is not used anywhere and is only here because JitPack has trouble if it sees it go. For reference: https://github.com/jitpack/jitpack.io/issues/3707.

We'll remove this submodule again, after the problem is fixed on the JitPack side of things.

To test:

Everything should work as normal, but this time `git submobule update --init` will pull in the "new" submodule. Let's do some testing with the demo app and the main WP apps against this one just to be sure. Thanks!

For wpandroid, there's a PR up that be used for testing: https://github.com/wordpress-mobile/WordPress-Android/pull/9052


hypest:
Oh no, 6840b35f4a was succeeding on JitPack but the follow up commit don't 🤦‍♂️. There might be an element of randomness here.

daniloercoli:
LGTM!

PR #488, opened 2019-01-23
Makes the media id used on uploads an Int.

SergioEstevao:


ios/gutenberg/MediaPickAndUploadCoordinator.swift
```diff
@@ -41,7 +41,7 @@ class MediaPickAndUploadCoordinator: NSObject, UIImagePickerControllerDelegate,
     }
     do {
       try data.write(to: url)
-      mediaCallback(mediaID, url.absoluteString)
+      mediaCallback(mediaID.hashValue, url.absoluteString)
```
mzorz: Was reading this:
 https://developer.apple.com/documentation/swift/hashable/1540917-hashvalue

> Hash values are not guaranteed to be equal across different executions of your program. Do not save hash values to use during a future execution


Does it have any implications then, for example if you're uploading something and then close the app and open it again?

PR #491, opened 2019-01-23
Fix/upload media file media

mzorz:
This PR just tidies up the passing of `mediaServerId`, making it clear that it's only known and used at the end of the upload, rather than passing a magic value of 0 it introduces a constant to make it more readable.

Side note: we should also rename that to `serverMediaId` to actually  mean what it carries, but this needs be done in coordination with iOS.


marecar3:
Nice work @mzorz! Thanks for the support on this one!

PR #493, opened 2019-01-24
Add Enter.key detection to Title block

daniloercoli:
This PR adds a new empty block when Enter.key is pressed in the Title block, and gives the ability to add a new block after it.

Note: This PR does hide the keyboard at the moment, it's a default behavior (at least) on Android., of the way we detected Enter.key. We may need to migrate the Title to RichText if there is no way to circumnavigate this problem.

Note 2: At the moment there is not split in place for the title filed. If you tap enter on the middle of the field there is no split yet.

We need to update the GB ref after the merge of the companion GB pull request here: https://github.com/WordPress/gutenberg/pull/13500


diegoreymendez:
@daniloercoli - I'm retargetting this one as the main one was already merged.

PR #495, opened 2019-01-24
Bump Aztec Android to v1.3.17

marecar3:
Pointing to the latest released Aztec-Android version.

### DIFF
--- a/react-native-aztec/android/build.gradle
+++ b/react-native-aztec/android/build.gradle
@@ -12,7 +12,7 @@ buildscript {
         wordpressUtilsVersion = '1.22'
         espressoVersion = '3.0.1'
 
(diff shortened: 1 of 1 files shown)


hypest:
Hmm, my impression is that the Aztec changes are not perhaps ready to be used in mainline gb-mobile yet. For example, there's an open PR still that tries to introduce one of the changes: https://github.com/wordpress-mobile/gutenberg-mobile/pull/275

I think we don't need to ship new Aztec code as part of a release without develop already using that release.

My comment assumes that this PR is part of the release process and not actually trying to introduce any functional changes.

marecar3:
I am closing this PR in order to create a new one.

Issue #372, opened 2018-12-13
Post title writing flow


koke:
As part of #182, we added the post title in the apps themselves, which has lead to some issues. For the beta, we'd want:

- [x] Title should have no border
- [x] Title shouldn’t be fixed (it should scroll with the content)
- [x] Title and content should be aware of each other in terms of focus. When the title is selected, the toolbar should only contain global tools (Inserter, Undo/Redo, Hide Keyboard)
- [x] The "hide keyboard" button should work on the title as well
- [x] Pressing enter on the title should move to the next block or split a new one if necessary [https://github.com/wordpress-mobile/gutenberg-mobile/issues/502]
- [x] Integrate the Noto font and use it on blocks
- [x] Make sure undo / redo works.
- [x] The title should be [saved in the global store](
https://github.com/wordpress-mobile/gutenberg-mobile/pull/450#discussion_r248672329) instead of part of `AppContainer.js`'s state.
- [x] The title placeholder should read "Add a Title". ([ref](
https://github.com/wordpress-mobile/WordPress-iOS/pull/10794#issuecomment-455329479))

hypest:
Allow me to add that it would be nice if we could timebox (the least) an experiment around this: render the title inside the RN side, effectively introducing the first mobile non-content block.

diegoreymendez:
All tasks are completed... this issue will be closed once all PRs are merged.

PR #501, opened 2019-01-25
Uses the global store for the post title and adds undo/redo support.

diegoreymendez:
## Description:

Following @Tug 's indications [here](
https://github.com/wordpress-mobile/gutenberg-mobile/pull/450#discussion_r248672329), this PR wires our title to the global store title.

This PR also adds undo/redo support for the title.

Moves forward #372 .

## Dependencies:

Gutenberg PR: https://github.com/WordPress/gutenberg/pull/13514

### Testing edits:

1. Launch the demo app.
2. Make sure the initial title is properly set up.
3. Edit the title, switch to HTML mode and back.

Make sure the title is still as you left it before switching modes.

### Testing undo redo:

1. Edit the title as you see fit.
2. Use the undo / redo buttons from the toolbar.

Make sure undo / redo works as it does in web-Gutenberg (ie: that's our baseline for comparing).


diegoreymendez:
Fyi, @daniloercoli

diegoreymendez:
Reassigned to @daniloercoli since he's working on the title feature with me.

src/app/AppContainer.js
```diff
@@ -92,7 +91,7 @@ class AppContainer extends React.Component<PropsType, StateType> {
 				toggleHtmlModeAction={ this.toggleHtmlModeAction }
 				setTitleAction={ this.setTitleAction }
```
pinarol: Hey @diegoreymendez 👋 it looks to me that we can also get rid of passing this.setTitleAction as a prop from component to component. So other components can also use editPost( { title: title } ) to via their own props with the help of withDispatch. what do you think?

diegoreymendez: That's a good observation.  Thanks @pinarol!

Considering this PR is working as expected right now though, I've created [an issue](https://github.com/wordpress-mobile/gutenberg-mobile/issues/512) that I will be tackling right after this is complete.

As usual my goal is to split the work into atomic steps where necessary.

PR #505, opened 2019-01-25
Media upload progress reattach android

mzorz:
This PR adds the needed interfaces to make Gutenberg start receiving upload progress updates when it's shown on the screen, for images that exist within the open Post that are currently being uploaded.

This can be tested with the WPAndroid PR https://github.com/wordpress-mobile/WordPress-Android/pull/9129

Related Gutenberg PR:
https://github.com/WordPress/gutenberg/pull/13516

#### Notes:
- refactored the `MediaUploadCallback` implementation a bit as it's needed to be passed to 3 interfaces  (`onCapturePhotoPressed`, `onUploadMediaPressed`, and now also `onImageQueryReattach`)
- this new interface allows the listener to be set on the bridge, updates were already being listened to
- also, the hosting app can start sending latest max progress updates once the Image component calls the new interface  from `componentDidUpdate()` in https://github.com/WordPress/gutenberg/pull/13516 instead of waiting for the next update (this is useful for slow connections where there would be a gap between the time the user opens the Editor and an update is actually seen)

### DIFF
--- a/react-native-gutenberg-bridge/index.js
+++ b/react-native-gutenberg-bridge/index.js
@@ -52,4 +52,8 @@ export function onCapturePhotoPressed( callback ) {
 	return RNReactNativeGutenbergBridge.onCapturePhotoPressed( callback );
 }
 
+export function onImageQueryReattach( callback ) {
+	return RNReactNativeGutenbergBridge.onImageQueryReattach( callback );
+}
+
 export default RNReactNativeGutenbergBridge;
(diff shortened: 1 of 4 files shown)


react-native-gutenberg-bridge/index.js
```diff
@@ -52,4 +52,8 @@ export function onCapturePhotoPressed( callback ) {
 	return RNReactNativeGutenbergBridge.onCapturePhotoPressed( callback );
 }
 
+export function onImageQueryReattach( callback ) {
```
SergioEstevao: Shouldn't we send up what is the ID of the media we want to reattach?

mzorz: It's not necessary to be passed here at all, as all we need to do is to tell the parent app which the new `callback` is so it can be set in [this line here](https://github.com/wordpress-mobile/gutenberg-mobile/pull/505/files#diff-f2e1dfe800569dfb4ee4830f88779890R110).
Later the events are being broadcasted and there we tell the listeners which progress is being made for which `mediaId`.

Issue #512, opened 2019-01-28
Standardize the mechanism for editing the post title.


diegoreymendez:
As suggested by @pinarol [here](
https://github.com/wordpress-mobile/gutenberg-mobile/pull/501#pullrequestreview-197044450), we should remove `setTitleAction` and make sure our code edits the title through calls to `editPost( { title: title } )`.

Edit: actually removing `setTitleAction` completely right now is not the best option, as `MainApp.js` can receive a title change action from native, and editing data from the storage is not within it's scope.  Check the referenced PR for the actual changes.

diegoreymendez:
Closed without merging as the changes didn't really reflect the initial intention, and were not as clean. as I'd hoped.

PR #514, opened 2019-01-28
Standardizes the post title editing logic.

diegoreymendez:
## Description:

Standardizes the mechanism for editing the post title.

Fixes https://github.com/wordpress-mobile/gutenberg-mobile/issues/512

## Testing:

1. Edit the title
2. Switch to HTML mode and back to visual.

Make sure the title is still as modified.

### DIFF
--- a/src/app/AppContainer.js
+++ b/src/app/AppContainer.js
@@ -18,6 +18,7 @@ type PropsType = {
 	title: string,
 	initialTitle: string,
 	initialHtml: string,
+	editTitle: string => mixed,
 	resetBlocks: Array<BlockType> => mixed,
 	setupEditor: ( mixed, ?mixed ) => mixed,
 	toggleBlockMode: ?string => mixed,
@@ -82,6 +83,7 @@ class AppContainer extends React.Component<PropsType> {
 		return (
 			<MainApp
 				rootClientId={ this.props.rootClientId }
+				editTitle={ this.props.editTitle }
 				serializeToNativeAction={ this.serializeToNativeAction }
 				toggleHtmlModeAction={ this.toggleHtmlModeAction }
 				updateHtmlAction={ this.updateHtmlAction }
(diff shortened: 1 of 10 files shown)


src/app/AppContainer.js
```diff
@@ -87,6 +87,7 @@ class AppContainer extends React.Component<PropsType> {
 		return (
 			<MainApp
 				rootClientId={ this.props.rootClientId }
+				editTitle={ this.props.editTitle }
 				serializeToNativeAction={ this.serializeToNativeAction }
 				toggleHtmlModeAction={ this.toggleHtmlModeAction }
 				setTitleAction={ this.setTitleAction }
```
pinarol: do we need to pass both setTitleAction and editTitle? it looks to me like setTitleAction can be eliminated. Also the function itself?
```
	setTitleAction = ( title: string ) => {
		this.props.editTitle( title );
	};
```

src/block-management/block-manager.js
```diff
@@ -291,13 +291,17 @@ export default compose( [
...
 			insertBlock,
 			replaceBlock,
 			selectBlock,
 		} = dispatch( 'core/editor' );
 
 		return {
 			insertBlock,
+			editTitle( title ) {
+				editPost( { title: title } );
+			},
```
pinarol: this might be moved to MainApp.js, what do you think? it looks like we don't call this directly from here, we just pass it to MainApp

diegoreymendez: This is called right above in line 169.  The code in here makes `editTitle` become a prop of this object.

src/app/AppContainer.js
diegoreymendez: Removed `editTitle` which wasn't meant to be there.  This PR got a bit confusing since I couldn't really completely remove `setTitleAction`.

diegoreymendez:
Closing without merging, as these changes didn't turn out to be as clean as I'd hoped at first.

PR #526, opened 2019-01-30
Bundle the Android JS file inside the bridge lib

hypest:
This PR removes the dependency on Travis+AWS to produce the Android JS bundle and instead, tasks JitPack to produce and include the JS file inside the Bridge library's assets folder. The parent app can then access the file automatically because of resources merging.

h/t to @jtreanor for the assets bundling idea!

To test:

Use this wpandroid side PR: https://github.com/wordpress-mobile/WordPress-Android/pull/9156

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,15 +1,3 @@
-addons:
-  artifacts:
-    debug: true
-    s3_region: "us-west-1"
-    target_paths:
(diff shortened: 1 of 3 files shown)


hypest:
Thanks for the review @jtreanor !

> node tools could be used directly

Aha, what do you have in mind there?

Issue #419, opened 2018-12-20
[WPAndroid] fonts are not the same as in iOS


mzorz:
First reported by @pinarol 

GB on Android is not loading Noto Serif apparently

<img width="677" alt="screen shot 2018-12-20 at 17 07 19" src="https://user-images.githubusercontent.com/6597771/50290138-86ab2e80-0449-11e9-82df-a88dec5b8304.png">

hypest:
This ticket was actually worked on in #530 so, moving this to in-progress.

hypest:
Fixed with https://github.com/wordpress-mobile/gutenberg-mobile/pull/513

PR #530, opened 2019-02-01
Remap font names for iOS for PR 372.

diegoreymendez:
## Description:

This PR simply remaps some font names for an open PR so that the work for iOS as well.

iOS unfortunately doesn't support font families `serif` and `monospace`.

Other changes will be needed but these are the basic changes before implementing the iOS native font properties.

## Testing:

The only testing needed is to make sure the font are good for Android too.

### DIFF
--- a/src/block-management/block-holder.scss
+++ b/src/block-management/block-holder.scss
@@ -70,7 +70,7 @@
 	padding-top: 24;
 	padding-bottom: 24;
 	font-size: 16px;
-	font-family: serif;
+	font-family: 'NotoSerif';
 }
 
 .aztec_container {


daniloercoli:
`NotoSerif` doesn't work on Android.
A viable solution is to set the font family by checking the current Platform. See the GIST below:
https://github.com/WordPress/gutenberg/pull/13628#issuecomment-459683383

diegoreymendez:
Tests are failing but I've no idea why.  Been digging into it without much success so far, so any idea is welcome.

hypest:
I think this PR pretty much addressed #419 too.

PR #531, opened 2019-02-01
Alt image setting

etoledom:
This PR updates the Gutenberg reference to test the Alt image setting.
The main PR is found here: https://github.com/WordPress/gutenberg/pull/13631

To test, refer to the `gutenberg` side PR.

**(Updated GIF)**
alt-setting

cc @iamthomasbishop


koke:
The right-aligned text field looked good when static, but feels weird while typing. @iamthomasbishop what do you think?

iamthomasbishop:
@koke I agree it's not perfect, but there are a number of considerations here and for the beta the alternatives aren't better, imo. I'd like to play with this in a test build before deciding to pivot. Some alternatives that come to mind:

* Left-align values: this causes other visual alignment issues
* Use two-line table-row style: sheet becomes much taller, not ideal

Long-term solution in my mind would be to iterate on this to utilize native table navigation/input patterns, for example:

* Android: when table row is tapped, show dialog for text entry
* iOS: show disclosure (`>` icon on right side of row, allow user to go down in hierarchy to a sub-page for text entry

Another option would be to stack a fresh sheet on top for entry, but that sounds like it could get visually disjointed if we start pulling sheets on top of sheets on top of sheets :)

Issue #203, opened 2018-10-29
i18n support


koke:
We want gutenberg-mobile to be localized to the languages we support in the apps. This includes a few challenges that we need to solve:

- Where are the translations coming from? Currently from the Gutenberg plugin, after 5.0 likely from core
- Do we combine strings with the apps or keep the translations separate?
- Do we need any special configuration to have React Native pick up the right language?
- The original/translated strings doesn't seem to be in the repo. What changes to build scripts do we need to test and release localizations?

Master PR: https://github.com/wordpress-mobile/gutenberg-mobile/pull/536

koke:
From @hypest:

Assumptions:

- stay close to GB-web
- pick strings from GB
- Minimal bridge interface, if needed at all
- Get device/site language from parent

Subtasks (all include some learning/research):

- script fetching translations from GlotPress (3d)
- Bridge interface to pass the device/site language setting (1d)
- convert .po from GlotPress to GB-usable translation data (2d)
- bootstrap GB (@wordpress/i18n packacge) with translations (5d)
- test out outputing some translating string in the UI (1d)

extra Subtask:

- make sure all current UI strings in GB-mobile pass through the translation function (2d)

Estimation: 14d of effort, closing in to **4 weeks**.

koke:
For this one, once we have more clarity on what we need to make it work, let's do a technical doc before implementing this so we're all clear on where translations come from and how they will be updated and loaded.

Tug:
> convert .po from GlotPress to GB-usable translation data (2d)

We don't need to, we can access JSON versions of the translations at `https://widgets.wp.com/languages/${ project }/${ locale }.json`

The data downloaded is ready to be passed down to `setLocaleData` here: https://github.com/WordPress/gutenberg/blob/master/packages/i18n/src/index.js#L45

So
> bootstrap GB (@wordpress/i18n packacge) with translations (5d)

Shouldn't be as long either

koke:
Also GlotPress allows export to JSON (in a few different flavors), https://translate.wordpress.org/projects/wp-plugins/gutenberg/dev/fr/default

PR #536, opened 2019-02-04
Use gutenberg translation files locally

Tug:
Addresses part of #203 
Requires https://github.com/wordpress-mobile/WordPress-Android/pull/9249
Requires https://github.com/wordpress-mobile/WordPress-iOS/pull/11056

~~Requires those 2 PRs to be merged first so the gb-mobile strings get translated in WPiOS and WPAndroid glotpress:
https://github.com/wordpress-mobile/WordPress-iOS/pull/10999
https://github.com/wordpress-mobile/WordPress-Android/pull/9214~~

### Summary
This PR adds a local script that downloads gutenberg translation files and cache them locally.

### Testing Instructions
- Load this branch locally
- Run `yarn start` and check that translation files are downloaded in `i18n-cache/data` and that a file is generated at `i18n-cache/index.native.js`
- Edit `src/app/App.js` and replace `setLocaleData( getTranslation( locale ) );` with `setLocaleData( getTranslation( 'es' ) );`
- Reload the app and check that parts of the app have been translated to Spanish

- Reset the code and run `yarn wpandroid` or `yarn ios`
- Change the app or the phone language
- Check that gutenberg editor is translated

A PR in both https://github.com/wordpress-mobile/WordPress-iOS and https://github.com/wordpress-mobile/WordPress-Android will be necessary to pass the new `locale` parameter.


react-native-gutenberg-bridge/ios/GutenbergBridgeDataSource.swift
```diff
@@ -24,4 +24,10 @@ public protocol GutenbergBridgeDataSource: class {
     ///
     /// - Returns: The locale slug value or nil.
     func gutenbergLocale() -> String?
+    
+    /// Asks the data source for the list of localized strings to be used by the editor.
+    /// Return `nil` if no localization file is present for the current locale
+    ///
+    /// - Returns: Gutenberg related localization key value pairs for the current locale.
+    func gutenbergTranslations() -> [String : String]?
```
Tug: Shouldn't it be `Any` as well here in the end?

ios/gutenberg/GutenbergViewController.swift
```diff
@@ -112,6 +112,15 @@ extension GutenbergViewController: GutenbergBridgeDelegate {
 }
 
 extension GutenbergViewController: GutenbergBridgeDataSource {
+    
+    func gutenbergLocale() -> String? {
+        return "en"
+    }
+    
+    func gutenbergTranslations() -> [String : String]? {
```
Tug: So I would guess here as well

react-native-gutenberg-bridge/ios/GutenbergBridgeDataSource.swift
pinarol: Didn't know values were expected to be arrays, just pushed a new commit bb606d4 to change this

ios/gutenberg/GutenbergViewController.swift
pinarol: addressed

i18n-cache/index.js
```diff
@@ -0,0 +1,128 @@
...
+	'ro', // Romainian
+	'ru', // Russian
+	'sk', // Slovak
+	'sq', // Albanian
+	'sv', // Swedish
+	'th', // Thai
+	'tr', // Turkish
+	'zh-cn', // Chinese (China)
+	'zh-tw', // Chinese (Taiwan)
+];
```
hypest: Heh, can we add `el` too? I'm biased :P

src/app/App.js
```diff
@@ -1,31 +1,63 @@
 /** @flow
  * @format */
+
+import '../globals';
```
hypest: Do we need to import `globals` here too? 🤔

src/app/App.js
```diff
@@ -1,31 +1,63 @@
...
-	return (
-		<AppContainer
-			initialHtml={ initialData }
-			initialHtmlModeEnabled={ initialHtmlModeEnabled }
-			initialTitle={ initialTitle } />
-	);
-};
+	setLocale( locale: ?string ) {
+		const translationsFromParentApp = this.props.translations || {};
+		const translations = Object.assign( {}, translationsFromParentApp, getTranslation( locale ) );
```
hypest: Hmm, I'm thinking, how about giving the WP apps priority here? Like, if there's a conflict between what the parent sends and what Gutenberg has then let's give the parent priority. That would also give us the ability to have each mobile platform decide to overwrite a string, for whatever reason (e.g. platform guidelines). WDYT?

Tug: It sounds sane. The only thing I'm afraid of though is mixing english and an other language in the case of using a language which is supported by the app but not by gutenberg for instance. 

If you don't think it's a problem I can give the parent app priority

src/app/App.js
Tug: Shouldn't be required, removing it 👍

src/app/App.js
hypest: Hmm, won't that be true the other way too? Like, if some strings are translated by GB but the native specific ones are not, right?

Tug: Yeah, I went with your approach in the end and decided to mix both and give priority to the parent app 👍

hypest: All in all, I think that giving priority to the native side strings should be a good start. We can revise if we see it be problematic.

pinarol:
hey @etoledom 👋 it'd be ok if you just review the iOS bridge and iOS example app changes?

src/app/App.js
koke: I checked the current Spanish translations in both apps and Gutenberg for common strings, and everything seems pretty consistent. The only difference I've spotted that's perhaps relevant is how media library is translated, and it makes sense that we're consistent with the app translation, so good call 😁

ios/gutenberg/GutenbergViewController.swift
```diff
@@ -116,6 +116,15 @@ extension GutenbergViewController: GutenbergBridgeDelegate {
 }
 
 extension GutenbergViewController: GutenbergBridgeDataSource {
+    
+    func gutenbergLocale() -> String? {
+        return "en"
```
etoledom: We could send the device locale directly. Since is the example app is not a blocker, but it would nice to test translations just by changing the device language in the run settings for fast tests.
```suggestion
        return Locale.preferredLanguages.first ?? "en"
```

hypest:
Even though this PR is ready to merge, please don't merge until the PRs in the parent apps are also ready to merge. cc @Tug @pinarol @koke

koke:
When I run `yarn install` I'm seeing this

```Could not find translation file https://widgets.wp.com/languages/gutenberg/en-ca.json
Could not find translation file https://widgets.wp.com/languages/gutenberg/en-au.json
Could not find translation file https://widgets.wp.com/languages/gutenberg/nb.json
```

Is that expected?

hypest:
Yes, I think I remember @Tug mentioning this to me on a voice chat that not all locales supported by the apps (where the original list of locales came from) are supported by Gutenberg @koke .

koke:
I was mentioning this to Tug on slack, that Norwegian seems to be on Glotpress so maybe this is more of a WordPress.com think. It might be a good idea to switch to fetching from glotpress directly before the release. This seems to download translations but I haven't tested if it's the same JSON format:

```
curl "https://translate.wordpress.org/projects/wp-plugins/gutenberg/dev/nb/default/export-translations?format=json"
```

PR #544, opened 2019-02-06
Bring placeholder text back on RichText powered blocks

daniloercoli:
The issue of missing placeholder text was noticed and reported since the Title block is now using the RichText component, but the problem was there on other blocks already. See Para and Heading.

This GB PR does fix the problem: https://github.com/WordPress/gutenberg/pull/13699

To test:
1. Start the demo app and try add blocks. The placeholder text should be visible on the para or heading block.

2. Use wp-android and point to this branch, and enable build from code ON. Start a new post. You should see the placeholder.


hypest:
By the way, I noticed that the vertical aligning of the placeholder is a bit off on iOS, when compared to the title text itself. See this gif:

title-placeholder-vertical-alignment

Not a blocker for this PR so, maybe @diegoreymendez can have a look?

Issue #202, opened 2018-10-29
Image settings


koke:
Implement the image inspector with image settings. Since we're not likely to have the block inspector in the beta, it's OK to present the image settings on a modal view or similar, we'll have to figure out  a temporary UI that looks good and it's easy to implement.

For reference, the image settings I'm seeing on the web:
- Alt text
- Image size
- Link settings

koke:
From @mzorz:

Assuming the UI is defined beforehand, I’d give this roughly 1 week (I’m actually thinking 1 to 2 days for implementation but I expect some back and forth with design iterations plus testing). What I’m thiking this internally does is:

- create the modal showing current settings for width/height and alt, and link settings
- with that, take the values and validate params, then re-write the html correspondingly when the user hits OK to accept any modifications they may have done
- re-render block if applies (probably this is done automatically, it’s React after all)

Estimation: **1 week**

etoledom:
> Create the modal showing current settings for width/height and alt, and link settings

Looking at the web version, that modal should look something like this:

![img_0028](https://user-images.githubusercontent.com/9772967/51486125-25f07800-1da0-11e9-97d1-ac16b21249e5.jpg)

I have the following question:

- Are all those controls required for this ticket?

And also looking at the web version, if we follow closely, the button that would make that modal appear would be a new "settings" button like this one:

<img width="347" alt="screen shot 2019-01-21 at 4 48 03 pm" src="https://user-images.githubusercontent.com/9772967/51486277-94cdd100-1da0-11e9-9273-90a1163fc886.png">

Leaving the edit "pencil" button to change the image.

cc @koke @hypest @iamthomasbishop

hypest:
We chatted over Slack with @iamthomasbishop and @etoledom and we think that the most "straightforward" route right now for the February Beta is to go full-RN with the image settings, since that will be less moving parts to wrangle and less developer resources to allocate. 🤞 

It's understood that full-RN will not give us the platform-native look and feel we'd like for this but, it's sacrifice we can iterate on afterwards.

etoledom:
First iteration of Image Settings is ready via all the PR mentioned [here](url
https://github.com/wordpress-mobile/gutenberg-mobile/issues/202#issuecomment-461617846)

PR #548, opened 2019-02-06
Add networking support

Tug:
This PR is required for task https://github.com/wordpress-mobile/gutenberg-mobile/issues/202

First it fixes how gutenberg is loaded on boot: we need to initialize some data middleware before calling our App component. Just like gutenberg is doing for the web in the plugin (server-side) part: https://github.com/WordPress/gutenberg/blob/master/lib/client-assets.php#L205

### TODO:
- [ ] Get an authentication token + a blog url param from the native app and use it to set up the apiFetch middleware. We can get some inspiration from calypso for this https://github.com/Automattic/wp-calypso/blob/master/client/gutenberg/editor/api-middleware/index.js#L199


Tug:
Thanks for reviewing @etoledom :)
Let's merge this and continue the work in another PR!

Issue #238, opened 2018-11-15
Pasting external content


koke:
Let's make sure we test pasting contents from another app and evaluate what we need to fix for the beta.

> Customer research has reveled that most people at least start their draft elsewhere

So far, pasting multiple paragraphs of text on iOS works visually, but:
- [x] they're not split into separate paragraphs
- [x] the block doesn't resize to fit the content until you type #438
- [ ] since I was pasting from a website, it inherited the custom font size. Should "Paste and match style" be the default for Gutenberg?

koke:
@daniloercoli @diegoreymendez as you work on split/merge, try to do some tests pasting content and log the issues here

hypest:
I think this ticket involves work on multiple levels so, allow me to split it up in smaller targets that might help us tackle it in steps:

# Target 1:
Be able to paste multi-paragraph *simple text* and having it split into multiple blocks. Think: copy multiple paragraphs of text that doesn't include rich text whatsoever (like bold/italics/headings/etc). Will need effort on Aztec (native) and RN the least.

# Target 2:
Be able to paste multi-paragraph *rich-text*. Think: paragraphs, headings, bold, italics. Blocks should pick up the styling from each of the copied sections. Will need effort on Aztec (native) and RN the least.

# Target 3:
Be able to paste rich-text along with images and more. Think: Targets 1, 2 plus images, unsupported blocks. Will involve effort on all layers (Aztec, RN, WP apps). Images pasted will need to get uploaded on the spot.

diegoreymendez:
Closing.  Let's open issues for any pending work.

PR #553, opened 2019-02-07
First steps to implement pasting external content

mkevins:
This PR aims to intercept a paste action in `ReactAztecText` and emit it as an event for consumption in JS. It is part of the work to address https://github.com/wordpress-mobile/gutenberg-mobile/issues/238. The goal is to serialize the clipboard contents (as text and html where possible). The event also includes the current text and the selection/cursor information to be handled in JS.

* `onTextContextMenuItem` method is called when the user selects paste from the context menu (typically after a long-press, or a touch of the cursor handle).
* This method is overridden to conditionally call `onPaste` if `shouldHandleOnPaste` is true.
  * `shouldHandleOnPaste` is mapped to JS and is truthy when the JS `onPaste` method is defined, and falsy otherwise.
* `onPaste` loops through all items in the primary clipboard, concatenating both text and html representations of them, and dispatches a JS event with the data.

I think it is reasonable to have `canCoalesce` return false as paste events will not be so frequent as to warrant it. _Frequent events from native code can be coalesced (merged) into one another to reduce their frequency (and load) on the JS side._

## Manual tests

There are currently no automated tests for this implementation, however, here are some manual tests that I've tried:

### Copy to the clipboard and paste as follows:

1. Paste into a new (empty) paragraph block
    * Expected behavior
      * Plain text
        * Inline: text inserted
        * Multi-line<sup>1</sup>: text inserted as a single paragraph
        * Multi-paragraph<sup>2</sup>: text split into multiple paragraph blocks
      * Styled text
        * Inline: styled text inserted
        * Multi-line: styled text inserted as a single paragraph
        * Multi-paragraph: styled text split into heading and paragraph blocks
        * Multi-paragraph (starting with heading): paragraph block replaced by heading block
      * Styled text with image
        * Multi-paragraph: styled text split into heading, image, and paragraph blocks
        * Multi-paragraph (starting with image): paragraph block replaced by image block
1. Paste into a paragraph block with content
    * Expected behavior
      * Plain text / Styled text
        * Inline: text inserted at cursor position
        * Inline (heading): text inserted as plain text at cursor position
        * Multi-line: text inserted at cursor as a single paragraph (splitting current paragraph if necessary)
        * Multi-paragraph: text split into multiple blocks at cursor position (splitting current paragraph if necessary)
      * Styled text with image
        * Multi-paragraph: styled text split into heading, image, and paragraph blocks (splitting current paragraph if necessary)
1. Paste into a paragraph block with content with some inner content selected
    * Expected behavior
      * Plain text / Styled text / Styled text with image
        * For all scenarios above: selected text removed and paste continued at cursor as above
      * URL
        * Selected text remains and is converted to link upon paste

1 - Multi-line => single paragraph - single '\n' characters
2 - Multi-paragraph => multiple consecutive '\n' characters

## Known issues

When bullet lists are copied to the clipboard and pasted to a paragraph block, the bullets are lost. Currently, list blocks are not supported. When support is added, behavior should be tested again.

### When pasting into heading blocks

When multi-paragraph / image content is pasted into an empty heading block, the heading block remains, and contains the text `null`.

When multi-paragraph / image content is pasted into a heading block with all text selected, the text remains, and the clipboard contents are added as blocks after the heading block.

When multi-paragraph / image content is pasted into a heading block with inner text selected, the selected text is removed, the text before the selection remains in the heading block, the clipboard contents are added as blocks after the heading block, and the text after the selection is added to a paragraph block below. I'm not sure whether this is the expected behavior.

### When pasting into code blocks

When multi-paragraph / image content is pasted into a code block, the block is not split, and new blocks are not created. Styled text (colored links) keep their color, and images are converted to an [object replacement character](https://en.wiktionary.org/wiki/%EF%BF%BC) (which renders as a little square).

### Cursor postition

After pasting, the cursor position is expected to be at the end of the pasted content. Currently, the cursor position does not follow this behavior. This may be related to https://github.com/wordpress-mobile/gutenberg-mobile/issues/602.

## Related PRs

* https://github.com/WordPress/gutenberg/pull/13841


react-native-aztec/android/src/main/java/org/wordpress/mobile/ReactNativeAztec/ReactAztecText.java
```diff
@@ -100,6 +105,15 @@ void addPlugin(IAztecPlugin plugin) {
         }
     }
 
+    @Override
+    public boolean onTextContextMenuItem(int id) {
+        if (id == android.R.id.paste && shouldHandleOnPaste) {
```
daniloercoli: Should we intercept the _Paste as plain text_ event here (`android.R.id.pasteAsPlainText`), and call `onPaste` passing the instructions to not send the HTML version of the content up to the event?

mkevins: Yes, good suggestion. We can pass a boolean to onPaste, and send `pastedHtml` as an empty string in that case.

hypest:
Thanks for the PR @mkevins ! This already starts to look being on the right path!

> * `onTextContextMenuItem` method is called when the user selects paste from the context menu

I wonder, will that also cover the case of the user pasting via the "Clipboard" button on Galaxy phones? If not can we add support? It doesn't have to be in this same PR though if it's not trivial to add. What do you think?

mkevins:
Hi @hypest, thanks for taking a look at this PR! Regarding your question:
> I wonder, will that also cover the case of the user pasting via the "Clipboard" button on Galaxy phones?
I don't think this PR will address that case, as this intercepts the selection of the standard Android TextView context menu. I think Samsung (and other third party keyboard / clipboard managers) interact through the input method framework in their own way. I think it's possible to add support, but I think it would be non-trivial. Probably best for a separate PR.

hypest:
> I think it's possible to add support, but I think it would be non-trivial. Probably best for a separate PR.

Indeed, makes sense to tackle that in a separate PR 👍

react-native-aztec/android/src/main/java/org/wordpress/mobile/ReactNativeAztec/ReactAztecText.java
```diff
@@ -321,6 +340,50 @@ private boolean onBackspace() {
...
+        disableTextChangedListener();
+        String content = toHtml(false);
+        int cursorPositionStart = getSelectionStart();
+        int cursorPositionEnd = getSelectionEnd();
+        enableTextChangedListener();
+        ReactContext reactContext = (ReactContext) getContext();
+        EventDispatcher eventDispatcher = reactContext.getNativeModule(UIManagerModule.class)
+                .getEventDispatcher();
+        eventDispatcher.dispatchEvent(new ReactAztecPasteEvent(getId(), content,
+                cursorPositionStart, cursorPositionEnd, text.toString(), html.toString())
```
diegoreymendez: Just one question on my end.

I see that when `isPastedAsPlainText == false`, we're sending both the text representation and the HTML representation with the event.  Will both be necessary simultaneously in our Javascript code?

The reason I'm asking is that if they're not necessary simultaneously, we can split this into two events (think `ReactAztecPasteEvent` and `ReactAztecPasteHTMLEvent`) to make the interface less ambiguous.

mkevins: Thanks for taking a look @diegoreymendez. Your question is a very good one!
> Will both be necessary simultaneously in our Javascript code?

The Javascript code has existing implementations for handling paste and that code expects both html and plain text data to be present. Splitting the event would clarify the interface, but there is also a benefit to maintaining common abstraction boundaries between the code bases. In this case, because there is already so much code in place on the Javascript side, I think it is best to mirror that implementation.

diegoreymendez: That sounds reasonable @mkevins.

diegoreymendez:
@mkevins - Since I'd need to implement the iOS side of this feature, I was wondering if the current state of it is such that I can go ahead and get started.  Can you let me know?

Also, could you add to the issue description:

1. The testing steps for different supported scenarios.
2. Any info on non-supported scenarios, or known issues.

Thanks!

mkevins:
> @mkevins - Since I'd need to implement the iOS side of this feature, I was wondering if the current state of it is such that I can go ahead and get started. Can you let me know?

Hi @diegoreymendez, I'd love to help you on getting the iOS side of this feature implemented. Yes, it is ready for you to implement, as it is working as expected on the Android side. Please don't hesitate if you have any questions regarding this.

> Also, could you add to the issue description:
> 
> 1. The testing steps for different supported scenarios.

I've considered creating tests for the Android ReactNative wrapper code, however, was advised to prioritize the implementation itself until a testing setup is in place. I am working on tests for the JS side, and I will add details to the issue description and related PR soon.

> 2. Any info on non-supported scenarios, or known issues.

Nothing that I'm aware of.

diegoreymendez:
> I've considered creating tests for the Android ReactNative wrapper code, however, was advised to prioritize the implementation itself until a testing setup is in place. I am working on tests for the JS side, and I will add details to the issue description and related PR soon.

I'm actually not referring to automated tests, but to a description of manual tests the reviewer can do, and what the expected behavior is.

mkevins:
Ah, I understand. I'll add that now. Thanks for the suggestion!

daniloercoli:
On a fresh checkout i was getting the error `gutenberg-mobile/gutenberg/packages/blocks/src/api/raw-handling/markdown-converter.js: Module showdown does not exist in the Haste module map` so i guess you need to add `"showdown": "^1.8.6",` to dependencies section of `package.json`.

hypest:
Thanks for updating the PR description with testing steps and Known Issues @mkevins !

It seems that the issues reported can be worked on in follow up PRs, no need to address them in this one.

src/globals.js
```diff
@@ -5,6 +5,9 @@ import { createElement } from '@wordpress/element';
 import jsdom from 'jsdom-jscore';
 import jsdomLevel1Core from 'jsdom-jscore/lib/jsdom/level1/core';
 
+// Import for side-effects: Patches for jsdom-jscore.
+import './jsdom-patches';
```
hypest: Hmm, this seems rather important since we're touching how `jsdom-jscore` is working. I'm sure we need the patches so, can you add some comments in the code on what do we need the patches for? Thanks!

mkevins: Good idea, I'll add some more comments to this to clarify.

mkevins:
> On a fresh checkout i was getting the error `gutenberg-mobile/gutenberg/packages/blocks/src/api/raw-handling/markdown-converter.js: Module showdown does not exist in the Haste module map` so i guess you need to add `"showdown": "^1.8.6",` to dependencies section of `package.json`.

Hi @daniloercoli , I was wondering if you were still having this problem. Is it possible that you have a `node_modules` folder present in the inner gutenberg directory (the submodule)? This directory is created when running `npm install` or `yarn install` from the inner directory; it may be related to that, and might be resolved by simply removing that directory: from within `gutenberg-mobile/gutenberg/`: `rm -rf node_modules`.

src/globals.js
hypest: Thanks @mkevins ! Can we expand on the reason(s) we replace each of the function prototypes? That insight will help anyone reading the code to understand why do we even replace these function.

hypest:
> Is it possible that you have a `node_modules` folder present in the inner gutenberg directory (the submodule)?

Hmm, I don't have an inner `node_modules` in the gutenberg submodule folder but I too had to `yarn add showdown` 🤔. Is a clean checkout working for you OK @mkevins ?

mkevins:
I ran a clean clone and had the same issue. Adding the `showdown` dependency as @daniloercoli suggested resolves it. Thanks @daniloercoli and @hypest for helping resolve the `showdown` issue.

mkevins:
Issues are being tracked here: https://github.com/wordpress-mobile/gutenberg-mobile/issues/624

src/globals.js
mkevins: I've added PR #826 to address this.

Issue #463, opened 2019-01-03
Scrolling glitch when splitting a long paragraph


pinarol:
There is a small scrolling glitch when splitting a long paragraph content on enter. It looks like it scrolls to the top of the block first and then down to the new block.

Note: It doesn't happen always, it is an intermittent issue, couldn't find the root cause yet

![50632271-fcd28980-0f4f-11e9-949f-67fcfde7d5d1](https://user-images.githubusercontent.com/5032900/50641954-bcd4cc00-0f7a-11e9-8c8c-59198748efc2.gif)

pinarol:
This also looks like the same issue, we need to address this one too https://github.com/wordpress-mobile/gutenberg-mobile/pull/474#pullrequestreview-193307699

pinarol:
I have opened a separate issue for the case of appending next text at the end of a long paragraph because that has a different root cause https://github.com/wordpress-mobile/gutenberg-mobile/issues/583

pinarol:
Most of the annoying issues are solved within https://github.com/wordpress-mobile/gutenberg-mobile/pull/554

But still there are some small glitching issues left on merge/split, but they are originating from RN layer. I'll describe the root cause below, and I am planning to open an issue to RN repo, I am also planning to open a PR to there because what I want is just a flag to turn off the automatic scroll adjustments they are making, so it'd be simple but there's a possibility of reject ofc. But even so, I am hoping we might have useful feedbacks.

So here's what happens: whenever content size changes scroll offset is also **sometimes** adjusted accordingly (only under some conditions, see RCTScrollView's calculateOffsetForContentSize:(CGSize)newContentSize to see exactly what conditions, long paragraphs don't have it always) If we could just delete the call for calculateOffsetForContentSize then everything could have worked perfectly(I tried and it worked perfectly). But, currently there's no way to do it, so I think it'd be good to request a ticket from RN folks to be able to turn it off. They put a comment explaining why they are calling this method and it looks like was a workaround in the first place:

>    // When contentSize is set manually, ScrollView internals will reset
>    // contentOffset to  {0, 0}. Since we potentially set contentSize whenever
>    // anything in the ScrollView updates, we workaround this issue by manually
>    // adjusting contentOffset whenever this happens

maybe this was needed in old iOS versions? not sure, but with iOS 12 it looks like we don't need that workaround imho. At least my tests were ok when I remove that call.

[...]

PR #554, opened 2019-02-07
Fix scrolling anomalies when splitting a paragraph or appending text to a long paragraph

pinarol:
To Fix: https://github.com/wordpress-mobile/gutenberg-mobile/issues/583

**Cases solved:**

1. appending text to a long paragraph:

Before:

scrolling-new-line

After:

glitch-followcaret1

2. splitting a long paragraph, scrolling down and never coming back issue

Before:

glitch1

After:

glitch2

**Cases couldn't be solved:**

Merge/split glitch issues couldn't be solved because they are caused by RN layer, whenever content size changes content offset is also **sometimes** adjusted accordingly (only under some conditions, see RCTScrollView's calculateOffsetForContentSize:(CGSize)newContentSize to see exactly what conditions, long paragraphs don't have it always) If we could just delete the call for calculateOffsetForContentSize then everything could have worked perfectly(I tried and it worked perfectly). But, currently there's no way to do it, so I am going to open a ticket to RN repo about it, it'd be good to be able to turn it off. They put a comment explaining why they are calling this method:

>    // When contentSize is set manually, ScrollView internals will reset
>    // contentOffset to  {0, 0}. Since we potentially set contentSize whenever
>    // anything in the ScrollView updates, we workaround this issue by manually
>    // adjusting contentOffset whenever this happens

maybe this was needed in old iOS versions, but with iOS 12 it looks like we don't need that workaround imho.
BTW I tried to make lots of workarounds for it but had no chance. Any idea is welcomed.

I should underline this, the problem with merging and splitting blocks is not related with the length of the block's content, sometimes long blocks split/merge pretty good, other times short blocks can have glitch. It is more related with the logic inside calculateOffsetForContentSize.

WPiOS PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/10968

Gutenberg PR: https://github.com/WordPress/gutenberg/pull/13705
react-native-keyboard-aware-scroll-view PR: https://github.com/wordpress-mobile/react-native-keyboard-aware-scroll-view/pull/4

**To Test**

Prerequisites
 
**Example app**

- Run `yarn install` <<important
- Run `yarn start`
- `yarn ios`
- `yarn android`

**WPiOS**

run yarn install, yarn start
Checkout the WPiOS branch in the [WPiOS PR](https://github.com/wordpress-mobile/WordPress-iOS/pull/10968)
Run `rake dependencies`
Open XCode
 
**WPAndroid**

run yarn install, yarn start

Test 1 - Append new text to a very long block
- Prepare a paragraph block that contains text larger than the viewable area
- Append new text at the end and until caret position goes under the keyboard
- Verify that the scrolling glitch problem doesn't occur as described in the [comment](https://github.com/wordpress-mobile/gutenberg-mobile/issues/463#issuecomment-455103187) here.

Test 2 - Insert new text to a very long block
- Prepare a paragraph block that contains text larger than the viewable area
- Choose a position in the middle and insert new text until caret position goes under the keyboard
- Verify that the scrolling glitch problem doesn't occur 

Test 3 - Splitting any long paragraph
- Try splitting any long paragraph in the first line
- Verify that caret won't go out of the viewable area

Test 4 - Test if we didn't affect existing functionality

( Repeat with both heading & paragraph blocks )

- Check if we don't have trouble in adjusting the block height as some text is added or removed
- Check that we have no problem focusing to the recently created block
- Check basic functionality on toolbar like bold/italic etc
- Check basic functionality on block toolbar like up/bottom arrows etc


pinarol:
@etoledom ready to review

pinarol:
hey @etoledom I made the silliest mistake by not pushing a part of the solution 😱I am very sorry. I'll update the PR and let you know. Although the 2nd screencapture you send(about merging the blocks) is in React Native layer. In RCTScrollView's calculateOffsetForContentSize:(CGSize)newContentSize method the scroll offset is reset to 0 in some conditions when content height changes :(

src/block-management/block-manager.js
```diff
@@ -128,6 +132,12 @@ export class BlockManager extends React.Component<PropsType, StateType> {
 		SafeArea.addEventListener( 'safeAreaInsetsForRootViewDidChange', this.onSafeAreaInsetsUpdate );
 	}
 
+	componentDidUpdate( prevProps: PropsType ) {
+		if ( this.props.blockCount > prevProps.blockCount ) {
```
etoledom: I'm curios about this one.
As I understand, this code fixes the issue of typing on a long paragraph block. How is the block count involved on that? 🤔

pinarol: you are %100 right to be curious about it. this is not related with appending to a long paragraph, it is related with the other issue about caret staying outside of viewable are when a split occurs. we are detecting if there's a new block added. I just added a safety belt here although this is not necessary %95 of the time, actually I tried now and couldn't reproduce the issue, so maybe it looks like we don't need this immediately. I'll remove it, we can add later if there's an upcoming issue about it, that way the intent will be more obvious.

PR #585, opened 2019-02-11
RTL layout fixes

### LINKED ISSUE #380: Make sure RTL text is handled correctly
We need to test more with RTL languages during Beta.

Known issues:

- [ ] Typing seems to work fine, but when you press enter, the block gets the wrong alignment. This doesn't happen if you add a new block via inserter.
- [x] Toolbar issue (Android). https://github.com/wordpress-mobile/gutenberg-mobile/pull/585
- [x] BottomSheet TextInput issue. https://github.com/WordPress/gutenberg/pull/13815 
- [ ] RichText placeholder position issue (iOS). https://github.com/wordpress-mobile/gutenberg-mobile/pull/591

etoledom:
Fixes #380 

This PR fixes two issues related to RTL layout.

rtl

**1. The block Toolbar on Android:**

The content of the scroll view on Android was aligned left, while on iOS it was correctly aligned right. Probably the issue has little to do with alignment itself but on how each platform rotate scroll view for RTL support.

This was fix by forcing the content to grow as big as the scroll view itself (allowing it to grow bigger if necessary).

**2. BottomSheet text input:**

Fixed in the `gutenberg` related PR (https://github.com/WordPress/gutenberg/pull/13815)

This was an issue since we are inverting the natural alignment for text inputs, so we had to express this intention on RTL layout too. Sadly `text-align: end` didn't work (it's not allowed).

---

I added a helper to test RTL layout, but I'm unsure if we should keep for continue testing RTL in the future. Ideally we would have a switch in the UI or use the RN Dev Menu, but this last option requires native development.

**To test:**

- Run the project.
- Go to `src/index.js` and switch `I18nManager.forceRTL(false);` to `true`.
- Reload the app (you might need to reload more than once for some reason)
  - Now the layout should look mirrored horizontally (RTL)
  - Note that elements like images and icons are not inverted 👍
- Check that the ToolBar displays correctly.
- Edit a Header block and check that the content of the ToolBar is still scrollable.
- Display the Image Settings BottomSheet. (Wheel icon on a non-empty image block)
- Check that the TextInputs are displayed properly.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -8,7 +8,7 @@ const gutenbergSetup = () => {
 	const apiFetch = require( '@wordpress/api-fetch' ).default;
 	const wpData = require( '@wordpress/data' );
 
-	I18nManager.forceRTL(false); // Change to `true` to debug RTL layout easily.
+	I18nManager.forceRTL( false ); // Change to `true` to debug RTL layout easily.
 
 	// wp-api-fetch
 	apiFetch.use( apiFetch.createRootURLMiddleware( 'https://public-api.wordpress.com/' ) );
--- a/src/index.js
+++ b/src/index.js
@@ -1,5 +1,5 @@
 // External dependencies
-import { AppRegistry } from 'react-native';
+import { AppRegistry, I18nManager } from 'react-native';
 
 // Setting up environment
 import './globals';
@@ -7,6 +7,8 @@ import './globals';
 const gutenbergSetup = () => {
 	const apiFetch = require( '@wordpress/api-fetch' ).default;
 	const wpData = require( '@wordpress/data' );
+
+	I18nManager.forceRTL(false);
 
 	// wp-api-fetch
 	apiFetch.use( apiFetch.createRootURLMiddleware( 'https://public-api.wordpress.com/' ) );
--- a/src/index.js
+++ b/src/index.js
@@ -8,7 +8,7 @@ const gutenbergSetup = () => {
 	const apiFetch = require( '@wordpress/api-fetch' ).default;
 	const wpData = require( '@wordpress/data' );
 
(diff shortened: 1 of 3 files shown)


src/index.js
```diff
@@ -8,6 +8,8 @@ const gutenbergSetup = () => {
 	const apiFetch = require( '@wordpress/api-fetch' ).default;
 	const wpData = require( '@wordpress/data' );
 
+	I18nManager.forceRTL( false ); // Change to `true` to debug RTL layout easily.
```
etoledom: Not sure if we should leave this here. But for the moment it could be useful for fast checking RTL layout in both iOS and Android, without the need of `AndroidStudio` or `Xcode` or changing settings on device.

etoledom:
Hey @daniloercoli - Can I bug you with a review?
Thank you! :)

daniloercoli:
> * Note that elements like images and icons are not inverted 👍

It seems that `H*` are inverted instead. Is it correct?

daniloercoli:
Tested this PR on both iOS and Android, and it seems to work fine. LGTM!

Note related to this PR, but i've noticed that the Image Settings BottomSheet is not dismissed when I tap back button on Android. I need to tap outside the sheet to dismiss it. (Tested on Android 6 - Nexus 5).

src/index.js
daniloercoli: I guess we can leave it here. It doesn't hurt, no?

etoledom:
Thank you @daniloercoli for the review!

> i've noticed that the Image Settings BottomSheet is not dismissed when I tap back button on Android

Good catch! If I recall correctly, @marecar3 also noticed it and was working on a fix. I'm not really sure if that's the case. If it's not, I can check that out. 👍

etoledom:
Thank you!

PR #587, opened 2019-02-12
Propagate componentDidMount to parent apps

koke:
Once Gutenberg mounts, notify the parent apps and includes whether it's showing
any unsupported blocks

For #556 

This will require some code to handle the events in the apps, so labeling this as not ready for review yet, since I don't have testing instructions

PR #589, opened 2019-02-12
Support for the bridge in the android demo app

hypest:
We haven't made an effort to augment the Android demo app to support the bridge interface since that could be tested directly with the main WP app. We avoided having code directly call any of the bridge functions without responding to a request _from_ the bridge anyway.

With #587 though, we have the editor send up an event upon initializing/mounting, and that requires the bridge to be in place to work on Android. This PR adds that support.

Test do fail in this PR because the Jest is missing a mock for `editorDidMount()`. After all, the bridge is still not there for Jest anyway so, we need to mock it. Leaving this change for the parent PR though since it's not related to this PR.

To test:

* Run the demo app. It shouldn't red-screen complaining that `editorDidMount` is missing from the bridge. The demo app should run fine.

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -167,6 +167,7 @@ dependencies {
     implementation project(':react-native-svg')
 	implementation project(':react-native-aztec')
 	implementation project(':react-native-recyclerview-list')
+    implementation project(':react-native-gutenberg-bridge')
     // compile fileTree(dir: "libs", include: ["*.jar"])
     implementation "com.android.support:appcompat-v7:$supportLibVersion"
     implementation "org.wordpress:utils:$wordpressUtilsVersion"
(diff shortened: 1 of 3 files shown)


daniloercoli:
This is working fine, once I changed
```
--- a/react-native-gutenberg-bridge/android/src/main/java/org/wordpress/mobile/WPAndroidGlue/WPAndroidGlueCode.java
+++ b/react-native-gutenberg-bridge/android/src/main/java/org/wordpress/mobile/WPAndroidGlue/WPAndroidGlueCode.java
@@ -140,7 +140,7 @@ public class WPAndroidGlueCode {
             }
 
             @Override
-            public void editorDidMount(Boolean hasUnsupportedBlocks) {
+            public void editorDidMount(boolean hasUnsupportedBlocks) {
                 mOnEditorMountListener.onEditorDidMount(hasUnsupportedBlocks);
             }
         });
```

to fix compile errors. I guess we need to wait until the parent PR does fix those Boolean types.

hypest:
Base PR has the change for the uppercase `Boolean` now so, this should be ready to merge if we want to.

hypest:
Base PR author is OK with merging so, will push the button.

PR #591, opened 2019-02-12
Fix placeholder position for RTL layout

etoledom:
This PR fixes an issue on RTL layout with the position of placeholders:

iOS doesn't manage RTL layout automatically on scroll views, so React Native will flip horizontally RCTScrollViews on RTL layout to get the Right to Left effect.

The problem is that we are inserting an view into an already flipped tree of views, and this created all sort of weird layout behavior, braking the `leading` and `trailing` anchored constraints:
screen shot 2019-02-12 at 7 55 09 pm

```
<RCTCustomScrollView: 0x7f99a3839c00; baseClass = UIScrollView; frame = (0 0; 375 646); transform = [-1, 0, 0, 1, 0, 0]; 
```
Note: `transform = [-1, 0, 0, 1, 0, 0]; `

The solution is a non-conventional horizontal constraint on RTL layout, since `rn-aztec`'s left is logically at the right and right is at the left, but visually they are positioned "correctly.

Since this solution is non-conventional, I added plenty of explanation in the code itself, please let me know if it is clear enough.

The result of this is a correct placement of the placeholder at the right:
(Tested with `right-to-left pseudolanguage`)
rtl-placeholder

**To test:**
- Checkout the corresponding [WPiOS PR](https://github.com/wordpress-mobile/WordPress-iOS/pull/11018)
- `rake dependencies`
- Holding `options ⌥` key, click on the `Play (build and run)` button on Xcode.
- On Run -> Options -> Application Language, choose: `Right-To-Left Pseudolanguage`
- Run the app.
- Do your best to open an instance of Gutenberg with everything flipped :)
- Check that the Aztec placeholders show properly as in the attached gif.

### DIFF
--- a/react-native-aztec/ios/RNTAztecView/RCTAztecView.swift
+++ b/react-native-aztec/ios/RNTAztecView/RCTAztecView.swift
@@ -29,9 +29,37 @@ class RCTAztecView: Aztec.TextView {
 
     private var previousContentSize: CGSize = .zero
 
+    var textHorizontalInset: CGFloat {
+        return contentInset.left + textContainerInset.left + textContainer.lineFragmentPadding
+    }
+
+    var isRTLLayout: Bool {
+        return reactLayoutDirection == .rightToLeft
+    }
+
     private lazy var placeholderLabel: UILabel = {
         let label = UILabel(frame: .zero)
+        label.translatesAutoresizingMaskIntoConstraints = false
+        label.textAlignment = .natural
+        label.font = font
         return label
+    }()
+
+    // RCTScrollViews are flipped horizontally on RTL. This messes up competelly horizontal layout contraints
+    // on views inserted after the transformation.
+    var placeholderPreferedHorizontalAnchor: NSLayoutXAxisAnchor {
+        return isRTLLayout ? placeholderLabel.rightAnchor : placeholderLabel.leftAnchor
+    }
+
+    // This constraint is created from the prefered horizontal anchor (analog to "leading")
+    // but appending it always to left of its super view (Aztec).
+    // This partially fixes the position issue originated from fliping the scroll view.
+    // fixLabelPositionForRTLLayout() fixes the rest.
+    private lazy var placeholderHorizontalConstraint: NSLayoutConstraint = {
+        return placeholderPreferedHorizontalAnchor.constraint(
+            equalTo: leftAnchor,
+            constant: textHorizontalInset
+        )
     }()
     
     // MARK: - Font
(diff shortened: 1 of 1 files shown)


react-native-aztec/ios/RNTAztecView/RCTAztecView.swift
```diff
@@ -29,10 +29,38 @@ class RCTAztecView: Aztec.TextView {
 
     private var previousContentSize: CGSize = .zero
 
+    var textHorizontalInset: CGFloat {
+        return contentInset.left + textContainerInset.left + textContainer.lineFragmentPadding
+    }
+
+    var isRTLLayout: Bool {
```
diegoreymendez: This is pretty small, but considering this is a property of `RCTAztecView`, a more precise naming for this property could be `hasRTLLayout`.

react-native-aztec/ios/RNTAztecView/RCTAztecView.swift
```diff
@@ -69,23 +97,33 @@ class RCTAztecView: Aztec.TextView {
...
-            ])
+        ])
     }
 
     // MARK - View Height: Match to content height
-    
+
     override func layoutSubviews() {
         super.layoutSubviews()
+        fixLabelPositionForRTLLayout()
```
diegoreymendez: This call here has me a bit confused because of the asymmetry between how we handle LTR and RTL: for LTR we have a fixed constraint setup at view-initialization time (well almost because it's lazy, but I digress), whereas for RTL we can update it dynamically whenever `layoutSubviews` is called.

In my mind if we need a dynamic solution, we need both updates here, whereas if a static solution is fine, we could move this code to the lazy initializer of `placeholderHorizontalConstraint` and make it something like:

```swift
    private lazy var placeholderHorizontalConstraint: NSLayoutConstraint = {
        return placeholderPreferedHorizontalAnchor.constraint(
            equalTo: leftAnchor,
            constant: isRTLLayout ? rightTextInset : leftTextInset
        )
    }()
```

Does that make sense?  Let me know if there's more to it.

react-native-aztec/ios/RNTAztecView/RCTAztecView.swift
```diff
@@ -29,10 +29,38 @@ class RCTAztecView: Aztec.TextView {
 
     private var previousContentSize: CGSize = .zero
 
+    var textHorizontalInset: CGFloat {
```
diegoreymendez: A suggestion for clarity / readability / maintainability:

```swift
var leftTextInset: CGFloat {
    return contentInset.left + textContainerInset.left + textContainer.lineFragmentPadding
}

var rightTextInset: CGFloat {
    return bounds.width - leftTextInset
}
```

diegoreymendez: Actually, the naming is off... the second one should be `leftTextInsetInRTLMode` or something like that.

diegoreymendez: I mean, naming aside my suggestion would be to have a calculated property for both options.

react-native-aztec/ios/RNTAztecView/RCTAztecView.swift
etoledom: This asymmetry is actually needed because of the horizontal flipping of the scroll views on RTL layout.
When we try to add a view on a horizontally flipped view, what happens is that the left and right right sides are also flipped logically, but RN flips back the child views so they don't look reversed.

When we try to constraint the placeholder's right anchor to the parent's right anchor (as we would normally do) this happens:

screen shot 2019-02-15 at 6 20 45 pm

Logically it's correctly anchoring it to the right, but it's visual positioned at the left. You can see the long constraint across the Header, that it's set to 0, and breaks constraints ¯\_(ツ)_/¯

That's why we set the right anchor of the left anchor of the parent, that will give us this:

screen shot 2019-02-15 at 7 07 24 pm

This is not perfect yet, but now the horizontal constraint will respond properly, so we can work with it.
What we have to do is to move the label all the way to the other side of the screen. That's when the dynamic value comes to play.
We set the constant of the constraint to be the width of the parent, and if the parent changes, this has to update accordingly. That's why it's in `layoutSubviews`.

In the other hand, for LTR, a static value will be enough, since constraining left to left anchors won't need modifications if the layout changes.

I hope this makes clearer.

If we just want consistency, we can always update "dynamically" on both RTL and LTR layouts, but LTR is not necessary, and I think it is good to remark this difference.

I see that my code comments weren't clear enough 😅

etoledom:
Thank you @diegoreymendez for the review!
I updated with two of the suggested changes, now I'll merge when ✅

PR #599, opened 2019-02-14
Focus on title when creating a new post.

diegoreymendez:
## Description:

Resolves https://github.com/wordpress-mobile/WordPress-iOS/issues/10594

This PR focuses the title when opening Gutenberg, if the post has no title and no content.  This is a simplified and initial approach, that seems to be enough to match the condition that we're dealing with a new post.

If we find this condition isn't enough we can expand incrementally.

## Related PRs:

Gutenberg: https://github.com/WordPress/gutenberg/pull/13874

### Test 1:

1. Run the demo app.

Make sure the focus is NOT in the title, since it's not empty and there are blocks.

### Test 2:

1. Edit `App.js`, and force `initialTitle = '';` and `initialData = '';`.
2. Run the demo app.
3. Make sure the title and content are empty.  If not, go back to step 1.

Make sur the focus is in the title, since the title is empty and there are no blocks.


src/app/App.js
```diff
@@ -9,6 +9,7 @@ type PropsType = {
 	initialData: string,
 	initialHtmlModeEnabled: boolean,
 	initialTitle: string,
+	focusTitle: boolean,
```
daniloercoli: It doesn't seem to be used or required at the moment.

src/block-management/block-manager.js
```diff
@@ -56,6 +56,7 @@ type StateType = {
 
 export class BlockManager extends React.Component<PropsType, StateType> {
 	scrollViewRef: Object;
+	titleRef: PostTitle;
```
daniloercoli: This doesn't seem to be used.

daniloercoli:
I've left just a couple of comments, but this PR is LGTM once those and GB conflicts are resolved!

Did you test this PR by integrating it into the main wp-ios or wp-android apps?

As discussed in Slack convo, we may want to iterate on this, and read the `isNewPost` parameter from the parent app (via the bridge/glue code) instead of manually checking it with `this.props.title === '' && this.props.blockCount === 0`

diegoreymendez:
> As discussed in Slack convo, we may want to iterate on this, and read the `isNewPost` parameter from the parent app (via the bridge/glue code) instead of manually checking it with `this.props.title === '' && this.props.blockCount === 0`

No, but I purposely left that out of the scope of this PR.  As soon as we close this I'll submit a PR to test the integration and make any necessary changes (which I expect will be none).

diegoreymendez:
Comments addresed @daniloercoli .

PR #610, opened 2019-02-18
Fix dictation issue on iOS

### LINKED ISSUE #606: [iOS] Voice-to-text: after input, content is removed
### Expected behavior

When using voice-to-text, if I input a string and tap to submit, I expect that the string will be applied in place.

### Actual behavior

As I'm speaking, the words are transcribed on the fly, but upon submitting, that content disappears.

Update: also tried Google's Gboard keyboard (on iOS) voice-to-text and that works as expected.

1. Start a post
2. Add a paragraph block
3. Tap the `microphone` icon on the keyboard
4. Speak to capture a phase
5. Submit captured phase
6. Observe that submitted phase is removed

etoledom:
fixes #606 

This PR fixes the dictation content disappearing from Paragraph blocks on iOS.

This is why the issue happens:
1. This happened just on empty TextViews (then explanation will assume that).
2. When the person is dictating, just `textViewDidChange` is called with the new text. (No `insertText`) is called.
3. When dictation finishes, `textViewDidChange` is called first with the text existing previous the dictation (an empty string in this case).
4. RNAztec propagates to the JS side the new content with that empty string.
5. The new text from dictation is inserted and `textViewDidChange` is called again with the full new text. (and RNAztec informs again the JS side)
6. RNAztec receives a `setContents` message from JS from the previous propagation of the empty string. Since JS resets `eventCount` to zero (undefined) when the string is empty, RNAztec refresh its state to an empty string, and the dictation string is lost.
7. RNAztec receives a `setContents` message from JS with the dictation string, but it's ignored since it has `eventCount == 1`

To fix the issue, I noticed that the step **3** is always called between `dictationRecordingDidEnd` and `insertDictationResult`.
Activating a flag in between those method calls to avoid propagating the empty string fixed the issue.

dictation

**To test:** 
- Build and run the project.
- Create an empty paragraph block.
- Test the iOS dictation feature.
- Check that it behaves properly.

### DIFF
--- a/react-native-aztec/ios/RNTAztecView/RCTAztecView.swift
+++ b/react-native-aztec/ios/RNTAztecView/RCTAztecView.swift
@@ -152,6 +152,7 @@ class RCTAztecView: Aztec.TextView {
         }
 
         super.insertText(text)
+        updatePlaceholderVisibility()
     }
 
     open override func deleteBackward() {
@@ -160,6 +161,7 @@ class RCTAztecView: Aztec.TextView {
         }
         
         super.deleteBackward()
+        updatePlaceholderVisibility()
     }
 
     // MARK: - Dictation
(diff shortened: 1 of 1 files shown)


etoledom:
Thank you for the review!

> I'm not sure the changes in this PR are the best way to resolve this issue as it feels a bit hackish.

I agree it's not an elegant solution. I tested other possibility and it seems to be the only way to fix it that doesn't involve touching JS and (probably) Android too. While not the best, it's probably is the simplest.

PR #611, opened 2019-02-18
Fix/Set focus on title for new posts on Android

daniloercoli:
This PR fixes [wordpress-mobile/WordPress-iOS#10594](https://github.com/wordpress-mobile/WordPress-iOS/issues/10594) for Android. 

Initially we've made a fully JS version of it [here](https://github.com/wordpress-mobile/gutenberg-mobile/pull/599), but does't seem to work fine on Android, due to probably the way Android handle the focus. 
I've spent some hrs trying to fix the issue on the Android side (native) without much luck, then added to the bridge the ability to send an event `focus on title` and used it in the android app.

In this PR we wait until the editor is mounted, listen for `editorDidMount`, and emit an event to set the focus on title when both content and title are empty.

**Test 1**
- Use `wp-android`
- Check out this branch and build the app from code
- Start the app
- Tap on the New Post button in the home screen
- When the editor is mounted the keyboard should appear and the focus should be on the title

**Test 2**
- Go to posts list
- Edit a post made in GB
- The editor should start but no focus on the title

Not sure if something is required on the iOS side. Tests are passing.

### DIFF
--- a/jest/setup.js
+++ b/jest/setup.js
@@ -5,6 +5,7 @@ jest.mock( '../react-native-gutenberg-bridge', () => {
 		subscribeParentGetHtml: jest.fn(),
 		subscribeParentToggleHTMLMode: jest.fn(),
 		subscribeSetTitle: jest.fn(),
+		subscribeSetFocusOnTitle: jest.fn(),
 		subscribeUpdateHtml: jest.fn(),
 		editorDidMount: jest.fn(),
 	};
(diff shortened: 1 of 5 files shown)


daniloercoli:
Not sure if something is needed on the iOS side, jest tests didn't report anything wrong, although I can't test it locally since `yarn ios` is broken again (** BUILD FAILED **).

Issue #592, opened 2019-02-13
Align text-label wording on block placeholders


iamthomasbishop:
## Currently

**When starting a new post**, the first placeholder block has a text label that says `Start typing or press (+) to add content`.

**When you add additional blocks**, those text labels say `Add text or type / to add content`.

## Proposal

Considering we don't currently provide slash commands, the second message isn't appropriate in our editor. 

I propose that we show 1 consistent message – `Start typing or press (+) to add content`.

This is related to: 
https://github.com/wordpress-mobile/gutenberg-mobile/issues/549

SergioEstevao:
@iamthomasbishop do you want "Start typing..." or "Start writting..." and can we keep the unicode character for now?

iamthomasbishop:
It should probably be "writing", considering we support various input methods (voice, Pasting, etc). Good catch.

I think @koke was working on the Unicode thing, so that should be fixed soon.

Issue #549, opened 2019-02-06
Clean up styling of starter paragraph block/placeholder


iamthomasbishop:
## Initial "starter" block

This is in reference to the first block that appears when creating a new post – the one that has the label of "Start writing...".

<img width="977" alt="image" src="https://user-images.githubusercontent.com/1200432/52372000-917f4a00-2a1c-11e9-86b6-d9d959ee21d2.png">

#### Punch list

- [ ] Change font-family to Noto Serif (it looks like it is currently the system default sans-serif)
- [ ] Swap in (+) icon if possible – the unicode character feels out of place
- [ ] Make sure placeholder text color is `$gray` aka `#87a6bc`

koke:
I've tried swapping the unicode character with the icon and it works, with some limitations:

- Since it's an inline image, it seems to be ignoring CSS, and I'm not sure if we can apply any styling to it
- The paragraph placeholder uses `RichText.placeholder`, which I doubt will support the attributed strings right away.

Issue #547, opened 2019-02-06
Refine inner block margins


iamthomasbishop:
Inner block paddings are a bit loose right now, let's make sure these are refined.

## Block Anatomy

By default, every block type (including Title) contains internal padding of 12pt top/bottom and 16pt left/right.

#### Static

<img width="620" alt="image" src="https://user-images.githubusercontent.com/1200432/52358801-47876b80-29fe-11e9-8956-5f346ec1f06e.png">

#### Focused

When a block is focused, we show the inline toolbar (44pt height) and collapse the margin-bottom, like so:

<img width="619" alt="image" src="https://user-images.githubusercontent.com/1200432/52358900-6f76cf00-29fe-11e9-9b2d-f59ec5c8c44e.png">

## Currently 

(proper paddings indicated with green, improper in red)

Android:

![image](https://user-images.githubusercontent.com/1200432/52362272-cd5ae500-2a05-11e9-8e3d-152108c08537.png)

As you can see, left/right paddings on Android are 👍 , however top/bottom paddings are a bit off.

iOS:

![ios](https://user-images.githubusercontent.com/1200432/52362263-c6cc6d80-2a05-11e9-9efe-aeb6231db43b.jpg)

## Punch-list

**Both Platforms**

- [ ] Ensure padding-top is always 12 (static + focused)
- [ ] Remove padding-bottom when block is focused

**iOS**

- [ ] Ensure padding on left/right sides is 16pt

iamthomasbishop:
Previous, related issue: https://github.com/wordpress-mobile/gutenberg-mobile/issues/393

SergioEstevao:
@iamthomasbishop I update the margins here, one of the issue we where having is that the text container had a minimum height of 50px so when we had a single line on it we were still getting 50px of height to it making it look to have bigger margins. Here is a screenshot with some extra colors:
![simulator screen shot - iphone xr - 2019-02-20 at 11 05 29](https://user-images.githubusercontent.com/651601/53087717-b46f2b00-34ff-11e9-81ff-8d210837a847.png)

iamthomasbishop:
Pulled this into Sketch to double check, and all of these margins look aligned _perfectly_.

![image](https://user-images.githubusercontent.com/1200432/53102055-94843b00-34f0-11e9-9fc3-922625b52e4b.png)

Thanks for taking care of this @SergioEstevao !

PR #616, opened 2019-02-19
Block placeholder fixes

SergioEstevao:
Refs: #592, #549, #547

Updates the GB reference to have the placeholder fixes. 
It also makes the title placeholder text o be consistent with the web and localized.

Related GB PR: https://github.com/WordPress/gutenberg/pull/13945

simulator screen shot - iphone xr - 2019-02-19 at 12 44 11


koke:
I haven't tested yet, but from your screenshot, the placeholder doesn't look left-aligned with the title

koke:
Also cc @Tug for i18n string changes

elibud:
Could we avoid saying "Welcome to Gutenberg", I don't think that would make sense to most of our users. Honestly I would stick with a neutral placeholder like "Add a title".

iamthomasbishop:
I think (iirc) the placeholder label text on the title does say `Add a Title`, unless that has changed.

One other thing that stands out to me is the margins of blocks on the initial canvas. The title feels too close to the navbar – maybe the title block is a little too tall, but it's hard to tell without seeing those boundaries.

For now, can we add a slight bit of space at the top of the canvas (above the title block) so it feels a little more balanced? Maybe 16pt? 

screen shot 2019-02-19 at 9 13 05 am

elibud:
@iamthomasbishop @SergioEstevao yes, looking at the code I can see we are keeping the "Add a Title" label, please ignore me 🤦‍♀️

SergioEstevao:
@iamthomasbishop here is a new screenshot with the new margins and some background color to make it easier to figure margins.

simulator screen shot - iphone xr - 2019-02-19 at 17 30 43

koke:
Keep in mind that those screens are for the example app, we should make sure that the margins look right on the WordPress apps as well

iamthomasbishop:
@SergioEstevao Thanks for the reference. Spacing looks really wonky there – is the title vertically aligned top? That's probably why it looked like a lot of empty space 😄 

In terms of spacing at the top of the canvas, I mocked this up just to get a visual reference, and 16pt/dp feels about right. It looks like this:

image

Left: title focused. Right: static.

I've also uploaded these to the Zeplin project for reference.

SergioEstevao:
@iamthomasbishop the component we use as the base for Aztec that is being used in the title always aligns the text on the top. I changed the dimensions of the component to make it more centred it now
looks like this: 
simulator screen shot - iphone xr - 2019-02-19 at 21 58 04

iamthomasbishop:
That looks a bit better already. Might still be worth adding 16pt of margin to the top of the canvas, but this at least looks cleaner.

Side note, and maybe this is something that can be addressed in a separate issue – I noticed the color on all of the placeholder text labels is different from the preferred `$gray` aka `#87a6bc` – can we match those? Thank you!

SergioEstevao:
@iamthomasbishop updated the colors: 

simulator screen shot - iphone xr - 2019-02-20 at 00 21 32

SergioEstevao:
@daniloercoli do you mind have a look in Android to see if things look ok after these changes?

daniloercoli:
It does seem to work fine on Android!

iamthomasbishop:
Thanks for all of the quick iterations/feedback, @SergioEstevao and @etoledom! When ready, can you share screenshots of the current state on iOS/Android for design review? TY!

PR #617, opened 2019-02-19
Adds support for pasting in iOS.

diegoreymendez:
## Description:

This PR builds on top of @mkevins 's work to implement pasting in iOS.

This is an initial version of it, which I'm submitting for review so we can try to spot issues.

This PR also implements the unsupported HTML renderer in gutenberg, which was crashing the app when pasing HTML that Aztec doesn't know how to render.

## Testing:

Copy paragraphs from another app and paste.
A good source is Safari.


diegoreymendez:
Thanks @etoledom, but again, I'd like to point out @mkevins has most of the merit here, as his work works very well almost out of the box.

I'll be opening issues for the things you spotted.

mkevins:
> Thanks @etoledom, but again, I'd like to point out @mkevins has most of the merit here, as his work works very well almost out of the box.

Thanks @diegoreymendez. I can't take all the credit - my work built on a lot of @iseulde's work on the web side, so credit to her and the Gutenberg team.

mkevins:
@etoledom thanks for testing it out. I'm getting the same two errors you mentioned above on the Android side as well.
Screenshot with some more info:

Regarding the second error, images from that source seem to only work when only the image is copied (without any surround text).

Issue #551, opened 2019-02-06
Prevent focus border from shifting block contents


iamthomasbishop:
When focusing a block, we add a 1px inset border. It looks like this is making the contents shift slightly. 

Example on Android, although this is happening on iOS as well:

![block-focus-shift](https://user-images.githubusercontent.com/1200432/52379390-01e39680-2a30-11e9-8b8d-f5cdfdb8947b.gif)

Are we able to use the  `box-shadow` CSS property? Adding a box-shadow of 1px would in theory align to the outside of the block.

Issue #540, opened 2019-02-05
Add border on Title when focused in order to match other block focus styles


iamthomasbishop:
When the Title is focused, it doesn't use the same styling (border) as all other blocks. We need to update this to match.

## Current State

Title focused:

<img src="https://user-images.githubusercontent.com/1200432/52288800-7b955a80-2932-11e9-84cd-5ae4f72810c7.jpeg" width="375">

vs. other type of block:

<img src="https://user-images.githubusercontent.com/1200432/52288954-b6978e00-2932-11e9-9ff4-a1e924e5d3c8.jpeg" width="375">

## Proposed

<img width="383" alt="image" src="https://user-images.githubusercontent.com/1200432/52289146-1e4dd900-2933-11e9-8a47-a636edc67e9e.png">

Note: The way the Title currently sits in the editor, the border-top might not be visible. For the sake of simplicity and consistency with all other block types, let's match the styling (`border: 1px solid ...` on all four sides).

iamthomasbishop:
@diegoreymendez I can't remember if we discussed already, but if possible (and not already solved as part of this issue, can we also make sure the placeholder text label color (when there is no title) is `$gray` aka `#87a6bc`? Thank you!

PR #622, opened 2019-02-20
Add border on Title when focused

pinarol:
Fix https://github.com/wordpress-mobile/gutenberg-mobile/issues/540
Fix https://github.com/wordpress-mobile/gutenberg-mobile/issues/551

This change adds borders to the Title when focused in order to match other blocks.
And also fixes a slight shift happening on Android when we focus to a block.

gutenberg PR needs to be merged first: https://github.com/WordPress/gutenberg/pull/13970

android-title-border

ios-title-border

**To Test**

Testing prerequisites

For WPiOS

run yarn install, yarn start
Open XCode WPiOS on the latest develop

For WPAndroid

run yarn install, yarn start

Test Steps:

WPiOS:

- Open a post and tap the title
- Verify you see focus indicating borders around the title just like the other blocks
- Verify there's no slight shifting on portrait mode when we focus on a block because there should be no left right borders

iPhone X and iPad special case:
- Verify that you do see the left right borders when there's empty space around the block like there's on iPhone X landscape mode, and iPad portrait and landscape modes.

WPAndroid:

- Open a post and tap the title
- Verify you see focus indicating borders around the title just like the other blocks
- Verify that the slight shift that is described here is fixed also: https://github.com/wordpress-mobile/gutenberg-mobile/issues/551


src/block-management/block-manager.scss
```diff
@@ -56,4 +57,20 @@
 	bottom: 0;
 	right: 0;
 	left: 0;
+}
+
+.blockHolderFocusedSemiBordered {
+	border-color: $gray-lighten-30;
```
iamthomasbishop: Would it make sense to use css `box-shadow` to apply these focus states? In theory that should prevent any shifting, just applying a shadow on the outside edge of the element.

pinarol: Unfortunately box-shadow property is not available in React Native and as I investigated it is not a very straight forward change to replace borders with shadows in RN. Because Android and iOS have very different methods for doing it and both methods have different side effects that needs to be addressed.

iOS has these props:
https://facebook.github.io/react-native/docs/shadow-props#shadowcolor

Android has the elevation option but there are no other options to fine-tune the shadow effect. So I am not even sure it'll give the results we want.
https://facebook.github.io/react-native/docs/view-style-props#elevation

I tried to apply these quickly but it messed up the child views' styles in different ways, so it didn't look like an ideal last minute change, but, if you want we can open a separate issue for this and try later? 

In this PR, I fixed the shifting of Android blocks. I don't expect any shifting will happen in cases where the block width is equal to window width, because in that case we don't have left&right borders at all.

src/block-management/block-manager.js
```diff
@@ -201,6 +226,7 @@ export class BlockManager extends React.Component<PropsType, StateType> {
 					keyboardShouldPersistTaps="always"
 					style={ styles.list }
 					data={ this.props.blockClientIds }
+					extraData={ [ this.state.isTitleFocused, this.state.isFullyBordered ] }
```
etoledom: I don't see this property being used in `KeyboardAwareFlatList` or somewhere else. What is it for? 🤔

pinarol: By passing extraData we make sure FlatList will re-render when extraData changes, otherwise changes won't get effect. If you look inside `keyboard-aware-flat-list.ios.js` or `keyboard-aware-flat-list.android.js` you can see all props are propagated to react-native's FlatList.

etoledom: Perfect! Thank you.

pinarol:
@etoledom 👋 thanks for the review. I am going to address that issue about the empty post but before that I found another issue unfortunately, it looks a bit annoying and still couldn't come up with a solution. On Android, the title focus is getting lost in the first tap:

android-title-focus-lost-on-first-tap

I am guessing the re-render of the title view in order to add the borders might be somehow triggering a blur event on Android side, I've seen the blur event coming from Android side but not sure about the root cause of it. This is not happening on iOS. Still looking into it.

pinarol:
Hey @etoledom 👋  This is ready for another look. 

It looks like Android can somehow re-renders the title if we add border to it and this makes a blur event triggered. So I changed the way we add borders, I updated the containers views with borders with transparent color. When we focus I just change the border color. So we will not only get rid of unnecessary re-renders but also we are removing possibilities of the shifting effects when we focus. This is how the shadow effect would work also, cc @iamthomasbishop 

ios-title-focus-landscape

android-titlefocus

src/block-management/block-manager.js
```diff
@@ -172,21 +188,21 @@ export class BlockManager extends React.Component<PropsType, StateType> {
...
-				<PostTitle
-					setRef={ ( ref ) => {
-						this.titleViewRef = ref;
-					} }
-					title={ this.props.title }
-					onUpdate={ this.props.setTitleAction }
-					placeholder={ __( 'Add title' ) } />
-			</View>
+			<PostTitle
+				onRef={ ( ref ) => {
```
etoledom: Nitpick: Can we name this `ref` instead of `onRef`? Since it seems to be more standard in React.

pinarol: done with 67b509a

PR #623, opened 2019-02-20
Links UI using BottomSheet component

etoledom:
This PR updates the Gutenberg ref. to test https://github.com/WordPress/gutenberg/pull/13972

links-ui

To test:
- Refer to the [Gutenberg side PR](https://github.com/WordPress/gutenberg/pull/13972).


iamthomasbishop:
I have a question regarding the divider lines: I understand they're not going to the edges because they're within a container, which makes sense. Is that on all sheets, or just this links one? I believe I've seen the dividers on the Image Settings sheet go to the right edge but maybe I'm dreaming 😄 I just want to make sure we're being consistent and that I'm generally aware of the differences.

etoledom:
> Is that on all sheets, or just this links one?

It's on all of them. Some time ago they were going until the edge, but that was because they were overflowing, drown outside the container. We didn't noticed it until we did the max-width 😁

It shouldn't be difficult to adjust the containers to allow de dividers to go until the edge.

iamthomasbishop:
> It's on all of them. Some time ago they were going until the edge, but that was because they were overflowing, drown outside the container. We didn't noticed it until we did the max-width 😁

Ok fair enough. Not a big concern as we'll sort it out soon. 😄

Because I've written so much css in a past life, my brain goes to utilizing the `calc` function, like `calc(100% + 16px)`, but I'm not sure if that is available to us.

etoledom:
> but I'm not sure if that is available to us

Just tested, and it doesn't work 😞

iamthomasbishop:
Curse you, RN! *shakes fist* – I would've been delightfully surprised had it worked 😆

iamthomasbishop:
Until we have a solution for extending the dividers beyond the containers, what do you think about extending the dividers to the icons as well, so that the dividers are 16pt from each edge of the sheet. It would look like this:

image

Until we are able to sort the dividers-to-the-right-edge thing, this might feel pretty balanced. In fact I don't even mind if we stick with this for a while 😄

etoledom:
Thank you @diegoreymendez !
This PR is just to update the Gutenberg reference, but it has been already updated by other PRs. So this is not needed anymore.

Issue #546, opened 2019-02-06
Update Inserter toggle icon style


iamthomasbishop:
The Inserter toggle icon ( `(+)` ) is using the outlined icon in gray. We need to update the following:

* Use filled icon variation: The original design proposed using Gridicons (which we're using elsewhere in the apps), but Material icons also offer a pretty much exact [alternative](https://material.io/tools/icons/?search=add&icon=add_circle&style=baseline) ("add_circle"), but I'm not sure if we have easy access to that via Core. If not, we should either propose adding that to the Core set, or add it manually for now.

* Change the icon color to the Primary WordPress blue that's in use: `$blue-wordpress` aka `#0087be`.

It looks like this: 

<img width="200" alt="image" src="https://user-images.githubusercontent.com/1200432/52357785-6553d100-29fc-11e9-8725-9849dd0a88dc.png">

Related to: https://github.com/wordpress-mobile/gutenberg-mobile/issues/55

PR #628, opened 2019-02-20
"Add Block" icon color

etoledom:
fixes: #546

This PR implements the changes made in https://github.com/WordPress/gutenberg/pull/14631 to change the default color of Dashicons (in this case on `ToolbarButton`).

img_1268

To test:
- Build the app.
- Check that the add-block button is filled with `$blue-wordpress` color.


iamthomasbishop:
Is this using a Dashicon? If so, can we either use a custom SVG or Material icon here? This icon feels a little off.

etoledom:
> Is this using a Dashicon? If so, can we either use a custom SVG or Material icon here? This icon feels a little off.

Yes it is Dashicon. The `ToolbarButton` component is made on top of Dashicon component.
We can have a custom SVG, we did it before for the hide-keyboard icon. But it's kinda hacky 🤔 

If the previous icon looks better compared to this one, maybe we can wait a bit for this change?

src/block-management/block-toolbar.js
```diff
@@ -49,8 +49,9 @@ export class BlockToolbar extends Component<PropsType> {
 					<Toolbar>
 						<ToolbarButton
 							label={ __( 'Add block' ) }
-							icon="insert"
+							icon="plus-alt"
 							onClick={ onInsertClick }
+							extraProps={ styles.addBlockButton }
```
Tug: We shouldn't be setting the icon's `color` attribute as it will be prone to error if we add css rules to this `addBlockButton` class. Also it's best practice to group styling in style or className.
So:
```
extraProps={ { style: styles.addBlockButton } }
```

etoledom:
Thanks @Tug and @iamthomasbishop for the review!

I implemented the suggested changes and tweaked a bit the look of the icon as Thomas suggested, adding a border to it. Now it looks 😎 

file

iamthomasbishop:
That looks _so much better_. Thank you! <3

etoledom:
Updated with the changes requested for a better clarity on the intention.
Thank you @Tug !

etoledom:
I updated the implementation of this, with a new `gutenberg` side PR: https://github.com/WordPress/gutenberg/pull/14631

Would you mind @Tug taking another look?
Thank you!

src/block-management/block-toolbar.js
```diff
@@ -59,8 +59,9 @@ export class BlockToolbar extends Component<PropsType> {
 					<Toolbar>
 						<ToolbarButton
 							label={ __( 'Add block' ) }
-							icon="insert"
+							icon={ ( <Dashicon icon="plus-alt" style={ styles.addBlockButton } color={ styles.addBlockButton.color } /> ) }
 							onClick={ onInsertClick }
+							iconProps={ { style: styles.addBlockButton, color: styles.addBlockButton.color } }
```
Tug: Do we still need this line?
I can't find any usage of `iconProps` in `ToolbarButton`

etoledom: Right! It's not needed anymore.
Thanks for noticing 🙏

etoledom:
Thank you!

PR #629, opened 2019-02-20
[Android] setSelection for Aztec wrapper

### LINKED ISSUE #602: [Android] Cursor moves to the end of the block after split
When you tap enter on a paragraph, the cursor goes to the end of the new block on Android. This doesn't make a difference if you're typing Enter at the end of a block, but it's noticeable when you split a block in two:

1. Type some text
2. Move the cursor to some place in the middle of the text
3. Press Enter

The block splits in two and moves the focus to the second block but the cursor is at the end.

split-focus 2019-02-15 10_41_14

hypest:
Fixes #602 

~This PR introduces a "setSelection" functionality to the Aztec wrapper component. It is currently wired up to the `RichText` component as a prop and updated every time the locally kept `selectionStart` and `selectionEnd` changes. See https://github.com/WordPress/gutenberg/pull/13978.

With this setting of selection, the caret position is set to 0 for the newly created block from the block split.~

This PR adds a `MovementMethod` that places the caret at the beginning of text when focusing the view. The Android default is to place is at the end, but in our case, placing it at the start is more appropriate.

To test:

1. In a paragraph block that has some text, place the caret in the middle of the text
2. Tap "Enter" to split the block
3. Notice the caret being placed at the beginning of the new block


daniloercoli:
I noticed that sometimes when tapping on another block, say you tap between two words, the caret is initially placed at the "point of the tap", but immediately after moved to the end of the text.
https://cloudup.com/ckcVwELlawk

To reproduce the issue
1. Tap on a Block_A and start adding new content to it
2. Tap in the middle of a different RichText block (the caret is properly placed)
3. Tap back in the middle of the Block_A
4. The caret is moved to the end of Block_A

(to replicate the issue you need to modify the content of the first block).

hypest:
Confirmed. Will need to fix that.

hypest:
I've revised the solution to take another route: Removed the setting of selection, and instead, added a MovementMethod that places the caret at the beginning of text when focusing the view. The Android default is to place is at the end, but in our case, placing it at the start is more appropriate.

Now, we might want to place the caret at the end in some cases, and we shall make the custom MovementMethod smarter then.

Ready for another pass @daniloercoli , thanks!

Btw, the Gutenberg-side PR is no longer needed and will close it when we finalize this PR.

PR #630, opened 2019-02-20
Fix for node.rel crash

hypest:
Addresses at least one case of "Pasting from some sources causes a red screen ([related comment](https://github.com/wordpress-mobile/gutenberg-mobile/pull/617#pullrequestreview-205427915))" from #624

To test:

0. On the mobile, open the browser and copy the "Gravity Forms" text (it's a link) from https://designsbytierney.com/2010/01/how-to-add-edit-format-text-in-a-wordpress-post-or-page/
1. Run the demo app
2. In a paragraph block, paste
3. It should work and the link be available in the block

### DIFF
--- a/package.json
+++ b/package.json
@@ -93,7 +93,7 @@
     "jed": "^1.1.1",
     "js-beautify": "^1.7.5",
     "jsc-android": "224109.x.x",
-    "jsdom-jscore": "git+https://github.com/wordpress-mobile/jsdom-jscore-rn.git#b1f06050503fc3a145d179188384d09583865043",
+    "jsdom-jscore": "git+https://github.com/iamcco/jsdom-jscore-rn.git#a562f3d57c27c13e5bfc8cf82d496e69a3ba2800",
     "jsx-to-string": "^1.3.1",
     "memize": "^1.0.5",
     "moment": "^2.22.1",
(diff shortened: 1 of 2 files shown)


hypest:
Thanks for the pass @diegoreymendez !

I will actually need to update this PR to host a different solution as, I closed [the Gutenberg side PR](https://github.com/WordPress/gutenberg/pull/13983) as invalid. Will report here again soon.

hypest:
Ready for another pass!

This time, the solution [uses this PR](https://github.com/iamcco/jsdom-jscore-rn/pull/1) opened against the `jsdom-jscore` library repo we're using.

mkevins:
> This time, the solution [uses this PR](https://github.com/iamcco/jsdom-jscore-rn/pull/1) opened against the `jsdom-jscore` library repo we're using.

I think this will make a lot of things easier. Does this mean we can move some (maybe all?) of the jsdom-patches?

hypest:
> Does this mean we can move some (maybe all?) of the jsdom-patches?

Most probably, but let's wait until after our release this coming days so we don't shake things up too much.

hypest:
`jscore-jsdom-rn` upstream accepted our PR so, will update this branch to use the upstream.

PR #633, opened 2019-02-20
Remove JS handling of title focus

etoledom:
There is a bridge message in place to let the parent app tell `gutenberg` when it should focus the title. Android seems to be using that mechanism to focus the title.
(Is that correct @daniloercoli ?)

This PR implement the title focus feature using this message from the parent app instead of letting the JS side do it by itself.

The main reason of this change is a glitch on the toolbar + SafeAreas.
Check the WPiOS related PR to see the native side implementation.

**Before:**
focus-bug

**Now:**
focus-fix

**To test:**
- Check out WPAndroid with this branch, and check that the autofocus on new posts still works.
- Checkout the [WPiOS related PR](https://github.com/wordpress-mobile/WordPress-iOS/pull/11097) running **metro on this branch**.
- Check that the autofocus on new posts works properly.
- Check that the toolbar is properly positioned.

cc @pinarol - This might also fix the title border issue on the initial autofocus mentioned here: https://github.com/wordpress-mobile/gutenberg-mobile/pull/622#pullrequestreview-205884150

### DIFF
--- a/src/block-management/block-manager.js
+++ b/src/block-management/block-manager.js
@@ -171,16 +171,11 @@ export class BlockManager extends React.Component<PropsType, StateType> {
 	}
 
 	renderHeader() {
-		const focusTitle = this.props.title === '' && this.props.blockCount === 0;
-
 		return (
 			<View style={ styles.titleContainer }>
 				<PostTitle
 					setRef={ ( ref ) => {
-						if ( focusTitle && ref ) {
-							ref.focus();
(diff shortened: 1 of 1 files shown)


etoledom:
@pinarol @daniloercoli - Thank you!

> The demo app doesn't set the focus on the title if both the content and the title are set empty in initial content

That's because the parent app has to send the message. I believe that it's a good thing, since it adds flexibility.

PR #635, opened 2019-02-21
Show the Title when in HTML mode

### LINKED ISSUE #609: Show the Title block when in HTML mode
Title has it all.

Currently, we only show the content section and no title at all.

daniloercoli:
This PR fixes #609 by adding the Title field to the HTML view, and updates the title on the store with the proper action when typing in the component.

This PR is still not ready for merging, since there is one weird bug with the default font we need to fix before. 
In short, everything seems to work fine on Android, except when I set the default font (`//font-family: $default-monospace-font;` ). Tried both the [CSS way](https://github.com/wordpress-mobile/gutenberg-mobile/compare/issue/609-Show-Title-block-when-HTML-mode?expand=1#diff-1044633f47e5946b5b4da209242d8e50R6), and the `fontFamily` property, but in both cases the TextInput component that holds the content of the post starts acting weirdly.
If some of you can take a look and check if there is a quick fix for this @etoledom @pinarol 

Another consideration, I see some functions in the `HTMLInputView` - Ref: `edit` , `stopEditing`, `onPersist`, that are used to reset blocks list, not sure if we need to follow the same approach for the title. I tested the current simple approach and it seems to work without problems.


hypest:
Had a look and it seems that there are layout/flexbox issues with the ScrollView and the TextInputs.

pinarol:
> Another consideration, I see some functions in the `HTMLInputView` - Ref: `edit` , `stopEditing`, `onPersist`, that are used to reset blocks list, not sure if we need to follow the same approach for the title. I tested the current simple approach and it seems to work without problems.

I didn't see any problems about this either. It seems that calling setTitleAction on every text change is doing the deal.

PR #638, opened 2019-02-21
Fixes a red sceen when pasting.

diegoreymendez:
## Description:

Fixes a red screen that would come up when trying to paste plain text at the end of a block or when text was selected.

## Dependencies:

Gutenberg PR: https://github.com/WordPress/gutenberg/pull/14011

## Known issues:

The text selection after pasting is not correct, but this is not a problem introduced by this PR.

### Test 1:

1. Copy some plain text.
2. Select some text from a block.
3. Paste

Make sure you don't get a red screen.

### Test 2:

1. Copy some plain text.
2. Place the caret at the end of a paragraph or heading block.
3. Paste

Make sure you don't get a red screen.


diegoreymendez:
Thanks @etoledom !

mkevins:
Thanks @diegoreymendez. I tried this on Android, and it fixes the issue.

Issue #479, opened 2019-01-17
Update design for Unsupported blocks


iamthomasbishop:
The current iteration of the Unsupported block isn't very helpful.

Current design: 

<img width="375" alt="image" src="https://user-images.githubusercontent.com/1200432/51355226-15788e80-1a7c-11e9-92f2-4c5882c23f9a.jpeg">

Proposed design:

* Show the block type name, rather than simply "Unsupported". This will tell the user quickly which blocks are on the canvas and they can move or delete them accordingly
* Add the block type icon as an additional visual cue
* Change styling of text to be system sans-serif and the darker shade ( #2e4453 aka $gray-dark ) 

(Below: an example of a focused Unsupported block)

<img width="379" alt="image" src="https://user-images.githubusercontent.com/1200432/51349796-0f7ab180-1a6c-11e9-9349-0ab272d87dcc.png">

Note: If need be, we can tackle the `question mark` icon separately, but here's how I would expect it to work: 

* User taps the `(?)` icon
* Dialog or sheet opens
* Short message explains that some blocks aren't yet supported, links to more information on a webpage

hypest:
Thanks for the iteration @iamthomasbishop!

Currently, we're using the "Unsupported" placeholder in the cases where the post has content not wrapped in Gutenberg block delimiters too. Think, text that is "orphan", not inside any block. In such cases, there is no block type/name, nor any icons to use. Do you have a proposal how to represent such cases?

On a different note, it seems that the new styling doesn't quickly convey the message that this block is actually unsupported. It looks, to me at least, as something went wrong with the rendering, perhaps a bug. User will get the proper message only after they tap on the "?" Is that two-steps approach deliberate perhaps?

etoledom:
> Do you have a proposal how to represent such cases?

Maybe it would be good to show them in a "classic editor" block using Aztec? (following the web behavior)

koke:
> Show the block type name

We can't reliably do this. We could hardcode some of the common core blocks for now, but otherwise all we know is the block's registered "name" (not title), as in `core/list` or `jetpack/contact-form`

iamthomasbishop:
> Currently, we're using the "Unsupported" placeholder in the cases where the post has content not wrapped in Gutenberg block delimiters too. Think, text that is "orphan", not inside any block. In such cases, there is no block type/name, nor any icons to use. Do you have a proposal how to represent such cases?

[...]

PR #643, opened 2019-02-22
Improve unsupported blocks

koke:
Show localized title and icon of unsupported core blocks. It also drops our custom `gmobile/unsupported` block and moves the functionality to a native variant of `core/missing`.

Gutenberg PR: https://github.com/WordPress/gutenberg/pull/14577

Simulator Screen Shot - iPhone XS - 2019-03-21 at 11 06 30

To test:

- Create a post with some blocks that are unsupported in the app.
- Open it in gutenberg-mobile and verify that the blocks show the right icon and localized block title. For non-core blocks, a generic icon is shown.

Part of #479


koke:
I haven't figured out how to set the right colors yet, and then I realized that translations weren't working for block titles, so I focused on that first, but cc @iamthomasbishop since you might want to keep track of this one

iamthomasbishop:
Awesome! I think I might need to update the colors here anyway, as the design changed slightly. IIRC, I just made the icon and text label `$gray-dark` aka `#2e4453`, but I'll double check.

iamthomasbishop:
Just for the sake of consistency and documentation, some specs at the end:

Static:
image

Focused:
image

* Icon and text label are indeed `$gray-dark` aka `#2e4453`
* Background has a border-radius of 4pt
* Text label uses system sans-serif font (SF Pro, Roboto) at 14pt
* Icon is 24pt
* Margin between icon and text label is 4pt

Tug:
Did a bit of refactoring, but this LGTM 👍 

Btw, this introduces some "Circle dependency" warnings when booting the app in dev mode. We indeed have circle dependencies in our code (often in gutenberg), I think those warnings didn’t show up before because we would load those components before our app component is even loaded but there were still there.

koke:
Since you're handling the `hasUnsupportedBlocks` issue in #790, should we go ahead and merge this one? (After resolving conflicts 😩)

etoledom:
Sure! One last smoke test.

PR #646, opened 2019-02-22
Implement pasteHandler with plain text fallback

mkevins:
## Description:
Moves forward #624

This PR aims to guard against errors in production during a paste event on mobile. As there are potentially many unexplored code paths (not yet covered by tests), the goal here is to call the original `pasteHandler` in a try/catch block, and fallback to plain text if an error is encountered.

Note that even in the plain text case, the `pasteHandler` may still call into some of the complex web code.

## Dependencies:

Gutenberg PR: https://github.com/WordPress/gutenberg/pull/14044

## Testing:

In order to test, we have to establish a state where a known error exists. A good source to trigger the error is here: https://github.com/wordpress-mobile/gutenberg-mobile/pull/617#pullrequestreview-205427915, but we need to rewind (or revert) the fixes introduced in #630.
1. Roll back to pre #630 (`git checkout 1fead08b5` in gutenberg-mobile and `git submodule update`)
2. `cd gutenberg` and cherry pick commit
3. Run demo (may need `yarn clean:install` first to clear cached dependencies)
4. Follow instructions here: https://github.com/wordpress-mobile/gutenberg-mobile/pull/617#pullrequestreview-205427915 copying text surrounding 'Gravity Forms'
5. Paste into a new paragraph block

Text should paste as plain text (losing the link and styles). It should not error out to the red screen.


koke:
Updated PR in #650

PR #647, opened 2019-02-22
Fix links ui on landscape iOS

etoledom:
## Description

This PR updates the Gutenberg ref to test https://github.com/WordPress/gutenberg/pull/14047.

It fixes an issue on KeyboardAvoidingView, that was adding extra bottom padding on specific circumstances:

- iOS.
- Landscape.
- Keyboard already present.
- BottomSheet presenting with a TextInput on Autofocus.
- It also had issues without Autofocus.

With some debugging I noticed a problem on the way the extra padding is calculated, so I made a simplified version of the component that will work well on the BottomSheet.

links-ui-landscape

**Note:** The BottomSheet is not scrollable (yet).

**To test:**
- Check that the Links UI presents properly on Landscape mode.
- If possible, test WPiOS and Android.
- Android should work exactly as before.


etoledom:
This issue doesn't happen on the WP apps, it seems to just affect the example apps, so I'm moving the Milestone to check it out again later.

Thank you @pinarol !

pinarol:
so the extra padding issue is still not reproducible on my side even on example app. but there's another issue I could repro and this PR seems to fix that:

Before this PR the bottom sheet sometimes don't show itself in portrait mode:
before-fix-bottom-sheet

So this looks fixed now.

There are other 2 small issues reproducible after the fix though:

1. The keyboard hide button stops working after the bottom sheet is closed until I manually go and tap to text in the block again.

2. Don't know if this is a bug but when I tap to an outside area just after opening the BottomSheet and with no changes made(because I just want to close the BottomSheet) we add the link anyway using an empty url. I think if the user hadn't entered a url text this means we shouldn't add a link and just close the bottom sheet. wdyt?

bottom-sheet-keyboard-close

etoledom:
Closing in favor of https://github.com/wordpress-mobile/gutenberg-mobile/pull/701

PR #651, opened 2019-02-22
Fix i18n initialization priority

koke:
Before this, the block titles were not being translated because the translation attempt happened before the locale data was set up.

This refactors the initialization process to ensure we setup localization before Gutenberg is imported

simulator screen shot - iphone xs - 2019-02-22 at 13 40 06

This was originally implemented as part of #643 but that one isn't ready to merge yet

## To Test

1. Switch your phone/simulator to another language
2. Try to insert a block and verify the title is translated

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -61,7 +61,7 @@ const setupLocale = ( locale, extraTranslations ) => {
 	}
 };
 
-class RootComponent extends React.Component {
+export class RootComponent extends React.Component {
 	constructor( props ) {
 		super( props );
 		setupLocale( props.locale, props.translations );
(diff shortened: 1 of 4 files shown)


src/index.js
```diff
@@ -86,3 +85,8 @@ export function registerApp() {
 
 	AppRegistry.registerComponent( 'gutenberg', () => RootComponent );
 }
+
+export function setupApp() {
```
hypest: Nit, let's call this `bootstrapEditor` or something. The "App" moniker starts to feel too vague, WDYT? :)

Tug: Agreed 👍

PR #655, opened 2019-02-22
Fix title padding on Android

pinarol:
This PR separates the title padding styles of iOS and Android and removes the vertical paddings of post title on Android because Android handles the paddings on native side.

Changes reside in gutenberg PR https://github.com/WordPress/gutenberg/pull/14057


pinarol:
@hypest ready for another look

PR #658, opened 2019-02-22
Fix yarn makepot

Tug:
This is not a priority for 1.0 release but it could use a fix. Not touching anything critical...
Just fixing the the pot file generation using the xgettext command line.

The current command assumes `gutenberg.pot` is generated in `gutenberg/languages/` and that bash is the default shell to run commands (sh is probably the default one actually)


package.json
```diff
@@ -59,7 +59,8 @@
     "bundle:ios": "mkdir -p bundle/ios && yarn react-native bundle --platform ios --dev false --entry-file index.js --assets-dest bundle/ios --bundle-output bundle/ios/App.js --sourcemap-output bundle/ios/App.js.map",
     "i18n-cache": "node i18n-cache/index.js",
     "postinstall": "yarn i18n-cache",
-    "makepot": "xgettext -f <( find . \\( -path './gutenberg/*/src/*.native.js' -or -path './src/*.js' \\) ! -path 'node_modules' -print ) --from-code=UTF-8 -k__ -k_n -k_x -k_nx -x ./gutenberg/languages/gutenberg.pot -o ./gutenberg-mobile.pot",
+    "premakepot": "npm run build --prefix gutenberg",
```
koke: Wouldn't this still need a `npm install` inside `gutenberg`, leaving a populated `node_modules` there?

Tug:
This is out of date since babel-plugin-makepot was removed from gutenberg in https://github.com/WordPress/gutenberg/pull/12559

PR #659, opened 2019-02-22
Update i18n cache translation urls and new languages

Tug:
This PR updates the `i18n-cache` module to fetch the translation files from https://translate.wordpress.org/projects/wp-plugins/gutenberg directly instead of wp.me.

It also adds new languages that seems to be supported.

### Testing Instructions

- Run `yarn clean:i18n-cache`
- Run `yarn i18n-cache`
- Check that no translation is failing to download and that they are all available in `i18n-cache/data`

### DIFF
--- a/i18n-cache/index.js
+++ b/i18n-cache/index.js
@@ -8,6 +8,8 @@ const fetch = require( 'node-fetch' );
 const supportedLocales = [
 	'ar', // Arabic
 	'bg', // Bulgarian
+	'bo', // Tibetan
+	'ca', // Catalan
 	'cs', // Czech
 	'cy', // Welsh
 	'da', // Danish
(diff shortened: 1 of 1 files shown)


i18n-cache/index.js
```diff
@@ -8,41 +8,56 @@ const fetch = require( 'node-fetch' );
...
 	'tr', // Turkish
+	'uk', // Ukrainian
+	'ur', // Urdu
+	'vi', // Vietnamese
 	'zh-cn', // Chinese (China)
 	'zh-tw', // Chinese (Taiwan)
 ];
 
-const getLanguageUrl = ( locale ) => `https://widgets.wp.com/languages/gutenberg/${ locale }.json`;
+const getLanguageUrl = ( locale ) => `https://translate.wordpress.org/projects/wp-plugins/gutenberg/dev/${ locale }/default/export-translations\?format\=json`;
```
hypest: Do we have a good understanding of which code commit is the `dev` (`Development` on GlotPress) channel corresponding to? And in particular, how can we know that it perfectly matches the code version of gutenberg-mobile?

hypest: Will go ahead and merge this PR. It'd still be useful if we can document the potential mismatch between our code version and the translations version on GlotPress.

hypest:
Tug is AFK these days and I'd prefer if we include this work in the code freeze today so, will merge this myself.

PR #663, opened 2019-02-22
Image placeholder moves out of its bounds once we tap on it

marecar3:
References : https://github.com/WordPress/gutenberg/pull/14066


SergioEstevao:
I tested this on iOS and now the placeholder no longer moves around over the other blocks when the block is selected.

hypest:
I moved this PR to target the `release/1.0.1` branch, so we include it in the "hotfix" release.

PR #669, opened 2019-02-25
Release v1.0.1b

mzorz:
This is just a small update to v1.0.1 that brings #668 into it.

**Note:** with it being such a nit, I'm leaving this PR open at this point to avoid messing the versioning. If we see a need to include the fix in the `v1.0.1` we can use this PR and update https://github.com/wordpress-mobile/WordPress-Android/pull/9324 to point to the merge commit accordingly, then update `develop` with `master` in this gutenberg-mobile repo as well. I intentionally avoided creating a `v1.0.2` hotfix for this small change, and preferred to leave the option open for choosing to include/not include this fix within the same window in the next few hours.

### DIFF
--- a/src/variables.android.scss
+++ b/src/variables.android.scss
@@ -6,6 +6,6 @@ $default-regular-font: serif;
 $title-block-padding-top: 0;
 $title-block-padding-bottom: 0;
 
-$min-height-title: 53;
+$min-height-title: 55;
 $min-height-paragraph: 50;
 $min-height-heading: 50;


hypest:
👋 @mzorz , thanks for opening this PR!

I tested #668 on an emulator and the block resizing "movement" is still there after setting the system font size to large so, I think our solution to have a fixed-valued block height is not bullet proof anyway.

So, trying to eliminate the small movement like this is not saving us much. The original fix (#664 ) was needed because the size difference was so big that the placeholder text was visibly clipped, thus ugly). I think that we can postpone a `1.0.1b` until after we meet in Málaga for our team meetup. For the WP apps freeze today we can live with the `v1.0.1` I think.

mzorz:
> I think our solution to have a fixed-valued block height is not bullet proof anyway.

Totally agreed! Going ahead with 1.0.1 and closing this one 🎉

PR #675, opened 2019-02-26
Fixes pasting in heading blocks.

diegoreymendez:
## Description:

Fixes pasting in header blocks, when the paste operation would result in a block split.

## Details:

In order to achieve this I copied the splitting code from the paragraph block into the heading block.  

There's a chance to unify code there, but I have purposely not stepped into that, as we would need to make sure all blocks that use rich text can use the same code (title?), or find a solution that would let us customize it for some blocks (this is probably what we'd want to do).

## Related PRs:

Gutenberg PR: https://github.com/WordPress/gutenberg/pull/14118

## Testing:

1. Copy two paragraphs of text.
2. Paste them into a header block.


daniloercoli:
Sorry @diegoreymendez if the review process took so long. 
I've left a comment in the companion PR, since there seems to be a small issue (easy to fix).

diegoreymendez:
@daniloercoli - No worries!

This is ready for another look.

Issue #376, opened 2018-12-13
Image block fails to load images on private sites


koke:
This is WordPress.com specific, but on private sites, Gutenberg won't be able to load any images in the editor as those require authentication.

We need to figure out how networking will work before tackling this one

etoledom:
I was thinking on letting the parents app manage Image downloading/uploading, authentication and caching with the code that is already in place, and then give back a local URL to RN to show the already downloaded image.

Not sure if it's the best approach, but it will save us from rewriting all that code.

What do you think?

koke:
I think for images we’ll have to do some of that for #206 anyway. 

I wonder if react native offers a way to customize their image loader instead of having to wrap the image component with custom logic

marecar3:
Hey @SergioEstevao, we need to tackle this one also.

koke:
I moved this to nice to have since its low reach and I expected not a trivial effort: we will have to figure out how to authenticate image requests. Unless this is much easier than I anticipated, I’d leave it as a nice to have

marecar3:
Hey @koke, thanks for the update.
@SergioEstevao do you maybe think of some rough estimation for this one? WDYT? Thanks.

hypest:
I added the "Needs Android dev" label since apparently we're only missing the functionality in the Android side now.

designsimply:
I ran into this problem while testing WPiOS 12.1 beta. I only had trouble viewing images added through the "Take Photo" flow for private sites on WordPress.com—the other upload flows such as "Choose photo from device" worked normally.

Steps to reproduce:

<ol>
<li>Start logged in to the app with an account that has a WordPress.com private blog.</li>
<li>Go to Me > App Settings and toggle "Use Block Editor" on.</li>
<li>Go to the private site > Media > + > Take photo.</li>
<li>Take a photo and wait for it to upload.</li>
<li>Tap on "Write Post."</li>
<li>Try to view the image in the post.</li>
<li>Give the post a title and publish it.</li>
<li>Try switching back and forth from the HTML to Visual views.</li>
<li>View the post in a browser and note the image is there.</li>
<li>View the post in Blog Posts and note the image is there.</li>
<li>Go back to the block editor and try to view the image in the post again.</li>
</ol>

[...]

PR #678, opened 2019-02-27
Allow gutenberg bridge to add custom modules.

SergioEstevao:
Refs: #376 

This PR changes the GB bridge class to support the customization of the bridge with extra modules. It also implements on the custom a sample custom image loader, to allow testing of loading of images.

How to test:
 - Start the demo app
 - Check that images are being loaded.

### DIFF
--- a/react-native-gutenberg-bridge/ios/Gutenberg.swift
+++ b/react-native-gutenberg-bridge/ios/Gutenberg.swift
@@ -9,6 +9,9 @@ import RNTAztecView
 
 @objc
 public class Gutenberg: NSObject {
+
+    private var extraModules: [RCTBridgeModule];
+
     public lazy var rootView: UIView = {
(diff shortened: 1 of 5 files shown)


react-native-gutenberg-bridge/ios/Gutenberg.swift
```diff
@@ -104,8 +108,9 @@ extension Gutenberg: RCTBridgeDelegate {
     public func extraModules(for bridge: RCTBridge!) -> [RCTBridgeModule]! {
         let aztecManager = RCTAztecViewManager()
         aztecManager.attachmentDelegate = dataSource.aztecAttachmentDelegate()
-
-        return [bridgeModule, aztecManager]
+        var modules:[RCTBridgeModule] = [bridgeModule, aztecManager]
+        modules.append(contentsOf: extraModules)
+        return modules
```
koke: I think you can just do `return [bridgeModule, aztecManager] + extraModules` here

PR #696, opened 2019-03-01
Add CircleCI config

jtreanor:
Adding a CircleCI config for this repo. I have moved `.travis/travis-checks-js.sh` to `bin/ci-checks-js.sh` so that it can be shared by Travis and CircleCI.

After this has been running for a while, I will turn off Travis got the repo.

To test:

- CircleCI checks are ✅ 

Update release notes:


.circleci/config.yml
```diff
@@ -0,0 +1,64 @@
...
+        default: false
+    docker:
+      - image: circleci/node:8
+    steps:
+      - checkout
+      - run:
+          name: Checkout Gutenberg
+          command: |
+            git submodule init
+            git submodule update
```
hypest: We are using the `--recursive` option to git submodule update in Travis so far and, I think it'd be better to replicate that here too. It would also give us the flexibility to add other submodules at any point the future without having to remember to come and change the CI configuration.

In case this here is taken from [the README page](https://github.com/wordpress-mobile/gutenberg-mobile#clone-the-project), I think we should probably update the README to include the `--recursive` too (but not necessarily in this PR).

jtreanor: Updated now 😄

PR #700, opened 2019-03-05
RNAztec iOS: Fix removing content after dictation

### LINKED ISSUE #673: [iOS] Dictation doesn't work
When you dictate some text on iOS, it will briefly show then get deleted.

Tested with WPiOS 11.9.0.20190225

etoledom:
Fixes #673 

This issue was originally fixed here https://github.com/wordpress-mobile/gutenberg-mobile/pull/610, but it resurfaced.

The original fix was blocking the empty strings to be sent from `textViewDidChange` to the RN side, when UIKit was inserting the dictation text. That empty string was finally overwriting the content when it was sent back from RN to Native, since with an empty string, the JS side nullifies the `eventCount`.

I noticed that in this case, the same was happening on `textViewDidChangeSelection`, so using the same strategy there solved the problem.

As discussed in the original PR, this is not the best fix, but a better solution will probably involve all platforms. If this continue being an issue in the future, I'd suggest to give some time to find that better solution.

dictation

**To test:** 
- Build and run the project.
- Create an empty paragraph block.
- Test the iOS dictation feature.
- Check that it behaves properly.

### DIFF
--- a/package.json
+++ b/package.json
@@ -108,7 +108,7 @@
     "react": "16.6.1",
     "react-native": "0.57.5",
     "react-native-hr": "git+https://github.com/Riglerr/react-native-hr.git#2d01a5cf77212d100e8b99e0310cce5234f977b3",
-    "react-native-keyboard-aware-scroll-view": "git+https://github.com/wordpress-mobile/react-native-keyboard-aware-scroll-view.git#gb-v0.8.4",
+    "react-native-keyboard-aware-scroll-view": "git+https://github.com/wordpress-mobile/react-native-keyboard-aware-scroll-view.git#gb-v0.8.6",
     "react-native-modal": "^6.5.0",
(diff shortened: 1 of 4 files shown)


pinarol:
hey @etoledom 👋 I tried testing this on the WPiOS app and there seems to be some problems:

1. it can’t write the whole word
2. caret position is messed up
3. dictation listening ends very early I don’t know why

ezgif com-video-to-gif 8

etoledom:
Thank you for the review @pinarol !

I updated the PR from develop and it seems to fix that issue you mentioned. Testing on WPiOS it seemed necessary to update the JS Bundle, since the app was crashing without metro running, so I did that too.

I also opened a WPiOS PR for easier testing with a description of the original issue:
https://github.com/wordpress-mobile/WordPress-iOS/pull/11198

etoledom:
Thank you!

PR #701, opened 2019-03-05
Fix links ui on landscape iOS (v2)

etoledom:
## Description

This PR updates the Gutenberg ref to test https://github.com/WordPress/gutenberg/pull/14240.

It fixes an issue on KeyboardAvoidingView, that was adding extra bottom padding on specific circumstances:

- iOS.
- Landscape.
- Keyboard already present.
- BottomSheet presenting with a TextInput on Autofocus.
- It also had issues without Autofocus.

With some debugging I noticed a problem on the way the extra padding is calculated, so I made a simplified version of the component that will work well on the BottomSheet.

links-ui-landscape

**Note:** The BottomSheet is not scrollable (yet).

**To test:**
- Check that the Links UI presents properly on Landscape mode.
- If possible, test WPiOS and Android.
- Android should work exactly as before.


etoledom:
Continuing the discussion from 

Thank you for your review @pinarol !

> The keyboard hide button stops working after the bottom sheet is closed until I manually go and tap to text in the block again.

That's right, I've noticed that too. Let's create a new issue for this one. 👍

> Don't know if this is a bug but when I tap to an outside area just after opening the BottomSheet and with no changes made(because I just want to close the BottomSheet) we add the link anyway using an empty url. I think if the user hadn't entered a url text this means we shouldn't add a link and just close the bottom sheet. wdyt?

I agree that we shouldn't try to add a link if the url text is empty. I'd say that this also deserves its own ticket.

Currently we don't have a `save` and/or `cancel` button, so we are `saving` on dismiss, even if it's tapping in the background or swiping down.
Not sure if this is the best way to "save" in general.

@iamthomasbishop what do you think?

pinarol:
Adding the gif that describes the issues:

bottom-sheet-keyboard-close

iamthomasbishop:
> I agree that we shouldn't try to add a link if the url text is empty. I'd say that this also deserves its own ticket.

Agreed.

> Currently we don't have a save and/or cancel button, so we are saving on dismiss, even if it's tapping in the background or swiping down. Not sure if this is the best way to "save" in general.

The settings/input of the bottom sheet should be applied immediately where possible, so regardless of the dismiss method (swipe to hide or tapping the scrim in the example given above) we should be saving the contents. It essentially acts the same as a popover on iOS would.

I also think we should consider a couple things:

* Tapping `return` should probably dismiss the sheet altogether. Right now, it just hides the keyboard. If we do this, we should probably rename this ^^ button to `done`.

* We should make sure the text input for URL use the URL input type/keyboard.

* (Maybe for another ticket): We are automatically adding `http://` to URLs when the user types a straight URL. If I type `wordpress.org`, I would expect it to stay as `wordpress.org`.

etoledom:
Thank you @pinarol and @iamthomasbishop !
Issues opened: #702 and #703

Issue #208, opened 2018-10-29
List block


koke:
Add support for the `list` block.

I don't know the difficulty of implementing the indent and outdent buttons, but I'd be fine leaving those out of the beta. I don't see anything else on the web version to leave out.

List-related issues:

- [x] Editor crash when undoing at list end #836
- [x] List crashes with extra spaces #871
- [x] List block split: item and caret at wrong lines when nested lists #886
- [ ] List block split problem with nested lists #874

Not blockers:

- [ ] List Block Ordered list number indicators don't respect alignment #846

koke:
This was estimated at 1-2 weeks by @daniloercoli, but as @diegoreymendez pointed out, we were not accounting for nested lists. Hopefully those are easy, but they aren't as critical for the beta, so we'll count *2 weeks per platform* for this task and re-evaluate when we know more

hypest:
Known issues after https://github.com/wordpress-mobile/gutenberg-mobile/pull/704:

### caret positioning when new item at the end
1. Do a long list, long enough to have a few empty list items here and there
2. Try to add a new item at the end of the list
3. Notice that the caret doesn't land on the new list item but rather, goes "back". The more empty lines from 2., the more chars the caret goes back.

koke:
Let's keep this open until we have nested lists, and the implementation is solid enough to enable on production builds.

koke:
To avoid cluttering the 2.0 Bugs column, I started a checklist on this issue with all the list-related bugs. Please keep adding there as you find them. When this is ready for production, let's make sure any remaining items have a card in the project

iamthomasbishop:
Just added my notes from a quick design review on the current (as of the current alpha) list block: https://github.com/wordpress-mobile/gutenberg-mobile/issues/897

mchowning:
Added a minor issue related to caret positioning when splitting list items: #929

hypest:
Bumping this to the next milestone.

koke:
Since we're shipping lists for both platforms in 1.4, and we have individual issues for every known bug, I think we can close this one?

hypest:
Will strikethrough the non-checked items in the ticket and will leave the individual tickets to have a life of their own 👍 .

PR #704, opened 2019-03-05
Feature/lists

SergioEstevao:
Refs #208 
~Gutenberg PR https://github.com/WordPress/gutenberg/pull/14249~
New Gutenberg PR https://github.com/WordPress/gutenberg/pull/14636

simulator screen shot - iphone xr - 2019-03-05 at 22 29 37

This PR implements the list block by only changing components and **not touching** the code of the original web block.

This achieved by improving the implementation of the block-edit context in order to send the isSelected and onFocus props to the RichText components.

For the moment it doesn't implement the indent and outdent buttons, because of a bug in Aztec for iOS that doesn't allow split to work properly when lists are indented.

If testing on iOS please run `yarn preios` to make sure you get the right version of Aztec. 

How to test:
 - Start the demo app 
 - Check the list display properly and you can interact with them and change type
 - Add new list block to the post
 - Check that the list block works properly.


react-native-aztec/src/AztecView.js
```diff
@@ -146,7 +146,7 @@ class AztecView extends React.Component {
           onContentSizeChange = { this._onContentSizeChange }
           onHTMLContentWithCursor = { this._onHTMLContentWithCursor }
           onSelectionChange = { this._onSelectionChange }
-          onEnter = { this._onEnter }
+          onEnter = { this.onEnter ? this._onEnter : null }
```
hotchemi: Might be really nitpicking but `undefined` is slightly better?

Tug: I guess you meant `this.props.onEnter && this._onEnter` here?

hotchemi: Ah yea!
..and I get used to TS and [only using undefined](https://github.com/Microsoft/TypeScript/wiki/Coding-guidelines#null-and-undefined) but found there's no standard equivalent option for eslint..ignore me😅

SergioEstevao:
Here is update of the current status of this PR:
 
- Insertion of a new list block
- Display of list block already present in the content
- Ability to write inside the lists
- Convert a block list from ordered `ol` to unordered 'ul' using the toolbar
- Indent in is working
- Indent out is not working
- When splitting the block an extra paragraph block is being inserted on the middle of the split.

All the above are working on iOS, but on Android we have some issues:

- Changing from bullets to number list is not working on Android, this is done on JS code so it's strange for me why is not working.
- Aztec android needs to implement the isMultiline prop added to the component. When this boolean parameter is on onEnter prop should only be invoked when an Enter is pressed and the previous line is an empty line.

@koke do you think after the Android issues we should merge this with a feature flag in place?

@hypest do you want to tackle the Android issue?

hypest:
As far as I can see on Android, adding a new item to the list by tapping Enter at the end of an item is broken. Instead of adding a new bullet, the block is actually split. Haven't debugged this yet but, @SergioEstevao are you referring to that too when you say `Aztec android needs to implement the isMultiline prop added to the component. When this boolean parameter is on onEnter prop should only be invoked when an Enter is pressed and the previous line is an empty line.`?

> @hypest do you want to tackle the Android issue?

Yeap, added it to my plans for this sprint 👍

SergioEstevao:
Yes it's exactly that! When the multiline property is true pressing Enter
shouldn't send an onEnter event up that causes the split.

koke:
> When splitting the block an extra paragraph block is being inserted on the middle of the split.

Can you explain this one?

I think it’s fine to merge with a feature flag when Android is working. If indent out is not working, I’d disable both buttons (in/out) until that works

SergioEstevao:
@koke here is a video:

ListSplits

koke:
I'm seeing the same on the web (except for the stray bullet in the second list):

lists-web

SergioEstevao:
> I'm seeing the same on the web (except for the stray bullet in the second list):
> 
> lists-web

Should I open a bug on the GB repo ? :)

koke:
No, I think that's the right behaviour. I just checked and Google Docs, Apple Pages, and Microsoft Word do the same thing.

SergioEstevao:
@Tug do you mind to test this. If testing on iOS please run `yarn preios` to make sure you get the right version of Aztec.

src/block-management/block-holder.js
```diff
@@ -238,6 +239,7 @@ export default compose( [
 			isLastBlock,
 			isSelected,
 			name,
+			clientId,
```
Tug: `withSelect` only add new props to the component, so it should not be necessary to explicitly add `clientId` which is already a prop given from the parent component here.

src/block-management/block-holder.js
```diff
@@ -181,6 +181,7 @@ export class BlockHolder extends React.Component<PropsType, StateType> {
 				insertBlocksAfter={ this.insertBlocksAfter }
 				mergeBlocks={ this.mergeBlocks }
 				onCaretVerticalPositionChange={ this.props.onCaretVerticalPositionChange }
+				clientId={ this.props.clientId }
```
Tug: That's good, it should have been added earlier 👍

react-native-aztec/src/AztecView.js
```diff
@@ -157,10 +157,13 @@ class AztecView extends React.Component {
...
           onSelectionChange = { this._onSelectionChange }
-          onEnter = { this._onEnter }
+          onEnter = { this.props.onEnter && this._onEnter }
+          // IMPORTANT: the onFocus events are thrown away as these are handled by onPress() in the upper level.
+          // It's necessary to do this otherwise onFocus may be set by `{...otherProps}` and thus the onPress + onFocus
+          // combination generate an infinite loop as described in https://github.com/wordpress-mobile/gutenberg-mobile/issues/302
           onFocus = { this._onAztecFocus } 
           onBlur = { this._onBlur }
-          onBackspace = { this._onBackspace }
+		  onBackspace = { this._onBackspace }
```
Tug: Minor but we're mixing tabs and spaces here, I think we should convert the whole file to tabs for more consistency

PR #709, opened 2019-03-07
Avoid to reset html to empty string if block is heading and platform is android

marecar3:
Following PR : https://github.com/WordPress/gutenberg/pull/14301


hypest:
The Gutenberg side PR is now merged so, let's update this PR and merge it.

PR #731, opened 2019-03-11
Use last gutenberg changes including the editor and block-editor packages refactoring

Tug:
Requires https://github.com/WordPress/gutenberg/pull/14375

https://github.com/WordPress/gutenberg/pull/13088 introduced a breaking change for us, we need to update our dependency to the editor store and components as it (the editor package) has been split in 2 packages: `core/editor` and `core/block-editor` for more flexibility.

This PR updates our code to make it work with those changes:

- The main change of this PR is the switch from `core/editor` to `core/block-editor` selectors and actions in `BlockManager` along with the use of `BlockEditorProvider` in the parent component which makes sure both stores are in sync.

- In `AppContainer`, `MainApp`'s is removed and its logic is moved to `AppContainer` as the component we're using `BlockEditorProvider` handles defining the `SlotFillProvider` already.

- `editorSetup()` is simplified in `src/index.js` and most of its logic is moved to `edit-post/src/index.native.js`. To achieve this we needed to also move `UnsupportedBlock` to gutenberg.
 
- Styles (open for discussion): `src/colors.scss` and `src/variables.(ios|android).scss` are replaced with `src/_colors.scss` and `src/_variables.(ios|android).scss` and now supersede the files in `gutenberg/assets/stylesheets/`. This means that, just like with gutenberg, we don't need to explicitly have `@import "variables.scss";` or `@import "colors.scss";` in our stylesheets. This is closer to what we have on gutenberg web. However it also means that we need to redefine all necessary variables when we decide to have our own version of `src/_{file}.scss`.

- `@wordpress/block-editor`, `@wordpress/edit-post` and `@wordpress/token-list` are added to the list of symlinked packages

### Testing Instructions
- Try the demo app on both android and ios and notice that there is no regression
- Try running the app inside its parent app and notice there is no regression


etoledom:
Hey @Tug - this is working super good! 🎉
I've been testing on iOS, and so far seems that there are no regressions.

I did find something that I wanted to share early:

It looks like there is a new formatting option on paragraph and heading blocks.
This button doesn't work, but the rest of the buttons work as in develop.

IMG_1461

Tug:
@etoledom Thanks for reviewing. About the new button that's interesting! I didn't notice it :)

Looking at it now 👍

Tug:
So it seems the `<code>` formatting was added in https://github.com/WordPress/gutenberg/pull/12249 even though it didn't display. For some reason it does now and it seems to work fine at first glance. Since we didn't test it yet I have removed it from our supported formats for now

etoledom:
Thanks for the update @Tug !

Testing on device it gives me the same error that is shown on the CI tests:
IMG_1462

Tug:
@etoledom Yep, I needed to make [an update on gutenberg](https://github.com/WordPress/gutenberg/pull/14375/commits/e6b27811e00d394fca86231155dede4de9ae5ae6) after the merge, our babel config doesn't support `export from`

etoledom:
Thank you @Tug for the update! It looks super good on iOS.
I'm still testing on WPiOS and then Android.

Just wanted to point out that the last commit on 
https://github.com/WordPress/gutenberg/pull/14375 seems to break tests there. 🤔

daniloercoli:
Tested on Android and was not able to reproduce the problem. 
For me this looks good and we can fix the problem on Android later if we notice it again.

hypest:
Nice work @Tug 🎉

PR #733, opened 2019-03-12
Fix for focus moved to the wrong block when pressing delete at beginning of a block and no merge

### LINKED ISSUE #345: Pressing delete at the beginning of a RichText block (and NO merge) moves the caret to another RichText block
Pressing delete at the beginning of a RichText block (when the merge action cannot happen due to different "unmergeable" block types) moves the focus to the block above (which is correct) but the caret is moved to another RichText block in the list. Usually the first at the top.

merge-and-caret

To reproduce this issue:
- Start the demo app
- Move the cursor at the beginning of the Para block `Lorem...`
- Tap `delete` 
- The block above the current, a "Page Break" block, is selected (correct), but the caret is moved to another RichText/ PlainText block on the list

Note that in GB the standard behavior is to move the focus to the previous block if the merge action cannot happen: https://github.com/WordPress/gutenberg/blob/master/packages/editor/src/store/effects.js#L165

daniloercoli:
This PR fixes #345 by checking whether the previous block is mergeable before calling the merge routine.

The default GB implementation *does* move the focus to the upper block, instead we want to keep the focus and caret to the actual position, otherwise we need to know - at mobile level - if the merge happened or not, and react by hiding the keyboard and remove the focus from the current block.


daniloercoli:
cc @Tug

Tug:
🤔 So it seems it’s actually a desired feature on gutenberg to shift focus to the previous block if it’s not mergeable. Don’t you think we should keep the same behavior on mobile?

daniloercoli:
Yeah, maybe. It's kind of difficult to hide the keyboard, hide the caret, and put the focus on a non textual block at the moment. So the easiest solution is to disable the behaviour on mobile.

Tug:
Oh right I didn't think about "non textual block"... Probably we could have this check in `BlockHolder.onFocus` but I'm not sure we can determine if a block is textual or not 👍

PR #739, opened 2019-03-12
Fix for Placeholder that is missing on heading block

daniloercoli:
This PR does fix the Placeholder text that is missing on the heading block (and probably other non-paragraph text blocks too).

In this PR we're passing down the `blockType` prop to the native wrapper, and use it to wrap the text the first time the user inserts a new value in the field. In this way Aztec properly render the text on the screen and the GB toolbar updated accordingly.

<del>I still need to re-check the code that does adds the tags around the text, since i'm not fully convinced it's safe, but the general idea is here.</del>

GB side PR available here: https://github.com/WordPress/gutenberg/pull/14404

**To test this PR**

__Test 1__
- Add a new heading block
- The placeholder text should appear
- Start typing in the block
- The styling should be there and the toolbar updated

__Test 2__
- Remove all the text in a Heading block
- The placeholder text should appear again
- Start typing in the block
- The styling should be there and the toolbar updated

__Test 3__
- Add a new heading block
- The placeholder text should appear
- Paste a text in the block
- The styling should be there and the toolbar updated


marecar3:
Nice work @daniloercoli! 
Seems that you found a robust solution which is good :)

While I was testing this PR on my Nexus 5x, Android: 8.0, I have noticed some glitch.
When you create a new heading block and after you type in a first letter you can notice that heading input field will resize very slow. 

ezgif com-video-to-gif (1)

I am wondering, what amount of work can it take to assign same formatting style (`<h1>`)for the `Placeholder` like iOS has or at least to increase the initial size of RichText when type is heading, WDYT ?

react-native-aztec/android/src/main/java/org/wordpress/mobile/ReactNativeAztec/ReactAztecManager.java
```diff
@@ -504,6 +513,11 @@ public void onTextChanged(CharSequence s, int start, int before, int count) {
                             oldText,
                             start,
                             start + before));
+
+            // Add the outer tags when the field was started empty, and only the first time the user types in it.
+            if (mPreviousText.length() == 0 && currentEventCount == 1 && !TextUtils.isEmpty(mEditText.getTagName())) {
+                mEditText.fromHtml('<' + mEditText.getTagName() + '>' + newText + "</" + mEditText.getTagName() + '>', false);
```
marecar3: Is there maybe some Util that we can use to generate the `html` from tag name and text in order to have one place for adding the `<,>` when there is a need for them?

react-native-aztec/android/src/main/java/org/wordpress/mobile/ReactNativeAztec/ReactAztecManager.java
```diff
@@ -504,6 +513,11 @@ public void onTextChanged(CharSequence s, int start, int before, int count) {
                             oldText,
                             start,
                             start + before));
+
+            // Add the outer tags when the field was started empty, and only the first time the user types in it.
+            if (mPreviousText.length() == 0 && currentEventCount == 1 && !TextUtils.isEmpty(mEditText.getTagName())) {
```
marecar3: If you agree we can probably put `mPreviousText.isEmpty()` instad of `mPreviousText.length() == 0`
also it would be to put `currentEventCount == 1` in some constant with proper name.

react-native-aztec/android/src/main/java/org/wordpress/mobile/ReactNativeAztec/ReactAztecManager.java
```diff
@@ -302,6 +303,13 @@ public void setColor(ReactAztecText view, @Nullable Integer color) {
         view.setTextColor(newColor);
     }
 
+    @ReactProp(name = "blockType")
+    public void setBlockType(ReactAztecText view, ReadableMap inputMap) {
+        if (inputMap.hasKey("tag")) {
```
marecar3: Maybe we can put `"tag"` in some constant in order to reuse it?

react-native-aztec/android/src/main/java/org/wordpress/mobile/ReactNativeAztec/ReactAztecText.java
```diff
@@ -66,6 +66,11 @@
     boolean shouldHandleOnSelectionChange = false;
     boolean shouldHandleActiveFormatsChange = false;
 
+    // This optional variable holds the outer HTML tag that will be added to the text when the user start typing in it
+    // This is required to keep placeholder text working, and start typing with styled text.
+    // Ref: https://github.com/wordpress-mobile/gutenberg-mobile/issues/707
+    String tagName = "";
```
marecar3: Any reason to not set it as private with prefix `m`?

react-native-aztec/android/src/main/java/org/wordpress/mobile/ReactNativeAztec/ReactAztecManager.java
daniloercoli: Unfortunately there are cases where Aztec does only contains "non-breaking-space" char and `isEmpty` does return true, but the length is not `0`. This is happening when you write and then remove all the content from the block.

Will address the other comments instead.

react-native-aztec/android/src/main/java/org/wordpress/mobile/ReactNativeAztec/ReactAztecManager.java
daniloercoli: Done!

react-native-aztec/android/src/main/java/org/wordpress/mobile/ReactNativeAztec/ReactAztecManager.java
daniloercoli: Not that i'm aware of, but in this case it's pretty simple the input string, that doesn't worth  introduce a new function for it. We can use StringBuilder, but it's a very simple string.

daniloercoli:
> I am wondering, what amount of work can it take to assign same formatting style (<h1>)for the Placeholder like iOS has or at least to increase the initial size of RichText when type is heading, WDYT ?

The block now uses a `minHeight` value read from `variable.scss`, defined in the mobile project, this way the rending issue is mitigated a bit.
Didn't find a quick way of stylizing the hint text on Android.

Issue #729, opened 2019-03-08
Strikethrough format button not highlighted.


etoledom:
Highlighting attribute buttons on caret positions works for **B**, _I_ and [link](ur#) but not for ~ABC~.
It does work when typing with ~strikethough~ format.

Tested on iOS. Not sure if is multiplatform issue.

![bug](https://user-images.githubusercontent.com/9772967/54055790-63914f00-41ee-11e9-81fd-c2a865eb84d1.gif)

hypest:
What is the exact html underneath that rich text content @etoledom ? Alternatively, can you add steps to reproduce?

I think that the format-lib doesn't support both `del` an `s` tags so, there might be a syncing issue with the toolbar when the html underneath is using the not supported tag.

etoledom:
This is the HTML generated:
```html
<!-- wp:paragraph -->
<p>Hello <strike>world</strike> <strong>bold</strong></p>
<!-- /wp:paragraph -->
```

Steps:
- Start on an empty paragraph block.
- Press the ~ABC~ attribute button (Strikethrough).
- Write something.
- Put the cursor over the written text (with the ~strikethrough ~ format)
- Check that the  ~ABC~ button is not active.

It happens for any ~strike~d text, is not necessary any combination of other attributes.
Tested on the last internal release (WPiOS 12.0, Hockey app)

Tested on WPAndroid beta (12.0-rc-1) and it does work well.

It would be good if someone can reproduce it on WPiOS.

JavonDavis:
Experiencing this on Pixel 3a - Android 9. I left a comment on https://github.com/wordpress-mobile/gutenberg-mobile/issues/757#issuecomment-505083797 with gifs showing the weird behavior I was seeing.

etoledom:
So, It happens that Gutenberg uses and recognizes ~strikethrough~ with `<s>` tags and not with `<strike>` tags.

I'm not sure that not recognizing `<strike>` tag is a bug. It is also not recognized in web, so I'm inclined to close this issue.

https://github.com/wordpress-mobile/gutenberg-mobile/issues/757#issuecomment-505083797 seems to be a different one.

etoledom:
Closing this since it works as expected with <s> tags

maxme:
>     Start on an empty paragraph block.
>     Press the ABC attribute button (Strikethrough).
>     Write something.
>     Put the cursor over the written text (with the strikethrough format)
>     Check that the ABC button is not active.
[...]

PR #741, opened 2019-03-13
Upgrade to React Native v0.59.0

hypest:
This PR tracks the effort to upgrade to the v0.59.x version of React Native.

Known issues:
1. ~On the android side of things, the demo app fails to keep the title block to its original value. It seems that there is a "stray" `AppContainer.setTitleAction()` call that empties the title, but trying to follow the callstack in the debugger shows things that don't make sense. Looks like the JS files could be mixed up somehow. Thing is, the RN app works fine inside wpandroid so, we can probably tackle this issue separately~. Seems fixed after updating with a recent `develop`.
2. When using a JS Debugger in Android, the stacktrace seems like it's pointing to the wrong source files. Example:
```
  * Place a `debugger;` statement in the beginning of the `SplitBlock()` in `paragraph/edit.native.js`
  * Run the android demo app and enable JS debugging
  * Place the caret inside a paragraph block and hit "Enter" to split it
  * Notice the debugger pausing on a "SplitBlock" method alright (check the stacktrace panel) but the file pointed at is `image/index.js` which makes no sense.
```


etoledom:
First tests looks good.
Wanted to highlight that this seems to fix: https://github.com/wordpress-mobile/gutenberg-mobile/issues/729 🎉

I'll also test integration with WPiOS. @SergioEstevao is there a branch for this?

etoledom:
WPiOS looks good too. I'll publish a branch if there's any.

marecar3:
Hey @hypest, nice work here :)

1. I found some glitch in the demo app. I have tested the application on my Nexus 5x, Android 8.0
and when I try to tap on the Header block it automatically switches focus to the upper next Paragraph block

ezgif com-video-to-gif (2)

2. On WPAndroid there is a similar glitch when keyobard is dismissed I want to focus the last paragraph but focus is lost.
ezgif com-video-to-gif (3)

hypest:
Thanks for taking this for a test drive @marecar3 !

Nice find about the focus. What I actually see is that the focus seems to move _as you scroll_ even! This can "explain" why the focus moves when the keyboard comes up and such. Quite frustrating. Will look into it.

etoledom:
I couldn't reproduce any of those issues on iOS. (both example and WPiOS apps).

I do think that is necessary a new JS Bundle, since currently WPiOS is broken without metro running.

etoledom:
Opent WPiOS PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/11271

hypest:
There's this issue on the ReactNative repo https://github.com/facebook/react-native/issues/23916 that is probably related to the scrolling/focus issue we're seeing here.

hypest:
I pushed c2d1889 that has a workaround for the losing-focus-on-scroll issue, by disabling the Flatlist's optimization of destroying the views when they get out of the viewport. That's a real bummer of course but, we apparently will need to implement a proper support for Views recycling that is more compatible with focus preservation.

For reference, this is the PR that sets that prop to default true for Android in React Native: https://github.com/Rachelmorrell/react-native/pull/170

hypest:
@marecar3 , I think the No1, No2, No3 findings you mentioned got fixed with c2d1889. Can you recheck?

Regarding finding No4 (`Missing padding on placeholder`), I'm not able to replicate. Tried Pixel 2XL (Android 9.0) and Nexus 5X Emulator (Android 8.0.0). Can you try again or maybe share more info about the device/OS you tried it on? Thanks!

marecar3:
@hypest nice that we fixed those, I will try the branch now and will come back with the new feedback.

marecar3:
Hey @hypest I have pointed this branch to the latest GB develop - 3f234faf8b811bdfd8ada5899836b3521d886163 but getting the below error : 

Screen Shot 2019-03-14 at 1 30 40 PM

Not sure if it's something wrong only on my laptop

hypest:
I've updated from current `develop` and pushed.

Fortunately, that fixed one known issue.

The Android demo app and wpandroid look good during my testing. @SergioEstevao can you do another round of smoke testing on the iOS side of things to make sure this will not disrupt anything if we merge? Thanks!

Issue #428, opened 2018-12-21
HTML Editor - Keyboard can not be closed


pinarol:
Keyboard can not be closed once it is opened in HTML mode.

<img width="312" alt="screen shot 2018-12-21 at 11 11 14" src="https://user-images.githubusercontent.com/5032900/50332054-fedd2700-0511-11e9-869d-e1ac52f1943c.png">

PR #743, opened 2019-03-13
Fix - HTML Editor - Keyboard can not be closed

pinarol:
Fix https://github.com/wordpress-mobile/gutenberg-mobile/issues/428

On iOS there was no way of closing the keyboard on HTML mode, now we are adding keyboardDismissMode option, so user can drag down to close the keyboard. 

Android behavior doesn't change. On Android there's already a keyboard down button. The 'interactive' option is supported for iOS only, so Android will continue having the default value: https://facebook.github.io/react-native/docs/scrollview#keyboarddismissmode

keyboard-drag

**To Test**

- Run metro on this branch
- Go to WPiOS app
- Open a post and switch to HTML mode
- Tap somewhere to open keyboard
- Verify that drag down closes the keyboard

### DIFF
--- a/src/components/html-text-input.js
+++ b/src/components/html-text-input.js
@@ -85,7 +85,7 @@ export class HTMLInputView extends React.Component<PropsType, StateType> {
 			<KeyboardAvoidingView style={ styles.container } parentHeight={ this.props.parentHeight }>
 				<ScrollView
 					style={ { flex: 1 } }
-					keyboardDismissMode='interactive' >
+					keyboardDismissMode="interactive" >
 					<TextInput
 						autoCorrect={ false }
 						textAlignVertical="center"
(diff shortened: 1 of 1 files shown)


daniloercoli:
Not sure if's a local problem or not, but when I switch to the HTML mode on the emulator I get part of the screen blank. See the picture below:
Screen Shot 2019-03-13 at 15 13 08

etoledom:
@daniloercoli I have seen that too on the example app, but I haven't been able to reproduce it on the WPiOS app.

I couldn't test Android since there is no option to switch to HTML in the Android example app 🤔

daniloercoli:
> I couldn't test Android since there is no option to switch to HTML in the Android example app 🤔

Will try in it the wp-android app.

etoledom:
@daniloercoli - I also noticed that the HTML textview does not display all the HTML text.
It's not related to this PR but maybe to the addition to the wrapper `ScrollView`?

daniloercoli:
Tested by integrating this PR in `wp-androind` and the HTML view works as expected.

pinarol:
Thanks @daniloercoli 🎉 
I have opened an issue for that https://github.com/wordpress-mobile/gutenberg-mobile/issues/744 
It looks like it is broken on develop branch also.

Issue #710, opened 2019-03-07
Introduce a Logger to send logs to parent apps


hypest:
Let's come up with a logging mechanism to use inside the RN app.

So far, we're using `console.log()` statements to print out messages to the JS debugger (if attached) or the parent app's ReactNative logging mechanism. While useful while developing locally, we'd ideally want to mimic what the main WP apps do and be able to log messages that are persistent across app runs and are even pushed online to services that we can inspect remotely.

The main apps already have a logging mechanism in place (Crashlytics/Fabric) so, it makes sense to try and wire up to that.

hotchemi:
Update: base PR(https://github.com/wordpress-mobile/gutenberg-mobile/pull/746) is merged and remaining task is to fix wp-iOS to address API interface change.
ref: https://github.com/wordpress-mobile/gutenberg-mobile/pull/746#issuecomment-478108948

hotchemi:
I suppose this issue is ready to close 🙇

hypest:
Completed via https://github.com/wordpress-mobile/gutenberg-mobile/pull/746 and https://github.com/wordpress-mobile/WordPress-iOS/pull/11380.

PR #746, opened 2019-03-13
Issues/710 Leverage existing console polyfill to send logs to native side

hotchemi:
## Issue

- https://github.com/wordpress-mobile/gutenberg-mobile/issues/710

## Overview

This PR introduces an interesting hook implementation for `console.debug/info/log/warn/error` which allows us to send logs to native side.
react-native provides a console polyfill if you expose a global function called nativeLoggingHook that takes two parameters, a message and a log level(debug: 0, info/log: 1, warn: 2, error: 3). This change utilizes the mechanism and make it possible to inject native code execution on iOS/Android native side.

ref: https://github.com/wordpress-mobile/gutenberg-mobile/pull/746#issuecomment-472732200

~~This PR introduces new Logger API which is alternative to [Console API](https://developer.mozilla.org/en-US/docs/Web/API/console), which not only runs `console.debug/info/log/warn/error` internally but also executes iOS/Android native Log API with React Native bridge API.~~

### Usage

just using `console.debug/info/log/warn/error` as usual😄

~~Basically I followed [Console API](https://developer.mozilla.org/en-US/docs/Web/API/Console/) design to make the API familiar for JS developer. `logger` object has 5 methods(debug, info, log, warn, error) methods which are corresponding to existing console logger methods.~~

#### iOS

- embed `console.log` in somewhere
- run example app on iOS
- confirm [nativeLogHook](https://github.com/wordpress-mobile/gutenberg-mobile/pull/746/files#diff-128eceb67546269e0aa3d3179662e60bR117) is called and there's a debug message on XCode console.

#### Android

- embed `console.log` in somewhere
- run example app on Android
- confirm `nativeLogHook` is correctly called and there's a log on Android debug console

### Note

- [ ] One different point between iOS/Android is that on iOS we use delegate so that we should implement actual Logger(with `CocoaLumberjack`) implementation on Wordpress-iOS side when bumping up gutenberg-mobile module version. But on Android we already have Logger(utils module) dependency inside so actual Logger implementation is implemented in `WPAndroidGlueCode`.


react-native-gutenberg-bridge/android/src/main/java/org/wordpress/mobile/WPAndroidGlue/WPAndroidGlueCode.java
```diff
@@ -149,6 +150,31 @@ public void editorDidMount(boolean hasUnsupportedBlocks) {
                     setFocusOnTitle();
                 }
             }
+
+            @Override
+            public void debug(String message) {
+                AppLog.d(AppLog.T.EDITOR, message);
```
hotchemi: Discussion points: Do we need to add new Enum case for `AppLog.T.EDITOR`?

react-native-gutenberg-bridge/logger.js
```diff
@@ -0,0 +1,30 @@
+/** @format */
+
+import { NativeModules } from 'react-native';
+
+const { RNReactNativeGutenbergBridge } = NativeModules;
+
+const logger = {
+	debug: ( message ) => {
+		console.debug( message );
```
hotchemi: Discussion points: Just in case I designed as executing console API internally but what do you think?

react-native-gutenberg-bridge/android/src/main/java/org/wordpress/mobile/WPAndroidGlue/WPAndroidGlueCode.java
mzorz: Good question, and thanks for opening the things that popped up in your mind and frame them as discussion points here. As to the specific question, I think it's ok to have a new enum to make a distinction between what happens on JS and what not, but maybe better try as is and see if we need one first? I don't have a strong opinion in this regard I think

hotchemi: Thanks! looks like `EDITOR` is used in other modules so may be better to have new one😄

ios/gutenberg/GutenbergViewController.swift
```diff
@@ -113,6 +113,26 @@ extension GutenbergViewController: GutenbergBridgeDelegate {
         alertController.popoverPresentationController?.permittedArrowDirections = .any
         present(alertController, animated: true, completion: nil)
     }
+
+    func debug(message: String) {
```
hotchemi: Memo: we'll inject actual logger in https://github.com/wordpress-mobile/WordPress-iOS/blob/14ba744f99517971c3a2afe1297ba6da9866616d/WordPress/Classes/ViewRelated/Gutenberg/GutenbergViewController.swift#L315 on WP-iOS side

koke:
> This PR introduces new Logger API which is alternative to Console API

Sorry, I should have spotted this before you did all the work, but I didn't realize until I saw the code. As far as I know, React Native [already sends](https://facebook.github.io/react-native/docs/debugging.html#accessing-console-logs) `console.log` to iOS/Android, it's just that we use some custom native logging that won't see those messages.

I did some research and came across [Libraries/polyfills/console.js](https://github.com/facebook/react-native/blob/master/Libraries/polyfills/console.js), which implements `console` for React Native, and there I discovered a nice (undocumented) function that seems to do exactly what we want: `global.nativeLoggingHook(str, logLevel)`.

I think we can define our own `global.nativeLoggingHook` that forwards messages to the bridge and avoid needing a new API.

hotchemi:
@koke Thank you for your feedback. I just opened PR to make discussion happened so it's not a problem  at all~!

> global.nativeLoggingHook

Ooh thank you I didn't know the function. I've got a feedback internally like overriding `console.log` but actually I deliberately prepared yet another API interface since I thought overriding global object is kind of too much to be done on application layer and kind of hacky way for most of developers..but yea I'd love to try to polish API as much as I can to make it nicer and make other members more productive 😄 Let me investigate and try to see how it'd be feasible.

Again thank you for your dedicated feedback~!

hotchemi:
@hypest @daniloercoli Can I ask a few points? 🙇

- as for https://github.com/wordpress-mobile/gutenberg-mobile/pull/746#discussion_r265253332

Seems I should open a PR for https://github.com/wordpress-mobile/WordPress-Utils-Android to add one case for `AppLog.T` but what the release process looks like? Can I suppose we'll be able to release new ver soon? Or should I use something from existing `AppLog.T`?

- as for https://github.com/wordpress-mobile/gutenberg-mobile/pull/746#issuecomment-472758203

I'd say there're two ways to implement, adding logger API like this PR or overriding `console.log` and on those lines API globally..what do you think about the point(Especially I'd love to hear @hypest opinion since he's the creator of the original issue).

hypest:
> Or should I use something from existing AppLog.T?

👋 @hotchemi , I think re-using `Applog.T.EDITOR` is fine for our purposes.

> >  as for #746 (comment)
>
> I'd say there're two ways to implement, adding logger API like this PR or overriding console.log and on those lines API globally..what do you think about the point

Seems to me that Koke's `nativeLoggingHook` suggestion doesn't actually override `console.log` per se, but goes deeper and utilizes a mechanism ReactNative has in place, albeit undocumented. Seems to me that it would be OK to use it in our case as well since we're not actually polyfilling the `console.log` call itself. What do you think about `nativeLoggingHook`? Do you think that it might be problematic down the road perhaps?

hotchemi:
@hypest Thank you for your quick response!

>  I think re-using Applog.T.EDITOR is fine for our purposes.

👍 

> What do you think about nativeLoggingHook

👍Let me check deeper!

hotchemi:
@hypest @koke Code became really simple! Thank you for your awesome proposal😄
79ecf11...6b6e911

Tug:
@hotchemi Just did an experiment to override `console` which seemed successful. I put the patch in a gist here https://gist.github.com/Tug/d42dc65dfbf163b62f05458048be746a you might be interested to look at it.
By the way, `console.log` (and others) accept multiple arguments and not always serializable objects, so we need to handle those case. My patch has a quick and dirty solution for this but we might want to improve on it.

hotchemi:
@Tug Thank you so much!
I'd say after moving to use `nativeLoggingHook`(79ecf11...6b6e911) we don't have to take care about those points on our side I guess. It's converted to one string and being passed to bridge method as a 1st argument(`message`).

react-native-gutenberg-bridge/index.js
```diff
@@ -7,6 +7,12 @@ const isIOS = Platform.OS === 'ios';
 
 const gutenbergBridgeEvents = new NativeEventEmitter( RNReactNativeGutenbergBridge );
 
+// Leverages existing console polyfill from react-native which is a global function called
+// `nativeLoggingHook` that takes two parameters, a message and a log level.
+global.nativeLoggingHook = function( message, logLevel ) {
```
koke: I have a couple suggestions here:

* I think it'd be better to keep all `global` replacements in `src/globals.js` to make it easier to find in the future. The bridge could just expose a method that returns the logging hook.
* The apps should not rely on magic numbers. We should either convert logLevel to a string, or expose some named constants for all platforms in the bridge to make it easier to know what each value means.

Tug: @koke it's not really magic, but part of `react-native`: https://github.com/facebook/react-native/blob/master/Libraries/polyfills/console.js#L376

hotchemi: Makes sense! I'll leave soon so let me address after a while 🙇

koke: > @koke it's not really magic, but part of `react-native`

Oh, I know. I meant that from the apps perspective, there's no named constant exposed, so if you look at [how it's being handled](https://github.com/wordpress-mobile/gutenberg-mobile/blob/6b6e91125e3b375b20098c64ea049ee66746aa69/ios/gutenberg/GutenbergViewController.swift#L117), it's just mapping numbers to specific log levels in the apps, and it's not obvious there what possible values there are.

By [magic number](https://en.wikipedia.org/wiki/Magic_number_%28programming%29) I just meant an "unnamed numerical constant" 😄

hotchemi: Yea we can have enum definition on Java and Swift side for sure.

hotchemi: Addressed 🙇 99e4314...dcf1213

hotchemi:
@koke @Tug @mzorz could you possibly check when you have time? 🙇

Tug:
@hotchemi I did some testing on Android but wasn't able to see the [default logs we have on boot](https://github.com/wordpress-mobile/gutenberg-mobile/blob/develop/src/index.js#L42) in the native app. Would you mind editing the PR description to add some testing instructions?

hotchemi:
@Tug Hi sorry I've added how to test in PR description🙇And your saying is true since when we run example app on Android [nativeLoggingHook](https://github.com/wordpress-mobile/gutenberg-mobile/pull/746/files#diff-790122ff7222773fbfe709696f8eedc5R60) in `MainApplication` which does nothing is actually called. To test with example app we should edit the callback method as [here](https://github.com/wordpress-mobile/gutenberg-mobile/pull/746/files#diff-f2e1dfe800569dfb4ee4830f88779890R155)🙇

daniloercoli:
> which does nothing is actually called. To test with example app we should edit the callback method as here🙇

What do you think about adding the missing logging code to `MainApplication.java`, so we can have logging in place in the demo app too?  (I guess a simple call to `Log.*` is OK).

hotchemi:
@daniloercoli thank you for your suggestion! I've added d00c1ac22c8429d3ca0313ab2399120d554dc4ec 🙇

react-native-gutenberg-bridge/android/src/main/java/org/wordpress/mobile/ReactNativeGutenbergBridge/GutenbergBridgeJS2Parent.java
```diff
@@ -17,10 +17,32 @@
         void onMediaFileUploadFailed(int mediaId);
     }
 
+    enum LogLevel {
```
hypest: Let's add a line comment pointing back to RN's definition of the constants (https://github.com/facebook/react-native/blob/master/Libraries/polyfills/console.js#L376) so we know we're trying to mirror those, and hopefully catch a discrepancy in the future if the RN side constants change. What do you think?

Tug:
Tested on Android and it works well 👍 
Will test on ios as well.

In the meantime I have tested logging global errors. I added the following in `src/globals.js`:
```js
const previousErrorHandler = global.ErrorUtils.getGlobalHandler();
global.ErrorUtils.setGlobalHandler( ( error, isFatal ) => {
	nativeLoggingHook( error.toString(), isFatal ? 3 : 2 );
	previousErrorHandler( error, isFatal );
} );
```

As well as threw an error in `BlockHolder` when a block is focused to test the whole thing:
```js
onFocus = ( event: NativeSyntheticEvent<NativeTouchEvent> ) => {
	throw new Error( 'This is a test error for logging' );
        ...
}
```

Feel free to consider this addition to your PR. I don't think it's necessary and we can always do it in a new PR.

react-native-gutenberg-bridge/android/src/main/java/org/wordpress/mobile/ReactNativeGutenbergBridge/RNReactNativeGutenbergBridgeModule.java
```diff
@@ -128,6 +128,11 @@ public void requestImageUploadCancel(final int mediaId) {
         mGutenbergBridgeJS2Parent.requestImageUploadCancel(mediaId);
     }
 
+    @ReactMethod
+    public void nativeLoggingHook(String message, int logLevel) {
```
hypest: Kinda a nit but, can we perhaps simplify the name of the method to just `log`? The "Hook" part in particular doesn't seem to be that relevant here anyway and I think it's too internal to RN to surface?

hypest:
👋 @hotchemi , tried the iOS demo app but I think I cannot see any log message from the app (in Xcode). I can normally see the same messages from the Android demo app though. Any particular steps I need to to for iOS perhaps? I've tried both `yarn ios` to launch the app and running it via Xcode directly.

react-native-gutenberg-bridge/android/src/main/java/org/wordpress/mobile/ReactNativeGutenbergBridge/GutenbergBridgeJS2Parent.java
hotchemi: 👍 a37fcf4

react-native-gutenberg-bridge/android/src/main/java/org/wordpress/mobile/ReactNativeGutenbergBridge/RNReactNativeGutenbergBridgeModule.java
hotchemi: After a consideration can I leave as it is? Because of the following reasons:

- `log` is a bit generalized name
- by keeping same method name developer can easily figure out the relationship between native and RN by grep

hotchemi:
@Tug Thank you for your suggestion! I suppose original goal of #710 is to just send log to native side so it'd be better to separate PR🙏
@hypest Oh really? Unfortunately I've upgraded my XCode so let me try tomorrow😞

hotchemi:
@hypest I've tested and I could see a log on XCode debug console. Let me know if I can help something🙇Below is steps that I've done🙇

#### Steps

- Add `console.log( 'this is test message for log' );` in `src/index.js`

Screen Shot 2019-03-28 at 20 04 02

- Run application on iOS
- You can see `this is test message for log` on XCode console after loading JS bundle
  - - Also make sure that debugger can be stopped at `nativeLoggingHook` in `GutenbergViewController`

Screen Shot 2019-03-28 at 20 01 25

Screen Shot 2019-03-28 at 20 06 24

react-native-gutenberg-bridge/ios/GutenbergBridgeDelegate.swift
```diff
@@ -49,6 +57,10 @@ public protocol GutenbergBridgeDelegate: class {
     /// Tells the delegate that the editor view has completed the initial render.
     ///
     func gutenbergDidMount(hasUnsupportedBlocks: Bool)
+
+    /// Tells the delegate that logger method is called.
+    ///
+    func nativeLoggingHook(message: String, logLevel: LogLevel)
```
pinarol: Could we update the name to be consistent with delegate method naming on iOS? some alternatives might be:
gutenbergDidLog(message: String, logLevel: LogLevel) 
gutenbergDidGenerateLog(message: String, logLevel: LogLevel)

react-native-gutenberg-bridge/android/src/main/java/org/wordpress/mobile/ReactNativeGutenbergBridge/RNReactNativeGutenbergBridgeModule.java
pinarol: (optional)It caught my attention that nativeLoggingHook is a noun but all other bridge methods are imperatives, might be something to improve also.

react-native-gutenberg-bridge/ios/RNReactNativeGutenbergBridge.swift
```diff
@@ -69,6 +69,12 @@ public class RNReactNativeGutenbergBridge: RCTEventEmitter {
         }
     }
 
+    @objc
+    func nativeLoggingHook(_ message: String, logLevel: Int) {
+        guard let logLevel = LogLevel.init(rawValue: logLevel) else { return }
```
pinarol: (optional)syntactic sugar: we can drop .init and use LogLevel(...)

react-native-gutenberg-bridge/ios/GutenbergBridgeDelegate.swift
hotchemi: Thx! 🙇 51183cf

react-native-gutenberg-bridge/ios/RNReactNativeGutenbergBridge.swift
hotchemi: Thx! 🙇 51183cf

react-native-gutenberg-bridge/android/src/main/java/org/wordpress/mobile/ReactNativeGutenbergBridge/RNReactNativeGutenbergBridgeModule.java
hotchemi: Finally I was convinced😄 a37fcf4...8f1e207

hotchemi:
@pinarol Thank you for your quick review! I've addressed and could you take a look again when you have time?🙇Thank you!

hypest:
Thanks for the review @pinarol .

Will now merge this PR, thanks @hotchemi !

hotchemi:
Thank you so much for your kind reviews!🙇
I'd say this PR contains breaking API interface change at least on [iOS](https://github.com/wordpress-mobile/gutenberg-mobile/pull/746/files#diff-a2b32a5592cc93f8d3d83829eb226720R63) side so let me know when next version being released. I'll open PR to address then🙇

hypest:
> I'd say this PR contains breaking API interface change at least on iOS side so let me know when next version being released. I'll open PR to address then🙇

@hotchemi , can you describe here what the breakage is? Does it mean that trying to now use `develop` in WPiOS will be broken? if yes, it'd be best if we open a PR to fix it as it will be a blocker for other PRs that want to test with WPiOS.

hotchemi:
@hypest Since I've added one method(editorDidEmitLog) in `GutenbergBridgeDelegate` just upgrading GM will cause a compile error I guess🙇

https://github.com/wordpress-mobile/WordPress-iOS/blob/b6e18c2a95aa1ec02882baff8ed8333d5d1dbd27/WordPress/Classes/ViewRelated/Gutenberg/GutenbergViewController.swift#L315

What to add is just one method in `GutenbergViewController`(and I suppose we can merge the change before upgrading GM) so let me open PR by Mon!

hotchemi:
@hypest It turned out we need to apply attached path file change at the same time when upgrading GM to upcoming version because `LogLevel` enum is not defined in current develop branch🙇Not sure who is responsible for updating GM version next time but apply the diff below or let me know then🙇

[diff.patch.txt](https://github.com/wordpress-mobile/gutenberg-mobile/files/3027194/diff.patch.txt)

hypest:
Thanks for the context @hotchemi .

@pinarol , since you had the chance to make a pass over this PR, can you wrangle adding the necessary support to WPiOS perhaps?

pinarol:
@hypest sure, I can open a PR on WPiOS side.

pinarol:
I've opened a PR for WPiOS side: https://github.com/wordpress-mobile/WordPress-iOS/pull/11380 it is fixing the build error but I couldn't see any logs are coming into editorDidEmitLog on WPiOS side. 

I am running metro on the latest gutenberg-mobile/develop. Example app didn't have any problem. I added a console.log("pinar") in the JS code and verified it is executed by debugging, but I can't see it coming to WPiOS. Any ideas? @hotchemi

hotchemi:
@pinarol thank you! Let me check on the working branch later 🙇

pinarol:
Update: @hotchemi I think you can ignore that. @etoledom could see the logs, it might be sth about my local.

hotchemi:
@pinarol Ooh thank you so much but can I check one point just in case? 🙇Don't we need to run `yarn bundle` and update App.js and App.js.map when we update js file? 🙇

etoledom:
Hey @hotchemi - We usually create new JS Bundles when there are breaking changes and for releases. To see the changes in WPiOS we simply build the WPiOS app and display the Gutenberg editor while the metro server is running in the branch you want to test 👍

pinarol:
@hotchemi we could have, but practically creating bundles is not critical at this point, we almost always keep metro open during development so we won't get blocked. (And bundles will be created on release day(this Friday) anyway.)

PR #750, opened 2019-03-15
Fix makepot script using wp-i18n makepot

Tug:
Address https://github.com/wordpress-mobile/gutenberg-mobile/issues/751

https://github.com/wordpress-mobile/gutenberg-mobile/pull/658 was a first attempt at fixing the mobile pot file generation.

Since then `babel-plugin-makepot` was removed from gutenberg and the official way is to use `wp i18n make-pot` instead. This PR updates our makepot script to use this.

### Testing Instructions
- Have `xgettext` installed
- Run `yarn makepot`
- Check that `./gutenberg.pot` and `gutenberg-mobile.pot` file were generated in the root folder
- `gutenberg-mobile.pot` should only contain the 4 following strings:
    - `ADD BLOCK HERE`
    - `Move block up`
    - `Move block down`
    - `Remove content`
All other strings in `./src` are already part of `gutenberg.pot`.

In a followup PR, we'll want to remove those 4 strings from this repository so all strings from this project are translated by the glopress gutenberg project


koke:
> * `Move block up`
> * `Move block down`

Note for future: I see Gutenberg has [similar labels](https://github.com/WordPress/gutenberg/blob/7964f381e07e06b2ac13698fd1465a4d0c16da1b/packages/block-editor/src/components/block-mover/index.js#L64), but they say "Move up" and "Move down". Perhaps we could reuse those.

Tug:
Yeah good idea, let's start the conversation here in the meantime 👍
@SylvesterWilmott What do you think of those copy changes in guteneberg-mobile?
- `Move block up` => `Move up`
- `Move block down` => `Move down`
- `Remove content` => `Remove Block`

Concerning `ADD BLOCK HERE` I haven't found equivalent string in gutenberg. There's `Add block` but it doesn't convey the information that the block will be inserted in place of this message.

SylvesterWilmott:
Actually we are using similar language in a future version of stats which allows a repositioning of blocks:

`Move up`
`Move down`
`Remove from insights`

In this case I would even consider simplifying:

`Remove Block` => `Remove`

Tug:
Ok I updated the strings, now we only have `ADD BLOCK HERE` in `gutenberg-mobile`.

I feel we should still keep this `yarn makepot` script in here to be able to check the strings we have that are only in gb-mobile.

Tug:
@koke not a priority but if you have some time to review this :)

package.json
```diff
@@ -56,7 +56,12 @@
...
     "i18n-cache": "node i18n-cache/index.js",
     "postinstall": "yarn i18n-cache",
-    "makepot": "xgettext -f <( find . \\( -path './gutenberg/*/src/*.native.js' -or -path './src/*.js' \\) ! -path 'node_modules' -print ) --from-code=UTF-8 -k__ -k_n -k_x -k_nx -x ./gutenberg/languages/gutenberg.pot -o ./gutenberg-mobile.pot",
+    "install:wpcli": "(test -x bin/wp-cli.phar || curl -Ls https://raw.githubusercontent.com/wp-cli/builds/gh-pages/phar/wp-cli.phar -o bin/wp-cli.phar && chmod +x bin/wp-cli.phar) && bin/wp-cli.phar --info",
+    "prewp": "yarn install:wpcli",
+    "wp": "php -d memory_limit=512M bin/wp-cli.phar",
+    "premakepot": "yarn makepot:gutenberg",
+    "makepot": "bash -c \"xgettext -f <( find . -path './src/*.js' ! -path 'node_modules' -print ) --from-code=UTF-8 -k__ -k_n -k_x -k_nx -x ./gutenberg.pot -o ./gutenberg-mobile.pot\"",
+    "makepot:gutenberg": "yarn clean:gutenberg && yarn wp i18n make-pot ./gutenberg --ignore-domain gutenberg.pot",
```
koke: Why is the `clean:gutenberg` needed?

koke:
> Have xgettext installed

Should we add some check to make sure xgettext is present and give a better error message if not? Is it worth mentioning in the README?

package.json
Tug: Because we want to avoid having source paths in the pot file for `build` and `build-module` files. I have tried to exclude those using the `make-pot` command but it's not very flexible so I'm not sure it's possible to do it cleanly.

Tug:
> Should we add some check to make sure xgettext is present and give a better error message if not? Is it worth mentioning in the README?

Well in theory we could use `wp i18n make-pot` but it doesn't seem to support flow syntax at the moment. I think using `xgettext` is a hack at the moment and since the process of updating the strings in WPAndroid and WPiOS is manual I don't think it's worth spending more time on it. The goal is to actually have an empty POT file generated here and remove the need for that feature.

Let's add a better error message though 👍

PR #752, opened 2019-03-15
Adding enzyme to empower unit-tests (with some examples)

etoledom:
This PR adds enzyme utility to facilitate testing of our components.

This is in part the code we worked on with @pinarol on the unit-tests spike. Our goal was to find a way to test components without depending on the environment (say the store, other components or modules, or native code).

We have found that `enzyme` helps quite a lot to build small unit tests without depending on the whole react tree/dependencies.

With it you can:
- `shallow()` to get a shallow render of the component. This will render elements just one level deep, avoiding complex trees (and eventually getting into NativeModules).
- `find()` a child of the rendered component.
- `dive()` into one level deeper, in case you need it (useful to access the component instances on HOCs and access the child component)
- `setProps()` on the fly.
- `instance()` to get the actual component instance and call methods on it.
- And [many other useful things](https://airbnb.io/enzyme/docs/api/shallow.html)

The idea of these unit tests is to run the smallest portion of JS possible, ideally just the method we want to test. For this, we set up its dependencies and test on its results.

Many time the "result" is more than just a return function, but those methods have side effects as modifying the state of the component itself or the state of the global store.

We will want to mock and intercept the calls that make those modifications, instead of testing on the result of the modified store.

We have aded some examples on a few components doing just that.
All comments from anybody are welcome 🎉

To test:
- `yarn test` 😁


koke:
These are looking nice 😍 I'm wondering if there are any blockers to have this in the gutenberg repo instead.

etoledom:
The "where to put the files" is something we are not sure yet 🤔 
Ideally would be to have the tests together with the component they are testing (I think).

The `jest.config.js` has a ignore rule to avoid running tests from `gutenberg`.
```
	testPathIgnorePatterns: [
		...,
		'/gutenberg/test/',
		'/gutenberg/packages/',
	],
```

Also, adding our tests in the `gutenberg` repo will make them be run by the `gutenberg` tests.
That sounds like something we might want to do, but will require more coordination.

Optionally we can try to configure this in a way that they don't run our tests (as a starting point) and we don't run theirs, still having our tests in the `gutenberg` repo.

But again, where to put the test files themselves is not something we have been focusing so far.

etoledom:
@koke we found a way to put our tests on `gutenberg` and run just ours tests by setting a `testMatch` configuration on `jest.config.js`.

But we prefer to do that change on a separated PR so we can have a discussion specific about it.

koke:
> we prefer to do that change on a separated PR so we can have a discussion specific about it.

That sounds good to me 👌

pinarol:
Additional note: Default export declaration of components usually exports components wrapped with composition of methods like withDispatch, withSelect etc., thus, makes it almost impossible to reach the real component from unittests(At least we couldn't find the way for it). But the good news is we we can just import the component as follows: 

✅ import { HTMLInputView } from '../components/html-text-input.js'; //This gives us the component itself directly

rather than the default export:

🚫import HTMLInputView from '../components/html-text-input.js';

hypest:
> Default export declaration of components usually exports components wrapped with composition of functions like withDispatch, withSelect etc., thus, makes it almost impossible to reach the real component from unittests(At least we couldn't find the way for it). But the good news is we we can just import the component as follows:
> 
> ✅ import { HTMLInputView } from '../components/html-text-input.js'; //Named export gives us the component itself directly, also note that named exports should be imported with the same name that it is exported rather than the default export:
>
> 🚫import HTMLInputView from '../components/html-text-input.js';

Let me take this opportunity and actually ask our web side friends: is there perhaps a reason the components modules are designed like that? Is unit testing without the HOCs discouraged perhaps? Any guidelines how and what to unit test? Maybe @gziolo has some insights there?

etoledom:
I personally think that it depends. There are occasions where the connections to the store are necessary to correctly run the code we want to test, like on [this example](https://github.com/wordpress-mobile/gutenberg-mobile/blob/2ee1620efa7c98caaa82019449febe129640a05e/src/app/AppContainer.test.js#L24).

But in most cases, we can test the component's JS code without depending  store's code, and just mock the small bits of it to give the needed data or to check if dispatch methods were called. That should make the tests simpler, run faster, and still being able to test what we want (atomic pieces of code / methods ).

etoledom:
Closed by mistake 😛

gziolo:
From our docs:
https://github.com/WordPress/gutenberg/blob/master/docs/contributors/testing-overview.md#pain-points

> Connected components are tricky to work with. To snapshot a connected component you'll probably want to export the unconnected component:

This is quite a common approach. I don't think we test often components wrapped with other components which hold any state. It's too much hassle to get them into a proper state. It's much easier to pass those props directly. In general, we prefer e2e tests for UI components over unit tests.

hypest:
> In general, we prefer e2e tests for UI components over unit tests.

Thanks for sharing that insight Greg!

For the native mobile project, we probably want balance unit testing and e2e testing, to avoid relying too much on the slow and highly platform dependent (integration) tests (which require the actual mobile platforms to be exercised).

In that sense, any tweaks we can make in order to facilitate unit testing (referring to the JS logic here) will be useful.

src/__tests__/html-text-input-test.js
```diff
@@ -0,0 +1,107 @@
...
+		textInput.simulate( 'changeText', 'text' );
+
+		//Check if the setTitleAction is called
+		expect( setTitleAction ).toHaveBeenCalledTimes( 1 );
+		expect( setTitleAction ).toHaveBeenCalledWith( 'text' );
+	} );
+} );
+
+// Utility to find the HTML TextInput from the wrapper
+const findHTMLTextInputFromWrapper = ( wrapper ) => {
```
Tug: I think we could move those helpers to the top (before the tests), it's a bit confusing having them here imo.
Also the search by attribute is misleading regarding the function name, I would expect a `findByType( TextInput )` maybe?

src/__tests__/paragraph.test.js
```diff
@@ -0,0 +1,91 @@
...
+		const before = null;
+		const after = null;
+
+		//When
+		instance.splitBlock( before, after, ...blocks );
+
+		//Then
+
+		// The block is deleted if before is null
+		expect( onReplace ).toHaveBeenCalledTimes( 1 );
```
Tug: Interesting, I didn't know jest would reset its mock functions on each tests!
I think it's more common to declare those inside our tests to avoid side effects.
```js
const onReplace = jest.fn();
const insertBlocksAfter = jest.fn();
const setAttributes = jest.fn();
const instance = getTestInstanceWithContent( '', { onReplace, insertBlocksAfter } );
...
expect( onReplace ).toHaveBeenCalledTimes( 1 );
```

Tug:
I love it ❤️!! Very good job @etoledom.
As @koke mentioned I think we should work towards having those in gutenberg, so circleci can run them on our PRs, we can choose to either run them along with web unit tests or separate, we have many options

src/__tests__/paragraph.test.js
etoledom: This is a setting:
https://github.com/wordpress-mobile/gutenberg-mobile/blob/d91cd0459eb0e31e314f504a517b90c0253ac3a5/jest.config.js#L17

I thought it would be safer to have them automatically cleared to avoid unexpected side effects.
Do you prefer to do it manually to make the intention clear?

Tug: Oh I see. I still think it's a little bit better to redefine those in each tests. Having self contained tests that only depends on the functions (including classes) we're testing is cleaner imo, it's closer to the functional programming paradigm.

Though it's very minor and I don't feel very strong about it, take it as my personal opinion :)

src/__tests__/html-text-input-test.js
pinarol: hey @Tug 👋 sure I can move them to the top if that's confusing.

But, I didn't quite follow the misleading part. `Search by attribute` is one way to find the HTML TextInput. There are different queries we can choose to achieve the same thing, but imho this wouldn't affect the function name?

src/__tests__/paragraph.test.js
etoledom: Cool! I removed the global function definitions and now they are recreated on each test case.
I usually prefer to minimize the setup code on the case itself by creating helpers, but good call on avoid globals 👍 

I did keep the setting just in case :)

etoledom:
Thank you @Tug for the review!
@pinarol and I have added a few commits to address your feedback 👍 

Also thank you @koke , @hypest and @gziolo for your feedback and comments.
Do you think this is on a good state to be merged soon?

src/__tests__/html-text-input-test.js
```diff
@@ -0,0 +1,108 @@
...
+import { HTMLInputView } from '../components/html-text-input.js';
+
+/**
+ * WordPress dependencies
+ */
+import { __ } from '@wordpress/i18n';
+
+// Utility to find the HTML TextInput from the wrapper
+const findHTMLTextInputFromWrapper = ( wrapper ) => {
+	const placeholder = __( 'Start writing…' );
```
mkevins: I wonder if it makes sense to have these placeholder text values as a constant somewhere. I see they are already "hardcoded" in [the implementation](https://github.com/wordpress-mobile/gutenberg-mobile/blob/bd1ebf52f1f9b40880b65ed584c89b9fefcbd6e0/src/components/html-text-input.js#L103). Any changes would probably be for a different PR, but, I'm wondering how likely it is this text will change in the future, and if that will require changes to the utilities so `find` will still return the correct value.

pinarol: Yes, might be! The question here is what is the most appropriate `somewhere` ? :) I sometimes want to do these kind of improvements but end up making things a bit more complicated on JS.
But I'd definitely be willing to update this according to suggestions and what most of the people think about it.

src/__tests__/html-text-input-test.js
Tug: Well `findHTMLTextInputFromWrapper` sounds like a generic helper that will find an TextInput element in any kind of wrapper element. So reading the code I thought that what it would do, but jumping at the definition I saw that it was actually matching on a specific attribute of our element. Let's say we want to reuse that function for another wrapper it won't work, or if we change the placeholder value it won't work either, or if for some reason we try to switch language in our tests...
In any case I think it would help if `findHTMLTextInputFromWrapper` was more generic, and maybe we can share those helpers across tests later.

pinarol: I think I got what you mean, you'd expect this function to act as a generic helper but it doesn't currently. Yes actually, I never thought it as a generic helper, this was only intended to be called by different tests from this file only. Could you describe a bit more about the generic helper you were imagining? What kind of query would it have instead?

etoledom:
Added a commit to properly test splitting paragraph block on Enter. Thanks @mkevins for spotting that 👍

src/__tests__/html-text-input-test.js
mkevins: > The question here is what is the most appropriate `somewhere`

I agree, this part is something that probably would benefit from input from others, even a quick look 👀 at [examples using Gutenberg block templates API](https://github.com/WordPress/gutenberg/blob/b944ed777b54c8fd93cc894040218550e0fd3c7f/docs/designers-developers/developers/block-api/block-templates.md) gives me the impression that these placeholders are intended to be set inline. I think any decisions / changes around this would not be relevant for this PR, as it seems to be a fairly pervasive pattern that could involve a lot of other areas in the code base.

I only wanted to mention it here to have guard up for something that could break in an unexpected way later on, from a seemingly cosmetic change. It may be that the complexity introduced by having some kind of "static" namespaced constants import (akin to `strings.xml` in Android) would outweigh the benefits derived from having centralized text assets and slightly DRYer code and tests. I'd love to hear others thoughts about it, if anyone has a strong opinion one way or the other.

src/__tests__/html-text-input-test.js
Tug: I added a commit to show what I meant: https://github.com/wordpress-mobile/gutenberg-mobile/pull/752/commits/2d45bc0b7f7d99485cda99dfa8aff8e0be923936
I think it's just the naming that was confusing for me, making it clear that the function returns the TextInput content part would have fixed it.

etoledom:
Thank you @Tug !
The commit looks great to me. Thanks for that 😁
✅ I'm happy to merge if @pinarol is ok with it too 👍

src/__tests__/html-text-input-test.js
pinarol: thank you @Tug that looks good 👍

pinarol:
@etoledom yes I am all good to go! ✅

etoledom:
Thank you all! 😁

Issue #744, opened 2019-03-13
[iOS]Text is invisible on HTML mode when content height is long enough


pinarol:
The html text is invisible on HTML mode when it is long enough. It is observable in the example app with iPhone X. The content is there, it can be selected/copied but it is not visible. I could repro the invisibility issue on iPhone X, but, other devices has also a trouble displaying the long text, for them text is visible but you can't see the whole text. So in general there's a problem in iOS devices when we try to display long text.

Example app:

![invisible-html](https://user-images.githubusercontent.com/5032900/54290728-8b9cfb80-45bc-11e9-9804-d14e3172e980.gif)

iPhone X - WPiOS (text becomes invisible):

![ios-longtext-html-iphonex](https://user-images.githubusercontent.com/5032900/54439218-71445880-4749-11e9-945a-471df1250531.gif)

iPhone 8 - WPiOS (can't display till the end of the text):

![ios-longtext-html](https://user-images.githubusercontent.com/5032900/54439518-fe87ad00-4749-11e9-94ac-194ddad27fab.gif)

**Root cause analysis:**

I found the almost exact same issue in react-native repo: https://github.com/facebook/react-native/issues/9077 The issue looks like it is related with iOS level.

This thread explains it very well: https://stackoverflow.com/questions/23582497/uitextview-text-becomes-invisible-after-height-reaches-8192-0

Looks like there's a kind of limitation that happens on iOS level and to overcome it we shouldn't put the TextView inside ScrollView and modify its height due to the content size. Instead, we should use a TextView smaller than the screen height and let the inner system handle scrolling on its own and make optimizations on it. 

So having said that, this is what we are doing on visual mode also. I tried it with a long enough text on paragraph block and the issue is fully reproducible there too:

(It is less likely to have such long text on visual side, and plus people will tend to split such a text in separate paragraphs so I'd not worry about it for now. But still, this is can be a warning about how we are getting away from iOS's recommended way of rendering long content) 

![ios-longtext-visualmode-iphonex-](https://user-images.githubusercontent.com/5032900/54440677-40195780-474c-11e9-98a4-8422c66b440b.gif)

**Solution Proposal:**

It looks like we need to sacrifice using the ScrollView on HTML mode, thus title and the html content won't be able to scroll altogether anymore, but we won't have anomalies. HTML TextInput will need to scroll inside of its own bounds. I gave this solution a quick try and it is working.

Any other ideas are truly welcomed!

[...]

PR #755, opened 2019-03-18
Fix text becoming invisible in HTML editor

pinarol:
To Fix: https://github.com/wordpress-mobile/gutenberg-mobile/issues/744

Root cause analysis and the solution proposal can be found in the issue. ^^^ 

To fix the anomalies we are removing the outer ScrollView and letting the TextView scroll in its own bounds, thus, letting iOS system make the required optimizations.

After the fix:

iOS:
ios-longtext-html-mode-after

Android:
android-longtext-html-mode-after

**Implementation Details**

Since we can not use keyboardDismissMode prop anymore I've added a pan gesture to detect when the user drags down and close the keyboard with it. This is active for iOS only, keyboardDismissMode='interactive' was also an iOS only behavior.

**To Test**

- Open the iOS example app with `yarn ios`
- Switch to html mode
- Verify that text is visible
- Scroll down to the bottom
- Verify that the whole content is visible
- Drag down to close the keyboard (iOS only)

Repeat tests for following platforms:

WPiOS
WPAndroid(Except for draggin down to close the keyboard)

Try using also a non iPhone X series device like iPhone 8.

cc @daniloercoli

### DIFF
--- a/src/components/html-text-input.js
+++ b/src/components/html-text-input.js
@@ -53,7 +53,7 @@ export class HTMLInputView extends React.Component<PropsType, StateType> {
 			onMoveShouldSetPanResponderCapture: ( ) => true,
 
 			onPanResponderMove: ( e, gestureState ) => {
-				if ( gestureState.dy > 100 && gestureState.dy < 110 ) {
+				if ( Platform.OS === 'ios' && ( gestureState.dy > 100 && gestureState.dy < 110 ) ) {
 					//Keyboard.dismiss() and this.textInput.blur() is not working here
 					//They require to know the currentlyFocusedID under the hood but
 					//during this gesture there's no currentlyFocusedID
--- a/src/components/html-text-input.js
+++ b/src/components/html-text-input.js
@@ -34,7 +34,6 @@ type StateType = {
 
 export class HTMLInputView extends React.Component<PropsType, StateType> {
 	isIOS: boolean = Platform.OS === 'ios';
-	textInput: TextInput;
 	edit: string => mixed;
 	stopEditing: () => mixed;
 	panResponder: PanResponder;
--- a/src/components/html-text-input.js
+++ b/src/components/html-text-input.js
@@ -9,7 +9,7 @@
 import { __ } from '@wordpress/i18n';
 
 import React from 'react';
-import { Platform, TextInput, View, UIManager, PanResponder } from 'react-native';
+import { Platform, TextInput, UIManager, PanResponder } from 'react-native';
 import styles from './html-text-input.scss';
 import KeyboardAvoidingView from './keyboard-avoiding-view';
 
@@ -30,7 +30,6 @@ type PropsType = {
 type StateType = {
 	isDirty: boolean,
 	value: string,
-	contentHeight: number,
 };
 
 export class HTMLInputView extends React.Component<PropsType, StateType> {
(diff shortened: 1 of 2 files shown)


pinarol:
hey @daniloercoli 👋When I first switch to html mode on WPAndroid app, the scroll position is at the bottom of the TextInput, I couldn't figure out why this might be happening, any ideas?

pinarol:
@daniloercoli There's another thing I couldn't figure out on Android side. There's an extra top padding on the html content, and I couldn't detect where it comes from. When I inspect it it shouldn't be there according to values, could I be missing sth?

Screenshot_1552908232

Screenshot_1552908210

src/components/html-text-input.js
```diff
@@ -38,13 +38,30 @@ export class HTMLInputView extends React.Component<PropsType, StateType> {
...
 
 	constructor() {
 		super( ...arguments );
 
 		this.edit = this.edit.bind( this );
 		this.stopEditing = this.stopEditing.bind( this );
 
+		this.panResponder = PanResponder.create( {
+			onStartShouldSetPanResponder: ( ) => true,
+			onStartShouldSetPanResponderCapture: ( ) => true,
```
etoledom: I think that just the `onStartShouldSetPanResponderCapture` is necessary for the pan responder to capture the event (and trigger `onPanResponderMove`).

src/components/html-text-input.js
```diff
@@ -83,9 +100,9 @@ export class HTMLInputView extends React.Component<PropsType, StateType> {
 	render() {
 		return (
 			<KeyboardAvoidingView style={ styles.container } parentHeight={ this.props.parentHeight }>
-				<ScrollView
-					style={ { flex: 1 } }
-					keyboardDismissMode="interactive" >
+				<View
+					{ ...( this.isIOS ? { ...this.panResponder.panHandlers } : {} ) }
+					style={ { flex: 1 } } >
```
etoledom: Do you think it's a good idea to remove this extra wrapper view and add the panHandlers directly to `KeyboardAvoidingView`?
It seems to work well but I wonder if there's a benefit on keeping this wrapper.

src/components/html-text-input.js
```diff
@@ -100,17 +117,13 @@ export class HTMLInputView extends React.Component<PropsType, StateType> {
 						ref={ ( textInput ) => this.textInput = textInput }
 						textAlignVertical="top"
 						multiline
-						style={ { ...styles.htmlView, height: this.state.contentHeight + 16 } }
```
etoledom: `contentHeight` is not being used anymore. It would be good to remove it from the state.

etoledom: And being at it, it looks like the `textInput` ref is also not being used anymore. This might be a good opportunity to clean it up :)

etoledom:
> When I first switch to html mode on WPAndroid app, the scroll position is at the bottom of the TextInput

I believe this has been the case from the beginning of times 🤔 I remember giving some time to this long time ago without good results.

src/components/html-text-input.js
pinarol: there's no benefit currently so I removed

src/components/html-text-input.js
pinarol: All true!

As far as I see textInput was even not used before this PR :)

pinarol:
@etoledom Ready for another look

daniloercoli:
> When I first switch to html mode on WPAndroid app, the scroll position is at the bottom of the TextInput

You mean that the caret is at the end of the text @pinarol ? If that's the case, it is the default behavior on Android, the caret is positioned at the end of the text on focus. We've made changes in the wrapper to move the caret at the beginning of a block onfocus/split.

pinarol:
> You mean that the caret is at the end of the text @pinarol ? If that's the case, it is the default behavior on Android, the caret is positioned at the end of the text on focus. We've made changes in the wrapper to move the caret at the beginning of a block onfocus/split.

I see the caret only if I tap to somewhere, and yes when I tap the caret appears at the end. If that's the default behavior on Android I wouldn't be worried though.

pinarol:
Will merge this since the padding issue in this [comment](https://github.com/wordpress-mobile/gutenberg-mobile/pull/755#issuecomment-473873004) doesn't look related with this PR. I removed the top padding of html content and the padding is still there on Android so I am thinking there can be an inset value on native Android level? @daniloercoli if you think the other way feel free to send your feedbacks and I can open a new PR.

Issue #302, opened 2018-11-30
Focus Loop when deleting content and blocks 


daniloercoli:
I can reproduce both on emulators and real devices a focus loop that makes the editor (and the app) unusable, since you cannot do much to recover from it, except looking at the screen and see the logic giving and moving the focus between blocks in the editor.

Made a video of the problem https://cloudup.com/chW89JFycRS

Steps to reproduce the issue:
- Start the demo app
- Tap into the first Para block (`Hello World!`)
- Tap Enter and add a new empty block after
- Write something in the block
- Repeat the step above by adding more blocks quickly

Sometime the issue start here, otherwise:
- on the last added block, press delete and keep it pressing for remove the content and the block.

I was also able to reproduce the issue by just adding one block and deleting it.

daniloercoli:
cc @mzorz

mzorz:
> I can reproduce both on emulators and real devices a focus loop

Interesting! Can you add steps to reproduce here please? thanks

daniloercoli:
Added testing steps in the description.

PR #764, opened 2019-03-19
iOS: Avoid block list auto-scroll when Links UI is present.

### LINKED ISSUE #702: Dismissing Links UI (BottomSheet) disables hide keyboard button.
When the Links UI BottomSheet is dismissed with the keyboard showing, the hide-keyboard button won't work until you focus on the paragraph block again.

References:
https://github.com/wordpress-mobile/gutenberg-mobile/pull/701#issuecomment-469733637

bottom-sheet-keyboard-close

### LINKED ISSUE #717: [iOS]Opening the links bottom sheet causes the focused text get out of the viewable area
This happens only in iOS. Opening the links bottom sheet causes the focused text get out of the viewable area. This is tested with iPhone XR simulator:

links-ui2

etoledom:
Fixes #717 
Fixes #702 

Related `gutenberg` branch: `rnmobile/717-block-list-scroll`
Related `gutenberg` PR: https://github.com/WordPress/gutenberg/pull/14522.

This PR fixes an issue where the block list scrolled when a Links UI's TextInput was focused.

### Problem:
This was happening because `KeyboardAwareFlatList` is always listening to keyboard events, without considering what TextField was selected. So the TextInput on Links UI was focused, a new keyboard event was emitted, and the flat-list scrolled trying to "avoid" the keyboard.

### Proposed solution:
Coming with a solution was tricky, since Links UI and BlockManager are not connected in any way. But this seemed to be a problem on BlockManager's flat-list scrolling when it shouldn't, and not something   related to Links UI (or BottomSheet) in particular.

I also tried to not modify the original implementation of `KeyboardAwareFlatList`.

Considering those points, in the proposed solution "BlockManager" sends a flag to `KeyboardAwareFlatList` via `enableAutomaticScroll`. When this flag is true, any kind of automatic scrolling is disabled.

The information used to enable auto-scroll is the focus state of the elements of the block list. If anything is focused on the block lists, means that the user is typing something in a block, thus we will want it to auto-scroll. This includes focusing a block for the first time, splitting and merging.

In any other case, i.e. user focusing a TextInput on a `BottomSheet` (or any other element outside the block list), we won't scroll the block list automatically.

### Challenges:

There were three main issues with this approach:

1. **`BlockHolder` was not receiving `onFocus` calls when Aztec was focused programmatically (i.e. creating a new block on enter, and selecting it)**

**Solution:** I rolled back an old `fix` where the onFocus from Aztec was silenced. (https://github.com/wordpress-mobile/gutenberg-mobile/blob/46f7b6bb4a5c3494dbe234f0e3435ed9d85f9b6c/react-native-aztec/src/AztecView.js#L159)
With this, we can get back the onFocus signal from the native Aztec, when the focus wasn't generated via a user tap.

I did add a check to avoid resurfacing the original fixed issue (https://github.com/wordpress-mobile/gutenberg-mobile/issues/302). It would be nice if @daniloercoli could double check if everything is fine.

2. **When splitting blocks, there's a short period of time in between onBlur (previous block) and onFocus (new block). And of course the `KeyboardAwareFlatList ` check of `enableAutomaticScroll` was right in between that change. So we were loosing scroll to the new block on split.**

**Solution:** Avoid to call onFocus/onBlur on `BlockHolder` when they are caused by a block split.
This solution uses a `_isSplitting` flag that I'm not happy with, but it works.

From the eyes of `BlockManager`, when splitting a block, the focus was never left from the block list itself. It just went from one block to the other.

3. **Android and iOS are considerably branched on how focus work. With a feeling that Android is quite unstable.**

**Solution:** No solution 😞

I did try to even up the ground and avoid branching between platforms, but when something was fixed, another issue appeared. The most fundamental problem that I saw that on Android, when splitting blocks, focus on the new block is called first, and then blur is called second, effectively hiding the keyboard after splitting. It seems to be an Android Native problem.

I finally ended up branching on the `hiddeKeyboard` toolbar button. But this solved an issue on Android where after pressing the hideKeyboard button, the keyboard was hidden but the block remained selected.

It would be probably good to revisit how Aztec focus/blur are working in both platforms.

### Extras:

- Added a few check on `BlockHolder` to avoid calling `focus (onSelect)/blur (clearSelectedBlock)` more than the necessary.
- Cleaned up some imports 

----

blocks-scroll

**To test iOS:**
- Check that the block list scrolls normally when selecting, splitting and merging blocks.
- Open the Links UI (as shown in the Gif).
- Write something in both TextInputs
- Check that the block list at the back doesn't scroll automatically.
- Dismiss the BottomSheet.
- Press the `hide keyboard` button.
- Check that the keyboard is dismissed and the block is deselected.

**To test Android:**
- Check that the block list scrolls normally when selecting, splitting and merging blocks.
- Select a Paragraph block.
- Press the `hide keyboard` button on the toolbar.
- Check that the keyboard is dismissed and the block is deselected.


etoledom:
Marked as `Not ready for review` since we are working on another solution with @pinarol

etoledom:
Closing in favor of https://github.com/wordpress-mobile/gutenberg-mobile/pull/767
Will make a new PR to fix #702

PR #765, opened 2019-03-20
Added a minimal delay to call setSelection() to enforce special focus handling cases

mzorz:
The title has it, when we want to override the caret selection (as in the case of deleting a block and coming back to the block above) we should do it _last_ to make sure any other positioning work has finished. Taking into account chained events would be super complex so, adding a delay seems to do the job while easy to understand. This is a try to improve and fix issue A described in https://github.com/wordpress-mobile/gutenberg-mobile/pull/756#issuecomment-474824104, stating here again for clarity:

**Issue A:** After two blocks are merged, a cursor will be placed on the last remembered position in the block. (Affected both iOS and Android platform)

Steps to reproduce:

1. Create a new Gutenberg post with 2 simple paragraph blocks
2. Put the cursor somewhere in the first block except its end
3. Go to the end of the second block and start backspacing to delete characters. Delete up until the very first character of the second block
4. Notice that the "Start writing" placeholder text appears
5. (PREVIOUSLY) Notice the caret is placed at the position set in step 2
5. (NOW) Observe the caret is correctly placed at the end of block as expected.

This was still happening for me on a Pixel 2, Android 8.1, but this change in code made it work for me.

cc @marecar3  @daniloercoli  interested in hearing your thoughts

### DIFF
--- a/react-native-aztec/android/src/main/java/org/wordpress/mobile/ReactNativeAztec/ReactAztecArrowKeyMovementMethod.java
+++ b/react-native-aztec/android/src/main/java/org/wordpress/mobile/ReactNativeAztec/ReactAztecArrowKeyMovementMethod.java
@@ -30,14 +30,23 @@ public class ReactAztecArrowKeyMovementMethod extends ArrowKeyMovementMethod {
         if ((dir & (View.FOCUS_FORWARD | View.FOCUS_DOWN)) != 0) {
             if (view.getLayout() == null) {
                 // This shouldn't be null, but do something sensible if it is.
-                Selection.setSelection(text, 0); // <-- setting caret to start of text
+                handleSelectionOnEnd(view, text);
             }
             else if (!reactAztecText.isTouched()) {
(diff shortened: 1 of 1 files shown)


marecar3:
Hey @mzorz, I have tested this PR on the `Nexus 5X 8.1.0`and `Motorola Moto G6 Android 8.0.0` and here are the results :

1. This PR against WPAndroid -> working
2. This PR against Demo app -> not working

1. [My PR](https://github.com/wordpress-mobile/gutenberg-mobile/pull/756) against WPAndroid -> not working
2. [My PR](https://github.com/wordpress-mobile/gutenberg-mobile/pull/756) against Demo app -> working

but as WPAndroid is more relevant then Demo app, I will say that this fix does the trick and we can go with it.

mzorz:
super interesting findings @marecar3 ! I'm with you, we could go ahead as it works on WPANdroid and check the demo app later 👍

PR #769, opened 2019-03-21
Implement readable content margins

### LINKED ISSUE #604: Tablet layout visual refinements
This issue covers various refinements related to layout/alignment.

Visualization for reference:

image

## Punchlist

- [ ] User should be able to scroll by dragging anywhere on canvas
- [ ] ~Center column container should have max-width of 580, which provides a nice readable balance/line-length~
- [ ] ~Block boundaries should extend beyond center column by 16 on each side (inner block 580 to edge of container)~
- [ ] Quick Toolbar should extend to edges of canvas

Note: Upon further inspection, the container-width might be dependent on readable-margins, so this is probably fine as-is. At least on iOS, when I change the OS-level font-size the container width adjusts (I didn't realize I had my OS-level setting to lower than default). However, the font-size doesn't actually adjust in the editor – I will create another separate issue for this.

etoledom:
fixes #604 

This PR implements a JS only version of the iOS Readable Content Margins. This help us to show a narrower content view when the screens are too wide.

In the original ticket, @iamthomasbishop opted to keep the native iOS Readable Content Guide, but I opted for a simpler first iteration with a fixed maximum width. And later on implement platform specific details.

ReadableContent

To test Android:
- Checkout this branch on WPAndroid project's `gutenberg-mobile` submodule.
- Run on a tablet (or tablet emulator)
- Check that the content is centered as in the image.
- Check that the selected blocks are boxed with a border.
- Run on a Phone
- Check that there are no side borders on selected blocks.

To test iOS:
- Check the steps mentioned here: https://github.com/wordpress-mobile/WordPress-iOS/pull/11316


etoledom:
Thank you both!

Issue #768, opened 2019-03-21
[Android]HTML editor scrolling isn't smooth


pinarol:
After fixing this issue https://github.com/wordpress-mobile/gutenberg-mobile/issues/744 this time it had a side effect on Android.

Before: 
https://cloudup.com/ckcs33vd-BP

After:
https://cloudup.com/c60oBwEiJ3B

Scrolling loses its momentum very fast. So to fix this we'll need to render iOS and Android html editors separately since the requirements are mutually exclusive.

PR #771, opened 2019-03-22
Fix android html mode scrolling

pinarol:
Fix: https://github.com/wordpress-mobile/gutenberg-mobile/issues/768

This PR separates the ui part of html editor for Android and iOS because for Android we need to wrap TextInput with a ScrollView, on the other hand on iOS this causes the following issue https://github.com/wordpress-mobile/gutenberg-mobile/issues/744 

**To Test**

WPAndroid:
- Switch to HTML mode
- Verify html content is scrolling with the title
- Verify scrolling is fast and smooth
- Verify when we edit title and html content and switch back to visual mode we see the edits

WPiOS:

Before starting:
Update podfile gutenberg commit with the last commit of this PR because there's a new bridge method that had been merged to develop which is unrelated with this PR, but WPiOS podfile wasn't updated. Make sure you run rake dependencies and clean build folder beforehands.

- Switch to HTML mode
- Verify html content is scrolling in its own bounds
- Verify scrolling is fast and smooth
- Verify when we edit title and html content and switch back to visual mode we see the edits


pinarol:
@daniloercoli I think we could make it work like that too. both has pros and cons, this time we'd need to add more platform checks for contentHeight calculation, scrollEnabled, panResponder etc. which can reduce readability a bit, but, it'd reduce the total code also.

etoledom:
I was thinking on a way that we could share more code. I did some tests on this branch: 
`try/html-editor`

On that branch there is more shared code, but I noticed that there are small differences that makes this challenging. So this solution on the mentioned branch is not too straightforward.

I'm happy with the solution proposed on this PR ✅

daniloercoli:
Let's get this merged then!

PR #772, opened 2019-03-22
Update Aztec to 1.5.0

loremattei:
This PR updates Aztec to 1.5.0

**To Test:**

- Run `yarn install` `yarn start` and `yarn ios` and verify that example app runs successfully.


pinarol:
Just a note: no need to update cart file unless we update the major version. we can just cd to react-native-aztec/ios/ and run `carthage update` and it will resolve to the biggest available version lower than 2.0.0, so it can be more practical. But this way is also fine.

PR #773, opened 2019-03-22
WIP / Experiment - detect block break or join from focus information

mzorz:
DO NOT MERGE - experiment

This is an experimental PR,  trying to identify "block breaks"  (tapping enter to produce a new block) or join (delete all text and appear at the end of the block immediately above) from the available information about focus to Aztec wrapper.

@marecar3 this is the approach I took but I think there's not enough information to decide on all cases. we can continue to work on this later after today's GB cut


mzorz:
Closing this - it definitely needs more information to tell where to put the caret, and this information lives at another level (the "block-concious" level) so, closing.

PR #774, opened 2019-03-22
Fix Android crash on device rotation

etoledom:
This PR updates the Gutenberg ref to test https://github.com/WordPress/gutenberg/pull/14576

To test: 
- Check out WPAndroid with the `gutenberg-mobile` submodule checkout on this branch.
- Run the app and instantiate an empty Gutenberg instance.
- Rotate the device/emulator.
- Check that gutenberg doesn't crash.


etoledom:
We are targeting the release branch since it's a safe change that will fix a crash on Android.
cc @pinarol @jtreanor

mzorz:
Confirmed works on Pixel 2 Android 8.1 - thank you @etoledom !

mzorz:
Merging as discussed and agreed with @etoledom

PR #784, opened 2019-03-26
[iOS]: Make Hide-Keyboard button work again after dismissing Links UI.

### LINKED ISSUE #702: Dismissing Links UI (BottomSheet) disables hide keyboard button.

etoledom:
fixes #702 

This PR fixes an issue that made the Hide-Keyboard toolbar button not work after the Links UI bottom-sheet was dismissed.
Additionally, now on both platform, dismissing the keyboard with this button will clear the block selection.

hide-keyboard

To test (iOS):
- Select a paragraph block.
- Present the Links UI (pressing the links button on the toolbar).
- Dismiss the Links UI bottom-sheet.
- Press the hide-keyboard button on the toolbar.
- Check that the button works and the keyboard is dismissed.
- Check that the block is now deselected.

To test (Android):
**Note:** On Android, the keyboard is already hidden after dismissing the Links UI.

- Select a paragraph block.
- Press the hide-keyboard button on the toolbar.
- Check that the keyboard is dismissed.
- Check that the block is now deselected.

### DIFF
--- a/react-native-aztec/src/AztecView.js
+++ b/react-native-aztec/src/AztecView.js
@@ -1,6 +1,6 @@
 import PropTypes from 'prop-types';
 import React from 'react';
-import ReactNative, {requireNativeComponent, ViewPropTypes, UIManager, ColorPropType, TouchableWithoutFeedback} from 'react-native';
+import ReactNative, {requireNativeComponent, ViewPropTypes, UIManager, ColorPropType, TouchableWithoutFeedback, Platform} from 'react-native';
 import TextInputState from 'react-native/lib/TextInputState';
 
 const AztecManager = UIManager.getViewManagerConfig('RCTAztecView');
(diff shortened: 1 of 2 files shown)


etoledom:
@daniloercoli or @mzorz could you please check the Android side?
Thank you!

marecar3:
Hey @etoledom, nice work here! :)

I just found one case on Android platform which is maybe interesting.
If you tap on Back button (Hardware/Software) on Android, it will dismiss the keyboard, but the block will be still in focus. Some users will probably click on the Back button, so maybe we should cover that case also, WDYT ?

etoledom:
@marecar3 - Thanks for the review!

> so maybe we should cover that case also, WDYT?

Sounds good! I'll check that out.

etoledom:
@marecar3 - I gave your suggestion a try, but it's more complicated than I thought.

The `react-native` [BackHandler](https://facebook.github.io/react-native/docs/backhandler.html) component won't help us here, since it works only when the Android Back button is on "Back mode` (with the arrow pointing to the left) and not when it's on "Hide Keyboard" mode (with the arrow pointing down).

I searched for a way to intercept that button on "Hide Keyboard" mode, and I couldn't find any. I also tried using keyboard events, but those are too generic, and we don't want to unselect the block for every time the keyboard is dismissed (i.e. presenting the Links UI).

I'd suggest to create a new ticket for this, since it's not actually related to our Hide Keyboard toolbar button, and it seems to not be a simple fix.

etoledom:
Thank you both! :)

PR #788, opened 2019-03-26
Intercept image request to add access token when site is private

marecar3:
This is the PR based on issue : https://github.com/wordpress-mobile/gutenberg-mobile/issues/376

It allows WPAndroid users who have a private site to access images when creating new blog Post.

Test Steps : 

Follow instructions on WPAndroid related PR : https://github.com/wordpress-mobile/WordPress-Android/pull/9483

### DIFF
--- a/react-native-gutenberg-bridge/android/src/main/java/org/wordpress/mobile/WPAndroidGlue/OkHttpHeaderInterceptor.java
+++ b/react-native-gutenberg-bridge/android/src/main/java/org/wordpress/mobile/WPAndroidGlue/OkHttpHeaderInterceptor.java
@@ -1,0 +1,35 @@
+package org.wordpress.mobile.WPAndroidGlue;
+
+import android.util.Log;
+
+import org.wordpress.mobile.WPAndroidGlue.WPAndroidGlueCode.OnAuthHeaderRequestedListener;
+
+import java.io.IOException;
+
+import okhttp3.Interceptor;
+import okhttp3.Request;
+import okhttp3.Response;
+
+public class OkHttpHeaderInterceptor implements Interceptor {
+
+    private static final String AUTHORIZATION_HEADER_KEY = "Authorization";
+
+    private OnAuthHeaderRequestedListener mOnAuthHeaderRequestedListener;
+
+    void setOnAuthHeaderRequestedListener(OnAuthHeaderRequestedListener onAuthHeaderRequestedListener) {
+        mOnAuthHeaderRequestedListener = onAuthHeaderRequestedListener;
+    }
+
+    @Override
+    public Response intercept(Chain chain) throws IOException {
+        Request.Builder builder = chain.request().newBuilder();
+
+        String authHeader = mOnAuthHeaderRequestedListener.onAuthHeaderRequested(chain.request().url().toString());
+        if (authHeader.length() > 0) {
+            builder.addHeader(AUTHORIZATION_HEADER_KEY, authHeader);
+        }
+
+        return chain.proceed(builder.build());
+    }
+
+}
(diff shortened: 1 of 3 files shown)


react-native-gutenberg-bridge/android/src/main/java/org/wordpress/mobile/WPAndroidGlue/OkHttpHeaderInterceptor.java
```diff
@@ -0,0 +1,35 @@
...
+    void setOnAuthHeaderRequestedListener(OnAuthHeaderRequestedListener onAuthHeaderRequestedListener) {
+        mOnAuthHeaderRequestedListener = onAuthHeaderRequestedListener;
+    }
+
+    @Override
+    public Response intercept(Chain chain) throws IOException {
+        Request.Builder builder = chain.request().newBuilder();
+
+        String authHeader = mOnAuthHeaderRequestedListener.onAuthHeaderRequested(chain.request().url().toString());
+        if (authHeader.length() > 0) {
```
mzorz: Can we maybe use `if (!TextUtils.isEmpty(authHeader))` here? this would avoid a NPE in case `authHeader` was null

react-native-gutenberg-bridge/android/src/main/java/org/wordpress/mobile/WPAndroidGlue/WPAndroidGlueCode.java
```diff
@@ -170,14 +182,27 @@ public void editorDidMount(boolean hasUnsupportedBlocks) {
                 }
             }
         });
+
+        ImagePipelineConfig imagePipelineConfig = getImagePipelineConfig(sOkHttpClient);
+        MainReactPackage mainReactPackage = new MainReactPackage(getMainPackageConfig(imagePipelineConfig));
+
         return Arrays.asList(
-                new MainReactPackage(),
+                mainReactPackage,
```
mzorz: We can probably just remove the local variables `imagePipelineConfig` and `mainReactPackage` and just replace this reference in this line here with ```new MainReactPackage(getMainPackageConfig(getImagePipelineConfig(sOkHttpClient)))```

react-native-gutenberg-bridge/android/src/main/java/org/wordpress/mobile/WPAndroidGlue/OkHttpHeaderInterceptor.java
marecar3: Hey @mzorz, yup we can do that.

react-native-gutenberg-bridge/android/src/main/java/org/wordpress/mobile/WPAndroidGlue/WPAndroidGlueCode.java
marecar3: Yup, we can do that :)

marecar3:
> LGTM @marecar3
> Left some comments to improve the code a bit but looks generally OK for me.
> 
> Also been testing it, looks to be working fine - I can see the code goes through the header request interface when I upload a new image to a Post, (or fwiw when I install the app from scratch) but if I open a Post with images already on it twice, I noticed the code is not run. Just to confirm, do you know whether that's because of fresco's cache?

Hi @mzorz, thank you very much for the comments, I have fixed them. 
I am not sure about the answer, so I also need to check and confirm that.

marecar3:
Hey @mzorz, I have created a Post from Gutenberg web and I tried to reload it from the mobile app.
First time when I have opened the Post, it loaded an image with the token, and a second time it took it from cache (like you experienced).

mzorz:
Excellent! Thanks for taking the time to confirm this @marecar3 

LGTM! Feel free to merge this one, then update the WPAndroid PR with the hash and finally merge that one too :D

PR #790, opened 2019-03-27
Fix `hasUnsupportedBlocks` check.

etoledom:
This PR fixes a problem where the `hasUnsupportedBlocks` sent to the native apps was always `false`.

The reason is that after [this refactor](https://github.com/wordpress-mobile/gutenberg-mobile/commit/e053284e6372afcdac19a7265c034dbed99036e2), the block list is not available on `AppContainer` until the "editor is ready", thus `this.props.blocks` will always be empty on `componentDidMount`.

The solution is to wait for `props.isReady` to be true to check and send `hasUnsupportedBlocks` with the appropriate value.

**Note:** `componentDidUpdate`  is the best place I found to do this check, but sending `hasUnsupportedBlocks` as part of the `editorDidMount` (from `componentDidUpdate`) feels not optimal. I'm open to suggestions (as always). But as a hotfix, it works.

- I'd rename the `editorDidMount` message sent to native, but I'd prefer to do it on a separate PR on develop, since it will require native code on both platforms to be modified.
- Other possibility is to split `editorDidMount` and create a separate message specific for `hasUnsupportedBlocks`.

**Note 2:** After it's approved, I'll create a new JS Bundle before merging.

To test:
- Run the project from Xcode (`open ios/gutenberg.xcodeproj`).
- Check in the Xcode console that the message `gutenbergDidMount(hasUnsupportedBlocks: true)` is written.
- Run this branch on WPAndroid. 
- Check that the proper `hasUnsupportedBlocks` is received.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "gutenberg-mobile",
-  "version": "1.1.1",
+  "version": "1.1.2",
   "private": true,
   "config": {
     "jsfiles": "./*.js src/*.js src/**/*.js src/**/**/*.js",
--- a/src/app/AppContainer.js
+++ b/src/app/AppContainer.js
@@ -148,7 +148,7 @@ class AppContainer extends React.Component<PropsType> {
 	componentDidUpdate( prevProps ) {
 		if ( ! prevProps.isReady && ( prevProps.isReady !== this.props.isReady ) ) {
 			const { blocks } = this.props;
-			const isUnsupportedBlock = ( { name } ) => ( { name } ) => name === UnsupportedBlock.name;
+			const isUnsupportedBlock = ( { name } ) => name === UnsupportedBlock.name;
 			const unsupportedBlocks = blocks.filter( isUnsupportedBlock );
 			const hasUnsupportedBlocks = ! isEmpty( unsupportedBlocks );
 
--- a/src/app/AppContainer.js
+++ b/src/app/AppContainer.js
@@ -87,10 +87,6 @@ class AppContainer extends React.Component<PropsType> {
 	}
 
 	componentDidMount() {
-		const blocks = this.props.blocks;
-		const hasUnsupportedBlocks = ! isEmpty( blocks.filter( ( { name } ) => name === UnsupportedBlock.name ) );
-		RNReactNativeGutenbergBridge.editorDidMount( hasUnsupportedBlocks );
-
 		this.subscriptionParentGetHtml = subscribeParentGetHtml( () => {
 			this.serializeToNativeAction();
 		} );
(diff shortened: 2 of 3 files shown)


etoledom:
cc @koke - This should fix what I mentioned on https://github.com/wordpress-mobile/gutenberg-mobile/pull/643#pullrequestreview-218286304

etoledom:
Thank you @hypest !
I'll push the new bundles to create the WP Apps PRs.

koke:
I'm fine with this as a hotfix, but also not super happy with the naming of things. I think it's OK if we need to abandon the `didMount` naming and have an `editorReady` event that we ensure fires just once as soon as the editor is initially set up.

I am concerned that `componentDidUpdate` is not called in the initial render, so the event won't be triggered unless the props/state change. With this refactor I understand this is happening right now (otherwise the code on `componentDidMount` would still work), but it doesn't seem future-proof. 

What if a future refactor changes the behaviour again and everything is ready on `componentDidMount`, and the component doesn't update until the user makes a change (which sounds like the desired behaviour)? In that case, the user could open a post, the `editor_session_start` even wouldn't get triggered right away, and they could potentially switch to classic/html without having triggered the session start.

etoledom:
I agree!
For those reasons I'd prefer to keep sending the `editorDidMount` event from the `componentDidMount` component lifecycle, and separate `hasUnsupportedBlocks` to be more flexible on when to emit that event. Whenever we emit it, we need to be sure that `props.isReady == true`, and to be sure we are sending it just once.

hypest:
I'd prefer if we don't expose/bubble RN/React lifecycle events to the parents, as it makes it hard to understand what the intention is by looking at the interface. Instead, the interface to the parents should be as close as possible to certain functionalities.

We already do various things in wpandroid based on the `editorDidMount` event ([sending analytics](https://github.com/wordpress-mobile/WordPress-Android/blob/1c0bc9653fb88e3fdca35ffe50699511f1adbb79/WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java#L3703), [hiding a progress bar](https://github.com/wordpress-mobile/WordPress-Android/blob/2ca115b2d4f2c8a2426275622a5f4fe8572195bc/libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/GutenbergEditorFragment.java#L261), [focusing the title bar](https://github.com/wordpress-mobile/gutenberg-mobile/blob/1a98224612f345800b2ed1074745c9897123c9e3/react-native-gutenberg-bridge/android/src/main/java/org/wordpress/mobile/WPAndroidGlue/WPAndroidGlueCode.java#L166) and [initiate a new media block](https://github.com/wordpress-mobile/gutenberg-mobile/blob/1a98224612f345800b2ed1074745c9897123c9e3/react-native-gutenberg-bridge/android/src/main/java/org/wordpress/mobile/WPAndroidGlue/WPAndroidGlueCode.java#L169)), which means we already overloaded the initial intent (signal the presence of unsupported blocks) so, separating the events into 2 makes sense. Since this PR is about a hotfix of an important regression though, it makes sense to keep it short and quick as is, and handle the events splitting in a separate PR.

hypest:
Ah oh, by the way, great job adding a test for the regression @etoledom ! 👏

Issue #476, opened 2019-01-17
Accessibility investigation.


etoledom:
At some point it will be necessary to investigate how to apply the native Accessibility in React-Native.

There is an articule about this here: https://facebook.github.io/react-native/docs/accessibility.html

With that investigation, we should be able to figure out what is the work we need to do to make Gutenberg Mobile accessible, and create a guideline for all future UI development work to comply to.

Related issues:
https://github.com/wordpress-mobile/gutenberg-mobile/issues/14
https://github.com/wordpress-mobile/gutenberg-mobile/issues/598
https://github.com/wordpress-mobile/gutenberg-mobile/issues/605
https://github.com/wordpress-mobile/gutenberg-mobile/issues/697

**Actions**, improve accessibility on:
- [x] Toolbar: PR #793
- [x] Image block (deselected): PR #794 
- [x] Image block (selected + toolbar): PR #920
- [x] Paragraph block: PR #923

rachelmcr:
Do we have a timeline or estimate for when we'll work on this? I started looking at accessibility in the WPiOS integration and ran into major accessibility issues (no labels for toolbar buttons, block-level buttons completely inaccessible). That's a concern for our users and a blocker for automating tests that use Gutenberg.

etoledom:
We will start working on this next spring 👍

rachelmcr:
Excellent, thanks for the update! @wordpress-mobile/platform-9 also agreed that we're happy to help with adding accessibility labels and identifiers as needed (especially for test automation but also as we come across issues). So once there's a documented/agreed-upon process for making Gutenberg accessible please let us know and we can contribute to any work that's needed in the existing UI!

etoledom:
Closing since the main issues have been solved.
A few [new independent tickets](https://github.com/wordpress-mobile/gutenberg-mobile/issues?utf8=✓&q=is%3Aopen+label%3AAccessibility+is%3Aissue) have been opened too.

PR #793, opened 2019-03-29
Accessibility Mark I - Toolbar Buttons

etoledom:
Part of #476 

This is a simple PR that implements the changes made on https://github.com/WordPress/gutenberg/pull/14697 to improve accessibility on the toolbar buttons.

### To test (iOS):
- Activate VoiceOver: https://github.com/wordpress-mobile/WordPress-iOS/wiki/Using-VoiceOver
- Select the "Add Block" button.
- Check that VoiceOver reads the label, that it reads the element type (button), and it reads the hint (after a couple of seconds).
- Select the "Unod/Redo" buttons.
- Check that it reads the label, the element type (button) and the state (dimmed)
  - For some reason it says dimmed instead of disabled, but it's a iOS system thing.
- Select a Paragraph or Heading block.
- Check that the format buttons are read correctly (label and type).
- Activate one of the format buttons (**b**, _i_, ~ABC~, H2, H3, or H4).
- Check that the button state is read correctly (selected)
- Select the Hide Keyboard button.
- Check that it reads properly (label and type).

### To test (Android):
- Activate TalkBack: https://www.wikihow.com/Enable-TalkBack-on-Your-Android
  - In my case, the system had a handy tutorial on how to use it.
  - If your version doesn't have it, it's actually very similar to [VoiceOver.](https://www.wikihow.com/Enable-TalkBack-on-Your-Android)
- Select the "Add Block" button.
- Check that VoiceOver reads the label, that it reads the element type (button), and it reads the hint (after a couple of seconds).
- Select the "Unod/Redo" buttons.
- Check that it reads the label, the element type (button) and the state (dimmed)

And that's it. On Android, blocks are still unusable by TalkBack :(

### DIFF
--- a/src/block-management/block-toolbar.js
+++ b/src/block-management/block-toolbar.js
@@ -82,7 +82,7 @@ export class BlockToolbar extends Component<PropsType> {
 				{ showKeyboardHideButton &&
 				( <Toolbar passedStyle={ styles.keyboardHideContainer }>
 					<ToolbarButton
-						title={ __( 'Hide keyboard' )}
+						title={ __( 'Hide keyboard' ) }
 						icon="keyboard-hide"
 						onClick={ this.onKeyboardHide }
 					/>
--- a/src/block-management/block-toolbar.js
+++ b/src/block-management/block-toolbar.js
@@ -48,7 +48,7 @@ export class BlockToolbar extends Component<PropsType> {
 		} = this.props;
 
 		return (
-			<View style={ styles.container }>
+			<View style={ styles.container } >
 				<ScrollView
 					horizontal={ true }
 					showsHorizontalScrollIndicator={ false }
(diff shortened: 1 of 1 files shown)


src/block-management/block-toolbar.js
```diff
@@ -58,18 +58,19 @@ export class BlockToolbar extends Component<PropsType> {
 				>
 					<Toolbar>
 						<ToolbarButton
-							label={ __( 'Add block' ) }
+							title={ __( 'Add block' ) }
 							icon={ ( <Dashicon icon="plus-alt" style={ styles.addBlockButton } color={ styles.addBlockButton.color } /> ) }
 							onClick={ onInsertClick }
+							extraProps={ { hint: __( 'Double tap to add a block' ) } }
```
etoledom: As we talked with @JavonDavis , we will try to change this on a future PR and pass directly an accessibilityHint prop.
For now I prefer to keep it simpler.

Tug: May I ask why the hint is "double tap"? A single tap seems to be enough to execute the action here, is it not?

etoledom: When VoiceOver is active, a single tap will select an element, and a double tap will trigger the action. It's kind of standard for hints.

Tug: Oh I see, indeed I needed to actually enable TalkBack to see it in action 👍

etoledom:
Thank you both!

hypest:
👋 @etoledom , I wonder if there are any guidelines we can extract at this point and add them to the Wiki for how the code of an accessible component should look. It can be a running/living document as we get better at this. WDYT?

etoledom:
@hypest - I have that in mind!

Not sure about best practices regarding gutenberg in specific, like where is best to add the accessibility related props (block files, controls, etc...), but there are more generics guidelines that I can definitely share right away, and we can improve them  in the way.

I can do something similar to this: https://github.com/wordpress-mobile/WordPress-iOS/wiki/VoiceOver-guideline#when-to-and-when-not-to with a RN language. 👍

hypest:
Sounds good, thanks @etoledom !

Issue #605, opened 2019-02-15
Ensure that OS-level font-size settings are honored in editor


iamthomasbishop:
While inspecting tablet layout refinements [here](https://github.com/wordpress-mobile/gutenberg-mobile/issues/604), I noticed that changes to font-size on the OS level aren't honored in the editor – both on iOS and Android.

## Steps to reproduce

1. Type a bunch of text in a Paragraph Block, preferably something in multiple lines.
2. Go to OS-level font-size settings, adjust to very large size.
3. Go back to Block Editor.
4. Notice that font-size is still 16.

hotchemi:
@iamthomasbishop Right now I'm investigating the issue and let me confirm that this is the issue only for iOS or both iOS/Android🙇I built WP-iOS/Android app with the latest commit hash on develop branch and what I saw is the following.

### iOS

- [head commit](https://github.com/wordpress-mobile/WordPress-iOS/commit/958cccc5cebab22018c87af51f07db89fa5c57d6)

After changing system level font size I could reproduce the issue.

<img src="https://user-images.githubusercontent.com/471318/54942481-85baf900-4f72-11e9-8aa7-01bac379deae.gif" width="250" />

### Android

- [head commit](https://github.com/wordpress-mobile/WordPress-Android/commit/389ebd4c3e3e3f854767bfd929f1ea785723dd31)

<img src="https://user-images.githubusercontent.com/471318/54942555-adaa5c80-4f72-11e9-80ff-6a4c8683cf1c.gif" width="250" />

So right now I'm wondering this is the only issue for iOS..am I missing something?🙇Thank you!

hypest:
👋 @hotchemi , judging from the gif you shared, I think that the editor there is Aztec, not Gutenberg. Have you perhaps tried Gutenberg too and the font sizing issue doesn't happen?

hotchemi:
@hypest Thank you for pointing out🙇I tried again on Android and found gutenberg-mobile and wp-android behave slightly differently.

### gutenberg-mobile

On example app it seems works.

<img src="https://user-images.githubusercontent.com/471318/55007953-9ecfb280-5023-11e9-949f-263fff84c475.gif" width="250" />

### wp-android

I could repricate the issue.

<img src="https://user-images.githubusercontent.com/471318/55008098-db031300-5023-11e9-8c49-0d7bbf54286e.gif" width="250" />

etoledom:
[...]

PR #795, opened 2019-03-31
[Issue/605][Android]Apply OS font size scale change in editor

hotchemi:
## Issue

- Fixes android part of https://github.com/wordpress-mobile/gutenberg-mobile/issues/605

## Overview

This PR fixes GM editor to apply system level font scale change immediately.

Honestly this change is a bit tricky as React Native has a [mechanism](https://facebook.github.io/react-native/docs/text#allowfontscaling) that allows `<Text>` component to scale font size according to system setting. **But the point is we're using GM component on top of fragment of which [retain flag is true](https://github.com/wordpress-mobile/WordPress-Android/blob/ff8a0692400f02286331f09a4f3705f8b2c54014/libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/GutenbergEditorFragment.java#L193)**, we can't rely on RN font scaling system.

After several investigations, I've noticed we can detect fontScale change on RN side by observing `Dimensions` change and added a function to calculate new font size to deal with the issue.

### Note

This PR doesn't affect to iOS side since the situation I explained above is a different on each platform.

### Links

- https://facebook.github.io/react-native/docs/text#allowfontscaling
- https://facebook.github.io/react-native/docs/dimensions

### Screenshots

| Before | After |
| ----------- | ----------- |
|||

### Test

- Run example app on Android
- Turn on `Use Block Editor` from app setting
- Open editor screen and type something
- Go to OS setting(Accessibility→Font Size) and change font size
- Back to editor screen and confirm content font size has been changed

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -16,7 +16,7 @@
       <activity
         android:name=".MainActivity"
         android:label="@string/app_name"
-        android:configChanges="keyboard|keyboardHidden|orientation|screenSize"
+        android:configChanges="keyboard|keyboardHidden|orientation|screenSize|fontScale"
         android:windowSoftInputMode="adjustResize">
         <intent-filter>
             <action android:name="android.intent.action.MAIN" />
--- a/react-native-aztec/src/utils.js
+++ b/react-native-aztec/src/utils.js
@@ -1,7 +1,10 @@
 import { Platform } from 'react-native';
 
+// Basically React Native has a mechanism to scale font size automatically.
+// But GM is rendered on top of fragment of which retain flag is true.
+// Hence we deal with fontScale change manually on our side.
 export function getScaledFontSize( fontSize, fontScale ) {
-	if ( Platform.OS == 'ios' ) {
+	if ( Platform.OS === 'ios' ) {
 		return fontSize ? fontSize : undefined; // to not to affect to iOS
 	}
 	const baseFontSize = 18;
--- a/react-native-aztec/src/utils.js
+++ b/react-native-aztec/src/utils.js
@@ -1,0 +1,9 @@
+import { Platform } from 'react-native';
+
+export function getScaledFontSize( fontSize, fontScale ) {
+	if ( Platform.OS == 'ios' ) {
+		return fontSize ? fontSize : undefined; // to not to affect to iOS
+	}
+	const baseFontSize = 18;
+	return ( fontSize ? fontSize : baseFontSize ) * fontScale;
+}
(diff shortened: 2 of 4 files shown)


android/app/src/main/AndroidManifest.xml
```diff
@@ -16,7 +16,7 @@
       <activity
         android:name=".MainActivity"
         android:label="@string/app_name"
-        android:configChanges="keyboard|keyboardHidden|orientation|screenSize"
+        android:configChanges="keyboard|keyboardHidden|orientation|screenSize|fontScale"
```
hotchemi: this is not directly related to the issue but to align condition with wp-android added `fontScale` to prevent activity recreation by fontScale change.

react-native-aztec/src/utils.js
```diff
@@ -0,0 +1,12 @@
+import { Platform } from 'react-native';
+
+// Basically React Native has a mechanism to scale font size automatically.
+// But GM is rendered on top of fragment of which retain flag is true.
+// Hence we deal with fontScale change manually on our side.
+export function getScaledFontSize( fontSize, fontScale ) {
```
hotchemi: currently there's no dependency of unit test in `react-native-aztec` but if ok I'd like to install jest and write an unit test for the function🙇

android/app/src/main/AndroidManifest.xml
hypest: 👋 @hotchemi , does this change make the demo app differ in behavior between Android and iOS?

I'm thinking that making the demo app behave closer to WPAndroid might not be a desirable goal. Instead, the demo app should be as simple as possible and ideally not hide any "mistakes" or "hacks" WPAndroid implements. Also, ideally the Android and iOS implementations of the demo app should be as identical as possible.

react-native-aztec/src/utils.js
hypest: Let's hold on a bit while we're still assessing the solution in this PR.

While at it though, what do you have in mind for how such a unit test would be run? Will it become part of the Jest testsuite already in place in gutenberg-mobile?

react-native-aztec/src/utils.js
```diff
@@ -0,0 +1,12 @@
...
+	if ( Platform.OS === 'ios' ) {
+		return fontSize ? fontSize : undefined; // to not to affect to iOS
+	}
+	const baseFontSize = 18;
```
daniloercoli: Should we use the default value set in RN for this? https://github.com/facebook/react-native/blob/1151c096dab17e5d9a6ac05b61aacecd4305f3db/ReactAndroid/src/main/java/com/facebook/react/uimanager/ViewDefaults.java#L15

hotchemi: This is a bit tricky but I did reverse-engineering to find out 18 is the base fontSize for paragraph block. If we use 14 you'll see the base font size is smaller than before.

| no fontSize specify | baseFontSize=18 |
| ----------- | ----------- |
|||

But yea it's too tricky...

hotchemi:
@daniloercoli @hypest Thank you for your review🙇

After opening PR I'm kind of wondering this is too workaround to merge to upstream..🤔Yes we can mimic plausible behavior with this change but I'm kind of skeptical this change is maintainable enough for team.

To resolve the issue fundamentally I think we should remove `setRetainInstance(true)` from GutenbergContainerFragment(and prepare something for screen orientation in other way) but what do you think?🤔If the issue is really crucial for user we need a quick workaround but seems it's not in this case. Or let me know if there're other ideas to deal with the issue🙇

hotchemi:
Seems like this PR is not a fundamental solution to address the issue, let me close for the time being 👋

PR #804, opened 2019-04-02
Revert "Revert "Include UI tests for blocks with Appium""

JavonDavis:
Reverts wordpress-mobile/gutenberg-mobile#803


jtreanor:
Jitpack is now working for this PR: https://jitpack.io/com/github/wordpress-mobile/gutenberg-mobile/f0692bc466/build.log

.circleci/config.yml
```diff
@@ -47,6 +77,87 @@ jobs:
...
+          command: |
+            git submodule init
+            git submodule update --recursive
+      - yarn-install
+      - run:
+          name: Accept Licenses
+          command: |
+            yes | sdkmanager --licenses > /dev/null
+            sleep 10
+            sdkmanager --update
```
jtreanor: Is this step still needed? Looking at the build log it doesn't seem to do anything.

.circleci/config.yml
```diff
@@ -47,6 +77,87 @@ jobs:
...
+    macos:
+      xcode: "10.0"
+    steps:
+    - checkout
+    - run:
+        name: Checkout Gutenberg
+        command: |
+          git submodule init
+          git submodule update --recursive
+    - run: yarn clean:install
```
jtreanor: Why can't we use the `yarn-install` command here with caching? `yarn clean:install` is quite slow.

.circleci/config.yml
```diff
@@ -47,6 +77,87 @@ jobs:
...
+        - react-native-aztec/ios/Carthage
+        - ~/.rncache
+    - build-ios
+    - run:
+        name: Zip up .app file
+        command: |
+          cp ./ios/main.jsbundle ./ios/build/Gutenberg/Build/Products/Release-iphonesimulator/gutenberg.app/main.jsbundle
+          cd ./ios/build/Gutenberg/Build/Products/Release-iphonesimulator/
+          zip -r ./Gutenberg.app.zip ./gutenberg.app
+          mv ./Gutenberg.app.zip ../../../../../../Gutenberg.app.zip
```
jtreanor: Instead of the several `../` here we can change this to:
```
mv ./Gutenberg.app.zip "$CIRCLE_WORKING_DIRECTORY/Gutenberg.app.zip"
```

.circleci/config.yml
```diff
@@ -47,6 +77,87 @@ jobs:
...
+          cd ../../../../../../
```
jtreanor: This has no effect since the next command always starts with a new shell in `$CIRCLE_WORKING_DIRECTORY`

.circleci/config.yml
```diff
@@ -47,6 +77,87 @@ jobs:
...
+  android-device-checks:
+    docker:
+    - image: circleci/android:api-27-node8-alpha
+    steps:
+      - checkout
+      - run:
+          name: Checkout Gutenberg
+          command: |
+            git submodule init
+            git submodule update --recursive
```
jtreanor: Can we extract this into a command since its repeated a few times? Something like this would work:

```
full-checkout:
  steps:
    - checkout
    - run:
        name: Checkout Gutenberg
        command: |
          git submodule init
          git submodule update --recursive
```

hypest: Using `git submodule update --init --recursive` directly should be fine too, right?

jtreanor: Yes, that would reduce the two lines to one, but I think its still useful to extract so that we don't repeat it in the config 3 times.

README.md
```diff
@@ -107,6 +107,23 @@ Then, open `chrome://inspect` in Chrome to attach the debugger (look into the "R
 
 This project is set up to use [jest](https://facebook.github.io/jest/) for tests. You can configure whatever testing strategy you like, but jest works out of the box. Create test files in directories called `__tests__` or with the `.test.js` extension to have the files loaded by jest. See an example test [here](https://github.com/wordpress-mobile/gutenberg-mobile/blob/develop/src/app/App.test.js). The [jest documentation](https://facebook.github.io/jest/docs/en/getting-started.html) is also a wonderful resource, as is the [React Native testing tutorial](https://facebook.github.io/jest/docs/en/tutorial-react-native.html).
 
+## UI Tests
+
+This repository uses appium to run UI tests. The tests live in `__device-tests__` and are written using Appium to run tests against simulators and real devices. To run these you'll need to check off a few things:
```
hypest: Nit: it seems that both lower-case and upper-case first letter `Appium` is found in this sentence. Perhaps we can use only one or the other unless they refer to different things. WDYT?

README.md
```diff
@@ -107,6 +107,23 @@ Then, open `chrome://inspect` in Chrome to attach the debugger (look into the "R
 
 This project is set up to use [jest](https://facebook.github.io/jest/) for tests. You can configure whatever testing strategy you like, but jest works out of the box. Create test files in directories called `__tests__` or with the `.test.js` extension to have the files loaded by jest. See an example test [here](https://github.com/wordpress-mobile/gutenberg-mobile/blob/develop/src/app/App.test.js). The [jest documentation](https://facebook.github.io/jest/docs/en/getting-started.html) is also a wonderful resource, as is the [React Native testing tutorial](https://facebook.github.io/jest/docs/en/tutorial-react-native.html).
 
+## UI Tests
+
+This repository uses appium to run UI tests. The tests live in `__device-tests__` and are written using Appium to run tests against simulators and real devices. To run these you'll need to check off a few things: 
+
+* For now you'll need run `yarn start`, and then either `yarn ios` or `yarn android` at least once before trying to run the tests on the respective platform
```
hypest: I wonder if we still need the `run-yarn-start-at-least-once` manual step. Do we? This PR includes scripts for CI and apparently things are done automatically there so, I wonder about the manual step in this case.

README.md
```diff
@@ -107,6 +107,23 @@ Then, open `chrome://inspect` in Chrome to attach the debugger (look into the "R
...
+* [Appium cli](https://github.com/appium/appium/blob/master/docs/en/about-appium/getting-started.md) installed and available globally, I'd also recommend using [appium doctor](https://github.com/appium/appium-doctor) to ensure all of appium's dependencies are good to go. You don't have to worry about starting the server yourself, the tests handle starting the server on port 4728, just be sure that the port is free or feel free to change the port number in the test file. 
+* For iOS a simulator should automatically launch but for Android you'll need to have an emulator fired up and running.
+
+After those are checked off to run the UI tests on ios run
```
hypest: Another nit but, let's align all references to iOS to be named `iOS` (capitalized O and S). Similar for Android where the first letter is capitalized.

README.md
```diff
@@ -107,6 +107,23 @@ Then, open `chrome://inspect` in Chrome to attach the debugger (look into the "R
...
+
+After those are checked off to run the UI tests on ios run 
+
+`yarn test:ui:ios`
+
+and for android run, 
+
+`yarn test:ui:android`
+
+Alternatively if you're experiencing problems that seem to be related to the tests starting the appium server, for example errors that say `Connection Refused`, `Connection Reset` or `The requested environment is not available`. Sorry about that this is still a WIP, you can manually start the appium server via [appium desktop](https://github.com/appium/appium-desktop) or the cli, then change the port number in the tests while commenting out the `beforeAll` and `afterAll` block.
```
hypest: Syntactic nit: I see the `if you're...` but I think there's nothing for the "then" of this sentence.

.circleci/config.yml
JavonDavis: It was failing to run the gradle command if I didn't include it, I'm not sure if it's still needed at this point, I'll give it a go though

.circleci/config.yml
JavonDavis: I included this when I was trying to get rid of the architecture error I had mentioned, 🤔not sure if it's necessary now, I'll remove and see

__device-tests__/helpers/caps.js
```diff
@@ -0,0 +1,25 @@
+exports.ios12 = {
+	browserName: '',
+	platformName: 'iOS',
+	platformVersion: '12.0',
+	deviceName: 'iPhone Simulator',
+	os: 'iOS',
+	deviceOrientation: 'portrait',
+	automationName: 'XCUITest',
+	appiumVersion: '1.9.1',
```
hypest: Not something that we need to address in this PR but, I wonder if we can have a way to match the version number here with the one in `package.json` so they are kept in sync, or maybe some warning raised if they don't match.

.circleci/config.yml
JavonDavis: 🤦🏽‍♂️Yeah you're right I'll change this

.circleci/config.yml
JavonDavis: 👍🏽

README.md
JavonDavis: They're the same. I'll go with the title case 🙂

README.md
JavonDavis: Locally for now yes... it uses the debug build locally and the release build in CI... Debug build isn't bundled with jsbundle and as a result, requires those steps.

README.md
JavonDavis: Sure I'll make the change

__device-tests__/pages/editor-page.js
```diff
@@ -0,0 +1,44 @@
...
+ * Internal dependencies
+ */
+import BlockInteraction from '../blocks/block-interaction';
+
+/**
+ * WordPress dependencies
+ */
+import { __ } from '@wordpress/i18n';
+
+export default class EditorPage {
```
hypest: Not sure what "Page" means in this context. Are you perhaps referring to a "screen" in the sense of multiple screens changing as the app moves from one to the other?

I think i'm a bit confused because "page" has special meaning within WordPress itself.

README.md
JavonDavis: Hmm this so instead of a full stop then new sentence more of a 

> then you can manually start...

?

bin/tmp/initial-device-tests-html.js
```diff
@@ -0,0 +1,6 @@
+/**
```
hypest: If I understand it correctly, the files in the `bin/tmp` folder are not exactly temporary or anything. Instead, they are the different source files to use to overwrite the `src/app/initial-html.js` module. This is nit but, can we rename the `tmp` folder to something else, maybe `example-content` or similar?

bin/load-initial-html.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+
+
+if [[ "${DEVICE_TESTS}" = true ]] ; then
+    cp ./bin/tmp/initial-device-tests-html.js ./src/app/initial-html.js
```
hypest: You probably have tried it in the past but, will ask just to be sure: instead of copying/overwriting files, have you perhaps tried using the `process.env.DEVICE_TESTS` environment variable directly in `src/app/initial-html.js` to export the desired html content?

src/app/App.js
```diff
@@ -29,6 +29,7 @@ export default class AppProvider extends React.Component<PropsType> {
 		if ( initialTitle === undefined ) {
 			initialTitle = 'Welcome to Gutenberg!';
 		}
+
```
hypest: Nit: let's remove this empty line diff. Also the same in the `App.test.js` file.

src/block-management/block-holder.js
```diff
@@ -199,12 +200,24 @@ export class BlockHolder extends React.Component<PropsType, StateType> {
 		const borderColor = isSelected ? focusedBorderColor : 'transparent';
 
 		return (
-			<TouchableWithoutFeedback onPress={ this.onFocus } >
+			// Needs to be false to access children
```
hypest: I'm assuming (and it's also deduced by the link provided) that you're referring to the `accessible` prop here so, let's name/specify it in the comment.

__device-tests__/helpers/caps.js
JavonDavis: Right, I've noted this... The need for the Appium version being specified is actually a result of the requirement of SauceLabs, otherwise there'd be no need and would just pick it up automatically.

__device-tests__/pages/editor-page.js
JavonDavis: Hmm it really is just referencing the only page in the Sample app with the Gutenberg Editor, figured it EditorPage was a good name but there can be some confusion based on the references within WordPress itself... Do you think we need to change this?

bin/tmp/initial-device-tests-html.js
JavonDavis: That's correct, we can rename it to a more appropriate name

bin/load-initial-html.sh
JavonDavis: Yes I have, the environment variable won't be available when the app is being bundled. I've also tried using [babel-plugin-transform-inline-environment-variables](https://www.npmjs.com/package/babel-plugin-transform-inline-environment-variables) but this also doesn't work during bundling for the release version.

src/app/App.js
JavonDavis: 👍

src/block-management/block-holder.js
JavonDavis: Yes that's right, I can mention it in the comment

src/block-management/block-holder.js
```diff
@@ -199,12 +200,24 @@ export class BlockHolder extends React.Component<PropsType, StateType> {
...
+			// https://facebook.github.io/react-native/docs/accessibility#accessible-ios-android
+			<TouchableWithoutFeedback
+				accessible={ false }
```
hypest: I'm still confused by the need to set `accessible={ false }` on any element. I know why it's done here but, it feels that we're doing something wrong with our element hierarchies. @JavonDavis can you open a ticket to revisit all instances of `accessible={ false }`?

README.md
hypest: I see, thanks.

No need to be part of this PR but, I wonder if trying to build a debug bundle and use that locally for device tests is a more robust strategy. Manual steps that only need to be done in cases that's not exactly clear when, tend to become gotchas and lead to a frustrating dev experience. Would be similar to the Jest tests anyway, since those don't use the running Metro instance either. Am I perhaps missing some other detail here?

src/block-management/block-holder.js
JavonDavis: I can understand the confusion... When creating custom elements that combine many child elements _it makes sense_ to set it to false in the sense that the default behaviour is once the element is accessible is true iOS will treat that element as a single thing vs if you needed the children to be accessible individually you'd need to counter that default behaviour

> can you open a ticket to revisit all instances of accessible={ false }?

Sure no problem!

__device-tests__/helpers/caps.js
hypest: > the Appium version being specified is actually a result of the requirement of SauceLabs

Oh, interesting. Let's add a line comment about this, it's quite a requirement apparently.

README.md
hypest: Ah, I see. Sorry, my coder's brain keeps wanting to see the "then" in the same sentence as the "if" 😊

May I suggest this revision: `Note, you might experience problems...`, so there is no if anyway?

__device-tests__/pages/editor-page.js
hypest: Cool, I see, thanks! I'd say there's no need to change the name 👍

bin/load-initial-html.sh
hypest: Interesting. I think I'd like to dig into that to understand how come the bundling doesn't pick up env variables. Nothing to block this PR though 👍

README.md
JavonDavis: Hmm I'm not sure but I don't think that bundle would be rebuilt when changes are made(?), which I think is a plus when developing the tests locally to not have to rebundle after small changes and allow for the hot reloading to update the app and simply run the test again after the change, I mean we could have it as a pre-step but then that adds a little more time to running locally.. If I'm wrong about my first statement then sure this could work well

bin/load-initial-html.sh
JavonDavis: Yeah it works when specifying the environment variables on `yarn start` so for example `VAR=foo yarn start` but otherwise no luck, and I basically spent a day trying to find a way to accomplish this but I didn't want to spend too much time on it right now. I did plan to circle back to it as well too so for sure let me know if you're able to come up with something here.

README.md
hypest: > I think is a plus when developing the tests locally to not have to rebundle after small changes and allow for the hot reloading to update the app

Fair point and one I can totally relate to.

Ultimately, my main concern here is the manual step to install the native binary. How about always building and installing the binary (effectively, always running `yarn android` or `yarn ios`) so any changes to the native side can be picked up too? To be clear, this is not something to be tacked necessarily in this PR, it's not a blocker.

README.md
hypest: Seems addressed with 8f6b24ef43eb613ff300b0ad228411e3fc40041a.

README.md
hypest: Seems addressed by 8f6b24ef43eb613ff300b0ad228411e3fc40041a.

README.md
hypest: Seems addressed by 8f6b24ef43eb613ff300b0ad228411e3fc40041a.

__device-tests__/helpers/caps.js
hypest: Seems addressed by 8f6b24ef43eb613ff300b0ad228411e3fc40041a.

bin/tmp/initial-device-tests-html.js
hypest: Seems addressed by 132202a7f809f14eef26ab6e0ba2f33da72a0e19.

src/app/App.js
hypest: Seems addressed by 7bd3b0b76a5d17a6d023467b07fb687d0f605191.

.circleci/config.yml
jtreanor: This change seems to have been reverted somehow 🤔

JavonDavis:
@etoledom I had to remove the hint from the toolbarbuttons. It seems on Android this hint is concatenated with the title to make up the content description attribute of the view... 

WDYT we should do here?

* leave the title as is
* replace the title with the same text that was in the hint
* in the test search for the concatenated title and hint

etoledom:
Oh no! Not having hints it's quite bad :(

I did notice on Android that the hints were read way too fast, and your explanation makes perfect sense with the actual behavior.

In general (or better said, on my experience with iOS) is super important the correctness of hints, where and where not to put them and what to write on them. Even though manually appending the hint to the title won't change the behavior on Android, on iOS it would be better to not have hints.

In my honest opinion, I wouldn't like to have to compromise the accessibility implementation in favor of UITests. I believe that we need to give the user the best experience possible, specially in this case.

But on this early stage, for sure we can try solution to make it work for both.

> leave the title as is

Do you mean removing the hint? 
If that's the case, this would be a last resource solution if a UITest level solution is not possible.

> replace the title with the same text that was in the hint

I wouldn't recommend this, since on iOS would be a bit chaotic.

> in the test search for the concatenated title and hint

Not sure if this will work, since on iOS the title still would be without hint, and just on Android with the hint appended (if I understood correctly). But if we can find a UITest level solution, I would go for that one.

src/block-management/block-holder.js
hypest: Seems addressed by 7bd3b0b76a5d17a6d023467b07fb687d0f605191.

JavonDavis:
> Not sure if this will work, since on iOS the title still would be without hint, and just on Android with the hint appended (if I understood correctly). But if we can find a UITest level solution, I would go for that one.

@etoledom we can search for a platform-specific identifier, I'll push an update with this to show how it could look

PR #812, opened 2019-04-03
e2e test improvements

Tug:
This PR brings some improvements to our Appium/e2e tests.
It bundles a test version of the app that it uses for testing instead of relying on a metro server running.
Doing so we are closer to a production build and we don't need a hack on the initial html part. We can just rely on the `__DEV__` global variable that RN has.
It also does a minor refactoring on how the test suite is run and remove one more dependency with an external script, `appium`, by installing it as a dev dependency and running it as an attached child process.


__device-tests__/helpers/utils.js
```diff
@@ -56,6 +64,18 @@ const setupDriver = async () => {
 		desiredCaps = _.clone( android8 );
 		if ( isLocalEnvironment() ) {
 			desiredCaps.app = localAndroidAppPath;
+			try {
+				const androidVersion = childProcess
+					.execSync( 'adb shell getprop ro.build.version.release' )
+					.toString()
+					.replace( /^\s+|\s+$/g, '' );
+				if ( ! isNaN( androidVersion ) ) {
+					desiredCaps.platformVersion = androidVersion;
```
JavonDavis: What's the result of this if there are multiple running Android devices with different versions? Do you think it's a good idea to accept maybe an ENV variable for this? I was also looking into the use of the `emulator` cli for firing up the Android emulator if it's not running?  WDYT?

src/app/App.js
```diff
@@ -23,7 +23,7 @@ export default class AppProvider extends React.Component<PropsType> {
 		const { initialHtmlModeEnabled } = this.props;
 		let initialData = this.props.initialData;
 		let initialTitle = this.props.initialTitle;
-		if ( initialData === undefined ) {
+		if ( initialData === undefined && __DEV__ ) {
```
JavonDavis: @hypest seems this does the trick

__device-tests__/helpers/utils.js
JavonDavis: Also another thing to note is that the specification of the Android platform version is more a requirement for the CI environment, locally if this isn't specified it'll just use the first running one it picks up

Tug: So you mean we could just clear it?
```
desiredCaps.platformVersion = undefined;
```

JavonDavis: Well, I think that'll still have the key in there and I'm not actually sure how it'd respond to that. `delete desiredCaps.platformVersion` would be better if we want it to just pick up any running emulator.

src/app/App.js
Tug: Yep, so just to be clear, doing so we disable running metro during our e2e tests and instead bundle a "production" version of the app (without minification to understand the stacktrace) that we install on our device when running local test (using `yarn test:e2e:android:local` for instance) or remotely on saucelabs using the circle ci script and running `yarn test:e2e:android`.

Tug:
Should be good to go now, @JavonDavis can you give it a go :)?

JavonDavis:
@Tug can you push an update to the README as well?

JavonDavis:
@Tug running locally with `test:e2e:ios:local` also fires up the app with the sample blocks still loaded... Was there a step I needed to do before?

__device-tests__/helpers/utils.js
```diff
@@ -28,15 +27,15 @@ const testEnvironment = process.env.TEST_ENV || defaultEnvironment;
 
 // Local App Paths
 const defaultAndroidAppPath = './android/app/build/outputs/apk/debug/app-debug.apk';
-const defaultIOSAppPath = './ios/build/gutenberg/Build/Products/Debug-iphonesimulator/gutenberg.app';
+const defaultIOSAppPath = './ios/build/gutenberg/Build/Products/Release-iphonesimulator/gutenberg.app';
```
JavonDavis: Changed to the Release folder since that's where the app is being built now for local runs as well

JavonDavis:
@Tug I'm just noticing the zip command in the build:ios step doesn't produce the zip you'd expect and as a result I think it's leading to some error on SauceLabs... When you unzip the file you'll see the  Gutenberg.app as expected but when that's unpackaged the contents are incorrect.

Issue #783, opened 2019-03-26
"Add new block" fails after first use


mkevins:
The `+` button to add new block works once, but does not work on subsequent button presses. This issue was discovered on the Android Emulator, and has not been reproduced on a real device (attempts were made, but the issue was not encountered).

**Steps to reproduce:**
* Press the `+` button and choose new paragraph block
* Write some text
* Press the `+` button again and choose a new paragraph block

**Result:**
A new paragraph block is not created. The focus changes to an already existing block.

**Screenshot:**
<img src="https://user-images.githubusercontent.com/8507675/54969087-b463be80-4fc8-11e9-9dd8-a80df264217f.gif" width="360">

mkevins:
After testing this issue further, I have discovered that the `+` button to create a new paragraph block fails when a Heading or Paragraph block is focused, but succeeds when a Page Break block or Code block is focused.

Also, it seems that using `+` to add either a Page Break block, or a Code block, succeeds, regardless of which block type had focus when tapping `+`.

Some other discoveries:
* If a heading block is created, with no text, and focus is changed to another block, the heading block remains. I'm not certain this is the desired behavior.
* When a Page Break block is created while a textual (Heading, Paragraph, or Code) block has focus, the Page Break block becomes the active block, but a (blinking) caret remains in the previously focused textual block, but the keyboard becomes hidden. I'm not certain this is the desired behavior, but if so, I think the keyboard should remain visible. Presently, it appears as though both blocks have focus simultaneously.

mkevins:
The recent upgrade (React Native to 0.59, React to 16.8.3) seems to have a source map issue, which has made it difficult to use the JS remote debugger. I am temporarily resorting to a bisection method to try to isolate the changeset that may have introduced this issue. I will test this issue manually using the following steps:

1. Place the cursor at the end of the "**Hello World!" block
2. Press the `+` button and add a new Paragraph block
3. Observe success or failure
4. Mark hashes / tags below with a [*] for success, and [ ] for failure

**Progress:**
- [ ] 2f145b

daniloercoli:
Fixed in https://github.com/wordpress-mobile/gutenberg-mobile/pull/857

Issue #745, opened 2019-03-13
Include UI tests for blocks through Sample App


JavonDavis:
We'll need to include an on-device testing solution to help check that everything's good to go as we make changes and include new blocks. I think once we have the following in place we'll be at a good place to start. 

## CI/CD

- [x] UI testing framework selection  #676
- [x] UI tests set to run in CI on PRs  #676

## Paragraph block

- [x] Create a short post with a single Paragraph block  #676
- [x] Create a long post with multiple Paragraph blocks #821 
- [x] Have a paragraph block with a word in it. Place the cursor in the middle of the work and tap "Enter". Verify that the block split into two paragraph blocks, each with the respective half word. #821 
- [x] Have two paragraph blocks with some content in each. Place the cursor at the beginning of the second and tap "Backspace". Verify that the blocks merged. #821 

## Image block

- [x] Create a post with a captioned Image block #1060
- [x] Create a post with a captioned Image block followed by multiple paragraph blocks #1060

## Heading block

- [x] Create a post with a heading and multiple paragraph blocks #1069 

## Inserting new blocks

- [x] Have a post with more than a few blocks. Focus the title block and insert a new block via the Inserter. Verify that the block was added at the end of the post and the view scrolled to it.(This does not happen if the post contains a heading block) #1139 
- [x] Create a post with a few blocks. Focus on the first block and insert a new block via the inserter. Verify that the block appears in its correct position in the post #1139 

I think if we can check off all of these we'll be in a good place to start here. 

--------

Additional coverage currently includes

- [x]  Copy/paste on plain and rich text content in a paragraph block
- [x] Rotating device while adding/removing blocks

hypest:
Thanks for compiling a list @JavonDavis !

Let's add some writing flow related items. Those are actually the things that are not easy to test in Jest (no access to Aztec itself with Jest) so it's crucial to add some as baseline to build the testsuite on.

For example:
- [ ] Have a paragraph block with a word in it. Place the cursor in the middle of the work and tap "Enter". Verify that the block split in two paragraph blocks, each with the respective half word.
- [ ] Have two paragraph blocks with some content in each. Place the cursor at the beginning of the second and tap "Backspace". Verify that the blocks merged.
- [ ] Have a post with more than a few blocks. Focus the title block and insert a new block via the Inserter. Verify that the block was added at the end of the post and the view scrolled to it.

What do you think?

JavonDavis:
[...]

PR #821, opened 2019-04-04
Add paragraph block writing flow tests

JavonDavis:
Part of #745 

This PR adds the following tests covering writing flows with the Paragraph block

- Create a long post with a multiple Paragraph blocks
- Have a paragraph block with a word in it. Place the cursor in the middle of the work and tap "Enter". Verify that the block split in two paragraph blocks, each with the respective half word.
- Have two paragraph blocks with some content in each. Place the cursor at the beginning of the second and tap "Backspace". Verify that the blocks merged.

Additional notes
-  ~~Tests on Android failing due to #783~~
- The PR removes the `BlockInteraction`s and uses functions from the editor-page class to interact with blocks on the editor screen
- Merged in improvements from https://github.com/wordpress-mobile/gutenberg-mobile/pull/812


src/block-management/inline-toolbar/index.js
```diff
@@ -52,18 +53,18 @@ export default class InlineToolbar extends React.Component<PropsType> {
 	}
 
 	render() {
+		const { testID } = this.props;
 		return (
 			<View style={ styles.toolbar } >
 				<ToolbarButton
-					accessibilityLabel={ __( 'Move up' ) }
-					label={ __( 'Move up' ) }
+					title={ __( `Move ${ testID } up` ) }
```
etoledom: Hey @JavonDavis !
These titles will be spoken out loud by VoiceOver. Probably is not the best to have a `testID` here 😞

JavonDavis: > though it's still not clear enough for a translator without a hint.

That's why I suggested including the block name as the variable in there but I also thought that we translated these. But we don't https://github.com/wordpress-mobile/gutenberg-mobile/blob/develop/src/block-management/block-picker.js#L70 ... Unless we get it translated here https://github.com/wordpress-mobile/gutenberg-mobile/blob/develop/src/block-management/block-picker.js#L34... 

Hmm does something more like `Paragraph block 1. Move up` or `Paragraph block 1. Move up to position 0` In this case it's two sentences and I'm not sure if I got this wrong but maybe this will be easier to translate? I think it's good for both VoiceOver and the tests if the name of the block is indicated in this identifier right? A user can be sure what block is there and being moved up or down or removed. This is just my thinking though, I'm not sure and we don't want to include anything that complicates the VoiceOver experience.

etoledom: Getting good labels for VoiceOver is not easy 😅
The best is to experience it by ourselves to have a general idea of what experience a VoiceOver user might have (including testing using VoiceOver blindfolded).

One important thing is to have the most meaningful part that describes the element at the beginning, so the user can move to the next one fast if that is not the element they are looking for.

So far I'm adding the text `Paragraph block" to describe the block itself, so if we start with the same string the Up button, the user will think that it is a new paragraph block already (unless they wait long enough to notice the difference).

 `Move up to row 0` sounds like the best to me so far. The current index is not really needed for VoiceOver and it would be an unnecessary extra load. Bit it seems to be necessary for UITest, right?
How does it sound something like: `Move up from row 2 to row 1`?

JavonDavis: It's not that the current index is needed because that can still be retrieved by looking after the from and before the to. I was just trying to make it easier to write selectors for the buttons but your point here, 

> One important thing is to have the most meaningful part that describes the element at the beginning, so the user can move to the next one fast if that is not the element they are looking for.

makes sense. There'll just be a more complicated selector here for finding the buttons, basically looking at the child and ensuring the numbers match, but definitely doable.

etoledom: Thank you for the effort!

rachelmcr: > What do you think? Maybe @rachelmcr could help us too :)

Sorry, I missed this ping earlier! I just want to note that for translation it's always better to use variables within a string (with a comment to the translators to explain what the variable contains), rather than string concatenation, so that translators can easily move a variable within the string to account for word order differences between languages.

JavonDavis:
@etoledom  I've gone ahead and updated the accessibility IDs based on what we discussed, can you have a look when you get a chance and let me know what you think? Thanks!

__device-tests__/helpers/utils.js
```diff
@@ -118,10 +126,92 @@ const stopDriver = async ( driver: wd.PromiseChainWebdriver ) => {
...
 
+// attempts to type a string to a given element, need for this stems from
+// https://github.com/appium/appium/issues/12285#issuecomment-471872239
+// https://github.com/facebook/WebDriverAgent/issues/1084
+const typeString = async ( driver: wd.PromiseChainWebdriver, element: wd.PromiseChainWebdriver.Element, str: string, clear: boolean = false ) => {
+	if ( clear ) {
+		await element.clear();
+	}
+
+	if ( isAndroid() ) {
```
JavonDavis: The need for this additional work for Android is because of how the type API works. It basically clears the field first before start typing. Although Appium implemented a workaround in their additional implementation on top of this to allow appending by copying the existing text before typing the new one, unfortunately, it still performs the 'clear' operation under the hood. Because of that when it clears it, it doesn't know when to stop properly and removes the block, as if the user had touched the backspace button on an empty block. 

The workaround uses `adb shell input <text>` instead to send the text.

__device-tests__/helpers/utils.js
```diff
@@ -118,10 +126,92 @@ const stopDriver = async ( driver: wd.PromiseChainWebdriver ) => {
...
+		if ( str in strToKeycode ) {
+			return await driver.pressKeycode( strToKeycode[ str ] );
```
Tug: We can probably remove this part since it will be handled by the paragraph logic (since `'\n'.split('\n')` gives us `["", ""]`)

JavonDavis: Hmm I think it wouldn't be handled in the paragraph logic for the case the string is `'\u0008'`. The split would cause the paragraphs length to be equal to 1 and then try to execute it like regular text instead of pressing the keycode. So this statement is just for the case is a single command, alternatively if the if statement is changed to `<=` it might handle it though right? 🤔

__device-tests__/helpers/utils.js
```diff
@@ -118,10 +126,92 @@ const stopDriver = async ( driver: wd.PromiseChainWebdriver ) => {
...
+	await action.perform();
+};
+
+// Starts from the middle of the screen or the element(if specified)
+// and swipes upwards
+const swipeUp = async ( driver: wd.PromiseChainWebdriver, element: wd.PromiseChainWebdriver.Element = undefined ) => {
+	let size = await driver.getWindowSize();
+	let y = 0;
+	if ( element !== undefined ) {
+		size = await element.getSize();
```
Tug: I feel this function could use an upgrade (maybe in another PR?) It seems to me that there are quite few edge cases where this approach would fail:
- What if the `element` height is bigger than 3 times the height of the screen
- What if `element` is not given and a single swipe is not enough
- What if the screen is only showing the top of `element`? `startY ` might be outside the screen boudaries. Also a swipe might not be enough to reveal the bottom of the element in this case

I think a "slightly" better approach would be to make this function a simple swipe up with an height parameter, it will try to center the swipe on the screen and do it in a loop if height is bigger than the size of the screen. We can also think of a `swipeUpUntilElementIsVIsible()` version which would do what its name says.

JavonDavis: For sure! I was saving improving this for another PR, do you think it's good enough for now? 

>  it will try to center the swipe on the screen

It originally just did this, but failed on the CI because the device it used was small and the keyboard took up more than half of the screen. 

> We can also think of a swipeUpUntilElementIsVIsible() version which would do what its name says.

Yep, for now I was hoping this was enough to use in a loop if needed before that get's implemented.

JavonDavis: Like what's done hear, https://github.com/wordpress-mobile/gutenberg-mobile/pull/821/files#diff-5c1289e37617de9b4adbba51cf493004R136

src/block-management/block-holder.js
```diff
@@ -189,6 +178,17 @@ export class BlockHolder extends React.Component<PropsType, StateType> {
...
 
+	getAccessibilityLabelForBlock() {
+		const { clientId, rootClientId } = this.props;
+		const order = this.props.getBlockIndex( clientId, rootClientId );
+		const name = this.props.getBlockName( clientId );
+		let blockType = getBlockType( name ).title;
+
+		blockType = blockType === 'Unrecognized Block' ? blockType : blockType + ' Block';
+
+		return `${ blockType }. Row ${ order + 1 }.`; // Use one indexing for better accessibility
```
Tug: Correct me if I'm wrong but I believe we want internationalization for the accessibility label. we can use `sprintf` from `@wordpress/i18n` to add variables:
```js
let blockTitle = getBlockType( name ).title;
blockTitle = blockTitle === __( 'Unrecognized Block' ) ? blockTitle : sprintf( __( '%s Block' ), blockTitle );
return sprintf( __( '%1$s. Row %2$d.' ), blockTitle, order + 1 );
```

src/block-management/block-holder.js
```diff
@@ -162,12 +150,13 @@ export class BlockHolder extends React.Component<PropsType, StateType> {
...
 		}
 
+		const order = this.props.getBlockIndex( this.props.clientId, this.props.rootClientId );
 		return (
 			<InlineToolbar
-				clientId={ this.props.clientId }
 				onButtonPressed={ this.onInlineToolbarButtonPressed }
 				canMoveUp={ ! this.props.isFirstBlock }
 				canMoveDown={ ! this.props.isLastBlock }
+				rowIndex={ order }
```
Tug: This is not necessary, we can connect `InlineToolbar` to the block-editor store to retrieve the block order:
```js
withSelect( ( select, { clientId } ) => {
	const {
		getBlockOrder,
	} = select( 'core/block-editor' );

	return {
		order: getBlockOrder( clientId ),
	};
} )( InlineToolbar );
```

src/block-management/inline-toolbar/index.js
```diff
@@ -52,18 +52,31 @@ export default class InlineToolbar extends React.Component<PropsType> {
 	}
 
 	render() {
+		const { rowIndex } = this.props;
+		const rowNumber = rowIndex + 1;
+		let moveUpButtonTitle = `Move up from row ${ rowNumber }`;
```
Tug: Same here, if we can use internationalization, it would be nice :)

__device-tests__/helpers/utils.js
Tug: Oh right, I missed the `if ( paragraphs.length > 1 ) {`!
I wonder why it's there though, is `paragraphs.length` the logic inside the loop still makes sense right?

Tug: > if the if statement is changed to <= it might handle it though right? 🤔

You mean `> 0` ? In that case yes :)

__device-tests__/helpers/utils.js
Tug: Yep it seems to work really smoothly for now (at least on the devices we tested :)). So I think we should be good 👍

src/block-management/block-holder.js
JavonDavis: > Correct me if I'm wrong but I believe we want internationalization for the accessibility label. 

Yep! You're right about this I'll push an update...

Hmm are you saying there's a sprintf function I can import like `import {sprintf} from '@wordpress/i18n'` or just `import sprintf from 'sprintf-js'`?

I take it for this to work I'll need to update the identifier in the tests as well to get that right but it's weird just changing it out locally doesn't work(which I'd expect it to since device language is English) or that's expected? 
Also is there a reason why `__( `${ blockTitle }. Row ${ order + 1 }.` );` wouldn't work?

__device-tests__/helpers/utils.js
JavonDavis: > You mean > 0 ? In that case yes :)

🤦🏽‍♂️ 😅 Yes that!

src/block-management/inline-toolbar/index.js
etoledom: One bad thing is that here we can't write a comment for the string, so it gets a little bit more complicated.

JavonDavis: > One bad thing is that here we can't write a comment for the string, so it gets a little bit more complicated.

@etoledom I'm not sure I understand what you mean here

etoledom: On iOS we can add a comment that is send together with the string to translate to the translators. As far as I know, we can't send a comment like that from string created in the JavaScript side (unless we can and I didn't know?).

JavonDavis: Ohh I didn't know that actually. You might be right, I'm not very familiar with how that works right now.

__device-tests__/helpers/utils.js
JavonDavis: Actually, we can't remove this part because of the Unicode characters, like the one used for backspace '\u0008', splitting on this will return an empty string and not send the keyevent.

src/block-management/block-holder.js
JavonDavis: I changed this to the clientId and updated the InlineToolbar, this what you meant? I used `getBlockIndex` instead since `getBlockOrder` wasn't returning what I expected?

src/block-management/inline-toolbar/index.js
JavonDavis: These are internationalized using `__` in the props passed to the `ToolbarButton` is that ok?

src/block-management/block-holder.js
Tug: Yes! `import { sprintf } from '@wordpress/i18n';` would work.
`__( ${ blockTitle }. Row ${ order + 1 }. );` doesn't because xgettext (the tool that extract strings to be translated and creates POT files) need a string literal not a template for it to work.
So using `sprintf` is the recommended way for gutenberg to add variables to your strings.
See https://github.com/WordPress/gutenberg/blob/master/packages/block-editor/src/components/block-mover/mover-description.js#L34 for an exemple of usage

src/block-management/block-holder.js
Tug: Oh right, getBlockOrder returns a list of clientIds, my bad, looks good now :)

src/block-management/inline-toolbar/index.js
Tug: Yes, to be clear you need constant strings here otherwise it won't work:
```
let moveUpButtonTitle = sprintf( __( 'Move up from row %d' ), rowNumber );
```

__device-tests__/helpers/utils.js
Tug: I'm not sure I understand, `'\u0008'.split( '\n' )` returns `[ '\u0008' ]`.
Then in your loop, `paragraph in strToKeycode` will be true and it will send the right keystroke.

That said it works fine at the moment and it's not bad per se so I don't mind too much if we keep it that way

src/block-management/block-holder.js
JavonDavis: Ahh I see, thanks!

src/block-management/inline-toolbar/index.js
JavonDavis: Yeah I understand now, thanks for explaining that in the previous comment!

hypest:
Anything we can help with to move this PR forward? Not exactly sure if there's a blocker or not. Wdyt @JavonDavis ?

JavonDavis:
@hypest I was away on vacation for a few weeks so that's the big reason it hasn't moved forward in a while and since I got back this week I haven't circled back to this yet. What's left is to fix the internationalization based on @Tug 's comments and I also wanted @etoledom's feedback on the new accessibility labels for the block

src/block-management/inline-toolbar/index.js
```diff
@@ -52,18 +54,30 @@ export default class InlineToolbar extends React.Component<PropsType> {
 	}
 
 	render() {
+		const { order } = this.props;
+		let moveUpButtonTitle = `Move up from row ${ order }`;
+		if ( this.props.canMoveUp ) {
+			moveUpButtonTitle += ` to row ${ order - 1 }`;
```
etoledom: So, we have learned some bits about best practices for accessibility labels and localization.
In my opinion, this construction for labels looks good, but it's not localized.

Check this for an example how are we localizing accessibility labels:
https://github.com/WordPress/gutenberg/blob/f257ed207fa7cb1dc28e4b1f277af000240fe992/packages/block-library/src/image/edit.native.js#L259-L264

Note the use of `sprintf` for formatted strings, the `/* translators: ...` comment (that will be readable by the translator in GlotPress), and that the whole sentence is written together to give the translator the best context.

JavonDavis: Yeah thanks @etoledom totally going to do this, I'm having trouble resolving issues with the merged in changes from develop, taking longer than I expected so that's why I haven't pushed that update yet... as soon as I push it I'll ping for the review...

JavonDavis: @etoledom I made the update in ca0a861, WDYT? 
cc @Tug

etoledom: Hey @JavonDavis !

I think is better to not append strings, since for the translators, the string ` __( ' to row %d' )` will be lacking some context. Beside there are no comments to help them.

I was thinking on something more like this: https://gist.github.com/etoledom/7162a547141490f76c885c0ecd7792ed

This proposed one in the gist looks a bit messy, but we ensure that the whole string gets to the translators and we add a comment that they can read via `/* translators: ...`.

Of course I'm open to suggestions and to improve this proposal too :)

JavonDavis: Ohh I see, I didn't consider that there'd be more context needed around this... Thanks! Your proposed change I think covers it well

JavonDavis: @etoledom made that update in dfe8851

src/block-management/inline-toolbar/index.js
```diff
@@ -52,19 +54,42 @@ export default class InlineToolbar extends React.Component<PropsType> {
...
 
 	render() {
+		const { order } = this.props;
+		const moveUpButtonTitle = this.props.canMoveUp ? sprintf(
+			/* translators: accessibility text. %1: current block position (number). %2: next block position (number) */
+			__( 'Move up from row %1$s to row %2$s' ),
+			order,
+			order - 1 ) : sprintf(
+			/* translators: accessibility text. %d: block position (number) */
+			__( 'Move up from row %1$s' ),
```
etoledom: It would be great if we can say here just `Move up`, for this case where the block can't be moved up.
The system will add a `dimmed | disabled` to let the user know that the action is disabled. And the user will knowledge the intent of the button.

src/block-management/inline-toolbar/index.js
```diff
@@ -52,19 +54,42 @@ export default class InlineToolbar extends React.Component<PropsType> {
...
+			order );
+
+		const moveDownButtonTitle = this.props.canMoveUp ? sprintf(
+			/* translators: accessibility text. %1: current block position (number). %2: next block position (number) */
+			__( 'Move down from row %1$s to row %2$s' ),
+			order,
+			order + 1 ) : sprintf(
+			/* translators: accessibility text. %d: block position (number) */
+			__( 'Move down from row %1$s' ),
```
etoledom: If we change Move up, it would be nice to change this one too. 👍

src/block-management/inline-toolbar/index.js
```diff
@@ -52,19 +54,42 @@ export default class InlineToolbar extends React.Component<PropsType> {
...
+			order );
+
+		const removeButtonTitle = sprintf(
+			/* translators: accessibility text. %1: current block position (number). */
+			__( 'Remove row %1$s' ),
```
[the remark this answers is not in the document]
JavonDavis: I think we had previously discussed taking the word block out of the text since the block name would have already been said in VoiceOver once the block is selected, WDYT?

src/block-management/inline-toolbar/index.js
JavonDavis: I think both messages are ok but, I agree removing the row number in this case this lines up better states the intent of the button when disabled and can be better for accessibility. As it relates to the tests this shouldn't have any impact so 👍

src/block-management/inline-toolbar/index.js
```diff
@@ -52,19 +54,36 @@ export default class InlineToolbar extends React.Component<PropsType> {
...
+
+		const moveDownButtonTitle = this.props.canMoveUp ? sprintf(
+			/* translators: accessibility text. %1: current block position (number). %2: next block position (number) */
+			__( 'Move block down from row %1$s to row %2$s' ),
+			order,
+			order + 1 ) : __( 'Move block down' );
+
+		const removeButtonTitle = sprintf(
+			/* translators: accessibility text. %s: current block position (number). */
+			__( 'Remove block at row row %s' ),
```
etoledom: Small typo.
```suggestion
			__( 'Remove block at row %s' ),
```

PR #835, opened 2019-04-05
don't allow setSelection on invalid range

mzorz:
Fixes the third case reported in this comment here: https://github.com/wordpress-mobile/gutenberg-mobile/issues/208#issuecomment-480246421

As expressed elsewhere, while the root cause of it may lay on how undo/redo and lists changes are working together, the extra check here should be harmless and fixes the issue for the time being.

### DIFF
--- a/react-native-aztec/android/src/main/java/org/wordpress/mobile/ReactNativeAztec/ReactAztecManager.java
+++ b/react-native-aztec/android/src/main/java/org/wordpress/mobile/ReactNativeAztec/ReactAztecManager.java
@@ -194,7 +194,11 @@ public class ReactAztecManager extends SimpleViewManager<ReactAztecText> {
         if ( selection != null ) {
             int start = selection.getInt("start");
             int end = selection.getInt("end");
-            if (view.getText().length() > end && start >= 0) {
+            // let's make sure the selection is within a valid range
+            // otherwise we'd crash here - first observed with lists
+            // and undo/redo states as expressed in
+            // https://github.com/wordpress-mobile/gutenberg-mobile/issues/208#issuecomment-480246421
+            if (view.getText().length() >= end && start >= 0) {
                 view.setSelection(start, end);
             }
         }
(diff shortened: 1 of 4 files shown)


react-native-aztec/android/src/main/java/org/wordpress/mobile/ReactNativeAztec/ReactAztecManager.java
```diff
@@ -194,7 +194,9 @@ private void updateSelectionIfNeeded(ReactAztecText view, @Nullable ReadableMap
         if ( selection != null ) {
             int start = selection.getInt("start");
             int end = selection.getInt("end");
-            view.setSelection(start, end);
+            if (view.getText().length() > end && start >= 0) {
```
hypest: I think we can add the equality to `end` in the condition since the caret can actually be placed right after the last char.

hypest: Also, let's add a line comment about why we're doing this here, wdyt?

mzorz: both addressed in 7b8955aa

mzorz:
Just changed the base branch as we won't be merging it into `release/12.1`-  we have some concerns about sync state between the native / JS side of things. We’ll continue looking into it.

hypest:
FWIW, the ticket this PR was trying to address (https://github.com/wordpress-mobile/gutenberg-mobile/issues/836) is closed now as it's not happening anymore so, we could concider closing this PR.

Though, the fix here is one that we might still consider doing if we end up not fixing the root causes of how come the RN side tries to set the selection to a position that is beyond the text length. Context: https://github.com/wordpress-mobile/gutenberg-mobile/issues/871#issuecomment-483975276

mzorz:
I think while the check looks harmless it could hide other problems so if the most urgent problem is not happening anymore I'd lean more to closing this one - also the references to this PR will still be there in case we need to grab n go with them in case we decide to not pursue the root cause further 👍

hypest:
> it could hide other problems

That's exactly what I'm struggling with at the moment 😬. This workaround works but masks all errors that stem from the RN side getting out-of-sync with Aztec. It will hide problems that we might miss or make hard to debug.

PR #840, opened 2019-04-08
Use react-native-cli with the Android debugging fix

hypest:
The [recent upgrade to React Native 0.59](https://github.com/wordpress-mobile/gutenberg-mobile/pull/741) had a regression that made us lose the ability to properly debug the JS source code under Android.

This PR introduces a workaround by using the fixes version of the `react-native-cli` package, before it becomes available directly via a future React Native release.

Relevant tickets: https://github.com/facebook/react-native/issues/23955, https://github.com/facebook/metro/issues/380

To test:

0. Add a `debugger;` statement somewhere in the JS code. Good examples can be the `block-holder.js` module and the `render()` function in `gutenberg/.../components/rich-text/index.native.js`.
1. Run the demo app on Android
2. Enable the Remote JS debugging mode via the debug menu, and notice Chrome opening up to debug
3. Notice the debugger stopping at the `debugger` statements and properly showing the source file

### DIFF
--- a/package.json
+++ b/package.json
@@ -146,5 +146,8 @@
     "tinycolor2": "^1.4.1",
     "turbo-combine-reducers": "^1.0.2",
     "underscore": "^1.9.1"
+  },
+  "resolutions": {
+    "@react-native-community/cli": "^1.5.2"
   }
 }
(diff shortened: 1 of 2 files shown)


mkevins:
I tested this on Android emulator w/ Chromium 73.0. It works 🎉!

I was able to set debug lines directly in the Chromium debugger, as well as in code with the `debugger` statement. Also, the variable scopes and closures are correct.

PR #842, opened 2019-04-09
Include test reporter for Jest

JavonDavis:
I thought it'd be good if we could see the test results without having to scroll to the step in CircleCi. This involves adding the jest-junit test reporter to our `jest.config.js` and specifying a path for the reports in Ci. Alternatively, it could be specified in the package.json when running but since we'll only use them in Ci I just took advantage of the environment option in the vi config. 

Another addition here is the `--runInBand` flag on the jest command. According to the [CircleCI docs](https://circleci.com/docs/2.0/collect-test-data/) in the Jest section: 

> When running Jest tests, please use the --runInBand flag. Without this flag, Jest will try to allocate the CPU resources of the entire virtual machine in which your job is running. Using --runInBand will force Jest to use only the virtualized build environment within the virtual machine.

To test: 

In Ci the test results should be shown in the test summary section at the top. This doesn't apply for the Check correctness tests since that's eslint and flow.


.circleci/config.yml
```diff
@@ -20,6 +20,14 @@ commands:
       - run:
           name: Checkout Gutenberg
           command: git submodule update --init --recursive
+  add-jest-reporter:
+    steps:
+      - run:
+          name: Install JUnit coverage reporter
+          command: yarn add --dev jest-junit
```
hypest: I think that will work on CI but it means that when running locally, the `jest-junit` package won't be found since we won't have it installed. Any particular reason to only install it on CI?

In any case, it'd be good if we can run `./bin/ci-checks-js.sh` locally too, both to verify it and to have an easy way to start all the tests. WDYT?

JavonDavis: > Any particular reason to only install it on CI?

I didn't think the reports would be useful locally but it doesn't hurt. As a fix I included it in the package.json and ignore the output in the gitignore.

JavonDavis:
> Also, not exactly related to this PR but, I notice that the tests fail locally if I have more than one device/emulator connected, which is something I usually have. I wonder if we can tweak the runner to pick one device and run the tests instead of failing.

@hypest Thanks for mentioning this, this isn't expected behavior and by default, it usually just picks one. I'll give it a go.

Issue #818, opened 2019-04-04
[Android] Sometimes keyboard suggestions remove the word entirely


koke:
To reproduce:

1. Create a new post
2. Type `Thankyou`
3. Tap on the suggested correction `Thank you`
4. Watch the text disappear

![suggest 2019-04-04 19_28_09](https://user-images.githubusercontent.com/8739/55575690-cf90a580-570f-11e9-9c72-543a7b8228c9.gif)

Tested on Moto G5 Plus, alpha-160, Android 8.1

PR #850, opened 2019-04-09
818 - Sometimes keyboard suggestions remove the word entirely

marecar3:
Fixes : https://github.com/wordpress-mobile/gutenberg-mobile/issues/818

To Test : 

1. Create a new post
2. Type Thankyou
3. Tap on the suggested correction Thank you
4. Watch the text **doesn't** disappear


marecar3:
I will write some explanation here so we have it documented. @hypest @daniloercoli 

When I worked on this issue I have found out that this line of code : https://github.com/wordpress-mobile/gutenberg-mobile/pull/850/files#diff-2243933cce9f1a4881db2d8a9610c054L521
is a starting point of why text disappears. 

> I was curious about why only the `h` tags

It's because only header block has initial syle in it, or to be more clear, header block can't start without style ( and if we don't set it by default, style will be missing when the user starts to write )  Conclusion from this is that we don't need to set the style initially for every block, but only for the header block. 

My solution for the missing text is to avoid calling `fromHtml` and to just toggle initial formatting which is missing from the header block - as a user doesn't click on the H2 | H3 | H4 .

If you have any other ideas then this, I am very open to listening as the current solution isn't best but I can't think of any better at this moment.

react-native-aztec/android/src/main/java/org/wordpress/mobile/ReactNativeAztec/ReactAztecManager.java
```diff
@@ -527,6 +531,36 @@ public void afterTextChanged(Editable s) {
         }
     }
 
+    private enum ReactAztecTextFormat {
+        H1("h1", AztecTextFormat.FORMAT_HEADING_1),
+        H2("h2", AztecTextFormat.FORMAT_HEADING_2),
+        H3("h3", AztecTextFormat.FORMAT_HEADING_3),
+        H4("h4", AztecTextFormat.FORMAT_HEADING_4),
+        H5("h5", AztecTextFormat.FORMAT_HEADING_5),
+        H6("h6", AztecTextFormat.FORMAT_HEADING_6);
```
daniloercoli: I suggest to add the PARA item here: `P("p", AztecTextFormat.FORMAT_PARAGRAPH);`

and maybe a comment for future reference, since it's not easy to understand why we need to toggle the formatting.

daniloercoli:
I've investigated the problem, and the GBoard seems to send a sequence of events (2/3 events in a row) when setting a suggestion. Our `setHTML` call does break the sequence in half, resulting in the new text being skipped.

Your PR does fix it without re.setting the text. Left a comment in the code, and going to approve it! 👍

react-native-aztec/android/src/main/java/org/wordpress/mobile/ReactNativeAztec/ReactAztecManager.java
marecar3: Thanks for the comments @daniloercoli, I have made those changes.

PR #853, opened 2019-04-10
Xcode 10.2 compatibility

jtreanor:
This breaks compatibility with Xcode versions < 10.2 so merging should be done carefully. The only change is an update to Aztec-iOS (see https://github.com/wordpress-mobile/AztecEditor-iOS/pull/1172).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -74,7 +74,7 @@ jobs:
           command: yarn device-tests
   ios-device-checks:
     macos:
-      xcode: "10.0"
+      xcode: "10.2.0"
     steps:
     - checkout
     - checkout-gutenberg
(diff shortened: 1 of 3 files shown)


hypest:
I wonder if there are any special instructions on how to enable v10.2 for the `yarn ios` builds from the command line, especially if there are multiple Xcode versions installed on the local machine. Any clues @jtreanor , @etoledom ?

etoledom:
@hypest , on Xcode settings (`cmd + ,` ) -> `Locations` tab, you can choose what command line tools to use. So any command line usage of Xcode (like on `yarn iOS`) will use that version.

Screen Shot 2019-04-16 at 4 09 45 PM

Issue #687, opened 2019-02-28
Video block


Tug:
Add support for Video block

pinarol:
TO DO:

- [ ] (First iteration)Make a basic video block that can upload & show media (from WP media library)
- [ ] Add unit-tests for touched areas( both image and vide block )
- [ ] Add support for Video settings
- [ ] Enhance video block to show videos with better aspect ratio rather than a fixed one
- [ ] Try and look for ways to make it cross platform

pinarol:
**Note about video playing differences between platforms**

We have limitations on react-native-video level. Having control buttons(like play, pause etc) in the video block are messing up the layout on iOS, so I had to hide them and add our own play button which opens the video on full page. Once the video opens full page then there's no problem about control buttons. A bit of a technical insight: I think this is the line responsible for it, they are setting the frame (as a workaround for some issue) when controls=true
https://github.com/react-native-community/react-native-video/blob/c30f246f5694452f4f081ff30ac4523dfd89f41d/ios/Video/RCTVideo.m#L593
And by doing that they are messing up the y position of our page. 

On the other hand on Android the full page preview just doesn’t work, the method call does nothing. But control buttons do work correctly. The video can play inline.

So now, we have different user experiences in both platforms in terms of video playing. If we want to make this equal in the future we'll need to either solve one of this: 

1. Android the full page preview doesn’t work
2. iOS controls=true messing up the layout

hypest:
The block has been released for both platforms. Closing this.

PR #854, opened 2019-04-10
Add video block ( first iteration )

pinarol:
To Fix: https://github.com/wordpress-mobile/gutenberg-mobile/issues/855

There are also some more TODO items in this issue: https://github.com/wordpress-mobile/gutenberg-mobile/issues/687#issuecomment-482011555 which are not handled in this PR

This PR introduces video block.

Gutenberg PR: https://github.com/WordPress/gutenberg/pull/14912

This is a breaking change for parent apps:

WPiOS PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/11454
WPAndroid PR: https://github.com/wordpress-mobile/WordPress-Android/pull/9644

iOS Example app:

video-block-ios2

Android Example app:
(Media option picker is not working but a sample video block is added to content)

video-block-android

## Types of changes

- Integrated react-native-video library to both iOS and Android example apps
- Updated bridge to pass media `filter` such as 'video', 'image'
- Added a sample video block to example app content

**To Test**

Test with both iOS and Android Example apps:

**iOS Example App Tests**

Note: Video playback isn't very smooth in iOS simulator but there's no problem on real devices as far as I can tell.

**Test 1** - Add Video block

- Open the media picker and add video block
<br>

**Test 2** - Insert video from device (successful) (currently available on iOS example app only)

- Tap on the Video block placeholder
- Choose option 'Choose from device'
- Verify that you can choose among videos only(in order to add videos to the ios simulator just drag a video from your computer and drop to the simulator)
- Choose a video
- Verify that spinner appears to show the progress:
Screen Shot 2019-04-15 at 18 27 44
- At the end of the progress you should be able to tap and play the video
<br>

**Test 3** - Insert video from device (failing) (currently available on iOS example app only)

- Select the video block
- Tap edit button in the toolbar
- Choose option 'Choose from device', then choose a video
- During the spinner upload continues, longpress somewhere in the screen(longpress will trigger mocking of upload failure)
- Verify you see the retry message:
Screen Shot 2019-04-15 at 18 33 49
- Tap on the gray area and verify that you see the retry option:
Screen Shot 2019-04-15 at 18 34 17
- Tap on Retry and see that upload starts again 
<br>

**Test 4** - Insert video from device (cancel) (currently available on iOS example app only)
- Select the video block
- Tap edit button in the toolbar
- Choose option 'Choose from device', then choose a video
- During the spinner upload tap on the video block 
- Verify that Cancel option appears:
Screen Shot 2019-04-15 at 19 04 09
- Choose Cancel and see that media is cleared and you see the placeholder again.
<br>

**Test 5** - Insert video from WP media library (currently available on iOS example app only)

- Select the video block
- Tap edit button in the toolbar
- Choose option 'WordPress media library'
(Since this is the example app we are only updating the block with a video link)
- Verify that video block is updated with another video and you can play the new video
<br>

**Test 6**

Verify that above tests are working for Image block also


hypest:
This PR is currently blocked by an issue with JitPack (the JS bundling fails there at the moment) so, moving it to the next milestone.

SergioEstevao:
Great work on this! Everything working great just a small design issue that I don't know if it's a problem that comes from the original GB code.

It looks that the when you insert a video to the block you get some margins on the top and bottom. Check the image bellow:

Screenshot 2019-04-24 at 16 35 07

pinarol:
@SergioEstevao thanks! That's because currently we use a fixed aspect ratio for every video, planning to enhance it on a separate PR, there's already a checklist item in the main issue: https://github.com/wordpress-mobile/gutenberg-mobile/issues/687#issuecomment-482011555

> Enhance video block to show videos with better aspect ratio rather than a fixed one

etoledom:
Besides the issues encountered while reviewing this PR, I noticed some few UX details that could be improved.

I don't expect any of this points to be revisited/implemented in this PR, but I'm documenting them to be analyzed and (maybe) implemented in the future.
I'd like to ask @iamthomasbishop's input to see if these details are worth exploring before creating ticket.

media_block_select
retry_options
retry
stop
take_video

etoledom:
One more issue:
As we were talking with @marecar3 , there seems to be a problem saving videos from sites with Jetpack, that creates a `videopress` block not recognized in mobile.

It is still a Video block, so mobile tries to read it, but its format is different, triggering a `Block validation` error.

The code from video press looks like this:

```html
<!-- wp:video {"guid":"kwKwpVsx","id":616,"src":"https://mysitefortestingetoledom.files.wordpress.com/2019/05/img_1413-1-1.mov"} -->
<figure class="wp-block-video wp-block-embed is-type-video is-provider-videopress">
    <div class="wp-block-embed__wrapper">
        https://videopress.com/v/kwKwpVsx
    </div>
</figure>
<!-- /wp:video -->
```

The code generated by gutenberg-mobile looks like this:

```html
<!-- wp:video {"id":588} -->
<figure class="wp-block-video">
    <video src="https://mysitefortestingetoledom.files.wordpress.com/2019/03/img_1338.mov"></video>
</figure>
<!-- /wp:video -->
```

The videopress save function can be found here:
https://github.com/Automattic/jetpack/blob/43f02648806d6a927445e528a007176a76e31875/extensions/blocks/videopress/save.js

I'm not sure how should we approach this issue.
cc @koke @hypest

etoledom:
This video issue seems to happen on sites `Premium` and above, where `videopress` is active.

Related to this, on a free plan, on web it doesn't let me to upload a video to the WP media library, it tells me to upgrade my plan. But I was able to upload a video from the mobile app. Maybe this is also worth investigating.

pinarol:
> Related to this, on a free plan, on web it doesn't let me to upload a video to the WP media library, it tells me to upgrade my plan. But I was able to upload a video from the mobile app. Maybe this is also worth investigating.

When I try to upload a video from a site with free plan using mobile, I get the following error on both block editor and aztec:

Screen Shot 2019-05-08 at 11 41 55

After the attempt I checked the WP library the video isn't there.

And when I try to do the same thing via web I get the following error but it doesn't tell me to upgrade my plan:

Screen Shot 2019-05-08 at 11 47 48

So I think we are doing things differently, could you be able to send more detailed steps or a video maybe?

SergioEstevao:
> One more issue:
> As we were talking with @marecar3 , there seems to be a problem saving videos from sites with Jetpack, that creates a `videopress` block not recognized in mobile.
> 
> It is still a Video block, so mobile tries to read it, but its format is different, triggering a `Block validation` error.
> 
> The code from video press looks like this:
> 
> ```
> 
> <figure class="wp-block-video wp-block-embed is-type-video is-provider-videopress">
>     <div class="wp-block-embed__wrapper">
>         https://videopress.com/v/kwKwpVsx
>     </div>
> </figure>
> 
> ```
> 
> The code generated by gutenberg-mobile looks like this:
> 
> ```
> 
> <figure class="wp-block-video">
>     <video src="https://mysitefortestingetoledom.files.wordpress.com/2019/03/img_1338.mov"></video>
> </figure>
> 
> ```
> 
> The videopress save function can be found here:
> https://github.com/Automattic/jetpack/blob/43f02648806d6a927445e528a007176a76e31875/extensions/blocks/videopress/save.js
> 
> I'm not sure how should we approach this issue.
> cc @koke @hypest

So this was a change that was done on the web side, so now when VideoPress is active instead of using a standard video block, they now use an embedded kind of thing.

Looking at the block code/attributes it looks we could try to intercept those, read the videopress URL from the block content, query the native part to solve the URL to a video URL,  and use the video component to render.

SergioEstevao:
> > Related to this, on a free plan, on web it doesn't let me to upload a video to the WP media library, it tells me to upgrade my plan. But I was able to upload a video from the mobile app. Maybe this is also worth investigating.
> 
> When I try to upload a video from a site with free plan using mobile, I get the following error on both block editor and aztec:
> 
> Screen Shot 2019-05-08 at 11 41 55
> 
> After the attempt I checked the WP library the video isn't there as expected.
> 
> And when I try to do the same thing via web I get the following error but it doesn't tell me to upgrade my plan:
> 
> Screen Shot 2019-05-08 at 11 47 48
> 
> So I think we are doing things differently, could you be able to send more detailed steps or a video maybe?

That is strange! A free site on WP.com when using the mobile apps should be able to upload a video. There is an exception on our API endpoints to allow this.

SergioEstevao:
There is being a lot of discussion around video uploads and video blocks so here is a small cheatsheet how things should work:

### WordPress.com Free Site:

 - On a free site, video uploads are disabled on the web browsers, but on the mobile, there is an exception on the media API upload endpoint that allows the clientID of our mobile Apps to upload videos.
- When a video is uploaded using GB editor mobile a video block is created that has the following format:

```
<!-- wp:video {"src":"https://chefventura.files.wordpress.com/2019/05/img_5692.mov"} -->
<figure class="wp-block-video">
  <video controls src="https://chefventura.files.wordpress.com/2019/05/img_5692.mov"></video></figure>
<!-- /wp:video -->

```
### WordPress.com Premium Site:
 - On premium sites, VideoPress is enabled on the web and mobile.
- The GB video blocks created should look like this:

```
<!-- wp:video {"guid":"GRlnOHUa","id":3870,"src":"https://sergioestevaosales.files.wordpress.com/2019/05/ams-boat-low.mp4"} -->
<figure class="wp-block-video wp-block-embed is-type-video is-provider-videopress">
<div class="wp-block-embed__wrapper">https://videopress.com/v/GRlnOHUa</div>
</figure>
<!-- /wp:video -->
```

Notice that this block has a different format from the block on free sites, it has an embedded URL instead of a video HTML element.

### WordPress.com Business/Atomic Site

 - Business/atomic sites have Jetpacks installed and by default, VideoPress is enabled on them. Both mobile and web should be able to upload videos, but the blocks look like this:

```
<!-- wp:video {"autoplay":false,"id":628,"loop":false,"muted":false,"src":"https://atomicsergio.blog/wp-content/uploads/2019/05/ams-boat-low.mp4"} -->
<figure class="wp-block-video">
<video controls src="https://atomicsergio.blog/wp-content/uploads/2019/05/ams-boat-low.mp4"></video>
</figure>
<!-- /wp:video -->
```

Notice that while video press is enabled for some reason on the web they are using the standard video block format, instead of the embedded format with the VideoPress URL

### WordPress.org self hosted without Jetpack:

- Videos are possible to be uploaded both on mobile and web
- Video Blocks should look like this:

```
<!-- wp:video {"src":"https://chefventura.files.wordpress.com/2019/05/img_5692.mov"} -->
<figure class="wp-block-video">
<video controls src="https://chefventura.files.wordpress.com/2019/05/img_5692.mov"></video></figure>
<!-- /wp:video -->
```

SergioEstevao:
At the moment we are only implementing the block using the format that has a `video` element inside. I think this is fine for a first version and it will work on all type of sites. 

The only problem is that we will not able to see/render on mobile video blocks that are created using the browser on Premium sites.

etoledom:
Thank you @SergioEstevao - That clarifies a lot!
> I think this is fine for a first version and it will work on all type of sites.

I definitely agree with that 👍 

>WordPress.com Premium Site:
>Notice that this block has a different format from the block on free sites, it has an embedded URL instead of a video HTML element.

To do this on Mobile, we might have to import the JetPack block, right?

> The only problem is that we will not able to see/render on mobile video blocks that are created using the browser on Premium sites.

Still for another PR but needed for release, I think we should show a message of some sort saying that this `videopress` block is not yet supported.

daniloercoli:
> To do this on Mobile, we might have to import the JetPack block, right?

I think we just need to read the VideoPress hash, query the native side (there is an API call that from a VP hash does return video info) and replace it with the video URL when the call returns.

daniloercoli:
How about self-hosted sites with JP installed @SergioEstevao ?

If i understand it right, we're fully supporting the Video block on self-hosted sites without JP, right?
(video blocks created on the web are successfully rendered on mobile, and vice-versa).

iamthomasbishop:
@etoledom @pinarol 
I spent some time yesterday going over the Video block anatomy to make sure I'm aligned on the potential states of the Video block (most of which also applies to the Image block and various other types of media blocks) before commenting. For documentation, here's what I came up with for states (and below, the actions that show when tapping on an item in that state):

Screen Shot 2019-05-08 at 10 09 41 AM

The core states are:

* Placeholder
* Uploading
* Error
* Uploaded (static/focused)

Some of this is relevant on the notes you wrote above, but I'll address each of your notes individually:

## Media blocks are difficult to select

I would expect the entire inner content of the block would be a single tap area that would fill the entire inner content. So on a placeholder, you could tap anywhere on the inside of the block to take the "add video" action – like how the Image block placeholder already works.

Some more detailed notes:

When a Video block is _static_ (not focused): 

* Tapping anywhere on the block should bring focus to the block itself, and outline/highlight the video.

When a Video block _is focused_: 

* Tapping on the inside (the video itself), should show an option/action sheet with options. 

* Tapping outside of that (on the inline toolbar/bottom space, for example) should essentially do nothing except keep the block focused.

* We _could_ remove the highlight from the video if we wanted to do that (this type of behavior might become helpful when we work on Gallery blocks or something with multiple media items inside a block).

## Retry options should have a `cancel` button

Agreed, and if we are to follow Aztec (which I think makes sense here), the options should be:

* `Remove Video` (essentially a cancel button)
* `Retry` or `Retry Upload`
* `Dismiss`

## Error state design differences

Agree, this should be cleaned up. See mockups above for overview, but here are the basic design specs (I've also added these video state designs to the Zeplin doc // cc @pinarol @etoledom) – _dependent on if we have a thumbnail image_ to show or not:

If we have a thumbnail image to show:

image

* **Overlay:** Black background at 40% opacity/alpha
* **Highlight**: 2pt border, color `#00AADC ` aka `$blue-medium`
* **Icon:** I'm using the Gridicon here for `refresh`, but we can continue using the current for now
* **Icon Color:** `$white` aka `#FFFFFF`
* **Text:** 12pt `$white` aka `#FFFFFF`

If we _do not have a thumbnail_:

image

* **Background:** `$gray-lighten-30` aka `#e9eff3`
* **Highlight**: 2pt border, color `#00AADC ` aka `$blue-medium`
* **Icon:** I'm using the Gridicon here for `refresh`, but we can continue using the current for now
* **Icon Color:** `$gray-dark` aka `#2e4453`
* **Text:** 12pt `$gray-dark` aka `#2e4453`

## Stopping an upload-in-progress

> There is no clear indication that tapping on the placeholder has the option to stop it

I'm not sure if this is a big concern because users instinctively tap on the video as it's uploading if they want to take an action – so for now I'd just match Aztec, where you tap to show options. But if we wanted to, we could add a text label in the center of the block that says `Cancel Upload` or `Stop Upload`, considering that's the only option we have in our actions for that state.

> If the upload finishes while the ActionSheet is present, `stop upload` will do nothing

I would expect when the upload finishes behind the ActionSheet that it would simply close the action sheet because those actions are no longer relevant. 

> We could pause the upload while the ActionSheet is visible

I wouldn't interrupt the upload, especially if we show the option to cancel/remove, etc.

## "Take a Video" label on ActionSheet

I think this is normal, at least for iOS – the label I've most often seen is something like "Take a Photo or Video", but I don't have strong feelings about this, just want to be consistent.

## Side notes

* Can we possibly update the Placeholder icons on the Video and Image blocks to use the proper color – `$gray-dark` aka `#2E4453`? // cc @marecar3

iamthomasbishop:
Some updated designs, based on conversations in Slack (@pinarol @etoledom @marecar3 @SergioEstevao):

Screen Shot 2019-05-08 at 1 50 59 PM

Notes:

* There seemed to be some uncertainty regarding if/when "posters" or video thumbnails would be available. I mocked up a variant for each relevant state (`uploading` and `error` primarily) to include those, so when we come to a conclusion, both are available.

* The no-thumbnail version of the `uploading` state contains a 40pt block type icon. This icon should be one shade darker of gray than the background (in the above mocks, the background is `$gray-lighten-30`, so the icon is `$gray-lighten-20`). I think the icon could also animate opacity between 0-100%, like this:

video-block-uploading-loader

pinarol:
hey @iamthomasbishop I've opened a separate issue to track those enhancements, we think they deserve a separate PR https://github.com/wordpress-mobile/gutenberg-mobile/issues/966 Please note that video block won't be available publicly after merging the current PR, we'll get these UI enhancements and some other fixes first.

pinarol:
I am going to open a separate issue for the unexpected videoo block html produced by Premium sites. We digged into that issue a lot yesterday but couldn't come up with a certain solution yet. It doesn't look directly related with our changes in this PR. But we'll definitely need to find a solution for that before we make video block publicly available.

pinarol:
> When I try to upload a video from a site with free plan using mobile, I get the following error on both block editor and aztec:

About my own comment here, It turned out I had a problem with my my mobile-secret settings that prevents me from doing this, you can ignore that @etoledom

pinarol:
> I am going to open a separate issue for the unexpected video block html produced by Premium sites. 

Update: we have added handling of invalid content to our project board, it will be resolved for v2. So I think I won't need to open a separate issue for this particular scenario. In the parallel we'll be working with Jetpack team to seek for a better solution.

pinarol:
I think your comments are addressed @etoledom but let me know if I missed anything. We'll handle the UI/UX enhancements here: https://github.com/wordpress-mobile/gutenberg-mobile/issues/966

pinarol:
Thanks @etoledom 🎉

> I think that probably the best would be to not save the local url in the url attribute, but on a new one dedicated for local urls.

I agree.

> To consider as a refactor in the future.

I think this is a good task to do after we have some unit and ui tests because it'll lead us to a kind of refactor. I added an item named `Don't save local uri in attributes` here: https://github.com/wordpress-mobile/gutenberg-mobile/issues/687#issuecomment-482011555

I will update branches from master/develop and resolve conflicts.

PR #857, opened 2019-04-11
FIX empty block creation on device running AOSP keyboard

### LINKED ISSUE #856: Android - Problem with the AOSP keyboard and new block creation
There is a problem on Android devices using the stock AOSP keyboard where trying to add new empty block to the editor does delete it immediately after.
Both the Inserter UI, and the Enter.Key detection on the current block, do insert the new block, and it's visible on the screen for a very short while, then it gets deleted.

The problem is on the InputFilter on Aztec that detects the Backspace pressed event, where for some reason it's fired when empty content is set to the block (AztecText component).
`sertTextFromJs` is called in this case with empty string.

This is only happening on AOSP keyboard (I guess it's visible on emulators, anyway better to fix this to be able to properly test the demo app).

Steps to repro:
- Start the demo app on emulator
- Insert a new empty para block
- Boom it gets deleted

AOSP keyaboard

daniloercoli:
This PR fixes #856 by making sure to check if text listeners are enabled before doing any text/block manipulating action `onEnter` or `onBackSpace`. 

In this particular case the BackSpace.Key detection code in Aztec is called when the initial empty HTML content is set to the block via `setTextFromJS`. Text listeners are disabled in this case, and those are not running, but InputFilters are not disabled.... By the way, I guess a better fix would involve changes to Aztec, but probably is easier to fix them here, since this change could also fix other potential problems in the future.

Steps to repro:

- Start the demo app on STOCK emulator (Check AOSP keyboard)
- Insert a new empty para block
- Make sure it stays on the screen

### DIFF
--- a/react-native-aztec/android/src/main/java/org/wordpress/mobile/ReactNativeAztec/ReactAztecText.java
+++ b/react-native-aztec/android/src/main/java/org/wordpress/mobile/ReactNativeAztec/ReactAztecText.java
@@ -64,6 +64,7 @@ public class ReactAztecText extends AztecText {
     // This is required to keep placeholder text working, and start typing with styled text.
     // Ref: https://github.com/wordpress-mobile/gutenberg-mobile/issues/707
     private String mTagName = "";
+    private String mEmptyTagHTML = "";
 
     private static final HashMap<ITextFormat, String> typingFormatsMap = new HashMap<ITextFormat, String>() {
         {
@@ -101,8 +102,7 @@ public class ReactAztecText extends AztecText {
                         return onBackspace();
                     }
                     else {
-                        String emptyTag = "<" + getTagName() + "></" + getTagName() + ">";
-                        if (!content.equals(emptyTag)) {
+                        if (!content.equals(mEmptyTagHTML)) {
                             return onBackspace();
                         }
                     }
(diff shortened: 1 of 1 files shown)


mkevins:
I just tested this on Android emulator, and it works. It also fixes this: https://github.com/wordpress-mobile/gutenberg-mobile/issues/783 . 🎉

react-native-aztec/android/src/main/java/org/wordpress/mobile/ReactNativeAztec/ReactAztecText.java
```diff
@@ -88,21 +89,20 @@ public ReactAztecText(ThemedReactContext reactContext) {
...
-                if (shouldHandleOnBackspace) {
+                if (shouldHandleOnBackspace && !isTextChangedListenerDisabled()) {
                     String content = toHtml(false);
                     if (TextUtils.isEmpty(content)) {
                         return onBackspace();
                     }
                     else {
-                        String emptyTag = "<" + getTagName() + "></" + getTagName() + ">";
-                        if (!content.equals(emptyTag)) {
+                        if (!content.equals(mEmptyTagHTML)) {
```
mkevins: I like this refactoring. This might be useful for other things.

PR #868, opened 2019-04-13
Use the Aztec-Android version with the #845 fix

### LINKED ISSUE #845: [Android] List Block's caret positioning is off when new item at the end
When there are a few empty list items, trying to add a new item at the end will not place the caret correctly.

1. Do a long list, long enough to have a few empty list items here and there
2. Try to add a new item at the end of the list
3. Notice that the caret doesn't land on the new list item but rather, goes "back". The more empty lines from 2., the more chars the caret goes back.

hypest:
Fixes #845 

This PR updates to a fixed Aztec Android version. No changes on the gutenberg-mobile side.

Aztec-Android PR: https://github.com/wordpress-mobile/AztecEditor-Android/pull/807
WPAndroid PR: https://github.com/wordpress-mobile/WordPress-Android/pull/9614

To test:

1. In the demo app, do a long list, long enough to have a few empty list items here and there
2. Try to add a new item at the end of the list
3. The new list item should be normally created at the end and the caret should be placed on it

### DIFF
--- a/react-native-aztec/android/build.gradle
+++ b/react-native-aztec/android/build.gradle
@@ -12,7 +12,7 @@ buildscript {
         wordpressUtilsVersion = '1.22'
         espressoVersion = '3.0.1'
 
-        aztecVersion = 'v1.3.24'
+        aztecVersion = '3a6ab4ffde163ec2f2ebd3732a20013ebd8723a4'
     }
 
     repositories {
--- a/react-native-aztec/android/build.gradle
+++ b/react-native-aztec/android/build.gradle
@@ -12,7 +12,7 @@ buildscript {
         wordpressUtilsVersion = '1.22'
         espressoVersion = '3.0.1'
 
-        aztecVersion = '3a6ab4ffde163ec2f2ebd3732a20013ebd8723a4'
+        aztecVersion = '1400adb886ca9372211f06b62dd24fbbbd252f42'
     }
 
     repositories {


daniloercoli:
I've tested this PR and it seems to work fine. 

The tests I run on this PR was to copy one emoji, that is made up by two characters plus the ZWJ between them, taken from here https://emojipedia.org/emoji-zwj-sequences/ , and pasting it in the editor ---> It worked without problem. Also tried rotating the device, adding and removing the emojis, end other tests.

When trying to paste content to list instead I noticed two problems (Not related to this PR since I was able to replicate them in develop):
- The option menu doesn't appear
- Tap long on the caret does make a selection of text (even it there is no visible text on the screen)
and then option menu appear on the screen.

hypest:
TIL about the use of ZWJs in emojis! 😰 

Great idea trying those out @daniloercoli 🙇  and glad they work 🎉 . Looping back, I added info about this in [this comment](https://github.com/wordpress-mobile/AztecEditor-Android/pull/807#issuecomment-483164872).

> When trying to paste content to list instead I noticed two problems 

Would you mind opening tickets for those @daniloercoli ?

hypest:
This PR works and is green on CI, but [the parent wpandroid PR](https://github.com/wordpress-mobile/WordPress-Android/pull/9614) fails because the JS bundling fails on JitPack. That's a problem not related to this PR though so, I'm going to merge and tackle the JitPack issue separately.

PR #844, opened 2019-04-09
Fixes issue in latest master

koke:
See https://github.com/WordPress/gutenberg/pull/14880 for description

PR #873, opened 2019-04-15
Hack: patch Metro to not watch files when bundling

hypest:
We seem to have reached some limits with file watchers in JitPack when building the Android JS bundle there (to use in WPAndroid) and that started after https://github.com/wordpress-mobile/gutenberg-mobile/pull/844 got merged.

See a failed JitPack log here: https://jitpack.io/com/github/wordpress-mobile/gutenberg-mobile/26013ee0d5/build.log

This PR introduces a nasty hack to fix this. I couldn't find a practical way to stop Metro from watching for files while bundling (not even sure why would it need do the watching at that phase, maybe a bug with the React Native bundle command?) and couldn't alter JitPack's default limit for inotify watchers. The hack here is to patch Metro by copying `DependencyGraph.js.patched` over the original, just to change the `watch: true` flag to `false`.

I will re-use the WPAndroid PR (https://github.com/wordpress-mobile/WordPress-Android/pull/9614) to test this out.

To test:

The WPAndroid PR should just be green as the build should succeed.


hypest:
Looping back, I [brought up here](https://github.com/facebook/metro/issues/355#issuecomment-483245534) the fact that we're affected by Metro's behavior to always watch the files, even when RN is performing the bundling procedure.

PR #880, opened 2019-04-17
Fix change list type crash

etoledom:
This PR updates the Gutenberg ref to test the [`Fix change list type crash` PR](https://github.com/WordPress/gutenberg/pull/15010). 

Screen Shot 2019-04-17 at 10 22 34 AM

**To test:**
- Run the app.
- Add a ListBlock.
- Change the list type to Ordered List.
- Check that it doesn't crash.


etoledom:
Thank you! 🙏

PR #885, opened 2019-04-17
Point to gutenberg hash with list block spaces crash fix

### LINKED ISSUE #871: List crashes with extra spaces
Adding extra spaces inside list blocks results in various crashes. I also observed similar issues when [working on adding multiline image captions](https://github.com/WordPress/gutenberg/pull/14883#issuecomment-483108111).

### Test Scenarios

1. (1) Begin editing new list, (2) add **two** spaces, (3) add any letter, (3) press Enter, and 💥
2. (1) Begin editing new list, (2) add **three** spaces, (3) press Enter, and 💥
3. (1) Begin editing new list, (2) add any letter, (3) add **three** spaces (second space creates a period), (4) press Enter, and 💥

All of these crashes are pretty similar, so just including a gif of the first scenario (space-space-letter) as an example:
output_spacespace

hypest:
Fixes #871 

Gutenberg side PR: https://github.com/WordPress/gutenberg/pull/15021

This PR brings in the fix, which is only on source code on the Gutenberg repo.

The internal representation of rich-text in Aztec and the format-lib are not fully compatible or in-sync and we also use html to establish the communication between the two. That can lead to cases like the issue linked, where Aztec will do some trimming to remove spaces that are unimportant for html rendering, but format-lib won't so, making the caret positioning logic to get out of sync.

This fix will try to detect when such spaces will be removed and avoid telling Aztec where to put the caret and hope for the best 🤞.

This fix is not "perfect" though. It only touches the Android side (iOS seems to handle the caret setting OK) and will also lead to the text entered by the user being mutated since the spaces will get trimmed by Aztec.

### To test A

1. Begin editing new list
2. Add **two** spaces, then any letter and then press Enter
3. The list should create a new list item and the spaces in the item above it will be trimmed

### To test B

1. Begin editing new list
2. Add **three** spaces and press Enter
3. The list should create a new list item and the spaces in the item above it will be trimmed

### To test C

1. Begin editing new list
2. Add any letter, add **three** spaces (second space creates a period) and press Enter
3. The list should create a new list item and the spaces in the item above it will be trimmed


mkevins:
I tested cases A, B, and C on Android emulator, and observed the results described. I noticed, though, that the period-creation after two spaces (described in test case C) did not always seem to happen. For example, if I:

1. Tap the `+` button, and create a new list block.
2. Type the letter 'T', followed by 3 spaces.

I do not observe the creation of the period. But if I:

1. Tap the `+` button, and create a new list block.
2. Type the letter 'T', followed by 1 space.
3. Type [Backspace].
4. Type 3 spaces.

I do observe the period being created.

In any case, the text in the list item is trimmed. I suspect that the creation of the period may have something to do with the keyboard (AOSP built in emulator keyboard, in this case), and may be unrelated to this PR.

Issue #888, opened 2019-04-18
[iOS] List block indent and outdent by Tab / Shift+Tab key commands.


etoledom:
Ref: https://github.com/wordpress-mobile/gutenberg-mobile/pull/863#issuecomment-484415445

On Aztec-iOS there are key shortcuts to indent/outdent list items. These are not working properly and breaking the BlockList.

As mentioned in the referenced thread, we want to:
- [ ] Deactivate these key bindings in the RNAztec level.
- [ ] Reactivate them using the `gutenberg` mechanism.

SergioEstevao:
Currently, on Gutenberg web the indent/outdent is not done using tab+shift tab. It works the following way:
 - if you are at the first character of a list item and press space it indents the list
 - on an outdent empty list item if you press backspace the list outdents one level.

hypest:
This ticket is not worked on at the moment so, I'll clear the assignee field to denote so.

PR #889, opened 2019-04-18
[iOS] Deactivate Tab and Shift+Tab key shortcuts

etoledom:
Fixes a part of https://github.com/wordpress-mobile/gutenberg-mobile/issues/888

Related Aztec PR: https://github.com/wordpress-mobile/AztecEditor-iOS/pull/1175

This PR removes the Tab and Shift+Tab key shortcuts.
On a second iteration of the ListBlock, we will want to reactivate them using gutenberg internal mechanism.

I also removed the `Tab` key functionality completely. It was adding tab characters to the content of blocks, and also breaking the ListBlock. This is different to how it works on Android and Web, so I don't see any harm on deactivating it until we handle it properly (cc @hypest )

IMHO, the functionality we want for the Tab key in this case is to move the focus to the next block, in the same way it works on web.

To test Tabs:
- Run the iOS example project on the simulator.
- Focus a RichText powered block.
- Press the Tab key (be sure to have the hardware keyboard connected to the simulator)
- Check that nothing happens.

To test Shift+Tabs:
- Replace the HTML content with this one: 
```html
<!-- wp:list -->
<ul><li>Hello<ul><li>world</li></ul></li></ul>
<!-- /wp:list -->
```

- Set the cursor in the indented list item.
- Press Shift+Tab in the hardware keyboard.
- Check that it doesn't modify the ListBlock.

### DIFF
--- a/package.json
+++ b/package.json
@@ -76,6 +76,7 @@
     "wpandroid": "yarn android --root $TMPDIR/gbmobile-wpandroidfakernroot --variant wasabiDebug --appIdSuffix beta --appFolder WordPress --main-activity=ui.WPLaunchActivity",
     "preios": "yarn preios:xcode10 && yarn preios:carthage",
     "preios:carthage": "cd react-native-aztec && yarn install-aztec-ios",
+    "preios:carthage:update": "cd react-native-aztec && yarn update-aztec-ios",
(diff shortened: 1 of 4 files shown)


etoledom:
Thank you!

PR #890, opened 2019-04-18
[iOS] Activate autoscroll on ListBlock

etoledom:
This PR updates the gutenberg ref to test https://github.com/WordPress/gutenberg/pull/15048

This activates autoscroll on iOS when the caret is out of the viewport area (i.e. goes behind the keyboard).

lists-autoscroll

To test:
- Create a list block.
- Add elements until the caret will go under the keyboard.
- Check that the block scrolls to show the caret.


mzorz:
Just tried this PR with words other than `Asdf` and it crashes with a red screen

hypest:
Moving this to next milestone as it needs some more testing (as per @pinarol ) and it can't make it to today's cut.

etoledom:
Thank you!

PR #892, opened 2019-04-18
Fix for history that's not empty on a fresh start of the editor

### LINKED ISSUE #887: History stack is not empty (and broke) on a fresh start of the editor, and on new RichText block
History stack is not empty on a clean start of the editor.

Problem 1:
- Start the demo app 
- Tap on a RichText block 
- Boom ---> you will see the Undo button available
- if you tap on it "an action" is actually executed, the Redo button slightly visible for a fraction of second, and the the "undo" is available again
- See the first seconds of the GIF attached below

Problem 2:
**If you add a new empty block it's hard to get rid of hit by tapping on `Undo`.** 

For the same reason of above, it seems two actions are executed in a sequence in a very small time. If you want to get rid of the block you need to tap "undo" many times, very fast. (The GIF below does show this behavior at the end).

history mess up

Investigated a bit, and I think the problem is due to the `onSelectionChange` called at init and every time the user adds new text -  `onSelectionChange` is called right after the `onChange` callback. So basically we're updating the JS side lot of times 🤔 
Not quite sure why we're calling `this.props.onChange( this.lastContent );` when the selection does change. 

cc @marecar3 @mkevins

daniloercoli:
This PR fixes #887 

All the changes are in the GB side companion PR https://github.com/WordPress/gutenberg/pull/15055

**Test 1**
- Start the demo app
- Tap on a RichText block
- Check the Undo/Redo button are not available

**Test 2**
- Start the demo app
- Tap on a RichText block
- Check the Undo/Redo button are not available
- Go to the end of the block
- Tap enter to create a new para
- Tap Undo twice* and the new block should be gone

* => there is another problem with history that I will tackle in another PR. Shortly, Aztec can modify the HTML tags order, so pressing enter on a block does result in sending back to the store an HTML content that could be slightly different from the one set at the beginning, resulting in a new history entry**


mkevins:
I tested this, and it does fix the issue, as described.

I have some additional findings documented in this comment: https://github.com/wordpress-mobile/gutenberg-mobile/issues/887#issuecomment-484765484

One other thing I did notice is that when I created a new paragraph block from the problematic italicized block, two taps on undo was sufficient to clear the new block, but a third tap was necessary to fully clear the undo history and see the undo button disabled.

daniloercoli:
> One other thing I did notice is that when I created a new paragraph block from the problematic italicized block, two taps on undo was sufficient to clear the new block, but a third tap was necessary to fully clear the undo history and see the undo button disabled.

Yes, this  is the issue I mentioned in the description (sorry if i was not clear, and still to open a new issue for it).
Basically when you tap enter on the first block the content of the block is retrieved from Aztec and stored in the store, and a history point made for it. Unfortunately Aztec does change the HTML given in input, even if you don't actually change the block, so the JS side notices the change and create the undo point.

Will open a new issue shortly with details.

PR #867, opened 2019-04-12
Fixed issue with re-focusing PostTitle

marecar3:
Fixes: https://github.com/wordpress-mobile/gutenberg-mobile/issues/865

Gutenberg PR : https://github.com/WordPress/gutenberg/pull/14969

**Steps to reproduce :**

1. Open WPAndroid/WPiOS app
2. Create a New post
3. Change focus on the Paragraph block
4. Click on the Post TitleR

**Result** -> Post Title should be focused

- [ ] Update gutenberg reference when PR https://github.com/WordPress/gutenberg/pull/14969 is merged.

PR #893, opened 2019-04-19
Fix title not focusing

etoledom:
This PR updates the gutenberg ref to test: https://github.com/WordPress/gutenberg/pull/15069
It fixes an issue where the post title is not being focused when the user tap on it.

All the description and test steps can be found in the `gutenberg` side PR.

This is the same fix proposed as a hotfix v1.2.1, but targeting develop instead.
Refs:
https://github.com/wordpress-mobile/gutenberg-mobile/pull/867
https://github.com/WordPress/gutenberg/pull/14969


etoledom:
Thank you !

PR #896, opened 2019-04-19
CircleCI: Publish 'react-native-aztec' and 'react-native-gutenberg-bridge' to Bintray

jtreanor:
This is a proposal for replacing Jitpack builds of `react-native-gutenberg-bridge` with a CircleCI job that publishes artifacts to Bintray on each push.


hypest:
Thanks for this PR @jtreanor ! Good to see alternatives to our JitPack pipeline!

Haven't yet had the chance to dive into the code itself but I wonder, how do you envision using CircleCI for the artifact with other jobs depending on it? Will the "waitForJitPack" [trick](https://github.com/wordpress-mobile/WordPress-Android/blob/develop/libs/editor/WordPressEditor/build.gradle#L79) work with Bintray too? Will we need to find another way to coordinate the builds of the different repos?

jtreanor:
@hypest It would be possible to do write a similar script to make Gradle wait for Bintray artifacts.

It is worth considering whether we would need it though. With CircleCI managing the artifacts, it is less of a black box so maybe it is fine to just wait for the build to complete before using the new version.

hypest:
👋 @jtreanor , one of the JitPack based flows that we use currently is to try any branch or commit (think: when git bisecting) of gutenberg-mobile against the main WPAndroid's develop. Do you think there's something we can use from CircleCI/Bintray to replicate that flexibility?

> With CircleCI managing the artifacts, it is less of a black box so maybe it is fine to just wait for the build to complete before using the new version

Hmm, I think that manually having to wait (checking CircleCI status on a browser's tab) is not a gradle/cli friendly flow and will degrade the development experience: Just issuing `./gradlew assembleVanillaRelease` for any combination of wpandroid+submodule and knowing that it will drive on its own (trigger and wait for artifacts) while you switch to another task is quite satisfying and productive and I'd like to work on preserving to be honest.

So, if we can replicate that self-driving on CircleCI it would be really cool :)

jtreanor:
> Do you think there's something we can use from CircleCI/Bintray to replicate that flexibility?

Here is the key difference between Jitpack and a CircleCI/Bintray solution:

- Jitpack builds are created on demand (i.e. the first time Gradle tries to find that build). This means any git ref works because if it doesn't exist already, it will be built.
- With a CircleCI/Bintray solution as in this PR builds and created on push or tag. We could absolutely build for every single commit or tag but I'm not sure if that would be a good idea (the bintray repo would balloon in size).

Can explain a little more about the workflow in question? Why are binary artifacts required for arbitray refs?

> So, if we can replicate that self-driving on CircleCI it would be really cool :)

We totally can reimplement the `waitForJitpack` functionality to work with this. There should be less waiting to do however since the builds happen on push rather than when the gradle command is run.

hypest:
> but I'm not sure if that would be a good idea (the bintray repo would balloon in size)

I agree, if it means keeping all those binaries on Bintray then it doesn't sound like a good idea.

> Can explain a little more about the workflow in question? Why are binary artifacts required for arbitray refs?

Being able to build with arbitrary refs helps when trying out branches that might not have a PR with them on the WPAndroid repo, which is the usual situation in our current flow where we only open WPAndroid ones when it's clear that a feature/fix should be tested on WPAndroid directly. It's also useful when debugging to find which commit introduced a regression in gutenberg-mobile. The binary pipeline, in particular, is needed to more properly assess the performance of the editor.

All in all, being able to mix and match branches in the two repos and in an arbitrary flavor helps keep the developer experience intuitive.

jtreanor:
Thanks for the context @hypest.

We can keep this workflow in mind as we try find a solution. I think we both feel that we may outgrow Jitpack at some point, but I don't think there will be a direct drop in solution.

jtreanor:
Closing as this has been sitting around for a long time. We can revisit if needed.

Issue #911, opened 2019-04-22
Improve Screen Reader support for Inner Toolbar shared buttons.


etoledom:
All blocks have the move Up/Down and the delete buttons.
These are not accessible.

They should read the state (if it's disabled), name, type.

PR #927, opened 2019-04-24
Improve Screen Reader support for Inner Toolbar

pinarol:
Fixes https://github.com/wordpress-mobile/gutenberg-mobile/issues/911

**To test (iOS):**

Activate VoiceOver: https://github.com/wordpress-mobile/WordPress-iOS/wiki/Using-VoiceOver

**To test (Android):**

Activate TalkBack: https://www.wikihow.com/Enable-TalkBack-on-Your-Android
In my case, the system had a handy tutorial on how to use it.
If your version doesn't have it, it's actually very similar to VoiceOver.

**Test Steps**

- [ ] Check in both platforms that you hear titles and hints for Move up, Move down, Trash buttons
- [ ] Check in both platforms that you hear disabled(a.k.a. dimmed in iOS) status on Move up, Move down buttons

### DIFF
--- a/src/block-management/inline-toolbar/index.js
+++ b/src/block-management/inline-toolbar/index.js
@@ -56,17 +56,19 @@ export default class InlineToolbar extends React.Component<PropsType> {
 			<View style={ styles.toolbar } >
 				<ToolbarButton
 					accessibilityLabel={ __( 'Move up' ) }
-					label={ __( 'Move up' ) }
+					title={ __( 'Move up' ) }
 					isDisabled={ ! this.props.canMoveUp }
 					onClick={ this.onUpPressed }
 					icon="arrow-up-alt"
+					extraProps={ { hint: __( 'Double tap to move up the block' ) } }
 				/>
 
 				<ToolbarButton
-					label={ __( 'Move down' ) }
+					title={ __( 'Move down' ) }
 					isDisabled={ ! this.props.canMoveDown }
 					onClick={ this.onDownPressed }
 					icon="arrow-down-alt"
+					extraProps={ { hint: __( 'Double tap to move down the block' ) } }
 				/>
 
 				<View style={ styles.spacer } />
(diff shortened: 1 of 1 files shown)


src/block-management/inline-toolbar/index.js
```diff
@@ -56,27 +56,30 @@ export default class InlineToolbar extends React.Component<PropsType> {
 			<View style={ styles.toolbar } >
 				<ToolbarButton
 					accessibilityLabel={ __( 'Move up' ) }
-					label={ __( 'Move up' ) }
+					title={ __( 'Move up' ) }
 					isDisabled={ ! this.props.canMoveUp }
 					onClick={ this.onUpPressed }
 					icon="arrow-up-alt"
+					extraProps={ { hint: __( 'Double tap to move up the block' ) } }
```
etoledom: > 'Double tap to move up the block'

I think that "Double tap to move **the block up**" would be more correct, but I'm not sure if the original is wrong or just another valid way to say it.

We could check with some native English speaker :)

pinarol: maybe @mattmiklic can help us on this :)

mkevins: The phrase "Double tap to move up the block" is technically correct, but may be ambiguous, because, in that case, it is not clear whether "the block" is the direct object, or the object of the preposition (i.e. the thing being moved, or where we are moving to, respectively).

With the phrase "Double tap to move the block up", it is more clear that "the block" is the thing that we will move.

pinarol: Cool, thanks for the explanation @mkevins ! 🎉 Will push an update

pinarol:
Fixed the hints @etoledom merging now :shipit:

src/block-management/inline-toolbar/index.js
JavonDavis: FWIW. In the PR @etoledom mentioned I included a row number in the hint. I'm thinking this will have a positive impact on clearing up any ambiguity(?).

PR #936, opened 2019-04-26
Move unselected block accessibility handling to block-holder

pinarol:
This PR moves accessibility handling of unselected blocks to block-holder, so:

- We don't have to handle blocks separately. Every block will have at least their name(including unknown blocks) read by screen reader without extra coding.
- If we wants to provide more information for a block we'll need to inject its own `getAccessibilityContent` function as its `attribute`.
- Some block content like Separator block have very little height, it is hard to select them otherwise.

gutenberg PR: https://github.com/WordPress/gutenberg/pull/15214

Note: This is still a kind of experiment, we can decide not to merge this based on your opinions and tests.

**To Test:**

It should be enough to test in only 1 platform.
Check below items with Voice Over on:

- [ ] Image block empty
- [ ] Image block with image - selected
- [ ] Image block with image - unselected
- [ ] Next page
- [ ] Paragraph block empty
- [ ] Paragraph block with text
- [ ] Post title empty
- [ ] Post title with text
- [ ] Unrecognized block
- [ ] Separator block (Should tell its name)
- [ ] Basic operations like Updating/Closing/Opening post

### DIFF
--- a/src/block-management/block-holder.js
+++ b/src/block-management/block-holder.js
@@ -26,8 +26,9 @@ import {
 import { withDispatch, withSelect } from '@wordpress/data';
 import { compose } from '@wordpress/compose';
 import { addAction, removeAction, hasAction } from '@wordpress/hooks';
-import { getBlockType } from '@wordpress/blocks';
+import { getBlockType, getUnregisteredTypeHandlerName } from '@wordpress/blocks';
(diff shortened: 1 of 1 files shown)


pinarol:
Closing this since we are handling it in https://github.com/wordpress-mobile/gutenberg-mobile/pull/937

PR #944, opened 2019-04-30
Add instructions for running simulator with other iOS devices to README

iamthomasbishop:
This PR adds simple instructions to the README doc for running the iOS simulator with other devices.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,13 +63,29 @@ yarn android
 
 The app should now open in a connected device or a running emulator and fetch the JavaScript code from the running packager.
 
-To compile and run the iOS variant of the app, use:
+To compile and run the iOS variant of the app using the _default_ simulator device, use:
 
 ```
 yarn ios
 ```
 
 which will attempt to open your app in the iOS Simulator if you're on a Mac and have it installed.
+
+### Running on Other iOS Devices
+
+To compile and run the app using a different device, open the target device in Xcode by going to `Hardware > Device`, then use:
+
+```
+yarn ios --simulator="DEVICE_NAME"
+```
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -63,14 +63,30 @@ yarn android
...
 
 ```
 yarn ios
 ```
 
 which will attempt to open your app in the iOS Simulator if you're on a Mac and have it installed.
 
+### Running on Other iOS Devices
+
+To compile and run the app using a different device, open the target device in Xcode by going to `Hardware > Device`, then use:
```
etoledom: I believe that having the simulator running prior the `yarn` command is not strictly necessary. `yarn ios` will launch it if it's not running already.

README.md
```diff
@@ -63,14 +63,30 @@ yarn android
...
-To compile and run the iOS variant of the app, use:
+To compile and run the iOS variant of the app using the _default_ simulator device, use:
 
 ```
 yarn ios
 ```
 
 which will attempt to open your app in the iOS Simulator if you're on a Mac and have it installed.
 
+### Running on Other iOS Devices
```
etoledom: Nit: Devices sound like the real hardware. Maybe we could say "Other iOS Device Simulators"?

README.md
iamthomasbishop: Ah ok, I was having issues with launching others straight-away and this worked, so I assumed this was the "proper" way to do it – but now launching straight-away is working so I'll remove the part specific to opening prior to running the command.

README.md
iamthomasbishop: Removed – good catch!

PR #948, opened 2019-05-02
Fix move selected text on iOS

### LINKED ISSUE #843: [iOS] Editor becomes unusable after moving selection in paragraph block.
[Video showing the issue](https://cloudup.com/c1vWeZ-lAFx)
Tested on 12.2 (WP Internal)

Steps:
- Have a paragraph block with text.
- Double tap on a word to select it.
- Long tap until you can drag the selection.
  - If you are having troubles with this, try to select a longer word (or more words) so the selection is bigger, and long press in the center of the selection.
- Drag the selection anywhere else in the paragraph block.
- Release and behold.

etoledom:
fixes #843 
Maybe this one too: https://github.com/wordpress-mobile/WordPress-iOS/issues/11583

This PR fixes an issue where the editor enters on a onFocus/onBlur infinite loop. This was easily reproduced by selecting text on a paragraph block, and moving it inside the same block.

I tracked the issue until this line:
https://github.com/wordpress-mobile/gutenberg-mobile/blob/6513ba88076009b883f8b708ec48dfcbe92793b5/react-native-aztec/src/AztecView.js#L149

But we need it in order to let Gutenberg know that the TextView was focused by iOS. Needed to solve this issue: https://github.com/wordpress-mobile/gutenberg-mobile/issues/702

With some investigation I noticed that `textViewDidBeginEditing` was sending one extra `onFocus` event that is not needed, and that triggered the loop. It was solved by sending the `onFocus` from `becomeFirstResponder`, with the proper checks.

move-selected

To test:

- Run the example project on iOS
- Select a long paragraph block.
- Select some text long enough.
- Long press to move it and release it somewhere else in the block.
- Check that the editor behaves as expected.
- Check that the Hide Keyboard button stays functional after presenting the Links UI. (check for #702 )
- Smoke test on Android.

### DIFF
--- a/react-native-aztec/ios/RNTAztecView/RCTAztecView.swift
+++ b/react-native-aztec/ios/RNTAztecView/RCTAztecView.swift
@@ -123,7 +123,7 @@ class RCTAztecView: Aztec.TextView {
         delegate = self
         textContainerInset = .zero
         contentInset = .zero
-        textContainer.lineFragmentPadding = 0        
+        textContainer.lineFragmentPadding = 0
         addPlaceholder()
     }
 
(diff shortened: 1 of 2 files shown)


etoledom:
Thank you @SergioEstevao !

@mkevins or @marecar3 would you mind smoke-testing this on Android? 🙏 

The change on `AztecView.js` remove some (supposedly unnecessary) calls to `onFocus`.
Checking that onFocus is called when it's necessary on RichText should be enough (that the RichText based blocks are being focused properly on different scenarios).

etoledom:
Thank you!

PR #949, opened 2019-05-02
Fix android rich text errors

Tug:
Gutenberg PR: https://github.com/WordPress/gutenberg/pull/15392

This fixes RichText errors on merge and split.
Most of the work is in the GB PR.

Important fix in this PR:
- We remove calls to `clearSelectedBlock` on focus that would reset the block selection and thus cause error in the MERGE_BLOCKS effect that need it initialized

Other fixes:
- Refactor `mergeBlocks` in `BlockHolder` to be closer to the web version
- Mock focusing RN elements for tests

### Testing Instructions (on both Android and iOS)
- Try splitting a header block
- Try splitting a paragraph block
- Try merging a paragraph block with a header block
- Try merging 2 paragraph blocks
- Try editing text into a paragraph or a header block
- Try converting a selection into bold and/or italic
- Try writing in bold
- Try writing in italic
- Try writing in both bold and italic
- Try removing a format without selecting any text by having the cursor on formatted text and pressing on the selected format button
- Try adding a link
- Try converting a word to a link
- Try copying a link and pasting it into a selection
- Try copy pasting text


hypest:
Tried out 95f7d9a and there's an event loop happening (apparently). It happens in various cases when typing but managed to capture one when typing and deleting some text in a paragraph:

Gutenberg-loop

I think the "key" detail in order to replicate is to have some typing/deleting happening fast enough and then the issue arises.

etoledom:
Testing on iOS I found a strange behavior with the list block:
- The list block doesn't get selected.
- Pressing "Enter" in the list block, duplicates the content of the block.

list_block

I was able to reproduce the same behavior on Android.

daniloercoli:
Tests run by using Google Board on my Nexus 5X running Android 7.1 

**Test 1**
I was able to replicate a loop easily, just split a para block and start writing into it.

demo-buy

**Test 2**
I was able to replicate a "deleting loop" by tapping on a para block, start deleting content by tapping backspace and keep it pressed, then release the backspace key.

demo-buy2

hypest:
Steps to repro a merge problem:

0. Start with this initial html:
```
<!-- wp:paragraph -->
<p>split, add a char in new para, backspace it and backspace again to merge</p>
<!-- /wp:paragraph -->
```
1. Put caret inside the `split` word and hit Enter to split the block
2. Type one char (e.g. `g`) at the start of the new block
3. Backspace to delete it
4. Backspace again to merge the blocks
5. Notice the `g` being there in the middle of the merged word/block

hypest:
We've been working on some fixes and @etoledom and I have some commits to share. Will merge to this branch now.

daniloercoli:
`Undo` doesn't work as expected.

Steps to repro:
- Start the demo app
- Put the caret in a para block, and tap Enter to split the block in two
- Tap `Undo` ---> Boom!
- The "original" block is back with the full content, but the block created after the splitting is not removed
output-scaled

etoledom:
Found an issue on split (iOS and Android):

Steps to reproduce:
- Select a paragraph block in the middle of the content.
- Split the block.
- Merge back the new block.
- (Android) Split again. 
  - (iOS) Move the cursor and then Split again.
- You will see that all the content stays in the first block.

phantom_split

etoledom:
I also noticed a regression on of this issue: https://github.com/wordpress-mobile/gutenberg-mobile/issues/699

It was related to the last commit. With some debugging I noticed that we are sending the correct selection back to Aztec here:
https://github.com/WordPress/gutenberg/blob/86894eaded3c4b283033c7fa438dbaa4a6c6863b/packages/block-editor/src/components/rich-text/index.native.js#L626

And the render is triggered with the correct selection, but the caret is not updated.

daniloercoli:
I noticed a minor(?) problem when trying to remove a block by tapping `Backspace`, and erasing all the content from it.
To reproduce the problem you can follow the steps below for simplicity, but I guess there are easier ways to reproduce it:
 
- Start the demo app
- On a para block tap enter and create a new block
- Paste something in the block
- Place the caret at the end of the block
- Backspace to delete all the content
- You will see that once the caret reached the beginning of the block, and the placeholder "Start Writing" is on the screen, you need to tap the Backspace 2 times to get the block deleted.

output-scaled

This seems like a minor issue, but better to not underestimate it, since it may uncover other issues we're missing from testing.

hypest:
Will now merge the branches with the fixes I was working on over the last few days, so we can coordinate on a single branch/PR.

hypest:
There's an undo bug where an extra character is appearing, which is a web-side issue. Tracked in https://github.com/WordPress/gutenberg/issues/15619.

hypest:
The undo issue reported by @daniloercoli in https://github.com/wordpress-mobile/gutenberg-mobile/pull/949#issuecomment-490915301 is happening on the web too (I tried https://frontenberg.tomjn.com/ as well as local WP on GB hash eea04cf20 and happens there too) so, not a native mobile related issue.

koke:
Not sure if it's a regression, but the moment you press backspace, the current format gets deselected:

format-backspace

koke:
I found some odd things splitting lists and trying to undo the changes:

odd-lists

react-native-aztec/android/src/main/java/org/wordpress/mobile/ReactNativeAztec/ReactAztecText.java
```diff
@@ -84,6 +84,8 @@
     public ReactAztecText(ThemedReactContext reactContext) {
         super(reactContext);
 
+        setGutenbergMode(true);
```
marecar3: Hey @Tug @hypest I have tried to test WPAndroid against this PR and it's crashing with this error: `java.lang.NoSuchMethodError: No virtual method setGutenbergMode(Z)V in class Lorg/wordpress/mobile/ReactNativeAztec/ReactAztecText`

I can see that we are targeting Aztec version `aztecVersion = 'gb-stray-selection-fix-try1'` but when I compare this branch: https://github.com/wordpress-mobile/AztecEditor-Android/tree/gb-stray-selection-fix-try1 with develop it seems they are identical: https://github.com/wordpress-mobile/AztecEditor-Android/compare/gb-stray-selection-fix-try1

I assume that we are missing some commits from that branch?

etoledom:
> I found some odd things splitting lists and trying to undo the changes:

This seems to be fixed by: https://github.com/wordpress-mobile/gutenberg-mobile/pull/973

> Not sure if it's a regression, but the moment you press backspace, the current format gets deselected:

Just tested this and it's working good now! 🎉 

With the last updates, I found a small issue on iOS: Sometimes the formats won't go away by pressing the format buttons.

It's easiest to reproduce in the lists block, but it also happens in paragraph. I couldn't reproduce it in Android, just iOS:

formats

I can take a look at this.

etoledom:
The format library doesn't seem to recognize bold with `b` tags, that's true for the web version too.
This means that our `<p><b>Hello</b> World!</p>` paragraph block in the initial example content is "buggy", but this issue is not generated in our end.

PR #957, opened 2019-05-03
Merge Release/1.4.0 into develop

loremattei:
This merges Release/1.4.0 into develop.


pinarol:
This PR won't change the gutenberg hash because I am reverting the gutenberg hash update with commit d31b4b6 . That's because we used a temporary branch for this particular release and don't want to refer to that branch from develop too.

PR #959, opened 2019-05-06
Improve accessibility on missing block

### LINKED ISSUE #950: Improve screen reader support for Unsupported Block
Screen reader should be able to select this block (currently it's not selectable).
It should read something like:
"Unsupported block. {Block name}". (i.e: Unsupported block. Gallery).

koke:
Fixes #950 

## To test

With VoiceOver/Talkback activated

1. Create a post with some unsupported blocks (gallery, contact form)
2. Tap on the unsupported gallery block, it should read `Unsupported block: Gallery`
3. Tap on the unsupported contact form (any non-core block), it should read `Unsupported block`


koke:
I’ve tested this on iOS but not yet on android

PR #964, opened 2019-05-07
CircleCI: Limit max Jest and Flow workers to avoid memory errors

jtreanor:
Recently there have been a significant number of both jest and flow failures on CircleCI. This limits parallelization of both so that memory problems are avoided.

I have run this many times already, and it seems to be working reliably (see [here](https://circleci.com/gh/wordpress-mobile/gutenberg-mobile/tree/circleci-memory-issues)).

Related to https://github.com/facebook/flow/issues/6242

### DIFF
--- a/bin/ci-checks-js.sh
+++ b/bin/ci-checks-js.sh
@@ -14,8 +14,8 @@ if [[ -z "${CHECK_CORRECTNESS}" ]] && [[ -z "${CHECK_TESTS}" ]] ; then
 fi
 
 if [ "$CHECK_CORRECTNESS" = true ] ; then
-  yarn run flow || pFail
-  yarn run lint || pFail
+  yarn flow || pFail
+  yarn lint || pFail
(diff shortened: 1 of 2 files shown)


bin/ci-checks-js.sh
```diff
@@ -14,8 +14,8 @@ if [[ -z "${CHECK_CORRECTNESS}" ]] && [[ -z "${CHECK_TESTS}" ]] ; then
 fi
 
 if [ "$CHECK_CORRECTNESS" = true ] ; then
-  npm run flow || pFail
-  npm run lint || pFail
+  yarn run flow || pFail
```
hypest: What do you think about using the `yarn <script name>` form instead of the `yarn run <script name>` one? The former is the one we typically use in the project.

jtreanor: Makes sense 👍

PR #967, opened 2019-05-09
Fix crash on undefined image url

mkevins:
## Description
This PR updates the Gutenberg reference to bring in the changes in this companion PR: https://github.com/WordPress/gutenberg/pull/15532.

This PR adds a check to ensure `attributes.url` exists before attempting to call `indexof`. Currently, there are scenarios which can result in `attributes.id` being truthy, while `attributes.url` is null or undefined for an image block.

When an image is cancelled before it is finished uploading, the `url` is set to `null`, but the `id` of the block remains. The next time the post is edited, the local draft will contain the state which triggers the crash.

## Testing steps:
This was tested using the steps here: https://github.com/wordpress-mobile/WordPress-Android/issues/9768#issuecomment-490852531

## Types of changes
This is a bug fix for the crash discovered here: https://github.com/wordpress-mobile/WordPress-Android/issues/9768


hypest:
Hey @mkevins , I guess the `v1.4.0` part of the branch name should be bumped to `1.4.1` since 1.4.0 is already deployed. One way to address this without creating a new PR is to make this PR target a `hotfix/1.4.1` branch instead of `master`. If and when we decide to deploy that as a "full" hotfix, we can open a PR from `hotfix/1.4.1` to `master` to track it. How does that sound?

mkevins:
I've created the hotfix branch with the correct version, and now merged this branch to that one. I have opened a PR on that branch targeting master.

Issue #874, opened 2019-04-16
List block split problem with nested lists


hypest:
List item creation has an issue that's not easy to describe. Look below for steps to reproduce.

To replicate:

(0. Using an external keyboard (like via Vysor) on Android)
1. Use an example html with nested lists like this one:
```
<!-- wp:list -->
<ul><li>line 1<ul><li>line 21</li><li>line 22<ul><li>line 31</li><li>line 32</li></ul></li></ul></li><li>line 3</li><li>line 4</li><li>line 5</li></ul>
<!-- /wp:list -->
```
2. Place the caret at the end of the "line 22" line and tap on Enter
3. Notice the caret going at the start of the "line 31" line.
4. Place the caret at the end of the "line 31" line and tap Enter again
5. Notice a new list item appearing above the "line 32" line but, having a `1` char in it, which is now missing from the line above leaving that as "line 3"

hypest:
👋 @koke , I think this issue is annoying but not a blocker for the code-freeze to re-enable lists, do you think otherwise perhaps?

koke:
I don't think it's a blocker (we have disabled the nested list buttons, right?), but let's put it on the project and get it sorted out for v2.

SergioEstevao:
@hypest I tested this on iOS and it now looks ok, are you still seeing this on Android?

hypest:
The issue is still happening on Android when using an external keyboard, like Vysor. But, step 3 doesn't happen when using the built-in virtual keyboard. Step 5 though does happen always on Android, even with the virtual keyboard.

On iOS, the issue still happens when running the demo app on an iPhone X 12.2 simulator. Gutenberg-mobile hash: 42d8443f

![nested-lists-enter-erroneous](https://user-images.githubusercontent.com/1032913/56969233-69137180-6b6d-11e9-8e10-6e3ed45cf107.gif)

SergioEstevao:
> The issue is still happening on Android when using an external keyboard, like Vysor. But, step 3 doesn't happen when using the built-in virtual keyboard. Step 5 though does happen always on Android, even with the virtual keyboard.
> 
> On iOS, the issue still happens when running the demo app on an iPhone X 12.2 simulator. Gutenberg-mobile hash: [42d8443](https://github.com/wordpress-mobile/gutenberg-mobile/commit/42d8443fd74a1837f70f20689182d700cc47f31f)
> 
[...]

PR #973, opened 2019-05-10
Issue/874 fix nested lists handling

SergioEstevao:
Fixes: #874

Related GB PR: https://github.com/WordPress/gutenberg/pull/15566
Relate Aztec PR: https://github.com/wordpress-mobile/AztecEditor-iOS/pull/1181

This PR updates the list code to support the following:
 - Activates the indent/outdent buttons on the UI for the list block
 - Improves the handling of the Enter character and make sure the format of text between Aztec and the format library is consistent 
 - Makes the handling of delete of new lines to be handled by the rich text format structure, instead of the native side.

How to test:
 - First of all try, the issue stated on #874 to see if it's working correctly
 - Then try to use the indent and outdent buttons and see if they work correctly
 - Try to delete list items and see if it works correctly


etoledom:
@SergioEstevao - Thanks for handling this!

I did a smoke test on the issue, and it seems to be resolved 🎉 
There is one detail that I found:
- Press enter on a nested list item to create a new empty item.
- Try to delete the new list, it removes the indentation (that looks OK), but also creates a new list item with the same indentation as before, next to the current one (that does not look good).

Check the gif for a better visual explanation:

delete_list_item

etoledom:
The web side has a similar (but not completely the same) behavior, so it's probably an issue in the web side that is affecting us.

web_list_issue

SergioEstevao:
I saw similar things, with the system showing two bullets in the same line,
I will check the HTML produced and if what the app is showing makes sense,
I will merge this.

SergioEstevao:
@etoledom I just printed the HTML being produced by the JS side and the HTML we are producing in Aztec looks right. Not sure what to do.

SergioEstevao:
Closing in favour of: https://github.com/wordpress-mobile/gutenberg-mobile/pull/991

PR #976, opened 2019-05-12
Fix crash on undefined image url

mkevins:
## Description
This PR updates the Gutenberg reference to bring in the changes in this companion PR: https://github.com/WordPress/gutenberg/pull/15532.

This PR adds a check to ensure `attributes.url` exists before attempting to call `indexof`. Currently, there are scenarios which can result in `attributes.id` being truthy, while `attributes.url` is null or undefined for an image block.

When an image is cancelled before it is finished uploading, the `url` is set to `null`, but the `id` of the block remains. The next time the post is edited, the local draft will contain the state which triggers the crash.

## Testing steps:
This was tested using the steps here: https://github.com/wordpress-mobile/WordPress-Android/issues/9768#issuecomment-490852531

## Types of changes
This is a bug fix for the crash discovered here: https://github.com/wordpress-mobile/WordPress-Android/issues/9768

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "gutenberg-mobile",
-  "version": "1.4.0",
+  "version": "1.4.1",
   "private": true,
   "config": {
     "jsfiles": "./*.js src/*.js src/**/*.js src/**/**/*.js",


mkevins:
I fixed the hotfix version number and merged [the previous PR](https://github.com/wordpress-mobile/gutenberg-mobile/pull/967) into this one as per this comment: https://github.com/wordpress-mobile/gutenberg-mobile/pull/967#issuecomment-491167524.

hypest:
Headsup, let's hold off merging this one since there is a PR that got merged to the main apps that brings in the latest of gutenberg-mobile develop (which is currently broken though).

Also, please open a WPAndroid PR as well so we can track this better, thanks!

mkevins:
I have opened the PR on WPAndroid to update the submodule reference there to this one. The issues blocking the develop branch should be non-blocking here, as this PR targets master as a hotfix (assuming we go forward with that).

PR #977, opened 2019-05-13
[WIP] Moving tests to gutenberg repo.

etoledom:
This modifies jest config to search for tests inside a `__native_test__` foder, and inside our local `src` folder


etoledom:
Update:

Now there is no need for `__native_tests__` folder.
- `gutenberg` side tests will need to be post-fixed with `.native.js` and inside a `test` folder. Wit this:
  - Our tests will be found by our jest instance.
  - Our tests won't be run by the web's `test-unit` suit.
  - We can use the current `test` folder already existing for many components.
- For local tests, the only new requirement is for them to be inside the main `src` folder. Beside either:
  - To be postfixed with `test.js`. or
  - To be inside a `__tests__` folder.

I don't particularly like that for local tests we have `__tests__` folder and for web we have `test` folder. In any case, we should move all those tests at some point to their corresponding component.

@hypest @koke @mkevins - any feedback?
If it looks good, I'll move the rest of the tests and prepare a proper PR.

etoledom:
I'm assuming for this step to not have our native test suit run by gutenberg CI.

On a next step we can set up a new test config on gutenberg and have an extra CI step running our tests.
We will be able to test that our components and the gutenberg code works well together.

koke:
So if I understand correctly, the code for the test lives in the Gutenberg repo, but we're still launching jest from gutenberg-mobile right?

I think it's a good first step: we move this and make sure nothing breaks in their toolchain, then do a second PR to introduce a native jest config to run the native tests. Is that the plan?

If I understood everything correctly, it sounds good to me 👍

etoledom:
> I think it's a good first step: we move this and make sure nothing breaks in their toolchain, then do a second PR to introduce a native jest config to run the native tests. Is that the plan?

Exactly, that's the plan!
I'll continue with this direction then 👍

koke:
I have one question though, because looking at [the paragraph test](https://github.com/WordPress/gutenberg/blob/59242f5edd421844a96c1d95f18957c59bf64422/packages/block-library/src/paragraph/test/paragraph.native.js), it looks like there's nothing specific to mobile there.

So I'm wondering how we can have tests that would eventually run in both the web and native test suites. Does it make sense to separate the two for unit tests, since they don't actually run in the simulator? What am I missing here?

etoledom:
> Does it make sense to separate the two for unit tests, since they don't actually run in the simulator?

I haven't really tried out to successfully run our tests with `test-unit`, but for example our jest config is quite different than the web's jest config.

For example, there is a jest preset special for `react-native`, and there are some dependencies on `node_modules/react-native`. I haven't looked in to this yet, so I'm not sure how much of this is necessary, it might depend on what component we are testing and how is it constructed (with other custom components, or with react-native components directly for example).

I personally think that it's also good to have them apart because they are different things, and it will make clear that a change on web is breaking something on mobile (for example).

@Tug was also in favor of having them apart.

If there are good reasons to have them all together, and it's technically possible, I'm not really agains that either.

Tug:
Eventually yes we might be able to run them simultaneously.
But I agree with what @etoledom said: we have a custom jest config and mocks at the moment so it's best to keep them separate.

mkevins:
> I haven't really tried out to successfully run our tests with test-unit, but for example our jest config is quite different than the web's jest config.

When I drafted this: https://github.com/WordPress/gutenberg/pull/15552/files, that was the approach I was using (i.e. using web's test pipeline to run the unit tests for mobile). My rationale is that since js unit tests will be totally isolated, with all dependencies mocked, they can assimilate well to the testing environment already present on `gutenberg`.

> So I'm wondering how we can have tests that would eventually run in both the web and native test suites. Does it make sense to separate the two for unit tests, since they don't actually run in the simulator? What am I missing here?

@koke I think this does make sense for unit tests (i.e. one test for both web and mobile) when we are using the same file (implementation). I think we also need separate unit tests for the native-specific implementations in those `*.native.js` files, which plays well with the conventions used in `gutenberg` where unit tests are organized by the filename.

koke:
> I think we also need separate unit tests for the native-specific implementations

I'm still trying to wrap my head around this. When there's a native JS file, it's more clear that we'll have a native test. For an example of the kind of thing I'm worried about, let's use the [Dashicon test](https://github.com/WordPress/gutenberg/blob/master/packages/components/src/dashicon/test/index.js) as an example.

- The test imports [Dashicon](https://github.com/WordPress/gutenberg/blob/master/packages/components/src/dashicon/index.js), which has the same implementation for web and mobile
- However there is a customization in [icon-class.js](https://github.com/WordPress/gutenberg/blob/master/packages/components/src/dashicon/icon-class.js)/[icon-class.native.js](https://github.com/WordPress/gutenberg/blob/master/packages/components/src/dashicon/icon-class.native.js)
- In this specific example, the behavior is still similar enough that the same [test case](https://github.com/WordPress/gutenberg/blob/master/packages/components/src/dashicon/test/index.js) would pass regardless of running on web or native, but there could be enough differences in `icon-class` to make it fail on just one platform.

Lookin at an example like this, I think we would eventually want most tests to run on both a web configuration and a mobile one on travis, but there should be room for web-only tests and native-only tests where it makes sense. Native-only we can do with the `.native.js` extension, but do we have a `.web.js` equivalent for web-only tests?

etoledom:
I see what you mean. I haven't  thought on a test suit that will trigger the same test file win a web and a mobile configuration. 
This idea was more like the web-only (test-unit), and mobile only (test-native) suits. Probably a 3th step?

The Dashicons is a good example of a test that will be beneficial to be triggered with both configs, there might not be many others thought (?). I was thinking about multiplatform blocks but they don't seem to have this same kind of tests. They have snapshots tests or don't have tests at all.

We can continue investigating these cases while we go into that direction.

mkevins:
> In this specific example, the behavior is still similar enough that the same test case would pass regardless of running on web or native, but there could be enough differences in icon-class to make it fail on just one platform.

I see what you mean. The Dashicon scenario would not exactly fit the convention of the PR I had drafted, and I believe in that case, the web pipeline would run the test and would ignore the `*.native.js` import. It is probably worth hashing out what exactly we mean by "unit" in "unit-testing", as I know there are many philosophies around that. On one end of the spectrum, the unit-under-test would mock _all_ dependencies, which in this case, would include `icon-class`. This level of granularity may be impractical in our case, but it would be interesting and I think helpful to define what is a "unit" in this case.

Either way, I think we need a solution for the different environments for integration testing. I think something can be achieved via some clever path/filename conventions combined with the use of regex in the jest configs, but I haven't explored this extensively.

etoledom:
Updated moving all the test files related to `gutenberg` side components to the `gutenberg repo.

cc @pinarol

mkevins:
@etoledom I have adjusted my PR to work with the test config you are bringing to `gutenberg` with this PR and the companion PR: https://github.com/WordPress/gutenberg/pull/15552/files. One issue I encountered is that any shallow mocks / stubs in a local `__mocks__` directory will "override" the mocks in our mobile config. Mostly, I don't see this as a "risk", since web folks wouldn't likely mock mobile specific implementations, but I am not sure if I've considered all possibilities. What do you think?

koke:
This looks good to me, but the Gutenberg PR is still marked as WIP, and this one has conflicts. Is this good to review/merge?

mkevins:
> This looks good to me, but the Gutenberg PR is still marked as WIP, and this one has conflicts. Is this good to review/merge?

@etoledom I think the changes to `src/__tests__/paragraph.test.js` are from the recent `onSplit` PR. Maybe these can simply be patched on gb side? The only other conflict is the gb hash, which can be resolved by merging develop to this branch first.

etoledom:
> but the Gutenberg PR is still marked as WIP
My mistake. Fixed!

> and this one has conflicts. Is this good to review/merge?
Conflicts fixed. It should be ready for review.

> Maybe these can simply be patched on gb side?
Thank you for the heads up!
I updated the test on `gutenberg` with the last paragraph test version seen on `gutenberg-mobile`

hypest:
Even though the work here seems in a good place, I'll bump the milestone to the next one to help unblock the 1.6 code freeze (due today).

etoledom:
@hypest @koke any insight here? Should we come back to this after v2?

Issue #828, opened 2019-04-05
After pasting, cursor should move to the end of the pasted content


koke:
Currently, the keyboard is dismissed but the caret remains at the beginning of the paragraph.

The keyboard should stay visible and the cursor should be at the end of the pasted content

hypest:
Related PR that may be affecting this issue: https://github.com/wordpress-mobile/gutenberg-mobile/pull/629

mkevins:
Thanks @hypest !

mkevins:
There are a few things going on in this issue that I'd like to separately identify.

1. When multi-line plain text is pasted, the pasted text _always_ begins in its own (new) paragraph block, even when the clipboard contents do not begin with a `\n` (newline).
2. The caret position after paste is not at the end of the pasted text.
3. The caret position after paste is not in the correct block (the wrong block is active / focused).

Here are some steps to reproduce:

1. Copy multi-line text to the clipboard (e.g. `"Paragraph 1.\n\nParagraph 2.\n"`).
2. Place caret at the end of the "**Hello** World!" block.
3. Tap the caret handle and paste the content.

**Expected result:**

* "Paragraph 1." is appended to the text in the "**Hello** World!" paragraph block to read: "**Hello** World!Paragraph 1."
* A new paragraph block is created below with the text: "Paragraph 2."
* The new paragraph block is focused / active
* The caret is in the new paragraph block, at the end: "Paragraph 2.|" (where '|' == caret).

**Actual result:**

* Text is not appended to "**Hello** World!" paragraph block.
* Two new paragraph blocks are created with the text "Paragraph 1." and "Paragraph 2.", respectively.
* The first of the new paragraph blocks is focused / active (i.e. the "Paragraph 1." block).
* The caret is at the beginning of the text: "|Paragraph 1.".

**Screencast:**

<img src="https://user-images.githubusercontent.com/8507675/55779305-6a122f80-5ae8-11e9-9d4d-57c6c55a78a5.gif" width="360">

mkevins:
A simpler case of this issue exists for pasting inline content as well.

Here are some steps to reproduce:

1. Copy inline text to the clipboard (e.g. "**Hello** World!").
2. Place caret just before the 'W' in the end of the "**Hello** World!" block: "**Hello** |World!".
3. Tap the caret handle and paste the content.

**Expected result:**

[...]

PR #995, opened 2019-05-17
Fix caret position after inline paste - after selection refactor

### LINKED ISSUE #903: Fix caret position after inline paste - is broken
As @mkevins reported :

After inline paste, the caret position should be "incremented" by the length of the inserted text. This is set explicitly during the paste handling via the forceSelectionUpdate method. A safety check was added to the render method to prevent an out of bounds exception that can occur when the caret is moved to a position that is larger than the text buffer, which may happen as a result of Aztec trimming spaces in the content.

For the purpose of checking this condition, the willTrimSpaces method https://github.com/WordPress/gutenberg/pull/15021/files#diff-4828a21853e899e5a36faecfa96d83e8R631 was introduced, which tries to evaluate whether Aztec will trim spaces, and if so, prevents setting the caret position. The regex that looks for outer spaces (that will be trimmed) also matches on any spaces adjacent to html tags (including inner tags). The result is that whenever the html contains inner tags adjacent to spaces (e.g. a bold word), the caret will not be in the correct position after paste.

mkevins:
This PR updates the `gutenberg` reference to bring in changes from the [companion PR](https://github.com/WordPress/gutenberg/pull/15701).

## Note
This PR is part of a series of PRs to fix the same issue. Most recently, a change in ownership of selection state has brought about an important refactor which regressed the previous PR before it was merged: https://github.com/WordPress/gutenberg/pull/15552. The solution there is no longer viable, but the unit test in that PR may still serve as a helpful tool to prevent more regressions.

## Description
This PR is intended to fix a regression of _part_ of this issue: https://github.com/wordpress-mobile/gutenberg-mobile/issues/828 , specifically, when inline content is pasted. This will resolve #903.

## How has this been tested?
This has been tested using the steps here:
https://github.com/wordpress-mobile/gutenberg-mobile/issues/828#issuecomment-481546078

## Types of changes
This is a bug fix, but currently only resolves a _part_ of the original issue. It serves as an incremental improvement. Also, this PR renames some local variables for clarity.


hypest:
Bumping to next milestone as this still needs a review and it's not a blocker for the v1.6 release.

PR #997, opened 2019-05-17
Fix crash occurring when we try to display block picker

pinarol:
This crash occurs when we we try to show block picker. It is not yet on the develop branch but it will come in when we update the gutenberg ref to the latest master. This looks like a side effect of: https://github.com/WordPress/gutenberg/pull/15462/files

Screen Shot 2019-05-17 at 20 03 20

To test:

- Tap on (+) and display the block picker
- Verify it doesn't crash
- Verify Heading block has a new icon

Simulator Screen Shot - iPhone X - 2019-05-17 at 21 15 52

Update release notes:

### DIFF
--- a/src/block-management/block-picker.js
+++ b/src/block-management/block-picker.js
@@ -12,7 +12,7 @@ import { FlatList, Text, TouchableHighlight, View } from 'react-native';
 /**
  * WordPress dependencies
  */
-import { SVG } from '@wordpress/components';
+import { SVG, Dashicon } from '@wordpress/components';
 import { BottomSheet } from '@wordpress/block-editor';
 import { Component } from '@wordpress/element';
 import { getBlockTypes, getUnregisteredTypeHandlerName } from '@wordpress/blocks';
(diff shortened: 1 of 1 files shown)


Tug:
I think it might be a good opportunity to switch to using the [BlockIcon](https://github.com/WordPress/gutenberg/blob/master/packages/block-editor/src/components/block-icon/index.js) component.
It has pretty much the same requirements. It accepts an `icon` props which can be both a string or an svg component.
And it's actually the component used to render the `icon` attribute of a block.
We should be able to override the foreground color for all blocks by passing a third argument `className` which will have the `color` we want.

koke:
> I think it might be a good opportunity to switch to using the BlockIcon component.

I agree, but if this solution works, we should merge it first to unblock master, then move to BlockIcon

pinarol:
hey @Tug 👋 `BlockIcon` doesn't seem cross platform yet. I also tried with Icon but `className` won't work on mobile too, also `style` just isn't passed to the inner Dashicon or SVG.

```
<Icon icon={ icon.src } size={ styles.modalIcon.width } style={ { fill: color } } className={ "components-picker_icon" }/>
```

So it looks like these components need a bit modification to be able to achieve the results we want. Please let me know if I am missing sth. 

I agree with @koke about not blocking the master, we can make other enhancements on a separate PR, wdyt?

PR #1002, opened 2019-05-20
Hotfix/v1.5.1

### LINKED ISSUE #1001: [Android] Crash on undo on v1.5.0
Steps to reproduce:

1. Run the demo app
2. Put the caret at the end of the "Hello world" paragraph block
3. Type in one character
4. Tap on "undo"
5. 💥  with:
```
Error while updating property 'text' of a view managed by: RCTAztecView

null

setSpan (13 ... 13) ends beyond length 12
```

hypest:
Fixes #1001 

Gutenberg side PR: https://github.com/WordPress/gutenberg/pull/15734

This is the PR to track the v1.5.1 hotfix release. It's a "hotfix" because we're building on top of v1.5.0 without bringing in any new developments from `develop` or Gutenberg's master.

To test the fix for the particular issue:
1. Run the demo app
2. Put the caret at the end of the "Hello world" paragraph block
3. Type in one character
4. Tap on "undo". App should work and the extra character should be removed
5. On debug modes, a yellow warning should appear, mentioning that the selection will be out of bounds and will skip setting it.

Update release notes:

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "gutenberg-mobile",
-  "version": "1.5.0",
+  "version": "1.5.1",
   "private": true,
   "config": {
     "jsfiles": "./*.js src/*.js src/**/*.js src/**/**/*.js",


hypest:
Even though the Gutenberg hash got updated, it only moved to point to an update of lint style fixes so no need to update the JS bundles. Actually, if one tries to rebuild the bundles they will notice that the exact same bundles are produced.

In the meantime, the sibling PRs on WPAndroid and WPiOS are now approved so, will go ahead and merge this one (but not the Gutenberg side one, see note in https://github.com/WordPress/gutenberg/pull/15734).

PR #1012, opened 2019-05-21
Fix react hook error on block selection

### LINKED ISSUE #1009: Error while trying to select an image block
We have some regressions with the [last PR](https://github.com/wordpress-mobile/gutenberg-mobile/pull/979) since it brought other changes from gutenberg `master`.

I’ve seen this one when trying to select a non-empty image block or picking an image from the library (and dismissing the modal since it wasn't working), tested on android:
```
Invariant Violation: Hooks can only be called inside the body of a function component. (https://fb.me/react-invalid-hook-call)

This error is located at:
    in FillComponent (at fill.js:73)
    in Fill (at slot-fill/index.js:13)
    in BlockControlsFill (at block-controls/index.js:14)
    in BlockControlsFill (at context.js:57)
    in IfBlockEditSelected(BlockControlsFill) (at edit.native.js:297)
    in RCTView (at View.js:45)
    in View (at edit.native.js:294)
    in TouchableWithoutFeedback (at edit.native.js:281)
    in ImageEdit (at edit.native.js:18)
    ...
```

Tug:
Fixes #1009

This PR cleans up our RN build and makes sure we don't include react twice. The actual fix is executing `gutenberg/node_modules` using `resolver.blacklistRE` in `rn-cli.config.js`

To test:
- Checkout this branch locally
- Run `cd gutenberg && npm install`
- Boot the demo app on android `yarn start:reset` and `yarn android`
- Select an image block
- The app should not crash


SergioEstevao:
Couldn't we just delete the react package from the node_modules folder?

hypest:
Do we know how the `gutenberg` subfolder ends up having a `node_modules` folder?

koke:
Also, should we add `gutenberg/node_modules` anyway to the ignore lists of metro/jest?

Tug:
> Couldn't we just delete the react package from the node_modules folder?

I tried (`@wordpress/element having all the required `react` and `react-dom` dependencies we could in theory remove those) but it doesn't work, somehow it can find both at runtime but at bundle time it fails.

> Do we know how the gutenberg subfolder ends up having a node_modules folder?

Yes, you need to `npm install` when you're developing for `gutenberg` and need to commit something, there are pre-commit hooks running. Also we might want to run tests.

> Also, should we add gutenberg/node_modules anyway to the ignore lists of metro/jest?

I tried to edit `module.system.node.resolve_dirname` in `.flowconfig` (I think that's what metro is reading right?) replacing `node_modules` with `./node_modules` but it didn't seem to work (I think node_modules is a special case, it works even if you don't configure it, so we would need a way to exclude it and I haven't found one). Regarding `jest` it seems to be working properly at the moment.

koke:
I was thinking of this, but I’m still not really sure how all the toolchain bits resolve things: https://github.com/wordpress-mobile/gutenberg-mobile/blob/25109f3d559c577af8e8605289c17a1ec55c1348/rn-cli.config.js#L15

Tug:
@koke Indeed it seems to be working 👍 

Updating the PR

etoledom:
Tested on iOS and Android after running `npm install` inside gutenberg submodule and everything seems to work.

A local `yarn test` seems to work as well and from this PR, CI is happy too ✅ 
`yarn test:e2e:ios:local` also runs happily.

android/app/build.gradle
```diff
@@ -145,7 +145,7 @@ repositories {
 
 configurations.all {
     resolutionStrategy {
-        force 'org.webkit:android-jsc:r236355'
+        force 'org.webkit:android-jsc:r241213'
```
hypest: 👋 @Tug , can you expand on the need to upgrade the JSC? Thanks!

Tug: No need other than for the sake of being up to date.
I explored that first because I thought there was some issue with react-native but found out it wasn't the cause. 
In any case I can revert that if you don't think it's worth upgrading.

hypest: > In any case I can revert that if you don't think it's worth upgrading.

Let's do that in a separate PR although I'd put low priority to that one anyway.

Should I assume that the other package version upgrades are also in a similar vein to upgrade in the sake of being up-to-date? If so, I'd also prefer those to be in separate PRs.

Long story short, if the only fix we need for the `hooks` is the deletion of the GB nested `node_modules` folder, let's only have that one in the PR. It makes things more clear, traceable and easier to merge.

Tug: >I'd also prefer those to be in the separate PR.
👌

Tug:
@hypest Ok I've reduced the PR to the only change that actually fixes the issue. Could you give it a go now?

PR #1017, opened 2019-05-22
Refactor icon customization in block picker

pinarol:
Fixes https://github.com/wordpress-mobile/gutenberg-mobile/issues/1004

gutenberg PR https://github.com/WordPress/gutenberg/pull/15778

To test:

Display the block picker and make sure all icons have the same color: `$gray-dark` and same size: `32px`

Screen Shot 2019-05-22 at 19 42 53

Update release notes:

- [x] If there are user facing changes, I have added an item to `RELEASE-NOTES.txt`.

PR #1021, opened 2019-05-23
UI for invalid block content

### LINKED ISSUE #969: UI for invalid blocks
Show a UI similar to the unsupported block if a block fails validation

For reference, this is what the web displays:

screen-shot-2019-05-11-at-15 30 58

For 2.0, we just want to show a non-actionable message.

etoledom:
Fixes #969 

This PR implements `BlockInvalidWarning` on `BlockHolder` to show a warning UI when the block content is invalid.

`gutenberg` side PR: https://github.com/WordPress/gutenberg/pull/15789

IMG_1853

To test:
- Run the example app.
- Display on gutenberg the HTML content from [here](https://gist.github.com/etoledom/a47cfe010a0f5c9f410d5c226b1560b3)
- You will see a red screen (since there is invalid content).
- Dismiss the red screen.
- Check that the invalid content UI is showing for all blocks. 

Update release notes:

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -1,2 +1,3 @@
 1.6.0
 ------
+* Added UI to display a warning when a block has invalid content.
--- a/src/block-management/block-holder.js
+++ b/src/block-management/block-holder.js
@@ -27,7 +27,7 @@ import { withDispatch, withSelect } from '@wordpress/data';
 import { compose } from '@wordpress/compose';
 import { addAction, hasAction, removeAction } from '@wordpress/hooks';
 import { getBlockType } from '@wordpress/blocks';
-import { BlockEdit } from '@wordpress/block-editor';
+import { BlockEdit, BlockInvalidWarning } from '@wordpress/block-editor';
 import { __, sprintf } from '@wordpress/i18n';
 /**
  * Internal dependencies
@@ -197,7 +197,7 @@ export class BlockHolder extends React.Component<PropsType, StateType> {
 						style={ [ ! isSelected && styles.blockContainer, isSelected && styles.blockContainerFocused ] }>
 							{ isValid && this.getBlockForType() }
 							{ ! isValid &&
-								<Text>ERROR</Text>
+								<BlockInvalidWarning />
 							}
 					</View>
 					{ this.renderToolbar() }
--- a/src/block-management/block-holder.js
+++ b/src/block-management/block-holder.js
@@ -175,11 +175,13 @@ export class BlockHolder extends React.Component<PropsType, StateType> {
 	}
 
 	render() {
-		const { isSelected, borderStyle, focusedBorderColor, isValid } = this.props;
+		const { isSelected, borderStyle, focusedBorderColor, isValid, name } = this.props;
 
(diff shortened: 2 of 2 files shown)


etoledom:
@iamthomasbishop - I decided by option 1 from https://github.com/wordpress-mobile/gutenberg-mobile/issues/969#issuecomment-493197167 since it's the simplest to implement for now.

I wasn't able to implement the exact style since I didn't found it on Zeplin, but I'm happy to update for any style change request 👍

iamthomasbishop:
Already looks really close design-wise, nice work @etoledom! And I figured option 1 would be the easiest, good call.

The design is the same as the inner part of the Unsupported/Placeholder block, but with an added description instead of the placeholder action button. I will drop you the link directly to the component in Zeplin via DM 😄

hypest:
Looks good @etoledom ! 🎉 

I'm only concerned about the combination of UI + red screen + yellow warnings I'm seeing with this. The UI part is fine and makes sense but, I'm a bit annoyed by the red screen (since we are already handling the error). We've discussed this over chat of course so not going to go deeper into this part. What it's not clear to me is what are the warnings popping up and whether, as a developer, I will need to dive into those to investigate when I see them occuring in the demo app. Here's a screenshot from the demo app with the test content (as described in the PR description):

screenshot-1558686843742

So, if the warnings there are expected, can we find a way to consolidate with the red-screen too? It's too verbose to have so many indications of the error if they actually stem from the same error.

etoledom:
Hey @hypest - thanks for checking it out!

The error comes from

https://github.com/WordPress/gutenberg/blob/68030994aedb67f448acb148af989d4fdb871e87/packages/blocks/src/api/validation.js#L641

And the warnings come from:

https://github.com/WordPress/gutenberg/blob/68030994aedb67f448acb148af989d4fdb871e87/packages/blocks/src/api/validation.js#L609

---

We could downgrade the error into a warning, getting this result:

IMG_1862

We still get multiple warnings, but it's explicit that it's about Block validation.
If we go this way, I'd prefer to do the change in a separated PR, so we can focus the conversation and feedback about that change in particular, and also it won't delay the completion of this PR.

---

Alternatively, we can deactivate the red screen for errors. It will still display for Exceptions.
The downside is that we won't have visual feedback of errors anymore, just in the JS console. I'd prefer to not go on this direction.

hypest:
Happy to have any improvement on this one in a separate PR @etoledom , no prob 👍 .

The yellow warning was something that I think we didn't discuss over chat when we talked about the redscreen, so it took me by surprise... not trying to bring the redscreen conversation up again.

etoledom:
Thank you @iamthomasbishop for the design specs!
I updated applying them. This is the final result:

invalid

Note that the media icons are black. This will be easily fixed after this PR is merged: https://github.com/wordpress-mobile/gutenberg-mobile/pull/1017

iamthomasbishop:
Perfect! Thanks for the attention to detail! :shipit:

hypest:
The failing Android device tests seem to fail due to some focus issues and are not related with the code and also happen on current develop (cdf14534) so, will ignore the failure.

koke:
I didn't get here in time to review, but 😍

PR #1022, opened 2019-05-23
List end test

hypest:
This PR adds an Appium test for testing the list block ending when inputting an "Enter" at an empty list item where the expectation is that that block should "end" and a new empty paragraph should be created underneath it.

### Things to note

1. The test only really tests the Android side of things as on the iOS we're still missing an implementation of the html check step. See https://github.com/wordpress-mobile/gutenberg-mobile/blob/876ebe714219f4f0ba3ab23e82e535747db041e6/__device-tests__/pages/editor-page.js#L272-L278
2. The new test is added in a _new_ testsuite module (a file separate to the other list test module) because we currently don't have a good way to clean up the state and blocks before starting the next test case. With a separate testsuite, the whole editor is cleaned up by tearing it down and re-starting it. cc @JavonDavis since you've extensively worked on the Appium tests and you might have some ideas and insights on this.

To test:

* Let the CI makes its pass and observe the SauceLabs output to verify that the list ending test runs fine. Edit: the CI run has finished for Android and here's a quick link to the video if the test: https://app.saucelabs.com/tests/340a973602ff40a28c39bd6462e9b778?auth=230dbdb64a11ce68fe91a0d14877cd11

Update release notes:

### DIFF
--- a/__device-tests__/gutenberg-editor-lists-end.test.js
+++ b/__device-tests__/gutenberg-editor-lists-end.test.js
@@ -1,0 +1,60 @@
+/**
+ * @format
+ * */
+
+/**
+ * Internal dependencies
+ */
+import EditorPage from './pages/editor-page';
+import {
+	setupDriver,
+	isLocalEnvironment,
+	stopDriver } from './helpers/utils';
+import testData from './helpers/test-data';
+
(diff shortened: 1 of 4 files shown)


hypest:
I'll go ahead and merge and in case @JavonDavis has any proposals to make this new test better we can do it in a follow PR.

Issue #1029, opened 2019-05-24
Regression: Virtual keyboard is hidden on paragraph split


hypest:
The virtual keyboard is closed right after a paragraph block split, but it shouldn't.

Steps to reproduce:

1. Run the demo app and place the caret in the middle (or the end for that matter) of a paragraph block
2. Tap "Enter" to split the block into two paragraph blocks
3. Notice that the virtual keyboard got hidden, even though the caret is in the newly created block

hypest:
FYI: `git bisect` lead me to 8cd754f9711f44bc6d5b589a796b8e3ebca415f0.

hypest:
This issue seems to have "evolved": Now, the caret is not placed in the new block at all. I've `git bisect`'d the new regression to happen with 36abf9b1 (https://github.com/wordpress-mobile/gutenberg-mobile/pull/963).

PR #1032, opened 2019-05-27
Fix caret lost regression

### LINKED ISSUE #1031: Regression: Caret is lost on paragraph split
Steps to reproduce:

1. Run the demo app and place the caret in the middle of a paragraph block
2. Tap "Enter" to split the block into two
3. Notice that the new block is created as normal but, the caret is lost

hypest:
Fixes #1031 , #1034 .

Gutenberg PR: https://github.com/WordPress/gutenberg/pull/15833

Fix for the regression where the caret is lost on paragraph split.

To test:

1. Run the demo app and place the caret in the middle of a paragraph block
2. Tap "Enter" to split the block into two
3. The block should split as normal and the caret be placed at the start of the new paragraph block. Note: #1029 still applies so, the virtual keyboard will be closed.

Update release notes:

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -1,3 +1,4 @@
 1.6.0
 ------
+- A new block is available: video block.
 * Added UI to display a warning when a block has invalid content.


mkevins:
I've tested this, and it fixes the issue as described. It also fixes https://github.com/wordpress-mobile/gutenberg-mobile/issues/1034 :tada:. I'm still waiting [to learn more about the intention of the original diff](https://github.com/WordPress/gutenberg/pull/15482/files#r287618893) that introduced the `else` statement. Also, it may be worth investigating further why the value of `isSelected` can affect a regression of the merging blocks crash.

hypest:
> Also, it may be worth investigating further why the value of isSelected can affect a regression of the merging blocks crash

I didn't get to the very bottom of that one but, what I had found while investigating was that because of https://github.com/wordpress-mobile/gutenberg-mobile/issues/1033, we are messing up with the value of `isSelected` via the RichText embedded in the title block. That block has a `null` `clientId` because the title block doesn't have a clientId. I think it's the then invalid combination of the global selection and local selection logic that leads to #1034.

PR #1035, opened 2019-05-27
Update Aztec version to 1.6.5

### LINKED ISSUE #938: [iOS] Unable to copy & paste text from mobile Gutenberg
Tested on iPhone (iOS 12.2) and iPad Pro (iOS 12.3b) using WPiOS 12.3 beta (this bug is not present in 12.2). Tested Android on Pixel 2, bug is not present.

I’m not exactly sure what the cause is, but text copied from _inside the editor_ isn’t always pastable. The only case that is working is `paste without formatting` Inside the editor. Pasting external text _into_ the editor works.

Example: https://cloudup.com/cd7zwlKVU2S

Steps to reproduce (inside editor):

1. Open a post or page that contains text
2. Highlight text, select `copy` from edit menu
3. Create a new block, tap to show edit menu, tap `paste`
4. Notice that *nothing is pasted*
5. Tap to show edit menu, tap `paste without formatting`
6. Notice text *is pasted*

Pasting from external text works:

1. Copy text from another editor, web page, etc.
2. Try pasting into GB
3. Notice the text *is pasted*, although the caret goes to the start of the pasted content

Pasting to external editor: 

Additionally, when trying to paste from the editor into an external text input/editor, the text from the clipboard is not able to paste. I tested this on (all iOS) Safari text input, Evernote, Apple Notes, Google Docs to confirm.

1. Copy text from post or page in mobile GB
2. Open external editor, such as Google Docs or Evernote, etc.
3. Try pasting clipboard
4. Notice text is not pasted

SergioEstevao:
Fixes #938 

This updates the current version of Aztec to 1.6.5. 
The main benefit of this version is to get copy/paste working across text blocks and copy/paste to external apps.

To test:

 - Start the demo app in iOS
 - Copy a bit of text from one block
 - Paste it on another block
 - Paste to another app in the system ( Notes app or Reminders)
 - See if the plain text is copied.
Update release notes:

### DIFF
--- a/react-native-aztec/ios/Cartfile
+++ b/react-native-aztec/ios/Cartfile
@@ -1,1 +1,1 @@
-github "wordpress-mobile/AztecEditor-iOS" "1.6.4"
+github "wordpress-mobile/AztecEditor-iOS" "1.6.5"
--- a/react-native-aztec/ios/Cartfile.resolved
+++ b/react-native-aztec/ios/Cartfile.resolved
@@ -1,1 +1,1 @@
-github "wordpress-mobile/AztecEditor-iOS" "1.6.4"
+github "wordpress-mobile/AztecEditor-iOS" "1.6.5"


koke:
For reference, these are the changes: https://github.com/wordpress-mobile/AztecEditor-iOS/pull/1192 https://github.com/wordpress-mobile/AztecEditor-iOS/pull/1193

hypest:
Aztec-iOS 1.6.5 is already in WPiOS so, let's move the milestone earlier, to v1.6 to have the Aztec versions match in the v1.6 release.

hypest:
I'll update from develop so to get up-to-date CI status.

PR #1054, opened 2019-05-31
Refactor Picker and fix state update on an unmounted component

### LINKED ISSUE #1004: Update Icon component to allow better customization for mobile
Currently we can't pass styles to `Icon` component for our own needs. It only accepts `className` and it is not applied for mobile. Let's update `Icon` component to allow this and then refactor `src/block-management/block-picker.js` to remove `iconWithUpdatedFillColor` function. 

This looks like a tech debt but actually it is more important than that. It can cause some side affects as https://github.com/wordpress-mobile/gutenberg-mobile/pull/997

cc @Tug

### LINKED ISSUE #1049: Media Picker - Warning - can't perform state update on an unmounted component
Notice the warning below when tried to upload a media in the wp-android app.

Screen Shot 2019-05-30 at 12 35 23

Tug:
Fixes #1049

Gutenberg PR: https://github.com/WordPress/gutenberg/pull/15937

### To test:
- Try on WPAndroid and WPiOS
- Try every components that use the Picker: the block picker, the link editor, the image add and edit... and check that the modal shows and dismiss as expected and the value picked is the right one


hypest:
Bumping to next milestone since it is still WIP.

Tug:
Closed per https://github.com/WordPress/gutenberg/pull/15937

PR #1056, opened 2019-05-31
FIX NPE in WPAndroidGlue while setting Auth header

daniloercoli:
This PR fixes https://github.com/wordpress-mobile/WordPress-Android/issues/9833 by making sure the Auth header requested listener is not null before using it. As described in this comment [here](https://github.com/wordpress-mobile/gutenberg-mobile/pull/1056#issuecomment-498183372) the auth header requested listener is unset on exit of the editor, but the network queue is still filled with "old" pictures request. When the queue process those pictures the listener previously unset is invoked leading to the crash.

To test:

- Load a post with lot of pictures
- Exit the editor asap
- Wait until all the request end and no crash

Update release notes:

- [ x ] If there are user facing changes, I have added an item to `RELEASE-NOTES.txt`.

### DIFF
--- a/react-native-gutenberg-bridge/android/src/main/java/org/wordpress/mobile/WPAndroidGlue/OkHttpHeaderInterceptor.java
+++ b/react-native-gutenberg-bridge/android/src/main/java/org/wordpress/mobile/WPAndroidGlue/OkHttpHeaderInterceptor.java
@@ -6,12 +6,12 @@ import org.wordpress.mobile.WPAndroidGlue.WPAndroidGlueCode.OnAuthHeaderRequeste
 
 import java.io.IOException;
 
+import okhttp3.HttpUrl;
 import okhttp3.Interceptor;
 import okhttp3.Request;
 import okhttp3.Response;
 
 public class OkHttpHeaderInterceptor implements Interceptor {
-
     private static final String AUTHORIZATION_HEADER_KEY = "Authorization";
 
     private OnAuthHeaderRequestedListener mOnAuthHeaderRequestedListener;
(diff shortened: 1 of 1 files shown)


mkevins:
I think that it is actually `mOnAuthHeaderRequestedListener` that is `null`, rather than the `HttpUrl` object. There seem to be two places where that is set (both in `WPAndroidGlueCode.java`): [in the attachToContainer method](https://github.com/wordpress-mobile/gutenberg-mobile/blob/bd73de7df18c858afd6569d7e24ce1798936d8be/react-native-gutenberg-bridge/android/src/main/java/org/wordpress/mobile/WPAndroidGlue/WPAndroidGlueCode.java#L294) and [in the onDestroy method](https://github.com/wordpress-mobile/gutenberg-mobile/blob/bd73de7df18c858afd6569d7e24ce1798936d8be/react-native-gutenberg-bridge/android/src/main/java/org/wordpress/mobile/WPAndroidGlue/WPAndroidGlueCode.java#L350).

I believe there are two possibilities for the `OnAuthHeaderRequestedListener` instance to be `null`.

1. A possible race condition after the `OkHttpHeaderInterceptor` is instantiated: https://github.com/wordpress-mobile/gutenberg-mobile/blob/bd73de7df18c858afd6569d7e24ce1798936d8be/react-native-gutenberg-bridge/android/src/main/java/org/wordpress/mobile/WPAndroidGlue/WPAndroidGlueCode.java#L81 and before its member (`mOnAuthHeaderRequestedListener`) is set.
2. Because `sAddCookiesInterceptor` is static, it lives outside the normal activity life-cycle. Its member is set to `null` when the activity is destroyed, so it may be that another thread is still "finishing up" from an async call, resulting in the NPE after the activity has been destroyed.

If it is indeed the latter case, and _only_ that case, maybe conditionally "nooping" the intercept when the listener is `null` is all that's needed, though I'm not entirely sure. It seems if it is the first case, maybe it's worth investigating why requests are intercepted before the listener is set. Any thoughts on that?

daniloercoli:
@mkevins you're right, apparently I've miss-read the crash report 🤦‍♂ 
Will add the check to avoid null call and see what happens in the next release.

daniloercoli:
I was able to reproduce the problem:
- Load a post with lot of pictures
- Exit the editor asap
- Wait until the crash

This is due to `onDestroy` being called here: https://github.com/wordpress-mobile/gutenberg-mobile/blob/bd73de7df18c858afd6569d7e24ce1798936d8be/react-native-gutenberg-bridge/android/src/main/java/org/wordpress/mobile/WPAndroidGlue/WPAndroidGlueCode.java#L350 and unset the listener, but network requests are still ongoing in the queue.

Updated the main description as well.

Issue #1066, opened 2019-06-04
[iOS] Video block has black background when upload is in progress or upload has failed


marecar3:
When a video is in progress or upload has failed, video block has a black background.

It is a regression caused by this PR: https://github.com/wordpress-mobile/gutenberg-mobile/pull/1046

<img width="317" alt="Screen Shot 2019-06-04 at 12 46 41" src="https://user-images.githubusercontent.com/28219092/58915312-b8aa0780-86ee-11e9-9ffd-3b923c0b5a53.png">

<img width="328" alt="Screen Shot 2019-06-04 at 12 47 28" src="https://user-images.githubusercontent.com/28219092/58915313-b8aa0780-86ee-11e9-9bb3-f799e0496a1f.png">

pinarol:
@iamthomasbishop setting the background color of the video somehow can lead to a crash on android. so it was reverted by this PR https://github.com/wordpress-mobile/gutenberg-mobile/pull/1046

Right now we are only setting the background to black on iOS but it causes difference between platforms:

Android (video doesn't have a background color so the container view's color is visible instead):

![Screenshot_1559726002](https://user-images.githubusercontent.com/5032900/58945130-f446df00-878b-11e9-9a65-9572b89238f0.png)

Portrait video:

![Screenshot_1559727413](https://user-images.githubusercontent.com/5032900/58946902-8b616600-878f-11e9-875b-acfd597d6603.png)


iOS (video has black background):

![Simulator Screen Shot - iPhone Xʀ - 2019-06-05 at 12 15 19](https://user-images.githubusercontent.com/5032900/58945149-fc068380-878b-11e9-8b01-1db026082fe7.png)

Portrait video:

![Simulator Screen Shot - iPhone Xʀ - 2019-06-05 at 12 43 15](https://user-images.githubusercontent.com/5032900/58946895-856b8500-878f-11e9-826b-2de350c0b336.png)

Just wanted know if you are OK with this.

Maybe it'd be better to equalize platforms and remove black background from iOS also? wdyt? 

We can play with container view's background color also but it is a reusable view between image/video blocks and I am a bit afraid to bring more side effects at last moment.

koke:
What if we set the background color on the [container view](https://github.com/WordPress/gutenberg/blob/4d5998d7b8558c94c6030d167111c11f92c5b46a/packages/block-library/src/video/edit.native.js#L213)?

pinarol:
[...]

PR #1068, opened 2019-06-04
Video block has black background when upload is in progress or upload has failed

marecar3:
Fixes : https://github.com/wordpress-mobile/gutenberg-mobile/issues/1066

Corresponding gutenberg PR : https://github.com/WordPress/gutenberg/pull/15991

To test :

Build WPiOS and WPAndroid against this PR.

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -1,3 +1,7 @@
+1.7.0
+------
+* Fixed keyboard flickering issue after pressing Enter repeatedly on the Post Title.
+
 1.6.0
 ------
 * Fixed issue with link settings where “Open in New Tab” was always OFF on open.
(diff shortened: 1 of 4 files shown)


pinarol:
Thanks Marko 👍 I tested with both WPiOS and WPAndroid and it is working fine now. I just added some non critical comments on gutenberg side.

I asked @iamthomasbishop some questions here https://github.com/wordpress-mobile/gutenberg-mobile/issues/1066#issuecomment-499014919 Let's wait for his answer before merging.

@daniloercoli could you be able to review and ✅ also if you are ok with the changes? we don't want to bring back the crash. I assume we are doing a safe fix since we aren't setting video background color on Android but it'd be good if you take a look also.

pinarol:
Thanks a lot @daniloercoli !

Issue #1074, opened 2019-06-05
[iOS]User needs to tap play again after going full screen on video block


pinarol:
Current behavior
- Tap on a video
- Full screen player opens in paused state
- Tap to play button to play it

Expected:
- Tap on a video
- Full screen player opens in a playing state

PR #1075, opened 2019-06-05
Start playback immediately after video goes full screen

pinarol:
Fixes wordpress-mobile/gutenberg-mobile#1074

This is an iOS only issue.

gutenberg PR https://github.com/WordPress/gutenberg/pull/15998

To test:

- Tap on a video
- Verify that full screen player opens in a playing state

Update release notes:


pinarol:
> There's a minor thing I noticed, which doesn't seem too important. When the video gets to the end, the controls still show in a playing state

Yes that seems a bit odd. Probably something's wrong on react-native-video level. I don't think it has a quick fix.

PR #1079, opened 2019-06-05
More block enhancements

### LINKED ISSUE #859: More block
We have a working implementation, but split/merge doesn't work.

We need to decide if we use RichText, or fix #324 to add split/merge support to PlainText

etoledom:
Fixes #859 

As mentioned in  https://github.com/wordpress-mobile/gutenberg-mobile/issues/859#issuecomment-498321249, I opted for modifying the More block to be read only, in order for it to be ready for the next release.

More info and test steps on the gutenberg side PR: https://github.com/WordPress/gutenberg/pull/16005

On the design side, I noticed on web that More block and Page Break block look the same, so I implemented the same design we use for Page Break on mobile too, but I'll wait for @iamthomasbishop 's input on this.

more-block

Update release notes:

### DIFF
--- a/__device-tests__/gutenberg-editor-lists-end.test.js
+++ b/__device-tests__/gutenberg-editor-lists-end.test.js
@@ -16,7 +16,7 @@ import testData from './helpers/test-data';
 
 jasmine.DEFAULT_TIMEOUT_INTERVAL = 240000;
 
-describe( 'Gutenberg Editor tests', () => {
+describe( 'Gutenberg Editor tests for List block (end)', () => {
 	let driver;
 	let editorPage;
 	let allPassed = true;
--- a/__device-tests__/gutenberg-editor-lists.test.js
+++ b/__device-tests__/gutenberg-editor-lists.test.js
@@ -16,7 +16,7 @@ import testData from './helpers/test-data';
 
 jasmine.DEFAULT_TIMEOUT_INTERVAL = 240000;
 
-describe( 'Gutenberg Editor tests', () => {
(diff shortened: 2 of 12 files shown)


iamthomasbishop:
Nice work! A couple of little things I wanted to document – although not specific to the Read More block (also applies to the Page Break block):

* The divider should extend to the ends of the canvas/container
* The label should be sans-serif
* Ideally the divider should be a dashed line, but I think that might've been a compromise bc of CSS limitations in the RN app

etoledom:
Thank you for the review @iamthomasbishop !

> The divider should extend to the ends of the canvas/container

I was able to extend the lines a bit more, until the edge of the block content.

We can extend them more using some negative magic margin values, but I'm not sure if that's what you had in mind, and it might not be a very good idea :) 

> The label should be sans-serif

✅ 

> Ideally the divider should be a dashed line, but I think that might've been a compromise bc of CSS limitations in the RN app

The lines are actually a normal (thin) view with a background color. Not possible to make them dashed with the current component we are using. I'd recommend this to be a future enhancement.

This is an updated screenshot:

more

koke:
> Not possible to make them dashed with the current component we are using. I'd recommend this to be a future enhancement

For reference, some initial effort going (on hold) here: https://github.com/WordPress/gutenberg/pull/14175

iamthomasbishop:
Looks great to me! Thanks for the updates @etoledom! :shipit:

etoledom:
Thank you both!

PR #1080, opened 2019-06-05
Add html verification on iOS

### LINKED ISSUE #1061: Testing: Toggle HTML mode on iOS
Currently, there's no way to verify HTML content for iOS, https://github.com/wordpress-mobile/gutenberg-mobile/blob/develop/__device-tests__/pages/editor-page.js#L272-L278 . We should find a way to implement this as a few tests depend on it.

JavonDavis:
Fixes #1061 

To test: Run List end test in `gutenberg-editor-lists-end.test.js`, ensure it toggles the HTML view on iOS as well as android

Update release notes:


__device-tests__/pages/editor-page.js
```diff
@@ -59,6 +59,44 @@ export default class EditorPage {
...
+		await toggleHtmlMode( this.driver );
+
+		const htmlContentView = await this.getTextViewForHtmlViewContent();
+		const text = await htmlContentView.text();
+		expect( text ).toBe( html );
+
+		await toggleHtmlMode( this.driver );
+	}
+
+	async verifyHtmlContentIOS( html: string ) {
```
hypest: `verifyHtmlContentIOS` seems identical to `verifyHtmlContentAndroid` at this point. Can we consolidate the two?

__device-tests__/helpers/utils.js
```diff
@@ -223,7 +223,15 @@ const toggleHtmlMode = async ( driver: wd.PromiseChainWebdriver ) => {
...
 		await driver.pressKeycode( 66 );
 	} else {
-		// TODO: toggle html mode in iOS
+		const menuButton = await driver.elementByAccessibilityId( '...' );
+		await menuButton.click();
+		let toggleHtmlButton;
+		try {
+			toggleHtmlButton = await driver.elementByAccessibilityId( 'Switch to HTML' );
+		} catch ( err ) {
+			toggleHtmlButton = await driver.elementByAccessibilityId( 'Switch To Visual' );
```
hypest: Trying to understand the logic, is the assumption here that `'Switch to HTML'` tap will fail only if we are already in html mode? If so, let's add a comment in the code about it. Reason is that since we'll be expecting the `try` to fail in half the cases, this is not an error case after all but rather, the way we detect which menu option is currently available. WDYT?

JavonDavis: I went ahead and wrote the extra code to get this more descriptive without the comments. Initially, it seemed the simpler way was just to go with the `try catch` but it's not that much additional effort to make this work without the assumption  so I did that

__device-tests__/pages/editor-page.js
JavonDavis: Right! done! 👍

JavonDavis:
Thanks, @hypest! I've resolved the conflicts and addressed the comments! I'll go ahead and merge this so we can use it in other PRs.

PR #1082, opened 2019-06-05
Quotes: Fix handling of insertion and removal of empty lines.

### LINKED ISSUE #207: Quote block
Add support for the `quote` block.

What we're *not* implementing in the beta:
- Text alignment
- Inspector settings: text size and color
- Quote styles. Not specific to quotes, we might need block transformations first.

SergioEstevao:
Fixes #207

Relate GB PR: https://github.com/WordPress/gutenberg/pull/16013

This PR fixes the following issues around the insertion and removal of empty lines inside quotes:
 - It was impossible to insert multiple empty lines inside a quote block
- After the insertion of an empty line at the end of the quote block, trying to remove it caused a crash or the merge of the block with the block above.

The main changes are the following:
 -  Aztec is updated in order that empty block elements are not removed. So when entering a new line in quote that results in this HTML: `<p></p>` isn't remove by Aztec
 - Checks are done in GB JS code to check if the selection range returned by iOS is inside the text length in the format structure. This could happen in scenarios where extra characters were added by Aztec in order to display paragraph styles in the last line.

To test:
 - Open the demo app
 - Select a quote block
 - Add a new line to the end of content on the quote part
 - Add another line, make sure it shows up
 - Try to delete both lines 
 - Check that no crash or unnecessary merge happens.

**Know issues :** 
 - When adding a new line to the end of quote, an extra line is added, this is because Aztec iOS adds an extra line in order to keep the paragraph.
 - When adding a brand new quote the quote part will have immediately a new line and the placeholder text will not be show.

I know the issues above are not ideal, but I think they are better than having the crash on removal, or not having the possibility of adding multiple empty lines inside the quote.

Update release notes:

### DIFF
--- a/react-native-aztec/ios/Cartfile
+++ b/react-native-aztec/ios/Cartfile
@@ -1,1 +1,1 @@
-github "wordpress-mobile/AztecEditor-iOS" "1.6.6"
+github "wordpress-mobile/AztecEditor-iOS" "fff3d4712c6971cd2673fce273c8e9d9330a15ca"
--- a/react-native-aztec/ios/Cartfile.resolved
+++ b/react-native-aztec/ios/Cartfile.resolved
@@ -1,1 +1,1 @@
-github "wordpress-mobile/AztecEditor-iOS" "1.6.6"
+github "wordpress-mobile/AztecEditor-iOS" "fff3d4712c6971cd2673fce273c8e9d9330a15ca"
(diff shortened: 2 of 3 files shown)


koke:
I tried this:

- Create a new post
- Add a quote
- Write a couple words and press enter, a new line inside quote is created
- Tap backspace, nothing happens

koke:
Ok, I had not realized I had to update Aztec for testing, the previous case works well.

I found a new issue, but I don't think we need to block on this one, we can handle it separately:

1. Add a quote
2. Write a few words
3. Press enter, it adds a new line
4. Press enter again, nothing happens

PR #1088, opened 2019-06-06
Fix crash when deleting all content of RichText based block

etoledom:
fixes: #1089

This PR updates the gutenberg ref to test https://github.com/WordPress/gutenberg/pull/16018

**EDIT:** This crash is reproducible only on iOS:
> The reason it didn’t crash in Android was that in Android Aztec does not send the onDelete event up, unless delete is pressed and the selection is on the zero

@SergioEstevao 

To test:
- Please refer to the gutenberg side PR: https://github.com/WordPress/gutenberg/pull/16018


etoledom:
Thank you!

Issue #1091, opened 2019-06-06
Regression: Paragraph block is half-visible after split


marecar3:
It seems that there is a regression when `Paragraph block` is split into two blocks, the first paragraph block is half-visible for a second.

The same problem is when two Paragraph blocks get merged, a newly created block is half-visible for a moment.

![ezgif com-video-to-gif (3)](https://user-images.githubusercontent.com/28219092/59068173-7a911d00-8881-11e9-9bed-1efb43d00234.gif)

hypest:
Not happening anymore after we have improved the layout delay with https://github.com/wordpress-mobile/gutenberg-mobile/pull/1076.

PR #1076, opened 2019-06-05
Additional alignment of focus and selection logic with the web's

hypest:
Fixes #1029 

Gutenberg PR: https://github.com/WordPress/gutenberg/pull/15999

This PR makes changes to the Aztec wrapper (Android) as well as the RN app to bring it closer to the web as far as focus and selection logic goes.

### Important changes:

1. Try to avoid having Aztec (at the wrapper level and down) request focus automatically. We rely on the JS logic to do that.
  1. `setFocusableInTouchMode`, `setFocusable` is set to `false` until Aztec needs to receive focus. Also setting it to false when Aztec loses focus.
  2. Completely removed the `isSelected()` prop since the component should not use it as a flag to request focus

2. Schedule a `requestRectangleOnScreen` to scroll into view
When creating a new RichText based block, Aztec receives focus OK but the container doesn't seem to get scrolled to make sure the edit box is in the viewport. https://github.com/wordpress-mobile/gutenberg-mobile/pull/1076/files#diff-5693404c4c8f8d757cd298cc5aabc561R185 asks the system to bring the caret into the viewport. In a future iteration of this logic, we should try to bring the whole _block_ view into the viewport.

3. Schedule the `showSoftKeyboard` to happen on a next frame

4. Remove the 500ms delay for onContentSizeChange and just schedule it for next frame. This will actually dramatically improve the performance perception of the editor on Android, making the writing flow smoother. There's still the visible glitch where blocks start with minimum height and then expand though and fixing that (separate PR) will also greatly benefit the performance.

5. No need to try blurring other components in `block-holder`'s onFocus. Component will self-blur via https://github.com/WordPress/gutenberg/pull/15999/files#diff-4828a21853e899e5a36faecfa96d83e8R753

6. Align block-holder's `withDispatch` with the web's https://github.com/wordpress-mobile/gutenberg-mobile/pull/1076/files#diff-2004b9687d71de0f0a9816269bb4a6a2R276

To test:
1. Run the demo app and place the caret in the middle (or the end for that matter) of a paragraph block
2. Tap "Enter" to split the block into two paragraph blocks
3. Notice that the virtual keyboard is still visible and the caret is normally placed at the beginning of the newly created block.

Also, test focus and selection in general, it should work as expected.

[...]

PR #1093, opened 2019-06-07
Implement the shadow node for proper laying out Aztec on Android

hypest:
Addresses #1091

This PR implements the shadow node logic for the Aztec wrapper custom React Native view. This is needed for better View calculations.

This work begun as an attempt to remove the need to specify the `minHeight` style  on the Aztec wrapper. I think that the view should be able to update its height automatically within RN and the Android layout system, without us manually having to set it as a style prop. In that respect, the effort here is not finished; the `minHeight` style is still needed (try removing it and you'll see that the wrapper doesn't update its height properly, although its now initial height is much better than before). Although, my experiments so far with a basic demo app (in repo: https://github.com/hypest/RN060-aztec-experiments) seem to work and the Aztec wrapper there seems fine without specifying any `minHeight`.

Even in this state, the behavior in this PR is better than before and that's why I'm opening this PR.

Known issue: the height of the component seems smaller than the content text but only by a little. The `onContentSizeChange` logic repairs it but it's still visible. Besides, the glitch is minimized due to the work on #1076 and that's why this PR builds on top of that, albeit in a separate PR so #1076 doesn't get blocked/complex.

To test:
1. Run the demo app on Android
2. Put the caret in the middle or the end of the "Lorem ipsum" paragraph block
3. Split the block
4. Notice that the visual glitch of the existing block resizing is quite small

I have also 
Update release notes:

### DIFF
--- a/react-native-aztec/android/src/main/java/org/wordpress/mobile/ReactNativeAztec/ReactAztecManager.java
+++ b/react-native-aztec/android/src/main/java/org/wordpress/mobile/ReactNativeAztec/ReactAztecManager.java
@@ -19,8 +19,9 @@ import com.facebook.react.bridge.ReactContext;
 import com.facebook.react.bridge.ReadableArray;
 import com.facebook.react.bridge.ReadableMap;
 import com.facebook.react.common.MapBuilder;
+import com.facebook.react.uimanager.BaseViewManager;
+import com.facebook.react.uimanager.LayoutShadowNode;
 import com.facebook.react.uimanager.PixelUtil;
-import com.facebook.react.uimanager.SimpleViewManager;
 import com.facebook.react.uimanager.ThemedReactContext;
 import com.facebook.react.uimanager.UIManagerModule;
 import com.facebook.react.uimanager.ViewDefaults;
@@ -31,6 +32,7 @@ import com.facebook.react.views.scroll.ScrollEvent;
 import com.facebook.react.views.scroll.ScrollEventType;
 import com.facebook.react.views.text.DefaultStyleValuesUtil;
 import com.facebook.react.views.text.ReactFontManager;
+import com.facebook.react.views.text.ReactTextUpdate;
 import com.facebook.react.views.textinput.ReactContentSizeChangedEvent;
 import com.facebook.react.views.textinput.ReactTextChangedEvent;
 import com.facebook.react.views.textinput.ReactTextInputEvent;
@@ -51,7 +53,7 @@ import java.util.Map;
 import java.util.ArrayList;
 import java.util.Arrays;
 
-public class ReactAztecManager extends SimpleViewManager<ReactAztecText> {
+public class ReactAztecManager extends BaseViewManager<ReactAztecText, LayoutShadowNode> {
 
     public static final String REACT_CLASS = "RCTAztecView";
 
@@ -95,6 +97,16 @@ public class ReactAztecManager extends SimpleViewManager<ReactAztecText> {
         aztecText.setFocusable(false);
         aztecText.setCalypsoMode(false);
         return aztecText;
+    }
+
+    @Override
+    public LayoutShadowNode createShadowNodeInstance() {
+        return new ReactAztecTextShadowNode();
+    }
+
+    @Override
+    public Class<? extends LayoutShadowNode> getShadowNodeClass() {
+        return ReactAztecTextShadowNode.class;
     }
 
     @Nullable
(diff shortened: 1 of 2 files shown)


marecar3:
Hey @hypest, I tested this branch a little bit and it seems that it's an improvement 👍

react-native-aztec/android/src/main/java/org/wordpress/mobile/ReactNativeAztec/ReactAztecTextShadowNode.java
```diff
@@ -0,0 +1,25 @@
+package org.wordpress.mobile.ReactNativeAztec;
+
+import android.support.annotation.Nullable;
+
+import com.facebook.react.bridge.ReadableMap;
+import com.facebook.react.uimanager.annotations.ReactProp;
+import com.facebook.react.views.textinput.ReactTextInputShadowNode;
+
+public class ReactAztecTextShadowNode extends ReactTextInputShadowNode {
+    private @Nullable
```
mkevins: Def not a blocker, just noticed that this got bumped to two lines for some reason. Your earlier work had it on one: https://github.com/hypest/RN060-aztec-experiments/blob/5bc23a59dfadc47b97ac5381f33c14829bda280e/react-native-aztec/android/src/main/java/org/wordpress/mobile/ReactNativeAztec/ReactAztecTextShadowNode.java#L10

react-native-aztec/android/src/main/java/org/wordpress/mobile/ReactNativeAztec/ReactAztecManager.java
```diff
@@ -496,6 +508,19 @@ public void onFocusChange(View v, boolean hasFocus) {
         // in case check ReactTextInputManager
     }
 
+    @Override
+    public void updateExtraData(ReactAztecText view, Object extraData) {
+        if (extraData instanceof ReactTextUpdate) {
+            ReactTextUpdate update = (ReactTextUpdate) extraData;
+
+            view.setPadding(
```
mkevins: I'm getting this warning in AS about these lines for RTL compatibility:

```
For RtL compatibility, use setPaddingRelativeor ViewCompat.setPaddingRelative() when setting left/right padding. less... (Ctrl+F1) 
Inspection info:For RtL compatibility, use setPaddingRelativeor ViewCompat.setPaddingRelative() when setting left/right padding.
```

react-native-aztec/android/src/main/java/org/wordpress/mobile/ReactNativeAztec/ReactAztecTextShadowNode.java
hypest: Ah, let me put them in a single line 👍

react-native-aztec/android/src/main/java/org/wordpress/mobile/ReactNativeAztec/ReactAztecManager.java
mkevins: This might be a local linting issue on my end.. I couldn't determine why it gave this warning when `ReactTextUpdate` merely inherits from `java.lang.Object` :thinking:

hypest: Good point. I've been actually mirroring the code in upstream's implementation of the TextInput shadow node here: https://github.com/facebook/react-native/blob/d88e4701fc46b028861ddcfa3e6ffb141b3ede3d/ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactTextInputManager.java#L193-L197

I'm kinda torn but, I think it makes more sense at the moment to follow upstream's lead on this one. Alternatively, we can even completely remove the code to set the padding since it's not a real goal for this PR anyway. WDYT?

react-native-aztec/android/src/main/java/org/wordpress/mobile/ReactNativeAztec/ReactAztecTextShadowNode.java
hypest: Done with c57d61b.

react-native-aztec/android/src/main/java/org/wordpress/mobile/ReactNativeAztec/ReactAztecManager.java
mkevins: > I think it makes more sense at the moment to follow upstream's lead on this one.

I agree, this seems like the best way forward. I haven't tested many scenarios without the code (only the "Lorem ipsum" test in the description), and I'm unsure if there are scenarios where it would be important to have it. So keeping it close to upstream seems the sensible approach. LGTM!

PR #1097, opened 2019-06-07
Issue/1045 paste of attributed strings

### LINKED ISSUE #1045: [iOS] Paste not yet implemented with NSAttributedString clipboard object
### Description
When styled text is copied in some applications in iOS, the clipboard contents may be populated with an `NSAttributedString` object for styled text, but may not have the contents available as html. When the styled text is pasted from the clipboard into the block editor, the content will lose its styles and fall back to plain text.

This issue will address the second part of this comment: https://github.com/wordpress-mobile/gutenberg-mobile/issues/938#issuecomment-487691560. [The first part of the issue is tracked here.](https://github.com/wordpress-mobile/gutenberg-mobile/issues/1044)

[This related issue](https://github.com/wordpress-mobile/gutenberg-mobile/issues/938) was closed, but further work is needed to fully implement copy / paste of styled text.

SergioEstevao:
Fixes #1045 

This PR improves the paste code by reading more UTI types when doing paste and processing them to HTML if possible.

**Note:** Make sure you run `yarn ios` to get the update of Aztec lib 

To test:
 - Start the demo app
 - Copy text with attributes (string, bold) from another app, for example Notes
 - Paste it on the demo app text block
 - Check that the content is pasted with attributes.

Update release notes:

### DIFF
--- a/react-native-aztec/ios/RNTAztecView/RCTAztecView.swift
+++ b/react-native-aztec/ios/RNTAztecView/RCTAztecView.swift
@@ -189,6 +189,13 @@ class RCTAztecView: Aztec.TextView {
 
     private func readHTML(from pasteboard: UIPasteboard) -> String? {
 
+        if let data = pasteboard.data(forPasteboardType: kUTTypeHTML as String), let html = String(data: data, encoding: .utf8) {
+            // Make sure we are not getting a full HTML DOC. We only want inner content
+            if !html.hasPrefix("<!DOCTYPE html") {
+                return html
+            }
+        }
+
         if let flatRTFDString = read(from: pasteboard, uti: kUTTypeFlatRTFD as String, documentType: DocumentType.rtfd) {
             return  flatRTFDString
         }
@@ -201,15 +208,11 @@ class RCTAztecView: Aztec.TextView {
             return  rtfdString
         }
 
-        if let data = pasteboard.data(forPasteboardType: kUTTypeHTML as String) {
-            return String(data: data, encoding: .utf8)
-        }
-
         return nil
     }
     
     private func readText(from pasteboard: UIPasteboard) -> String? {
-        return pasteboard.string        
+        return pasteboard.string
     }
 
     func saveToDisk(image: UIImage) -> URL? {
--- a/react-native-aztec/ios/RNTAztecView/RCTAztecView.swift
+++ b/react-native-aztec/ios/RNTAztecView/RCTAztecView.swift
@@ -178,7 +178,7 @@ class RCTAztecView: Aztec.TextView {
     }
     
     // MARK: - Paste handling
-    private func read(from pasteboard: UIPasteboard, uti: String, documentType: DocumentType) -> String? {
+    private func read(from pasteboard: UIPasteboard, uti: CFString, documentType: DocumentType) -> String? {
         guard let data = pasteboard.data(forPasteboardType: uti as String),
             let attributedString = try? NSAttributedString(data: data, options: [.documentType: documentType], documentAttributes: nil),
             let storage = self.textStorage as? TextStorage else {
@@ -196,15 +196,15 @@ class RCTAztecView: Aztec.TextView {
             }
         }
 
-        if let flatRTFDString = read(from: pasteboard, uti: kUTTypeFlatRTFD as String, documentType: DocumentType.rtfd) {
+        if let flatRTFDString = read(from: pasteboard, uti: kUTTypeFlatRTFD, documentType: DocumentType.rtfd) {
             return  flatRTFDString
         }
 
-        if let rtfString = read(from: pasteboard, uti: kUTTypeRTF as String, documentType: DocumentType.rtf) {
+        if let rtfString = read(from: pasteboard, uti: kUTTypeRTF, documentType: DocumentType.rtf) {
             return  rtfString
         }
 
-        if let rtfdString = read(from: pasteboard, uti: kUTTypeRTFD as String, documentType: DocumentType.rtfd) {
+        if let rtfdString = read(from: pasteboard, uti: kUTTypeRTFD, documentType: DocumentType.rtfd) {
             return  rtfdString
         }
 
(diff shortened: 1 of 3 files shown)


react-native-aztec/ios/RNTAztecView/RCTAztecView.swift
```diff
@@ -175,26 +175,34 @@ class RCTAztecView: Aztec.TextView {
...
         }
-        
-        return String(data: data, encoding: .utf8)
-    }
 
-    private func cleanHTML() -> String {
-        let html = getHTML(prettify: false)
-        return html
+        if let data = pasteboard.data(forPasteboardType: kUTTypePlainText as String) {
+            return String(data: data, encoding: .ascii)
```
koke: Why `.ascii` and not `.utf8`?

react-native-aztec/ios/RNTAztecView/RCTAztecView.swift
```diff
@@ -175,26 +175,34 @@ class RCTAztecView: Aztec.TextView {
...
+        }
+
+        if let data = pasteboard.data(forPasteboardType: kUTTypeText as String) {
+            return String(data: data, encoding: .ascii)
```
koke: Why `.ascii` and not `.utf8`?

SergioEstevao: when the UTI is TypeText you don't have any guarantees of the encoding. So I was trying to put the lowest denominator there. It's a pity the platform doesn't have an auto-discover mode for this.

koke: I think my slack comment got lost in the threads, so replying here. Are you checking for `kUTTypePlainText` and `kUTTypeText` for a reason?

https://developer.apple.com/documentation/uikit/uipasteboard/1622091-strings makes me think that `kUTTypeUTF8PlainText` is the only one relevant

SergioEstevao:
So you see the HTML field getting sent to JS side?

etoledom:
> So you see the HTML field getting sent to JS side?

Yes, it get's to the JS side. But it seems to be ignored.

SergioEstevao:
@etoledom improve the paste method to better check HTML imports.

It should now be able to import from Notes, Pages and Safari correctly.

Thanks for your tests!

react-native-aztec/ios/RNTAztecView/RCTAztecView.swift
```diff
@@ -178,26 +178,41 @@ class RCTAztecView: Aztec.TextView {
...
-    
-    private func text(from pasteboard: UIPasteboard) -> String? {
-        guard let data = pasteboard.data(forPasteboardType: kUTTypePlainText as String) else {
-            return nil
+
+    private func readHTML(from pasteboard: UIPasteboard) -> String? {
+
+        if let data = pasteboard.data(forPasteboardType: kUTTypeHTML as String), let html = String(data: data, encoding: .utf8) {
+            // Make sure we are not getting a full HTML DOC. We only want inner content
+            if !html.hasPrefix("<!DOCTYPE html") {
```
SergioEstevao: I'm doing this check here to check if we are not importing a full HTML document because on my tests I found that the when copying from the Notes app, it was sending the full HTML document, using `span` and styles for bold/italic styles instead of using just `<strong>` and `<em>` element. 

My assumption is that when importing HTML if a full document is there, maybe there is content there is not valid or importable to GB so instead we should try to read RTF.

If think that this is not a good assumption we can remove this line, but then import from the Notes app will not work.

mkevins: > using span and styles for bold/italic styles instead of using just <strong> and <em> element.
> 
> My assumption is that when importing HTML if a full document is there, maybe there is content there is not valid or importable to GB so instead we should try to read RTF.

FWIW, [code exists on GB side to handle inline styles on span tags](https://github.com/WordPress/gutenberg/blob/master/packages/blocks/src/api/raw-handling/phrasing-content-reducer.js#L9), however, the limitations of our DOM implementation (jsdom-jscore) prevent it from working correctly.

To "enable" this code path, we'd need the style property on `ELEMENT_NODE`s to be implemented, as well as `ATTRIBUTE_NODE`s.

> If we think that this is not a good assumption we can remove this line, but then import from the Notes app will not work.

It's possible that implementing the necessary DOM spec will require significant time / effort. I think your approach is good to mitigate the underlying problem (that our DOM implementation is insufficient). I wonder if wrapping the html in a "full document" is correlated with using inline styles on spans, or if Notes.app happens to behave this way, and other apps have differing behavior regarding these two characteristics.

Is there some way to directly check if the pasteboard html generated by the external app uses inline styles on span tags?

etoledom: >  I think your approach is good to mitigate the underlying problem

I agree with this!

>  I wonder if wrapping the html in a "full document" is correlated with using inline styles on spans, or if Notes.app happens to behave this way, and other apps have differing behavior regarding these two characteristics.

I just tested the Google docs app, and it happens that it does use inline styles on spans, but without wrapping it up on a full HTML document. So this approach doesn't work for it.

But it does work for the Notes app.

I'm OK with shipping this improvement to support the Apple default's Note app, and check for a more wide alternative later on.

What do you think @SergioEstevao ?

react-native-aztec/ios/RNTAztecView/RCTAztecView.swift
```diff
@@ -178,26 +178,41 @@ class RCTAztecView: Aztec.TextView {
     }
     
     // MARK: - Paste handling
-    
-    private func html(from pasteboard: UIPasteboard) -> String? {
-        guard let data = pasteboard.data(forPasteboardType: kUTTypeHTML as String) else {
-            return nil
+    private func read(from pasteboard: UIPasteboard, uti: String, documentType: DocumentType) -> String? {
+        guard let data = pasteboard.data(forPasteboardType: uti as String),
```
etoledom: Small detail:
`uti` is already `String` so this extra `as String` is not needed.

Maybe we can change `uti`'s type to be `CFString` to make it simpler on the caller side?

PR #1100, opened 2019-06-07
Add Paste UI Tests for Android

mkevins:
### Description

This PR aims to implement UI tests for copying and pasting plain text and styled text from one paragraph block into another.

Since the copy and paste context menu items are not part of the application itself, the approach used to tap those buttons relies on specific x,y offsets from the active UI element. This may be somewhat brittle, and will likely require some tweaking for different platforms / OSs, however, in lieu of an alternative, I think there is some value in using such a workaround.

### To test:
Run a device or emulator with Android 8.
(It may be necessary to specify the exact platform version of in the `caps.js` file https://github.com/wordpress-mobile/gutenberg-mobile/blob/develop/__device-tests__/helpers/caps.js#L16).
Run the test:
`TEST_RN_PLATFORM=android yarn device-tests __device-tests__/gutenberg-editor-paste.test.js`

### Screencast:

There are no user facing changes in this PR.


__device-tests__/helpers/utils.js
```diff
@@ -183,6 +183,54 @@ const clickBeginningOfElement = async ( driver: wd.PromiseChainWebdriver, elemen
...
+};
+
+// press "Copy" in floating context menu
+const tapCopyAboveElement = async ( driver: wd.PromiseChainWebdriver, element: wd.PromiseChainWebdriver.Element ) => {
+	const location = await element.getLocation();
+	const action = await new wd.TouchAction( driver );
+	const x = location.x + 220;
+	const y = location.y - 50;
+	action.press( { x, y } );
+	//action.wait( 1000 );
```
JavonDavis: We can probably remove this right?

__device-tests__/helpers/utils.js
```diff
@@ -183,6 +183,54 @@ const clickBeginningOfElement = async ( driver: wd.PromiseChainWebdriver, elemen
...
+	action.release();
+	await action.perform();
+};
+
+// press "Paste" in floating context menu
+const tapPasteAboveElement = async ( driver: wd.PromiseChainWebdriver, element: wd.PromiseChainWebdriver.Element ) => {
+	const location = await element.getLocation();
+	const action = await new wd.TouchAction( driver );
+	action.press( { x: location.x + 100, y: location.y - 50 } );
+	//action.wait( 1000 );
```
JavonDavis: And this one

__device-tests__/helpers/utils.js
mkevins: Yes, I'll push a commit to remove those. Initially, I wasn't sure if they were required, as I'd seen similar waits on examples online.

__device-tests__/helpers/utils.js
mkevins: I will remove this.

mkevins: Ok, I've removed them.

__device-tests__/helpers/utils.js
mkevins: Ok, I've removed them.

PR #1086, opened 2019-06-06
Issue/697 change color of text and links

SergioEstevao:
Fixes #697 

Related GB PR: https://github.com/WordPress/gutenberg/pull/16016

Simulator Screen Shot - iPhone Xʀ - 2019-06-06 at 14 40 39

This PR updates the RCTAztecView component to support the setting of base text color and the color for links.

To test:
 - Run the demo app
 - Check that the default body text on text blocks (paragraph/list/quotes) is using the new gray 900 color.
 - Check that links are being draw with the new blue-500 color

Update release notes:

- [ ] If there are user facing changes, I have added an item to `RELEASE-NOTES.txt`.

PR #1101, opened 2019-06-07
Set link text color and link underline from JS

marecar3:
It is Android update for PR : https://github.com/wordpress-mobile/gutenberg-mobile/pull/1086

With this PR we are allowing JS side to change the link text color on Android, which will also turn on the underline option on the link.

In order to achieve this, I have created a branch on `AztecEditorAndroid` repo with change which allows us to set link style: https://github.com/wordpress-mobile/AztecEditor-Android/compare/fix/make_link_style_public

Currently, there is a problem with how props are propagated to native code. With this line of code https://github.com/WordPress/gutenberg/pull/16016/files#diff-4828a21853e899e5a36faecfa96d83e8R817 , Android native code will receive `linkTextColor` prop after `text` prop, which means it's too late to change link style as text is already drawn on the screen.

The only solution that came on my mind was to add `linkTextColor` as property in `text` prop: https://github.com/WordPress/gutenberg/pull/16050/files .

@SergioEstevao suggested that we try to set text again when we receive `linkTextColor` prop, which I tried and it's working, but I think it's not the best option for Android regarding performance, as we will force to setText on components that don't have a link (as `linkTextColor` prop is always passed from JS side) and if we want to set a text only on components that has a link, we would need to check if each component has it which also seems an impact on performance.

Let's open discussion here @daniloercoli @hypest , maybe we can figure out a better solution than this.

### DIFF
--- a/react-native-aztec/android/build.gradle
+++ b/react-native-aztec/android/build.gradle
@@ -12,7 +12,9 @@ buildscript {
         wordpressUtilsVersion = '1.22'
         espressoVersion = '3.0.1'
 
-        aztecVersion = 'v1.3.27'
+        aztecVersion = '26aaa61059661eb83db84111841577e99870adf6'
+
+
     }
 
     repositories {
--- a/react-native-aztec/android/src/main/java/org/wordpress/mobile/ReactNativeAztec/ReactAztecManager.java
+++ b/react-native-aztec/android/src/main/java/org/wordpress/mobile/ReactNativeAztec/ReactAztecManager.java
@@ -171,7 +171,9 @@ public class ReactAztecManager extends SimpleViewManager<ReactAztecText> {
 
     @ReactProp(name = "text")
     public void setText(ReactAztecText view, ReadableMap inputMap) {
+        String linkTextColor = inputMap.getString("linkTextColor");
         if (!inputMap.hasKey("eventCount")) {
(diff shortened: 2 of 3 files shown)


SergioEstevao:
I don't know the details of Aztec implementation in Android, but do we need to do this for links but not for the regular text color?

daniloercoli:
> Currently, there is a problem with how props are propagated to native code. With this line of code https://github.com/WordPress/gutenberg/pull/16016/files#diff-4828a21853e899e5a36faecfa96d83e8R817 , Android native code will receive linkTextColor prop after text prop, which means it's too late to change link style as text is already drawn on the screen.


Try to move `linkTextColor` before `text` prop, right below `Style`, it should work. 
Not quite sure this can be considered as a valid solution, but worth to try. 

Otherwise we may want to redraw the Aztec editor when linkTextColor is set on the wrapper.

hypest:
I'll try @daniloercoli 's idea and try to move this PR forward, in an attempt to make it to today's code freeze.

hypest:
For the release, we opted for a hardcoded setting of the link color https://github.com/wordpress-mobile/gutenberg-mobile/pull/1109

marecar3:
> For the release, we opted for a hardcoded setting of the link color #1109

In that case, probably we can close this one.

PR #1102, opened 2019-06-07
Add device rotation tests

JavonDavis:
Part of #745 

To test: Run `gutenberg-editor-rotation.test.js`

Update release notes:

### DIFF
--- a/__device-tests__/gutenberg-editor-rotatation.test.js
+++ b/__device-tests__/gutenberg-editor-rotatation.test.js
@@ -10,13 +10,14 @@ import {
 	setupDriver,
 	isLocalEnvironment,
 	stopDriver,
-	isAndroid, rotateDevice,
+	isAndroid,
+	toggleOrientation,
 } from './helpers/utils';
 import testData from './helpers/test-data';
 
 jasmine.DEFAULT_TIMEOUT_INTERVAL = 240000;
 
-xdescribe( 'Gutenberg Editor tests', () => {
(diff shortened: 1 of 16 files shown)


__device-tests__/gutenberg-editor-rotatation.test.js
```diff
@@ -52,6 +52,9 @@ describe( 'Gutenberg Editor tests', () => {
 		await editorPage.sendTextToParagraphBlock( paragraphBlockElement, testData.mediumText );
 
 		await toggleOrientation( driver );
+		if ( isAndroid() ) {
+			await driver.hideDeviceKeyboard();
```
hypest: Let's add a comment here about why we need to do this, especially on Android.

We'll actually want to remove that requirement down the road by not having the RichText component trigger the extracted EditText UI (`flagNoExtractUi` and friends). But for now, adding a line comment and a "TODO" marker should be OK. WDYT @JavonDavis ?

hypest:
Hey @JavonDavis , great to see orientation change tests coming to the project 😍 !

May I suggest complementing the writing test with a check that the _content_ of the block is still there after rotation? That'd be a super valuable check. WDYT?

__device-tests__/gutenberg-editor-rotatation.test.js
JavonDavis: Sounds good I'll add that! 👍🏽

PR #1117, opened 2019-06-11
Update video player to open the URL by browser on Android

pinarol:
Fixes https://github.com/wordpress-mobile/gutenberg-mobile/issues/1119

gutenberg PR: https://github.com/WordPress/gutenberg/pull/16089

This should change only Android. We are canceling out the react-native-video's player controls and replacing it with system browser:

video-android

iOS should remain same:

video-ios

To test:

WPAndroid:

- Open a post with video
- Verify that videos have a single play button and controls are invisible
- Tap on the video
- Verify that video opens outside of the app, probably in browser(I think it might also be opened by another app if user has selected it as the default app to open videos)

WPiOS should remain same:

- Open a post with video
- Tap on the video
- Verify that video opens fullscreen INSIDE the app

Update release notes:


hypest:
I can confirm that this makes the flow better, at least for the time being until the inline previewing works. I haven't given a look at the code side of things though so, I'm just adding this comment without aprooving the PR.

hypest:
We decided to hide the video block on Android for the 1.7 release (https://github.com/wordpress-mobile/gutenberg-mobile/pull/1127) so, I'll remove this PR from the milestone and re-target it towards develop. Let's decide later if we want to work some more on this one or note.

pinarol:
I think this PR brings a nice enhancement and worth merging it even if it doesn't solve the original overflow problem. Because the browser has a much better player. But even if we don't merge this, let's not forget to add the bottom item on a separate PR:

- Showing the play button even if video is not ready to play(this will be valid only for only iOS in case of discarding this PR)

daniloercoli:
I'm ok on merging this one on Android, since external player(s) (apps or browser) is way better. considering also the block has limited dimension, and the play'ed video inside them has a small viewport.

About the overflow problem, I suggest to wait until the AndroidX migration is completed.

If we agree to merge this PR we need upgrade the GB side companion PR to bring in latest changes.

PR #1112, opened 2019-06-10
Experiment - Android X migration

marecar3:
Issue: https://github.com/wordpress-mobile/gutenberg-mobile/issues/1062

With this PR we are migrating `gutenberg-mobile` android application to [AndroidX](https://developer.android.com/jetpack/androidx).

RN already provided AndroidX support so we are targeting[ RN60 with @jtreanor's patch for cocoapods](https://github.com/jtreanor/react-native#v0.60.0-patched) in this PR 

Libraries that are also migrated to AndroidX:

1. [Aztec-Android ](https://github.com/wordpress-mobile/AztecEditor-Android/pull/825/files)

2. [react-native-video](https://github.com/wordpress-mobile/react-native-video/commit/7d5905c76229e71c3ec519952a5fc3572c79feb4)

3. [react-native-recyclerview-list](https://github.com/wordpress-mobile/react-native-recyclerview-list/compare/master...wordpress-mobile:experiment/android_x_migration)

[WPAndroid](https://github.com/wordpress-mobile/WordPress-Android/pull/10020) against this PR.

**Note:** As we are using the newest version of `jsc-android` 241213, should we push it on : https://dl.bintray.com/wordpress-mobile/react-native-mirror/org/webkit/android-jsc/ ? @hypest @jtreanor (not sure)

~~**Not working:**~~  Working ✅ 

1. Gutenberg mobile tests, `enzyme` complains that adapter isn't set, I assume that it's something with RN60 version, maybe @etoledom can help here with some hint?

2. WordPress-Android tests -> jitpack.io can't build `react-native-video-wordpress` as it can't find RN60 at https://dl.bintray.com/wordpress-mobile/react-native-mirror/com/facebook/react/react-native/

@hypest I am curios should we push RN60 there or maybe I just wrongly configured `react-native-video` https://github.com/wordpress-mobile/react-native-video/compare/enable-android-build-of-standalone-package...wordpress-mobile:experiment/android_x_migration#diff-3a56ae3cf3e05886fec69653e96f2456R53

I must admit that I am not sure how build.gradle file should be configured as I saw that jitpack is trying to call this command: `./gradlew clean -Pgroup=com.github.wordpress-mobile -Pversion=57ad4e840a -xtest -xlint install` in `react-native-video`.
  
Update release notes:

- [x] If there are user facing changes, I have added an item to `RELEASE-NOTES.txt`.

PR #1125, opened 2019-06-12
Fix Android device tests for AndroidX migration

mkevins:
This PR fixes e2e Android device tests for the AndroidX migration.

It seems that UI element locators were broken by a change in the way accessibilityIds are being assigned . `content-desc` has an extra comma and space in each element, described here: https://github.com/wordpress-mobile/gutenberg-mobile/pull/1112#issuecomment-501165250.

The approach in this PR is to proxy calls to the driver, and selectively append the necessary characters (`, `) to the call for any functions that need it. Additionally, where `accessibilityIdXPathAttrib` is tested for equality, an accessibilityIdSuffix is added to the appropriate ids. The suffix is not required when `accessibilityIdXPathAttrib` is tested via `contains`, as the lack of the suffix will not affect the match.

To test:
`yarn test:e2e:android:local`


marecar3:
Hey, @mkevins  👋I have fixed iOS tests so it seems that we are ready to merge this one in AndroidX branch.

marecar3:
Merging this one in https://github.com/wordpress-mobile/gutenberg-mobile/pull/1112

Issue #1124, opened 2019-06-12
Paste multi-paragraph text in Post Title does nothing


etoledom:
Pasting a multi-paragraph text into the Post Title currently does nothing.

This happens because PostTitle is based on RichText, and PostTitle doesn't have proper support for `onSplit` and `onReplace`.

The behavior on WEB is to paste all text as a single line in the Post Title, even if the copied text had multiple paragraphs.

etoledom:
**As a proposed solution**, I think the best would be to not use RichText on PostTitle, since it is coupled with many block-specific behavior, including https://github.com/wordpress-mobile/gutenberg-mobile/issues/1033.

**As an alternative**, I implemented multi-paragraph pasting on PostTitle, on a similar way as it is implemented on Paragraph block: https://github.com/WordPress/gutenberg/compare/rnmobile/multi-paragraph-paste-on-post-title

The result is interesting:
- Pasting multi-paragraph content on Post Title will update the title field with the first paragraph, and create appropriate blocks for the rest one the paragraphs:

![paste-title](https://user-images.githubusercontent.com/9772967/59350701-093be900-8d1d-11e9-97f2-abc911487468.gif)

I think it might be a good thing, but it has its catches, like:
- Pasting multi-paragraph text in the middle of the post title, will split the post title content, leaving just the first half in the title. (as splitting a paragraph block). This might not be expected for users.
- Pasting multi-paragraph text at the end of the title with content, it won't add anything to the title and just add blocks.

**Another possible solution** would be to continue using RichText but to try to emulate the behavior on web. This will probably mean to add more custom code on RichText, which is not ideal.

---

I'd prefer to have the same behavior than web, by replacing RichText with other component (maybe a plane `TextInput`).

What do you think?
@hypest @daniloercoli @SergioEstevao

SergioEstevao:
> I'd prefer to have the same behavior than web, by replacing RichText with other component (maybe a plane TextInput).

This for sure! We are introducing a lot of custom behaviour in RichText because of this. If we can get rid of it it will be great!

daniloercoli:
> I'd prefer to have the same behavior than web, by replacing RichText with other component (maybe a plane TextInput).

This was discussed on Slack not long time ago, but thanks to open the discussion here after your findings.
[...]

Issue #1033, opened 2019-05-27
Title block alters global selection state


hypest:
We're basing the title block on `RichText` on mobile but, the `RichText` component has logic that among other things, updates the global selection state as the caret changes position. That's probably an unwanted side effect and maybe we need to reconsider using the RichText component for the title block as is.

PR #1134, opened 2019-06-14
PlainText based on RCTAztecView

etoledom:
Fixes https://github.com/wordpress-mobile/gutenberg-mobile/issues/1124
Fixes https://github.com/wordpress-mobile/gutenberg-mobile/issues/1033

This PR updates the gutenberg ref to test https://github.com/WordPress/gutenberg/pull/16178

It also add changes to Aztec Wrapper to handle plain text, avoiding all html handling.

### The changes on Aztec are:

- New `plainText` prop:
This prop has the same structure than the `text/contents` prop, but by using it, Aztec will set the text content directly to its non-attributed text property, avoiding any html handling.

- onChange bubble event also changed:
When `plainText` is set, the Aztec wrapper will take the text content directly from the text property, again without any formatting or html handling.
It will be sent in the same structure than before, but with the name `plainText` for the text property.

These changes will need to be reproduced on Android if we decide to go with this solution.

Everything else is managed directly on PlaneText JS side.

Update release notes:

### DIFF
--- a/react-native-aztec/ios/RNTAztecView/RCTAztecView.swift
+++ b/react-native-aztec/ios/RNTAztecView/RCTAztecView.swift
@@ -36,6 +36,9 @@ class RCTAztecView: Aztec.TextView {
 
     @objc var plainText: NSDictionary? {
         didSet {
+            guard plainText?["eventCount"] == nil else {
+                return
+            }
             self.text = plainText?["text"] as? String
             if let content = plainText {
                 setSelection(from: content)
--- a/react-native-aztec/ios/RNTAztecView/RCTAztecView.swift
+++ b/react-native-aztec/ios/RNTAztecView/RCTAztecView.swift
@@ -34,6 +34,16 @@ class RCTAztecView: Aztec.TextView {
         }
     }
 
+    @objc var plainText: NSDictionary? {
+        didSet {
+            self.text = plainText?["text"] as? String
+            if let content = plainText {
+                setSelection(from: content)
+            }
+            updatePlaceholderVisibility()
+        }
+    }
+
     override var textAlignment: NSTextAlignment {
         didSet {
             super.textAlignment = textAlignment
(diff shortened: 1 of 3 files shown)


SergioEstevao:
While this approach works, I really think we should avoid use RCTAztecView for this. Can't we just do another component based on UITextInput and wrap around that?

@daniloercoli is it hard in Android to create a simple text input component that has the Enter problem sorted? or could we send a PR for RN where both enter are properly intercepted?

etoledom:
>  is it hard in Android to create a simple text input component that has the Enter problem sorted

That would be better! I'm not sure about the maintenance cost though.
I'm open to experiment in that proposal and open another PR too!

daniloercoli:
> @daniloercoli is it hard in Android to create a simple text input component that has the Enter problem sorted? or could we send a PR for RN where both enter are properly intercepted?

Starting from scratch it's basically re-creating the default implementation of TextInput available in RN and gives it a proper Enter.key detection mechanism. 
RN TextInput implementation, just to say one thing, has the `eventCount` "hack"  already place, and it's needed for performance reasons, so we will need it in our new simple implementation... We also need to add the code that detects Backspace.Key, and maybe other custom events.

Before start think to create a new RN component, that will bring in new maintenance cost,  I think we can time-box a tentative of looking at the default TextInput implementation of RN and see if we can patch it properly, or find a way to use it and have key pressed events properly recognized.

etoledom:
Created a new PR trying @SergioEstevao's approach and taking into account @daniloercoli's concerns by subclassing TextInput (both JS and Natice[iOS] sides) to add custom functionality on top of it:

https://github.com/wordpress-mobile/gutenberg-mobile/pull/1148

hypest:
Bumping to next milestone since this is still WIP.

etoledom:
Removed the milestone since this is a proof of concept, not intended to be merged.

etoledom:
Replaced by this other approach: https://github.com/wordpress-mobile/gutenberg-mobile/pull/1148

PR #1139, opened 2019-06-14
Add block insertion tests 

JavonDavis:
Part of #745 

To test: Run `TEST_RN_PLATFORM=android yarn device-tests __device-tests__/gutenberg-editor-block-insertion.test.js`

The PR also updates the filename used for the app to include the branch name as a unique identifier. There's a little concern about branches that might be running at the same time using the same filename and end up testing the wrong versions of the app. This change eliminates that

Update release notes:


__device-tests__/gutenberg-editor-block-insertion.test.js
```diff
@@ -0,0 +1,114 @@
...
+		} else {
+			for ( let i = 4; i > 0; i-- ) {
+				paragraphBlockElement = await editorPage.getParagraphBlockAtPosition( 1 );
+				await clickMiddleOfElement( driver, paragraphBlockElement );
+				await editorPage.removeParagraphBlockAtPosition( 1 );
+			}
+		}
+	} );
+
+	it( 'should be able to insert block at the end of post from the title', async () => {
```
mchowning: _Someone_ 😒 merged [a PR](https://github.com/wordpress-mobile/gutenberg-mobile/pull/1201) earlier today that is going to break this test because blocks inserted from the post title are now placed at the _top_ of the post.

__device-tests__/helpers/test-data.js
```diff
@@ -3,7 +3,7 @@ exports.shortText = `Rock music approaches at high velocity.`;
 exports.mediumText = `The finer continuum interprets the polynomial rabbit. When can the geology cheat? An astronomer runs. Should a communist consent?`;
 
 exports.longText = `Beneath the busy continuum blinks the ineffective husband. Why a metric bow outside the official subway? How can the prompt crop exhaust his tree 
-Does this chord crowd my emptied search? A theory bubbles under the cartoon. The discontinued speaker cracks every thick epic. Its extraordinary twin shifts behind
+Does this chord crowd my emptied search? A theory bubbles under the cartoon. The discontinued speaker cracks every thick epic. extraordinary twin shifts behind
```
mchowning: Was this change intentional? If so, I'm just curious to learn why it was needed.

bin/sauce-pre-upload.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+
+# check if CIRCLE_BRANCH variable is set
+if [[ -n "$CIRCLE_BRANCH" ]] ; then
+    # replace / with - and assign to new var SAUCE_FILENAME
+    export SAUCE_FILENAME=${CIRCLE_BRANCH//[\/]/-};
+ else
+    echo "Expected CIRCLE_BRANCH env variable";
```
mchowning: Would it be useful/approrpiate to have something like `exit 1` here?

__device-tests__/helpers/caps.js
```diff
@@ -13,8 +13,8 @@ exports.ios12 = {
 exports.android8 = {
 	browserName: '',
 	platformName: 'Android',
-	platformVersion: '8.0',
-	deviceName: 'Android Emulator',
+	platformVersion: '9.0',
+	deviceName: 'Google Pixel 3 GoogleAPI Emulator',
```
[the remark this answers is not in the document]
JavonDavis: Right! So this is ignored locally because https://github.com/wordpress-mobile/gutenberg-mobile/blob/add/tests-add-block-insertion/__device-tests__/helpers/utils.js#L85, this is to change was mainly a change to the device on CI to a more commonly used one, you can look at one of the video runs to see that change, it's not a super important detail so if you think we need to rever that I have no problem with that.

__device-tests__/gutenberg-editor-block-insertion.test.js
JavonDavis: 😅 working on the update now

__device-tests__/helpers/test-data.js
JavonDavis: The test using the text was being auto corrected otherwise, I saw this as a simple solution.

__device-tests__/helpers/caps.js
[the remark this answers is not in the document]
mchowning: No need to revert, I was just wondering. Thanks for explaining!

bin/sauce-pre-upload.sh
JavonDavis: 🤔 Hmm I guess the added value would be other scripts could then use the exit code as a check, I don't feel too strongly about it so I can go ahead and make that update if you think it's best to make that distinction.

mchowning: I didn't know if exiting immediately here might make the cause of the failure more obvious. But I'm not familiar enough with how this script is used to know if that change would actually help, so if you don't think it would be helpful, I wouldn't bother making the change.

JavonDavis: It's only used here https://github.com/wordpress-mobile/gutenberg-mobile/pull/1139/files/579773229a76b2ab75999b04669fc9ff92d5f743#diff-1d37e48f9ceff6d8030570cd36286a61R141 locally it's expected you can prefix the command with the env var if you're triggering a CI run

mchowning: 👍  I'm fine if you want to leave this as-is.

__device-tests__/helpers/test-data.js
```diff
@@ -1,10 +1,10 @@
 exports.shortText = `Rock music approaches at high velocity.`;
 
-exports.mediumText = `The finer continuum interprets the polynomial rabbit. When can the geology cheat? An astronomer runs. Should a communist consent?`;
+exports.mediumText = `The finer continuum interprets the polynomial rabbit. When can the geology runs? An astronomer runs. Should a communist consent?`;
```
JavonDavis: Changes here were made to avoid autocomplete causing validation errors.

JavonDavis:
It seems an upgrade to the packages cause the tests to stop running in parallel and the length of the runs went up back to about 30 mins, I made a few additional changes to command revert that.

__device-tests__/gutenberg-editor-block-insertion.test.js
mchowning: I see you took care of this with 5ab6353 and 2cf5fe7. 👍

mchowning:
Hey @JavonDavis ! The code changes look good to me, but the tests are failing to run on CI. Looks like the missing file being referenced [was added recently](https://github.com/WordPress/gutenberg/pull/16404/files#diff-8c287010c3d8d7a91eab78ab0a22dfad), but neither `develop` or this PR are referencing a `gutenberg` commit from after that PR was merged. So you might just need to update the submodule ref for `gutenberg`.

PR #1148, opened 2019-06-18
PlainText as TextInput subclass

etoledom:
This PR is an alternative to https://github.com/wordpress-mobile/gutenberg-mobile/pull/1134

Fixes https://github.com/wordpress-mobile/gutenberg-mobile/issues/1124
Fixes https://github.com/wordpress-mobile/gutenberg-mobile/issues/1033

`gutenberg` side PR: https://github.com/WordPress/gutenberg/pull/16208

This PR introduces a new Native Component called `RNPlainText`.

### On JS:
RNPlainText is a subclass of TextInput. This give us all the functionality of TextInput with the ability of being extended in any way we want. On this example, I added the `onEnter` functionality, applied to `PostTitle`.

I'm not sure if this kind of subclassing is good or not recommended. But it seems to work great! (@tug what do you think?)

Currently, `Code Block` works exactly as it was when using PlainText based on TextInput, and PostTitle is now also using PlainText handling `onEnter` events.

### On iOS:

To be able to reuse TextInput functionality, we need to have three new classes:
- `RNPlainTextManager` (subclass of `RCTBaseTextInputViewManager`)
- `RNBackedPlainTextView` (subclass of `RCTBaseTextInputView`)
- `RNPlainText` (subclass of `RCTUITextView`)

All three subclasses works magically with the minimum implementation, and working together with a TextInput subclass on JS side, we get all this functionality for free.

`RCTBaseTextInputViewManager` also uses the shadow view `RCTBaseTextInputShadowView`, so we don't need to worry about sizing either.

We can send UITextView events via `RNPlainTextDelegate` to `RNBackedPlainTextView`, to send them to JS via `RCTBubblingEventBlock` (as we do on Aztec, just with one extra step because of the wrapping view).

(cc @SergioEstevao )

**Drawbacks:**
This new RNPlainText library depends on `RCTText` library, which is independent of `React` library.
This might be tricky to manage via Cocoapods (or not). I remember having some difficulties with Aztec.

### On Android:
~I don't know if this subclassing could be problematic on the Android side. I hope this PR helps to figure that out. Feel free to push changes directly to this branch with the Android implementation if needed (cc @daniloercoli @hypest @marecar3 )~

EDIT: @daniloercoli did the basic Android implementation and seems to work well 🎉 

---

This is a proof of concept and feedback is encouraged to go forward :)


etoledom:
BTW, I think we can use a similar strategy for Aztec too, and use:
- A backed view to hold Aztec, based on `RCTBaseTextInputView`
- A `RCTBaseTextInputViewManager` subclass
- We won't be able to subclass `RCTUITextView`, since it would be replaced by Aztec, but that class is not that big. It adds placeholder and some other few things to UITextView.

In this way we will gain most of the TextInput behavior, including shadow view implementation and size handling, but powered up by Aztec.

We would need to coordinate closely iOS and Android for this though.

@hypest, I remember you were doing something (maybe) similar on Android. Does this sounds familiar?

react-native-plain-text/src/PlainText.js
```diff
@@ -0,0 +1,55 @@
+/**
+ * External dependencies
+ */
+import React from 'react';
+import { requireNativeComponent, TextInput, StyleSheet } from 'react-native';
+
+const emptyFunctionThatReturnsTrue = () => true;
+
+class PlainText extends TextInput {
```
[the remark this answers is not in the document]
etoledom: Sure! That sounds like a good idea.

Should we rename everything then? Even the library `react-native-wp-text-input`? or similar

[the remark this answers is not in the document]
Tug: Subclassing a component isn't a problem per say, but in this case it's a bit problematic since we don't maintain `TextInput`. It's internal implementation can easily change which would break our code each time we update react-native. Could you tell me a bit more about the requirements for `PlainText`, what does it have that `TextInput` doesn't?

etoledom:
>  Could you tell me a bit more about the requirements for PlainText, what does it have that TextInput doesn't?

@tug - On this PR, we are adding an `onEnter` event, that does two things:
- Send an event when the Enter key is pressed.
- Cancels the default behavior (it won't add a new line) when onEnter prop is set.

There is an `onKeyPress` event on TextInput, but it has two problems:
- It doesn't cancel the default implementation.
- On Android, it only works with soft keyboards.

I see your concern about breaking changes. One noticeable fragile section is the `render()` method of our custom TextInput subclass, where we call a bunch of internal methods, and those might easily  change, but since it's basically a copy paste without mixing with our custom code, those changes might not be difficult to overcome.

On the native side we use a clean subclass (at least on iOS), were we don't depend on internals directly, but we do depend on the structure. If Facebook were to refactor the structure of the native side of TextInput, we would need to adapt as well, but that probably won't happen on every version update.

For our custom events, at least with `onEnter` we don't depend on internals directly, so those shouldn't break that easily.

react-native-wp-text-input/ios/WPBackedTextInputView.swift
```diff
@@ -0,0 +1,38 @@
+import UIKit
+
+class WPBackedTextInputView: RCTBaseTextInputView {
+    lazy var _backedTextInputView: WPTextInptu = {
```
SergioEstevao: Typo on the variable name... should be `WPTextInput`

etoledom: Absolutely!
Thanks for spotting that out :)

react-native-wp-text-input/ios/WPBackedTextInputView.swift
```diff
@@ -0,0 +1,38 @@
...
+        let textInput = WPTextInptu(frame: self.bounds)
+        textInput.autoresizingMask = [.flexibleWidth, .flexibleHeight]
+        textInput.backgroundColor = .clear
+        textInput.textColor = .black
+        textInput.textContainer.lineFragmentPadding = 0
+        textInput.scrollsToTop = false
+        textInput.isScrollEnabled = false
+        textInput.textInputDelegate = self
+        textInput.wpTextInputDelegate = self
```
etoledom: Not a fan of this name, but `textInputDelegate` is already used.

react-native-wp-text-input/ios/WPBackedTextInputView.swift
etoledom: Apparently I copied one name with the typo and pasted it all over the place 😛
Thanks @SergioEstevao for fixing the file names!

etoledom:
@hypest what do you think about this approach for a PlainText component?

@SergioEstevao also agreed that it looks good.

We'd like to know if this approach looks good enough to dedicate more time to it a finish a proper PR.

hypest:
Bumping to next milestone as it is still WIP.

hypest:
> @hypest what do you think about this approach for a PlainText component?

I like it and I think it's workable! It adds a very narrow additional interface to TextInput, which means that we might even find a way to contribute this back upstream.

My vote is 👍 for moving ahead with this approach.

Tug:
@etoledom Thanks for the explanation!

Sorry if I'm misunderstanding the goal of the PR, but I'm finding this approach a bit overkill (maybe?)

I spent a bit of time trying to find a solution using composition instead of inheritance but even if we manage to make it work nicely, the fact that TextInput's `onKeyPress` only works with soft keyboards on Android is indeed problematic.

It sounds like we're trying to patch react native limitations here and I find it a bit worrying. The main reason is that we're adding quite a complex (and fragile imo) piece of code that we'll have to maintain on the long run.

What would be the alternatives you think?
Couldn't we use Aztec for the title instead for instance?

etoledom:
> Couldn't we use Aztec for the title instead for instance?

Hey @Tug !

As you know We are currently using `RichText` that is based on Aztec. This has been giving us problems, and this is one of the big motivations to look for alternatives. So my first idea was to just take out RichText from the equation.

On a [previous proposal](https://github.com/wordpress-mobile/gutenberg-mobile/pull/1134), I investigated the idea of using `RCTAztecView` to power PlainText.
Although it works pretty well, the idea didn't seem to look good enough to the team. 

I had to introduce some small custom code to the native side of the Aztec wrapper to make it work nicely as a PlainText component. 

I believe that the [PlainText based on RCTAztecView](https://github.com/wordpress-mobile/gutenberg-mobile/pull/1134) alternative is the simplest to start with, and the custom changes introduced to Aztec Wrapper are less flaky than the changes on RichText needed to be used as a PlainText-like component, but it will probably hit us back over time.

Another alternative was to start from scratch, but we would have to re-implement AztecWrapper with all our patches, or re-implement React-Native's implementation, and both ways are too much extra work and maintenance. But since we can subclass, I got to the alternative present on this proposal.

I agree that this currently looks like an overkill, but I think this can be helpful as a starting point.
The native part, since is object-oriented in nature, looks quite elegant and extensible (I can talk for iOS). It's a proper extension of RN's implementation without messing with its internal code. We only will have problems if they change the architecture of the whole RN Text package.

What really worries me is [WPTextView.js](https://github.com/wordpress-mobile/gutenberg-mobile/blob/a6744f503a1d90c30b8e32f9bba77bad70fe4981/react-native-wp-text-input/src/WPTextInput.js), and it's probably what worries you too, so I agree with that too!
I didn't expected the need to copy-paste so much code.

I still have to upgrade this PR with the RN upgrade to 0.60. I think this will be a good exercise to know how easy/difficult will be to maintain this component as it is presented now.

hypest:
Bumping to the next milestone since this is still under review.

hypest:
Is this PR still relevant? Can we close it if not @etoledom ?

etoledom:
Closing since there's no plan to continue working on this any soon.
I'd love to come back to this at some point! (It was working pretty well, and it might help and simplify RichText if we implement a similar strategy on it)

PR #1168, opened 2019-06-20
📦 Add support for Haul packager

koke:
We have been experimenting with ways of loading external code that lives outside the project root, and [metro](https://facebook.github.io/metro/) seems to be a limiting factor there.

Metro [doesn't support symlinks](https://github.com/facebook/metro/issues/1), which is a highly requested feature, and it seems all our attempts to resolve modules outside the project root have failed so far. We are still experimenting with that option, since metro is the officially supported packager, and will likely cause less trouble than the alternative with future React Native upgrades. At least for now, since Haste will be [deprecated in 0.61](https://github.com/facebook/react-native/issues/24316), opening the door to better official support for other packagers, but we can't wait a few months for that to happen.

A good looking alternative is [Haul](https://callstack.github.io/haul/). Its main benefit is that it's based on Webpack, so it's easier to reuse existing configurations from Gutenberg and other projects we might want to bring in as plugins. And its documentation is exponentially better than Metro's.

In this PR I add support for bundling gutenberg-mobile with Haul as an option. If you run the same commands that you used to, it shouldn't affect your workflow.

## Usage

The choice of packager is controlled by the `PACKAGER` environment variable. For Haul, you should set `PACKAGER=haul`, anything else will default to Metro.

To start the server:

```sh
PACKAGER=haul yarn start
```

To generate the bundles

```sh
PACKAGER=haul yarn bundle
```

## Implementation

I created a tiny wrapper in `bin/cli.js` to either call the `haul` or `react-native` tools depending on the `PACKAGER` variable.

⚠️ I had to disable Hot Module Reloading, because the way it works was breaking block validation. It's something we can investigate later, but `react-hot-loader` was wrapping each component inside a proxy component, including `RichText.Content`. The Gutenberg element serializer was expecting a function instead of a component at one point, and rendering the contents differently, which created validation issues (and would probably have broken the saved content as well).

⚠️ I added a couple of custom loaders for importing SASS files in React Native code, which should become packages instead of living in the root folder. I should clean that up before merging.

Using Webpack allows us to use the `resolve.alias` configuration, meaning that we can have Haul find the gutenberg packages where they are.

## Testing

For now, all my testing has been just getting to the point where gutenberg launches. It seems to work so far, but I haven't done extensive testing yet.

Besides checking that nothing is broken, I'm looking for feedback on how this makes the development workflow better or worse, so we can take that into account when we evaluate options.


mkevins:
Thanks for creating this branch @koke. I have been using this as a basis for some additional experiments: [Nested plugin repo built from source](https://github.com/wordpress-mobile/gutenberg-mobile/compare/try/haul...try/haul-matt), [Nested plugin repo build from pre-built bundle (with externals)](https://github.com/wordpress-mobile/gutenberg-mobile/compare/try/haul-matt...try/haul-matt-externals).

It seems with Haul (and Webpack) a lot of additional configuration options become available.

> Its main benefit is that it's based on Webpack, so it's easier to reuse existing configurations from Gutenberg and other projects we might want to bring in as plugins. And its documentation is exponentially better than Metro's.

I agree with this, especially about the documentation.

> In this PR I add support for bundling gutenberg-mobile with Haul as an option. If you run the same commands that you used to, it shouldn't affect your workflow.

I very much like this approach in general, and especially here. The set of possibilities becomes a superset, rather than a subset, or an overlapping set.

While exploring the tangents mentioned above, along with @pinarol's [PR exploring using metro](https://github.com/wordpress-mobile/gutenberg-mobile/pull/1170) to build the plugins from an outside directory, I realized there are a few combinations or "hybrid" setups that may be possible. In Pinar's PR, gutenberg-mobile is still built with metro, with the plugins being built (externally) using the `wp-scripts` (w/ Webpack). Then, metro is configured to watch those changes. I wonder if `wp-scripts` could be tweaked to use externals, and build a bundle that metro could consume (which I think would alleviate the need to use the exposeGlobals function).

The downside to developing that way is that development would be slightly cumbersome, since we'd need to run two commands in separate terminals. We could hypothetically find a way to wrap that in a script, but I think logging would get "mixed up" there, and I don't know if it'd be worth the effort to sort out.

Issue #1173, opened 2019-06-25
Video Block doesn't continue upload after losing and restoring network connection


JavonDavis:
This specifically happens if the user tries to save a draft offline while the post contains a video block on iOS. This also leads to some data loss if there's other content in the post. 

Steps to reproduce
1. Create a post with some content
2. Add a video block and try to add a video
3. before the upload completes switch to Airplane mode
4. Try to save the post as a draft

At this point you should see a white space where the video block should be

![IMG_0046](https://user-images.githubusercontent.com/6520461/60063917-45c7f700-96c4-11e9-89ce-3ef2fd9e8dba.PNG)

5. Turn off airplane mode and connect to the network 
6. Try to save the post as a draft

At this point, you should see something like this,

![IMG_0047](https://user-images.githubusercontent.com/6520461/60063930-5b3d2100-96c4-11e9-838f-97d91cb47a01.PNG) with the HTML matching up with this

Both these screenshots are from the same post, and the content shown in the first one was replaced with older content in the post the save completed. 

The preview shows something completely different,

![IMG_0050](https://user-images.githubusercontent.com/6520461/60064068-d9012c80-96c4-11e9-81e0-3725f2528d14.PNG)

and on the web it lines up with the preview

![Screen Shot 2019-06-24 at 9 14 21 PM](https://user-images.githubusercontent.com/6520461/60064136-15cd2380-96c5-11e9-8928-e0ea6cfe3eed.png)

The video also wasn't uploaded properly and does not play on web or mobile.

pinarol:
hey @JavonDavis 👋  Could you also switch to the html mode and add the html output here? That'll be some useful info when debugging.

JavonDavis:
> with the HTML matching up with this

@pinarol I had mentioned it here but luckily I did take a screenshot too 😄 

![IMG_0051](https://user-images.githubusercontent.com/6520461/60101168-e2b87d80-9720-11e9-9058-b10126395971.PNG)

PR #1182, opened 2019-06-27
[iOS]Fix: Video Block doesn't continue upload after losing and restoring network connection

pinarol:
Fixes https://github.com/wordpress-mobile/gutenberg-mobile/issues/1173

gutenberg PR: https://github.com/WordPress/gutenberg/pull/16331

There are multiple issues need to be addressed:

1. ~~Local URL remains in the video block when upload fails.~~ Update: It looks like current upload mechanism depends on this local url to update the post if upload finishes when the block editor is closed. So we can't just remove it. 
2. Upload progress events are not received from the main app as they used to, when we close/re-open a post during an upload we need to wait a lot to receive the upload progress event, which wasn't the case before. So this might be a side effect of changes on iOS side. 

To test:

WPiOS (@etoledom)

- [ ] Video upload from device library
- [ ] Video upload by capturing 
- [ ] Select video from WP Media library
- [ ] During the upload close/re-open the post and see it continues the progress
- [ ] Video upload with no network, let it fail & retry upload
- [ ] Video upload cancel
- [ ] Try uploading multiple videos, close/re-open the post, wait until they complete and see everything is OK.
- [ ] Try uploading multiple videos, close the post, wait until uploads are finished, reopen the post and see everything is OK.

WPAndroid (@marecar3)

- [x] Video upload from device library
- [x] Video upload by capturing 
- [x] Select video from WP Media library
- [x] During the upload close/re-open the post and see it continues the progress
- [x] Video upload with no network, let it fail & retry upload
- [x] Video upload cancel
- [x] Try uploading multiple videos, close/re-open the post, wait until they complete and see everything is OK.
- [x] Try uploading multiple videos, close the post, wait until uploads are finished, reopen the post and see everything is OK.

Update release notes:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,15 +6,16 @@ commands:
       - restore_cache:
           name: Restore Yarn Cache
           keys:
-            - yarn-v2-{{ arch }}-{{ checksum "yarn.lock" }}
+            - yarn-i18n-v1-{{ .Environment.CIRCLE_JOB }}-{{ checksum "yarn.lock" }}
       - run:
           name: Yarn Install
-          command: YARN_CACHE_FOLDER=~/.cache/yarn yarn install --frozen-lockfile
+          command: yarn install --frozen-lockfile --prefer-offline
       - save_cache:
           name: Save Yarn Cache
-          key: yarn-v2-{{ arch }}-{{ checksum "yarn.lock" }}
+          key: yarn-i18n-v1-{{ .Environment.CIRCLE_JOB }}-{{ checksum "yarn.lock" }}
           paths:
-            - ~/.cache/yarn
+            - node_modules
+            - i18n-cache/data
   checkout-gutenberg:
     steps:
       - run:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -101,6 +101,13 @@ jobs:
         command: |
           echo 'export TEST_RN_PLATFORM=ios' >> $BASH_ENV
           echo 'export TEST_ENV=sauce' >> $BASH_ENV
+    - run:
+        name: Prepare build cache key
+        command: find yarn.lock ios react-native-aztec/ios react-native-gutenberg-bridge/ios -type f -print0 | sort -z | xargs -0 shasum | tee ios-checksums.txt
+    - restore_cache:
+        name: Restore Build Cache
+        keys:
+          - ios-build-cache-{{ checksum "ios-checksums.txt" }}
     - restore_cache:
         name: Restore Dependencies Cache
         keys:
@@ -109,8 +116,8 @@ jobs:
         - dependencies-{{ checksum "react-native-aztec/ios/Cartfile.resolved" }}
         - dependencies-
     - run:
-        name: Yarn preios
-        command: yarn preios
+        name: Yarn preios (if needed)
+        command: test -e ios/build/gutenberg/Build/Products/Release-iphonesimulator/gutenberg.app || yarn preios
     - save_cache:
         name: Save Dependencies Cache
         key: dependencies-{{ checksum "react-native-aztec/ios/Cartfile.resolved" }}-{{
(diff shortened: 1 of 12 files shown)


pinarol:
> Upload progress events are not received from the main app as they used to, when we close/re-open a post during an upload we need to wait a lot to receive the upload progress event, which wasn't the case before. So this might be a side effect of changes on iOS side.

@diegoreymendez 👋I see that recently some changes were made to the media upload on WP-iOS app. Could you help me check if there are some side effects related to those ? You might want to use this branch to experiment because I made a change on gutenberg level to make sure video block is registering for upload events when component is mounted.

diegoreymendez:
@pinarol - The changes we did are just to trigger uploads when the App is opened / comes to foreground and when a connection is re-established.  They are new triggers to make the uploads happen but they don't really change any existing logic for uploads.

pinarol:
Good catch @etoledom ! I had to revert the changes that I made to prevent keeping setting the local url in the attributes.src. It turns out the main app uses that info to update the post content when upload finishes during block editor is open. 

So this PR is fixing 2 things only:
1. Upload was not continuing when we close / open post
2. In this situation the video block seemed blank but now it has a placeholder:

Before:
Screen Shot 2019-07-04 at 15 55 21

After:
Screen Shot 2019-07-04 at 15 41 48


Ready for another look

pinarol:
I still can't make sense of why we are receiving upload events with a little delay compared to the past. When I open a post with an ongoing upload the spinner doesn't get visible right away(because no upload events are received from the main app). I remember like it was getting visible quickly in the past. Do you observe the same thing @etoledom ?

etoledom:
@pinarol I updated from develop hoping that the last commit would fix the broken UI Tests

etoledom:
> Do you observe the same thing @etoledom ?

Yes I did notice a small delay. It's usually quite small, but it looks like when I re-open the post with the ongoing upload on the final stage, the block won't receive any event until it's finished. That last process is usually longer and feels like a bigger delay.

etoledom:
Thank you @pinarol for the update!

The first half of the original Issue https://github.com/wordpress-mobile/gutenberg-mobile/issues/1173 it's solved.

And the details I noticed on https://github.com/wordpress-mobile/gutenberg-mobile/pull/1182#pullrequestreview-256171788 are both fixed 🎉 

I did notice two issues: One seems to be exclusive from this branch, while the second is still reproducible on develop.
I also experienced 2 different crashes related to media upload, but I find it unlikely that they are related to this PR, and they are also difficult to reproduce (they are related to CoreData)

**Issue 1**:
On many of my tests, uploading and replacing video blocks, I ended up with the post being duplicated. And it's not local core data problem since they got duplicated on the web too.

Screen Shot 2019-07-05 at 8 46 13 AM

I wasn't able to reproduce this on develop, but I wonder how can this be affected by this PR.

Steps:
- Add a video block
- Add a video from camera roll
- While it's updating, turn off internet connection to make it fail
- Save as draft
- Turn ON internet connection again
- Wait until the video is uploaded
- Wait for the toast message "Post draft updated"
- Close the post and refresh the post list.

I was able to check on web that 2 posts are created at the same time upon saving draft.

**Issue 2**:
Following the second part of the original issue https://github.com/wordpress-mobile/gutenberg-mobile/issues/1173 
After step 6, pressing the Save As Draft, the app will get stuck with the blocking `Saving...` Popup forever. Here is a video with the whole sequence https://cloudup.com/cvx_J8gNlf2

As I mentioned, this also happens on develop, and we can create a new ticket.

**Crashes:**
I've got one on video but I'm not able to reproduce it following the same exact steps.
I'll try to investigate them a bit more and I'll post a stack trace to get some idea about what might be happening.

But again, I find it unlikely that those crashes are exclusive from this branch.

pinarol:
@etoledom 👋Thanks for the detailed testing. I could repro the `Issue 1` on develop branch also. Although I doubt that it is a bug. Offline save of a draft is maybe just taking a snapshot of the post and duplicating it to avoid data loss.

About `issue 2`: 
> After step 6, pressing the Save As Draft, the app will get stuck with the blocking Saving... Popup forever.
I saw this one too, but after waiting a minute or so it is displaying an error message. So in my case it was not forever but definitely needs a shorter timeout.

Let's open issues on WPiOS repo if you think those are not related with this PR. These all seem like good tasks for Groundskeeping.

pinarol:
@marecar3 Could you also check this on WPAndroid? Thanks in advance.

etoledom:
>  I could repro the Issue 1 on develop branch also

Thanks for trying it out! Now that you mentioned I was able to reproduce it on develop too.

Since all the issues I found are not related to this PR, I think it's ✅ to :shipit: on the iOS side of things.
@marecar3 let me know if you think the Android side is good to go as well 👍

marecar3:
Hey sorry for missing this one. 
I will take a look at it right away! :)

Issue #981, opened 2019-05-14
Remove logging of translations on Gutenberg initialization


koke:
At least on WPiOS, all the translations are logged every time Gutenberg opens, which makes it super noisy for development and support.

etoledom:
Fixed via #1183

PR #1183, opened 2019-06-28
Reduce iOS bridge logging

etoledom:
This PR introduces a way to modify the level of logging. It also modifies the default log level threshold to `trace` if the packager is running (as it was before), and to `error` otherwise (the default for release builds). This help non-gutenberg devs to avoid the gutenberg verbose logs when they interact with the block editor.

**Note:** The sockets error messages are not managed by this PR :( 

To test:
- Launch the demo app from Xcode (ios/gutenberg.xcodeproj) with metro running.
- Check that you see the `Running application gutenberg` logs
  - There is one from JS side sent through the bridge and one directly from the native side.
- Check that you see some warnings when the content appears.
- Rebuild and lunch the app again.
- Stop the metro server while the app is building.
- Check that you don't see any of the previously mentioned logs.

**Note 2:** I'm able to launch the demo app without metro server. I'm not sure if that's the default for anybody now or is because I already launched the UI test suit locally.

### DIFF
--- a/react-native-gutenberg-bridge/ios/Gutenberg.swift
+++ b/react-native-gutenberg-bridge/ios/Gutenberg.swift
@@ -29,6 +29,15 @@ public class Gutenberg: NSObject {
         return !bridge.isLoading
     }
 
+    public var logThreshold: LogLevel {
+        get {
+            return LogLevel(RCTGetLogThreshold())
+        }
+        set {
+            RCTSetLogThreshold(RCTLogLevel(newValue))
+        }
+    }
+
     private let bridgeModule = RNReactNativeGutenbergBridge()
     private unowned let dataSource: GutenbergBridgeDataSource
 
@@ -61,6 +70,8 @@ public class Gutenberg: NSObject {
     public init(dataSource: GutenbergBridgeDataSource, extraModules: [RCTBridgeModule] = []) {
         self.dataSource = dataSource
         self.extraModules = extraModules
+        super.init()
+        logThreshold = isPackagerRunning ? .trace : .error
     }
 
     public func invalidate() {
@@ -103,6 +114,10 @@ public class Gutenberg: NSObject {
         bridgeModule.sendEventIfNeeded(name: EventName.setFocusOnTitle, body: nil)
     }
 
+    private var isPackagerRunning: Bool {
+        let url = sourceURL(for: bridge)
+        return !(url?.isFileURL ?? true)
+    }
 }
 
 extension Gutenberg: RCTBridgeDelegate {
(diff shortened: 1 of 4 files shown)


koke:
I think I'd prefer to have this logic in the apps instead. I think it's best to pass everything to the app and let `GutenbergBridgeDelegate.gutenbergDidEmitLog(message:logLevel:)` decide what to log.

Besides, if the logging is too verbose maybe we are logging too many things? See #981

etoledom:
> See #981

Yes, this PR was motivated at first from that issue. The translations are logged as props on the `Running application gutenberg` log. This is coming from two sides:

From the Native side:
https://github.com/facebook/react-native/blob/3b88f0784050631ec309358782e1c4b1cb5579fe/React/Base/RCTRootView.m#L293

From the JS side:
https://github.com/facebook/react-native/blob/3b88f0784050631ec309358782e1c4b1cb5579fe/Libraries/ReactNative/AppRegistry.js#L193

If you deactivate all the logs suppression "hacks" that I mentioned [here](https://github.com/wordpress-mobile/WordPress-iOS/pull/12005#issue-291899167), you will see the `Running application gutenberg` log with all the translations being logged two times.

The JS side log can be silenced at the delegate implementation, but it would be nice to keep logging trace` and `info` when we are on gutenberg dev mode. 
In the other hand, the native side log can just be silenced by setting the RCTLogThreshold manually.

The idea of the new bridge property `logThreshold` is to both modify `RCTLogThreshold` and keep the RN native side and JS side log thresholds in sync.

What do you think about keeping the bridge property `logThreshold`, but let the client modify it when they want?
I'd like to make public the `isPackagerRunning` property in that case too, to be able to make logging more verbose when we need it (gutenberg dev). We can rename it to `isDevMode` or something similar.

Do you still prefer that the client manages both JS side logs and RN Native side logs manually and independently?

react-native-gutenberg-bridge/ios/Gutenberg.swift
```diff
@@ -61,6 +70,8 @@ public class Gutenberg: NSObject {
     public init(dataSource: GutenbergBridgeDataSource, extraModules: [RCTBridgeModule] = []) {
         self.dataSource = dataSource
         self.extraModules = extraModules
+        super.init()
+        logThreshold = isPackagerRunning ? .trace : .error
```
koke: Can we do `.warn` instead of `.error` our would that still be too verbose?

etoledom: Yeah we can, we would see all the YellowBox messages though. 
And `.error` is the default for for Release builds, so if we change it, we would have to check for `#if DEBUG` to select between `.warn` and `.error`.

As you mentioned, maybe we can leave it as it is now and tweak it later if needed.

etoledom:
Thank you @koke ! Let's keep improve it :)

Issue #632, opened 2019-02-20
"Add New Block" indicator isn't shown when title field is focused


iamthomasbishop:
### Expected behavior

If I tap the Inserter toggle to choose a new block while focused on the title, I expect the "add new block" indicator to show, just like when in the body of the post.

### Actual behavior

Indicator isn't shown while focused on title.

Example:

![screenshot_20190220-151630](https://user-images.githubusercontent.com/1200432/53125130-cca47180-3522-11e9-8f0c-3210f5931009.png)

Steps to reproduce (iOS and Android):

1. Start new post in mobile GB
2. Add or focus title
3. Tap (+) Inserter toggle
4. Notice that there is no position indicator

Potentially related issue: https://github.com/wordpress-mobile/gutenberg-mobile/issues/543

### Outstanding Tasks

- [ ] Show ABH indicator when immediately adding a block to a new post ([context](https://github.com/wordpress-mobile/gutenberg-mobile/issues/632#issuecomment-513213114))

mchowning:
This was closed automatically by the PR #1219, but I'm reopening it because there is still a remaining issue to be resolved. 

When an empty post is first opened, until the user interacts with the "Start writing..." portion of the post in some way (i.e., tapped on it), there will be no "Add Block Here" indicator shown when the user opens the add block bottom sheet—this includes if the user has only interacted with the post title. The technical reason for this is that the ABH indicator is attached to our block holders, but until the user taps on the "Start writing..." placeholder text in a new post, the post does not actually contain any blocks (it just has a RichText field that is replaced with a block onTouch), so there is nothing to show an ABH indicator.

This can be fixed (possibly without too much difficulty), but before going down that path I wanted to see how we wanted to approach this. On one hand, not having the indicator when a new post is first created does not seem like a big deal since the entire post content ("Start writing...") will be replaced with whatever block is added. On the other hand, this is inconsistent with how this is handled in other cases. For example, if the user has a post with only a single empty paragraph block (which is what it _looks_ like when you open a new post, but the empty paragraph block is only added once you actually interact with the post content) then the ABH indicator will appear. 

To be more precise, now that the initial fix for this issue is merged we have this behavior:
1. Open new empty post in WPAndroid/WPiOS
2. Observe that the post title is selected, but do not select the content section ("Start writing...")
3. Click the ➕ icon to open the add new post bottom sheet
4. **Observe that the ABH indicator IS NOT shown**
[...]

PR #1201, opened 2019-07-05
Insert blocks at top of post when post title is selected

mchowning:
### Summary
This PR updates the behavior when inserting a block which the post's title is selected. Previously, the block would be inserted at the bottom of the post, but now it will be inserted at the top of the post. This work is being done in advance of addressing #632 (missing "Add New Block" indicator when inserting a block from the post's title) so that we're inserting the post at the proper location before we add an "Add New Block" indicator for the insertion position.

[Related gutenberg PR](https://github.com/WordPress/gutenberg/pull/16440)

Before | After
---- | ----
before mp4 | after mp4

#### Post Title selected scenario
1. Load demo app
2. Tap on the post Title to select it
3. Add any block
4. Observe that the newly added block is inserted at the **top** of the post

#### No block selected scenario
1. Load demo app
2. Do **NOT** select any block
3. Add any block
4. Observe that the newly added block is inserted at the **bottom** of the post

#### Any block selected scenario
1. Load demo app
2. Select any block **other than the post's Title**
3. Add any block
4. Observe that the newly added block is inserted below the previously selected block

### Update release notes

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -2,6 +2,7 @@
 ------
 * Tapping on an empty editor area will create a new paragraph block
 * Fix content loss issue when loading unsupported blocks containing inner blocks.
+* Adding a block from the Post Title now inserts the block at the top of the Post.
 
 1.8.0
 ------


mchowning:
>Since this PR targets v1.10, let's wait until the code freeze is finished to merge this 

Hey @etoledom ! Based on the conversation we just had, I'm going to go ahead and merge this for 1.9.

Issue #1214, opened 2019-07-10
Run mobile tests in Gutenberg CI


etoledom:
Since our mobile unit tests are already on the gutenberg repo, the next step is to make them run on the Gutenberg CI.

- This new CI task should be allowed to fail.

PR #1215, opened 2019-07-10
Fix issues generated from "Run tests on gutenberg repo" PR

etoledom:
fix https://github.com/wordpress-mobile/gutenberg-mobile/issues/1214

This PR is the companion of https://github.com/WordPress/gutenberg/pull/16404

This fixes the breaking test caused by changes and files moved to `gutenberg` repo.

Now all native mocks are defined in the `gutenberg` repo, so we don't need them in this repo anymore. Also the jest `setup.js` and `enzyme.config.js` were moved to `gutenberg`.

There was a weird error caused by a mismatch of `react` and `react-dom` versions. We need to keep them synchronized.

**Note:**
- Running local tests after have run `npm install` in the `gutenberg` submodule will make the tests fail.

**To test:**
- Run `yarn test`
- All tests should pass
- All warnings should have been solved


etoledom:
Marked as Not ready for review until I update from develop.

etoledom:
I believe this is ready for review 🙏

etoledom:
Thank you both! 🎉 

> Just a note that we should be able to remove jest.config.js completely once we've ported the last remaining test (src/index.test.js).

I completely agree!
The tests described in that file actually look like good candidates for the integration tests we want to run on Gutenberg CI.

PR #1245, opened 2019-07-25
Prevent always dismissing keyboard when switching away from empty image caption

### LINKED ISSUE #1221: Virtual keyboard dismisses when focus moves from image/video caption to paragraph
Steps to reproduce:

1. Run the demo app
2. Tap on the image (or video) block to select it
3. Tap in its caption. The virtual keyboard should come up.
4. Tap in a paragraph or heading block
5. Notice the virtual keyboard becoming dismissed while the focus and caret is in the paragraph block

mchowning:
Fixes #1221 

Detailed description and testing steps are in the [related gutenberg PR](https://github.com/WordPress/gutenberg/pull/16759)

There are user facing changes, so I have updated `RELEASE-NOTES.txt`.


RELEASE-NOTES.txt
```diff
@@ -4,6 +4,7 @@
 * Deselect post title any time a block is added
 * Fix merging of empty paragraph blocks when added from a post's title
 * Fix loss of center alignment in image captions
+* Prevent keybaord dismissal when switching between caption and text block
```
hypest: s/keybaord/keyboard 😺

mchowning: 🤦‍♂️

Thanks for catching that @hypest ! Fixed now.

mchowning:
I updated this PR to no longer update the README. That will be taken care of in [this PR](https://github.com/wordpress-mobile/gutenberg-mobile/pull/1254).

PR #1246, opened 2019-07-26
Release v1.10.0

Tug:
This merges `release/1.10.0` to `master` for the next releases of the WPAndroid/WPiOS apps.

Gutenberg branch: https://github.com/WordPress/gutenberg/pull/16775
WPAndroid PR: https://github.com/wordpress-mobile/WordPress-Android/pull/10286
WPiOS PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/12210


RELEASE-NOTES.txt
```diff
@@ -1,3 +1,10 @@
+1.10.0
+------
+* Adding a block from the post title now shows the add block here indicator.
+* Deselect post title any time a block is added
+* Fix merging of empty paragraph blocks when added from a post's title
```
mchowning: @tug In light of your comment (https://github.com/WordPress/gutenberg/pull/16677#issuecomment-515484612) that we may not want this change 
 and since we're likely going to lose it with your imminent PR merge, we might want to pull this line out of the release notes. Alternatively, I could revert the entire change from https://github.com/wordpress-mobile/gutenberg-mobile/pull/1227. What do you think?

RELEASE-NOTES.txt
```diff
@@ -1,3 +1,10 @@
...
+* Fix loss of center alignment in image captions
```
mchowning: Looking back at this now, seems like I perhaps should have called this out as being an Android-specific issue. I'm happy to add that change if you agree.

RELEASE-NOTES.txt
```diff
@@ -1,3 +1,10 @@
+1.10.0
+------
+* Adding a block from the post title now shows the add block here indicator.
```
mchowning: There is one final edge case for this (adding from the post title on a new post when there is just the placeholder text and no paragraph block yet) is still not fixed. Should we pull this from the release notes until that is fixed?

RELEASE-NOTES.txt
Tug: If you think we can easily revert it, let's do this. Otherwise keeping it shouldn't impact the user experience too much I think

RELEASE-NOTES.txt
Tug: Ok, I'll update and remove it from https://github.com/wordpress-mobile/WordPress-iOS/pull/12210/files#diff-a3d054590c32de57d87be233927e755dR9

RELEASE-NOTES.txt
Tug: I think this is fine as it's included in this release

mchowning: Aww, but that screws up my plan of changing *all* the release notes. 😉

RELEASE-NOTES.txt
mchowning: Reverted.

PR #1248, opened 2019-07-26
Revert when inserting block from title replace block if appropriate

mchowning:
Reverting change that made it so that multiple empty paragraph blocks could not be inserted from the post title. [Related gutenber PR](https://github.com/WordPress/gutenberg/pull/16773).

Update release notes:


mchowning:
Hold on, the release notes got messed up. Fixing that now.

PR #1251, opened 2019-07-29
Enforce Prettier (add CI and pre-commit hook)

### LINKED ISSUE #816: Prettier not being used consistently
The [README](https://github.com/wordpress-mobile/gutenberg-mobile/blob/develop/README.md) states that this project uses prettier to enforce our code style, but if you run `yarn prettier` on the develop branch there are currently >30 files that are changed.

This seems like a problem because:
- We are not getting the benefits of a consistent code style across the project
- New developers to the project who run prettier as the README directs may think that all of the changes means there is an issue with their project setup.
- Developers will either (1) avoid running `yarn prettier` when they make changes because of the large number of unrelated style changes they will need to separate out from their intended changes or (2) will run `yarn prettier` and commit those changes, resulting in PRs that are very hard to review because of the unrelated style changes.

A few options come to mind for what we can do to handle this:

1. **Drop our usage of prettier.** I am not a fan of this because I think there is a lot of value in having a consistent code format that no one has to think about. With that said, this is the easiest solution to implement and would slightly simplify the project. Also, we **[just](https://github.com/wordpress-mobile/gutenberg-mobile/pull/7)** added prettier, we can't give up this soon! :-)
2. **Add a [git hook to the project that runs prettier automatically](https://prettier.io/docs/en/precommit.html).** I have worked on another project that has taken this approach and it has worked well. It does make me feel a bit weird when there is something that changes my code automatically, but I have not seen any issues with it since prettier seems to be a very solid tool.
3. **Have our CI reject PRs that are not 'pretty'.** This does the most to make sure that all committed code is pretty, but it will also add some friction to the development process. (push change -> build failed because you forgot prettier -> run prettier -> push prettier changes -> get PR comments, and repeat).

It appears that Option 3 was the intended path when prettier was added to the project by @maxme since one of the commit messages is "[Add style check command (to be used in travis)](https://github.com/wordpress-mobile/gutenberg-mobile/pull/7/commits/e439b3f48259c34c1e0b72eac2665d3f6cb3c8aa)". Maybe this decision was already made and we just need to update our build job on Travis. Of course, we could do both Options 2 & 3, so that prettier is automatically run and if you somehow manage to skip the git hook then Travis will catch it.

Once we've picked an option (assuming we don't pick Option 1), we also need to do a PR that just catches the project's code style up.

mchowning:
Fixes #816 by adding a prettier check to our CI builds and as a pre-commit hook.

This was tested by verifying that the "Check Correctness" check on CI [failed](https://circleci.com/gh/wordpress-mobile/gutenberg-mobile/10121?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) before updating the code in this repo to comply with prettier, and the commit adding those changes caused the check to then [succeed](https://circleci.com/gh/wordpress-mobile/gutenberg-mobile/10124?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link).

It may be easier to review this PR by looking at the commits so you can see the changes that are not solely prettier formatting updates separately. The vast majority of the changes in this PR are just formatting changes.

One mildly annoying aspect of this is the running prettier with the `--list-different` flag (which is needed to get a usable exit code) is that the console output seems rather unclear to me about whether it failed or succeeded ("success!" -> "error!"):

Screen Shot 2019-07-29 at 5 06 25 PM

That is why I decided to also echo some error messages to the console and make it clear what was occurring. Certainly open to other thoughts though.

### Update release notes
Not applicable because there are no user facing changes.

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -1,25 +1,22 @@
+/** @format */
+
 module.exports = function( api ) {
 	api.cache( true );
 	return {
-		presets: [
-			'module:metro-react-native-babel-preset',
-		],
+		presets: [ 'module:metro-react-native-babel-preset' ],
 		plugins: [
 			'@babel/plugin-proposal-async-generator-functions',
 			'@babel/plugin-transform-runtime',
 			[
 				'react-native-platform-specific-extensions',
 				{
-					extensions: [
-						'css',
-						'scss',
-						'sass',
-					],
+					extensions: [ 'css', 'scss', 'sass' ],
 				},
 			],
 		],
 		overrides: [
-			{ // Transforms JSX into JS function calls and use `createElement` instead of the default `React.createElement`
+			{
+				// Transforms JSX into JS function calls and use `createElement` instead of the default `React.createElement`
 				plugins: [
 					[
 						'@babel/plugin-transform-react-jsx',
@@ -31,7 +28,8 @@ module.exports = function( api ) {
 				],
 				exclude: /node_modules\/react-native/,
 			},
-			{ // Auto-add `import { createElement } from '@wordpress/element';` when JSX is found
+			{
+				// Auto-add `import { createElement } from '@wordpress/element';` when JSX is found
 				plugins: [
 					[
 						'./gutenberg/packages/babel-plugin-import-jsx-pragma',
@@ -48,9 +46,7 @@ module.exports = function( api ) {
 		],
 		env: {
 			development: {
-				plugins: [
-					'@babel/transform-react-jsx-source',
-				],
+				plugins: [ '@babel/transform-react-jsx-source' ],
 			},
 		},
 	};
--- a/bin/ci-checks-js.sh
+++ b/bin/ci-checks-js.sh
@@ -16,7 +16,7 @@ fi
 if [ "$CHECK_CORRECTNESS" = true ] ; then
   yarn flow || pFail
   yarn lint || pFail
-  yarn prettier:check || { echo "ERROR: 'yarn prettier:check' found a problem"; pFail; }
+  yarn prettier:check || pFail
 fi
 
 if [ "$GUTENBERG_AS_PARENT" = true ] ; then
--- a/bin/ci-checks-js.sh
+++ b/bin/ci-checks-js.sh
@@ -16,6 +16,7 @@ fi
 if [ "$CHECK_CORRECTNESS" = true ] ; then
   yarn flow || pFail
   yarn lint || pFail
+  yarn prettier:check || { echo "ERROR: 'yarn prettier:check' found a problem"; pFail; }
 fi
 
 if [ "$GUTENBERG_AS_PARENT" = true ] ; then
--- a/jest.config.js
+++ b/jest.config.js
@@ -33,7 +33,10 @@ module.exports = {
 		'/__device-tests__/',
 	],
 	testURL: 'http://localhost/',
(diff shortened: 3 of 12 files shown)


mchowning:
Thanks @JavonDavis !

>This looks good! just a quick question, should this pre-check auto format? The message says something like success formatting 1 file with prettier-eslint so it made me think the file the should have been updated but it wasn't

Yes, that is confusing, I completely agree 😀! This sounds similar to what I was talking about with respect to the screenshot in the issue description. It seems like this is the behavior when using the `--list-different` flag, which is needed to make the command return a non-zero exit code if there are any formatting changes. If I add the `--write` flag to also change the files, then the exit code becomes `0`, so the build doesn't fail. 🤦‍♂️

Because of this issue, I did make the script on husky and circleci print an additional error message, so that it would be clear(er) that `yarn prettier:check` did not actually succeed:

```
"husky": {
      "hooks": {
          "pre-commit": "yarn prettier:check || { echo '\nERROR: `yarn prettier:check` found a problem\n'; exit 1; }"
      }
  }
```

I'm betting that you didn't see that because you ran `yarn prettier:check` directly (instead of having it run by husky or circle-ci). Since I only had the additional error messaging at the level of husky and circle-ci, you wouldn't have seen it running the command directly. I just pushed a commit (08b1d7c) fixing that so the additional error messaging is contained with the `prettier:check` script itself. Thanks for catching that!

FWIW, I also added a bit more detail to the error message to (hopefully) make it clearer. WDYT?

JavonDavis:
@mchowning thanks for clearing that up for me, yeah I saw the error message and it was definitely clear but was just checking if I misunderstood what was entirely suppose to happen because of that first message I mentioned. 

Now I'm wondering why we didn't go with the `--write` flag in the pre-commit hook since it would've done the auto-formatting and fix the errors, on CI we could leave as is to fail in case code code pushed without the pre commit hook being ran.
-  Is it that we want to not change someone's code automatically? - They'd have to do it themselves anyway to pass the CI check so why not?
- Is it that the `--write` can't auto fix everything, in that case I'd still expect to fix what it can and return a non-zero exit code right? 

I'm not sure but I think we could benefit more if we use the `--write` flag locally in the pre-commit hook, let me know if I'm making sense here or missing something, I haven't used prettier in a little while 😅

mchowning:
>I'm not sure but I think we could benefit more if we use the --write flag locally in the pre-commit hook

I chatted about this with some of the other gutenberg-mobile devs and I think the preference at this time is to not automatically make the changes. A few of the reasons were (1) consistency with the `gutenberg` project currently just rejects the commit instead of making the changes automatically; and (2) @koke raised a good point about there possibly being unexpected behavior when not all changes are staged for commit. Personally, I'm not a fan of pre-commit hooks making changes to non-generated files, but that is probably more of just a personal preference.

I prefer to not automatically make any changes in the hook at this time. We can, of course, always update this if we decide we want that in the future. Let me know what you think @JavonDavis , and thanks for bringing this up!

JavonDavis:
@mchowning Sounds good! I just caught up the conversation internally around it as well

> Personally, I'm not a fan of pre-commit hooks making changes to non-generated files, but that is probably more of just a personal preference.

I kinda feel the same I mainly raised the question here since we'll be failing on CI if the check doesn't pass, so just thinking why not, in any case don't have strong feelings about this vs the way you currently have it so I think we're good to go for the merge IMO! 👍

maxme:
🎉

PR #1262, opened 2019-08-06
Hide block if being replaced

### LINKED ISSUE #1177: [Empty blocks] Tiny refinement to "add new block" placement/flow
Related to my original proposal on #1165, which has been merged (thanks @marecar3 for the quick fix!).

While reviewing, I caught one thing that I should have mentioned in my original proposal – we should probably also **hide the empty block** when the inserter/library bottom sheet is open.

Screen Shot 2019-06-25 at 12 26 42 PM

mchowning:
Fixes #1177 by hiding a block that would be replaced when the add block bottom sheet is presented.

Further description and testing steps are available in the [related gutenberg PR](https://github.com/WordPress/gutenberg/pull/16931).

Update release notes:

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -2,6 +2,7 @@
 ------
 * Add rich text styling to video captions
 * Prevent keyboard dismissal when switching between caption and text block on Android
+* Blocks that would be replaced are now hidden when add block bottom sheet displays
 
 1.11.0
 ------


hypest:
> The only thing that I probably change here is to put some prefix in front of the branch name, e.g. fix/hide_block_if_being_replaced or issue/hide_block_if_being_replaced as it's a good practice here.

Good point. We technically follow the naming guidelines of the main WP apps anyway (see https://github.com/wordpress-mobile/WordPress-Android/wiki/Pull-Request-Guidelines) which point back to https://make.wordpress.org/mobile/handbook/general-guides/git-branching/#feature-branches. In this case, the typical branch name we'd use is: `issue/1177-hide-block-if-being-replaced`

Issue #1268, opened 2019-08-07
[Android] Content is missing on long posts in html mode


marecar3:
If there is some long post that should be presented in HTML mode e.g. https://gist.github.com/hypest/25e6d96b9c45ce92770f3c4c7ccafa9d

HTML editor won't show any content as a result of: `ReactEditText not displayed because it is too large to fit into a software layer (or drawing cache), needs 18616320 bytes, only 18432000 available`

marecar3:
There is an active PR: https://github.com/wordpress-mobile/gutenberg-mobile/pull/1277
which solves the issue.
But as it's explained in this comment: https://github.com/wordpress-mobile/gutenberg-mobile/pull/1277#issuecomment-521997921
we would need to go deeper and then to see if we can solve it in a better way.

daniloercoli:
I've investigated the issue as part of the current sprint and think i found something interesting:

- **The issue is only available on Android 8.X**
I was not able to reproduce it on devices running Android 7, and Android 9.

- As already mentioned in the  original proposed PR, even if we set the whole ReactView to "software layer", the warning `W/View: ReactEditText not displayed because it is too large to fit into a software layer (or drawing cache), needs 112216320 bytes, only 8294400`  is still available in the log.

- The ReactEditText component already has a workaround that sets its layer type to "software". [See here](https://github.com/facebook/react-native/blame/6991e28653e4ca44c811c9ad4564d064c90264b1/ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactEditText.java#L119).

With those above in mind I started to break our app by changing the native/web code, until I found that the culprit is the `ScrollView` that wraps the TextView used to display html code, added to give to TextView scrolling capabilities. [See here](https://github.com/WordPress/gutenberg/blob/master/packages/components/src/mobile/html-text-input/container.android.js#L14).

- If you replace `ScrollView` with the simpler `View` component, the HTML code is then correctly displayed, although no scrolling abilities. Very hard to scroll down.
- I tried with no luck to keep  `ScrollView` in place, and change the CSS styles applied to it.

**My conclusions:**
[...]

PR #1269, opened 2019-08-07
Hotfix/v1.10.1

marecar3:
Fixes https://github.com/wordpress-mobile/gutenberg-mobile/issues/1268

To test:

1. Open new Post
2. Toggle HTML editor mode
3. Copy content: https://gist.github.com/hypest/25e6d96b9c45ce92770f3c4c7ccafa9d
4. Paste into HTML editor

Result: Content should be visible

Update release notes:

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -1,6 +1,6 @@
 1.10.1
 ------
-* Fix missing content on long posts in html mode
+* Fix missing content on long posts in html mode on Android
 
 1.10.0
 ------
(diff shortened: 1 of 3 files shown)


hypest:
Hmm, I think we shouldn't rush this into a hotfix, especially this late in the v13.0 Beta lifecycle (we are now in the middle of the second week as Beta). Let's instead target `develop` and we can target the 13.1 of the apps.

marecar3:
Thanks for the review @hypest!
Will change base branch.

PR #1277, opened 2019-08-08
Fix/1268 content is missing on long posts in html mode

marecar3:
Fixes https://github.com/wordpress-mobile/gutenberg-mobile/issues/1268

To test:

1. Open WPAndroid app
2. Create new Post
3. Toggle HTML editor mode
4. Copy content: https://gist.github.com/hypest/25e6d96b9c45ce92770f3c4c7ccafa9d
5. Paste into HTML editor

Result: Content should be visible

Update release notes:

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -1,9 +1,7 @@
-1.12.0
-------
-* Fix missing content on long posts in html mode on Android
 1.11.0
 ------
 * Toolbar scroll position now resets when its content changes
+* Fix missing content on long posts in html mode on Android
 
 1.10.0
 ------
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -1,4 +1,4 @@
-1.11.0
+1.12.0
 ------
 * Fix missing content on long posts in html mode on Android
 
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -1,3 +1,7 @@
+1.11.0
+------
+* Fix missing content on long posts in html mode on Android
+
 1.10.0
 ------
 * Adding a block from the post title now shows the add block here indicator.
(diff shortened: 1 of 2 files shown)


daniloercoli:
I wasn't able to reproduce the original issue on a Nexus 5X running Android 7.1. The long content is correctly shown on screen after having pasted it into the HTML editor.

Worth to have another set of eyes testing this PR. cc @hypest

marecar3:
Thanks for the update @daniloercoli!
Any chance that you have time to try this fix @hypest? Thanks.

hypest:
I can see the original issue on my Galaxy S7 (Android 8.0.0) when running WPAndroid develop (https://github.com/wordpress-mobile/WordPress-Android/commit/6a6c100217cdcec2ef396d4dec9866fba73f56c6) but not seeing the issue when using this PR. The issue is that the html view of a large-ish post is empty.

Here's the thing though:
1. I can still see `W/View: ReactEditText not displayed because it is too large to fit into a software layer (or drawing cache), needs 112216320 bytes, only 8294400 available` lines even with this PR
2. The ReactEditText component already has a workaround that sets its layer type to "software". [See here](https://github.com/facebook/react-native/blame/6991e28653e4ca44c811c9ad4564d064c90264b1/ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactEditText.java#L119). Which means, the fix in this PR might actually be fixing something else and not the issue mentioned in the original ticket ("HTML editor won't show any content as a result of: ReactEditText not displayed because it is too large to fit into a software layer (or drawing cache), needs 18616320 bytes, only 18432000 available").
3. I'm not feeling comfortable withe broadness of the solution in this PR, were we set the layer type to "software" for the whole ReactNative view and for every device and every Android version. If I understand it correctly, this would mean disabling hardware acceleration for the React Native view, a rather drastic measure just to show the html view correctly.

I think we should dig deeper for a solution on this one, possibly finding the specific root cause of the issue, or managing to apply the "software" layer type to the html mode only, if that ReactEditText is still the root cause.

marecar3:
Hey @hypest, thanks for the details, it helps a lot at this stage.

Generally, I agree with your thoughts. 
Solution for this issue demands a better understanding of the problem which is causing it, so investing more time would be a good path for us.

daniloercoli:
We can close this PR since it was replaced by https://github.com/wordpress-mobile/gutenberg-mobile/pull/1401

PR #1292, opened 2019-08-14
Update Aztec version on Android

daniloercoli:
This PR updates the Aztec version used in the Aztec wrapper for Android.
Changes made to Aztec are minimal, and added a couple of methods to keep track of toolbar button pressed, and undo/redo events.  https://github.com/wordpress-mobile/AztecEditor-Android/pull/843

**Note**: Undo/Redo buttons live outside Aztec at the moment of this PR, so we could track tapping on those directly in the host apps, but we've agree to consider those buttons as they were part of the aztec-editor toolbar. We wanted to be "location" agnostic and be ready to move the undo/redo into the toolbar quickly if necessary.

To test:
Use this PR here: https://github.com/wordpress-mobile/WordPress-Android/pull/10389

Update release notes:

- [ x ] If there are user facing changes, I have added an item to `RELEASE-NOTES.txt`.

### DIFF
--- a/react-native-aztec/android/build.gradle
+++ b/react-native-aztec/android/build.gradle
@@ -12,7 +12,7 @@ buildscript {
         wordpressUtilsVersion = '1.22'
         espressoVersion = '3.0.1'
 
-        aztecVersion = 'v1.3.29'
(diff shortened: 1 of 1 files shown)


daniloercoli:
Updated to the tagged version of Aztec

PR #1295, opened 2019-08-16
Remove duplicate code when removing blocks after each test

RachelleT:
Fixes # Instead of manually removing blocks after each test, code is placed in the AfterEach block to automatically remove blocks after each test.

To test: The blocks are removed by the code in the afterEach block.

Update release notes:


hypest:
👋 @RachelleT and sorry for the delayed response here.

I'm a bit confused about the changes in this PR. Can you expand the PR description with individual explanation of each change? Thanks in advance!

Also, I see that the PR includes a new application (`MyApplication`). Was that included in the PR by intention? What does the new application offer? If it's needed functionality, perhaps we can chat about adding it to the existing demo app in this repo.

__device-tests__/helpers/caps.js
```diff
@@ -1,7 +1,7 @@
 exports.ios12 = {
 	browserName: '',
 	platformName: 'iOS',
-	platformVersion: '12.2',
+	platformVersion: '12.4',
```
JavonDavis: Any specific reason for this change? This updates the version used on CI and I'm not sure if that's available as yet

__device-tests__/pages/editor-page.js
```diff
@@ -46,7 +46,7 @@ export default class EditorPage {
 	// Finds the wd element for new block that was added and sets the element attribute
 	// and accessibilityId attributes on this object and selects the block
 	// position uses one based numbering
-	async getBlockAtPosition( position: number, blockName: string ) {
+	async getBlockAtPosition( position: number, blockName: string = '' ) {
```
JavonDavis: Hmm I'm not seeing where you used this default value, is there a reason you included it here?

__device-tests__/gutenberg-editor-block-insertion.test.js
```diff
@@ -38,6 +39,28 @@ describe( 'Gutenberg Editor tests for Block insertion', () => {
 		editorPage = new EditorPage( driver );
 	} );
 
+	afterEach( async () => {
+		let blockExist = await editorPage.hasBlockAtPosition( 1 );
```
JavonDavis: This block of code is duplicated across the tests that you used them in. Might be best if we put it into a function and call it, maybe `utils.js` could be a could spot, let me know what you think? Also is there a reason it's not used in the other tests?

JavonDavis:
@RachelleT also seems like there's a few merge conflicts with develop, could you resolve those? Thanks!

__device-tests__/gutenberg-editor-block-insertion.test.js
RachelleT: I get what you're saying and it makes sense. Do you think its better if the code goes in utils or editor-page? The code uses functions from editor-page, if it goes in utils I would have to import from editor page.

RachelleT:
Closing in favor of #1585

PR #1305, opened 2019-08-20
Fix HTML mode error on iOS

### LINKED ISSUE #1303: Crash on HTML mode
There's a crash when switching to HTML mode.

On the example app, it's triggered when trying to edit the HTML content (the iOS example app has the UI to switch to HTML mode)

On WPiOS, it's triggered just by switching to HTML mode.

This is the error:

IMG_2962

mchowning:
Fixes #1303 . Additional details and testing instructions are in the description of the [related gutenberg PR](https://github.com/WordPress/gutenberg/pull/17115).

Update release notes:

I only observed this red screen error was only visible on dev builds, so there are no user facing changes. Therefore, the release notes have _not_ been updated.


mchowning:
:wave: @etoledom . You opened #1303 against the 1.11 milestone. When I checked the most recent internal build for iOS I didn't observe any crash or other problems. If I'm not missing something, what do you think about targeting this fix for 1.12 since the release branches have already been cut. Admittedly, this seems like a pretty safe fix, but if there's no user-facing reason we need to push this fix in right away, seems like we're better off just

etoledom:
So, first it was my mistake since Tickets should not really have milestones 😁

As far as I understand, our current frozen branch is 1.11, right? This one will go together with WP apps v13.2. 
Being that the case, if we don't target our current frozen branch, we will be shipping this crash on the next release cycle.

Let me know if I messed up the numbers :)

mchowning:
>As far as I understand, our current frozen branch is 1.11, right? This one will go together with WP apps v13.2.

That all sounds right to me.

>if we don't target our current frozen branch, we will be shipping this crash on the next release cycle.

My point was that I don't think this is a crash. I believe it's just a red-screen error, which is not shown on release builds. For example, if I build the release branch _and do not have the bundler running_, I'm not seeing any issues. I could definitely be missing something though, so let me know.

mchowning:
After more investigation, I'm _pretty sure_ this isn't a crash on release builds, but I'm not 100% sure, so I'm swayed to your view @etoledom . 😀 Let's include this in v1.11.

I've updated the target branches appropriately.

PR #1309, opened 2019-08-22
iOS DarkMode 

etoledom:
This PR adds the necessary changes to implement DarkMode on iOS.

Gutenberg PR: https://github.com/WordPress/gutenberg/pull/17067

Prerequisites to test:
- Xcode 11 beta is needed. 
- Run Xcode 11
- open the iOS project via `open ios/gutenberg.xcodeproj`
- Run `yarn clean:install` to install the new dependency 
- Build and run on a simulator
- Go to the iOS Settings app on the simulator
  - -> Developer -> Dark Appearance -> Switch ON

To test:
- Go through the app and check that Dark Mode is implemented correctly.

**Note:**
- The toolbar buttons are not yet updated. I believe that the IconButon component will need a deeper refactor to be able to select different colors on all its states.
- For some reason, some components won't change of color mode dynamically (bottom-sheet, some elements in the tool bar). I'll investigate this and try to update. But I think this fix can be added as a future PR.
- The horizontal SafeArea is gray, and that's not great. My first try to fix this broke the tool bar avoiding view behavior. Same than the previous point, I'll investigate this but can be left for a future PR.


etoledom:
Thank you! 🙏

Issue #1298, opened 2019-08-19
Cross Platform MediaUpload component


pinarol:
We already have MediaUpload for mobile and web but they have different props. We need to refactor the mobile implementation(gutenberg/packages/block-editor/src/components/media-upload/index.native.js) to work with same props with web.

Note that some props might be web specific and not needed by mobile.

We'll also need to update the unittests: gutenberg/packages/block-editor/src/components/media-upload/test/index.native.js

MediaUpload component is currently used by image, video blocks

Here are the mobile side edit implementations for them:

gutenberg/packages/block-library/src/image/edit.native.js
gutenberg/packages/block-library/src/video/edit.native.js

MediaUpload component opens the media options picker on mobile. And then it calls the media picker from the main apps:

<img width="334" alt="Screen Shot 2019-08-19 at 15 55 03" src="https://user-images.githubusercontent.com/5032900/63266955-bab94680-c299-11e9-8baa-bb49d6b86dd6.png">

![media-upload](https://user-images.githubusercontent.com/5032900/63266489-b0e31380-c298-11e9-96cc-1cd2f465efa8.gif)

To investigate/debug the web part:

cd to gutenberg

npm install
npm run dev
At another terminal:
docker-compose up

[This page](https://github.com/WordPress/gutenberg/blob/master/docs/contributors/getting-started.md) has more detailed instructions in case you might need.

Keep in mind that mobile unit-tests(`yarn test`) might fail if gutenberg-mobile/gutenberg/node_modules has modules in it. So the ideal way is checking out another gutenberg repo and running the code from there.

**To test**

**Testing prerequisites**

Media upload isn't supported by free sites so if you only have a free site let us know your WordPress.com user names and we can arrange a test site for you.

For WPiOS

Checkout the PRs branch to any arbitrary folder and cd .. to it
run yarn install, yarn start
Open XCode WPiOS on the latest develop
Clean build folder on Xcode, and then run the app

For WPAndroid

open grade.properties at WordPress-Android folder
add wp.BUILD_GUTENBERG_FROM_SOURCE = true to grade.properties
checkout the PRs branch in the subrepo of WordPress-Android repo
cd to WordPress-Android/libs/gutenberg-mobile
run yarn install, yarn start
yarn wpandroid on a separate terminal in the same directory

---

Make sure nothing is broken by running these test cases on WPiOS and WPAndroid main apps:
[...]

PR #1331, opened 2019-08-27
Feat: support of multiple selection in media upload

dratwas:
In this PR I added support for multiple selections of media in MediaUpload component.  It is a breaking change and apps will crash without changes from listed PRs.

Associated issue - https://github.com/wordpress-mobile/gutenberg-mobile/issues/1298

Related gutenberg PR - https://github.com/WordPress/gutenberg/pull/17397
Related Wordpres-iOS PR - https://github.com/wordpress-mobile/WordPress-iOS/pull/12404
Related Wordpress-Android PR - https://github.com/wordpress-mobile/WordPress-Android/pull/10556
To test:
- set multiple prop to true in image/video block
- run Wordpress-iOS or Wordpress-Android
- add image/video block - multiple select should be enabled for selecting media from the device or WP Media
- the array of selected media should be passed to the callback function

NOTE: Multiple selections of media from the device on iOS doesn't work since this functionality is implemented by the client app. Our example app from `ios` directory doesn't support multiple selections from the device because it uses build-in `UIImagePickerController`. In order to make it works in our example app, I would need to add an external library that supports multiple selections.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,6 @@
 [submodule "gutenberg"]
 	path = gutenberg
-	url = https://github.com/callstack/gutenberg.git
+	url = ../../WordPress/gutenberg.git
 [submodule "react-native-aztec"]
 	path = react-native-aztec-old-submodule
 	url = ../react-native-aztec.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,6 @@
 [submodule "gutenberg"]
 	path = gutenberg
-	url = ../../WordPress/gutenberg.git
+	url = https://github.com/callstack/gutenberg.git
 [submodule "react-native-aztec"]
 	path = react-native-aztec-old-submodule
 	url = ../react-native-aztec.git
--- a/ios/gutenberg/GutenbergViewController.swift
+++ b/ios/gutenberg/GutenbergViewController.swift
@@ -45,6 +45,9 @@ class GutenbergViewController: UIViewController {
 }
 
 extension GutenbergViewController: GutenbergBridgeDelegate {
+    func editorDidAutosave() {
+        print("➡️ Editor Did Autosave")
+    }
 
     func gutenbergDidLoad() {
         gutenberg.setFocusOnTitle()
--- a/ios/gutenberg/GutenbergViewController.swift
+++ b/ios/gutenberg/GutenbergViewController.swift
@@ -70,9 +70,17 @@ extension GutenbergViewController: GutenbergBridgeDelegate {
             print("Gutenberg did request media picker, passing a sample url in callback")
             switch currentFilter {
             case .image:
-                callback([(id: 1, url: "https://cldup.com/cXyG__fTLN.jpg")])
+                if(allowMultipleSelection) {
+                    callback([(1, "https://cldup.com/cXyG__fTLN.jpg"), (3, "https://cldup.com/cXyG__fTLN.jpg")])
+                } else {
+                    callback([(1, "https://cldup.com/cXyG__fTLN.jpg")])
+                }
             case .video:
-                callback([(2, "https://i.cloudup.com/YtZFJbuQCE.mov")])
+                if(allowMultipleSelection) {
+                    callback([(2, "https://i.cloudup.com/YtZFJbuQCE.mov"), (4, "https://i.cloudup.com/YtZFJbuQCE.mov")])
+                } else {
+                    callback([(2, "https://i.cloudup.com/YtZFJbuQCE.mov")])
+                }
             default:
                 break
             }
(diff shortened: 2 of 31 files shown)


react-native-gutenberg-bridge/ios/GutenbergBridgeDelegate.swift
```diff
@@ -63,7 +63,7 @@ public protocol GutenbergBridgeDelegate: class {
     ///     - source: the source from where the picker will get the media
     ///     - callback: A callback block to be called with an upload mediaIdentifier and a placeholder image file url, use nil on both parameters to signal that the action was canceled.
```
pinarol: I know this is still draft but let's not forget to update `callback` definition also.

react-native-gutenberg-bridge/ios/RNReactNativeGutenbergBridge.swift
```diff
@@ -22,12 +22,23 @@ public class RNReactNativeGutenbergBridge: RCTEventEmitter {
...
+                    let formattedMedia = media.map { (id, url) in
+                        return ["id": id, "url": url]
+                    }
+                    callback([formattedMedia])
+                } else {
+                    guard let (mediaID, mediaURL) = media.first else {
+                        callback(nil)
+                        return
+                    }
+                    callback([["id": mediaID, "url": mediaURL]])
```
pinarol: we can define constants for "id" and "url"

react-native-gutenberg-bridge/ios/RNReactNativeGutenbergBridge.swift
```diff
@@ -39,12 +50,12 @@ public class RNReactNativeGutenbergBridge: RCTEventEmitter {
             return
         }
         DispatchQueue.main.async {
-            self.delegate?.gutenbergDidRequestImport(from: url, with: { (mediaID, url) in
-                guard let url = url, let mediaID = mediaID else {
+            self.delegate?.gutenbergDidRequestImport(from: url, with: { media in
```
pinarol: I think `media` is an array now? so how about renaming to sth like `mediaList`?

pinarol:
This is looking pretty good overall, I just wanted to point out some things early on.

So what are some pending work here?

dratwas:
Prop `multiple` works on iOS with these changes so basically I need to check how the `cancelation` mechanism works and make it works with multiple as well and add tests.

pinarol:
thanks @dratwas ! did you have chance to try multiple uploads also?

pinarol:
> and add tests.

really appreciate that!

dratwas:
> thanks @dratwas ! did you have chance to try multiple uploads also?

AFAIK this is a responsibility of the `client` app. I implemented it in `Wordpress-iOS` here https://github.com/wordpress-mobile/WordPress-iOS/pull/12404/files#diff-0992f0530230d5beb61d0569d50f4459R40
I tried it and native side uploads media and sends all "mediaUpload"  events related to it but it is not handled on JS side because at this moment JS implementation of `image/edit.native.js` assumes that we can get only one image. So it is prepared to support multiple media upload but needs to be handled in `gallery` block.

pinarol:
> I tried it and native side uploads media and sends all "mediaUpload" events related to it 

That's actually enough for this particular task thanks! Sorry If I confused you, the rest will be handled in `gallery` block. I just wanted to make sure we are sending the "mediaUpload" events generated by simultaneous uploads. 👍

pinarol:
Could you be able to update the branch from `develop` ?

pinarol:
Could you also add the associated issue to PR description?

ios/gutenberg/GutenbergViewController.swift
```diff
@@ -70,9 +70,9 @@ extension GutenbergViewController: GutenbergBridgeDelegate {
             print("Gutenberg did request media picker, passing a sample url in callback")
             switch currentFilter {
             case .image:
-                callback(1, "https://cldup.com/cXyG__fTLN.jpg")
+                callback([(id: 1, url: "https://cldup.com/cXyG__fTLN.jpg")])
             case .video:
-                callback(2, "https://i.cloudup.com/YtZFJbuQCE.mov")
+                callback([(2, "https://i.cloudup.com/YtZFJbuQCE.mov")])
```
pinarol: Looking inconsistent compared to line 73 since it has `id:`, `url:` labels for tuple items

pinarol:
Could you be able to update the branch resolve merge conflicts one more time? Sorry for being late for reviewing, I'll start after this. Thanks!

mkevins:
I tested this on Android, and it is working nicely. :tada:

.gitmodules
```diff
@@ -1,6 +1,6 @@
 [submodule "gutenberg"]
 	path = gutenberg
-	url = ../../WordPress/gutenberg.git
+	url = https://github.com/callstack/gutenberg.git
```
SergioEstevao: Should this stay with this value after the merge?

dratwas: No, it should be reverted before the merge

PR #1357, opened 2019-09-06
iOS DarkMode toolbar buttons

etoledom:
`gutenberg` PR: https://github.com/WordPress/gutenberg/pull/17356
## Description

This PR allow us to add different styles to DashIcons inside IconButton component depending on the the style theme (Dark/Light mode).

It also enhance the DarkMode theme for Toolbar buttons (and the toolbar separator)

With this change, is not needed to send `ariaPressed` prop from `IconButton`. This is a change that [was requested](https://github.com/WordPress/gutenberg/pull/13977#issuecomment-468731745) long ago, (even though we still are sending it button states).

A second idea is to send styles directly to DashIcon from the Button, to avoid having to declare styles for every new class. I'm not sure if this is something we want to change though.

As always any moment is welcome! 

toolbar

To test:
- Run the example app with Xcode 11 beta
- Check that the toolbar colors are correct on the toolbar (following the screenshots)
- Check that the rest of the icons in the app shows properly
  - In particular the BottomSheet icons (Add link / Image settings)
- Check that the icon Subscript is colored properly (Numbers on H1, H2, etc...)

cc @iamthomasbishop

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -1,3 +1,13 @@
+1.15.0
+------
+* Fix issue when multiple media selection adds only one image or video block on Android
+* Fix issue when force Touch app shortcut doesn't work properly selecting "New Photo Post" on iOS
+
+1.14.0
+------
+* Fix a bug on iOS 13.0 were tapping on a link opens Safari
+* Fix a link editing issue, where trying to add a empty link at the start of another link would remove the existing link.
+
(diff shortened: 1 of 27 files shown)


iamthomasbishop:
Beyond stoked to see this! Well done @etoledom!

etoledom:
Thank you!

PR #1360, opened 2019-09-09
Intercept paste without formatting command.

### LINKED ISSUE #1311: [iOS] Placeholder text remains after "Paste without Formatting"
Pasting text on iOS using `Paste without Formatting` doesn't remove the block placeholders. (A regular `Paste` works fine, and on Android `Paste as plain text` also works fine.)

Steps to reproduce:

1. Copy text (from within the editor or from another app).
2. In the title field or any block, use `Paste without Formatting` to paste the text.

Result: The placeholder overlaps the pasted text.

Tested on iPhone XS and iPad Mini, iOS 12.4, WPiOS 13.0.1 and 13.1.0.20190819.

SergioEstevao:
Fixes #1311

This PR changes the code in the RNAztecView in order to override the `pasteWithoutFormatting` the same way we do with `paste` and do the paste action in GB instead of relying in the Aztec paste.

To test: Follow the instructions on the ticket #1311.

Update release notes:

### DIFF
--- a/react-native-aztec/ios/RNTAztecView/RCTAztecView.swift
+++ b/react-native-aztec/ios/RNTAztecView/RCTAztecView.swift
@@ -245,14 +245,23 @@ class RCTAztecView: Aztec.TextView {
     }
     
     override func paste(_ sender: Any?) {
-        let start = selectedRange.location
-        let end = selectedRange.location + selectedRange.length
-        
         let pasteboard = UIPasteboard.general
         let text = readText(from: pasteboard) ?? ""
         let html = readHTML(from: pasteboard) ?? ""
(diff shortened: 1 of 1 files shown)


SergioEstevao:
@etoledom good find, I dig up a bit the code and found out that in your example the pasteHandler tries to see if  the plainText can be markDown, and then translates it to HTML and then for some reason we get that red screen further down the line.

On you specific example the '-' are translated to a nested list elements and then make the system break.

this was not happening before, because the paste without format was handled by Aztec, and no tentative of converting from markdown was done.

SergioEstevao:
@etoledom I fixed the issue you found on this PR for GB: https://github.com/WordPress/gutenberg/pull/17550

PR #1363, opened 2019-09-11
Generate installable builds for demo app on Android

JavonDavis:
Adds installable builds for demo app on Android

To test: 
1. Peril should generate a url to the apk for the demo app
2. Download the apk and make sure it's ok 

Update release notes:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -78,6 +78,11 @@ jobs:
           name: Bundle Android and Generate debug .apk file for testing
           command: yarn test:e2e:build-app:android
       - run:
+          name: Attach artifacts
+          command: |
+            mkdir -p Artifacts
+            mv ./android/app/build/outputs/apk/debug/app-debug.apk "Artifacts/Gutenberg-$SAUCE_FILENAME.apk"
+      - run:
           name: Upload apk to sauce labs
           command: |
             source bin/sauce-pre-upload.sh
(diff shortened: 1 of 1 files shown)


hypest:
I wonder if we should add a comment somewhere (the Peril message template maybe?) about what to expect from the APK. In this case, the demo app will try to connect to a Metro instance and fail if the reverse proxy is not already set up. WDYT @JavonDavis ?

JavonDavis:
> I wonder if we should add a comment somewhere (the Peril message template maybe?) about what to expect from the APK.

🤔 I'm thinking it's assumed it's a normal APK that can be installed and used. 

> In this case, the demo app will try to connect to a Metro instance and fail if the reverse proxy is not already set up.

This isn't the case, it's a release build so no need for a metro instance running.

jtreanor:
The text in the comment is defined [here](https://github.com/Automattic/peril-settings/blob/master/org/pr/android-installable-build.ts#L77). It would be possible to define this on a per-project basis with a bit of work.

JavonDavis:
> It would be possible to define this on a per-project basis with a bit of work.

Sure I'm still open to doing this but I'm thinking the use case would be more to clarify that it's the demo app for testing blocks.

SergioEstevao:
Now that we have the APK available on the main apps, we no longer need this for the demo apps.

Issue #1332, opened 2019-08-27
Crash on the latest master on gutenberg repo


pinarol:
Following crash occurs when we update gutenberg hash to point to master.

<img width="344" alt="Screen Shot 2019-08-23 at 12 02 32" src="https://user-images.githubusercontent.com/5032900/63786950-ab6e7480-c8fb-11e9-9f61-673b53cd8074.png">

It looks like [this PR](https://github.com/WordPress/gutenberg/pull/16932) caused this regression.

etoledom:
We have this PR with a partial fix: https://github.com/WordPress/gutenberg/pull/17228

It solves the error referenced on this ticket, but it exposes an underlying problem given the nature of the changes from https://github.com/WordPress/gutenberg/pull/16932

WIP.

PR #1366, opened 2019-09-13
Fix store crash with gutenberg master

Tug:
Fixes https://github.com/wordpress-mobile/gutenberg-mobile/issues/1332

This PR brings us back to targeting gutenberg `master`

Gutenberg PR: https://github.com/WordPress/gutenberg/pull/17228

To test: 
- Check that tests are all green
- Try to boot the demo app and perform regular actions
- Try to save and reload a post in particular

Update release notes:


mchowning:
This is looking good! Did a quick smoke test to see how this was progressing and everything seems to be working.

This branch does show some problems that have been fixed on `rnmobile/master` since this branch doesn't include those fixes (i.e., separator block crash, ios crash when viewing html). This obviously isn't a problem since `rnmobile/master` has these fixes, but I wanted to note this in case anyone else ran across them while testing.

src/index.test.js
```diff
@@ -22,17 +22,20 @@ describe( 'RootComponent', () => {
 
 	it( 'renders without crashing', () => {
 		const app = renderer.create( <RootComponent /> );
+		renderer.act( () => {
+			jest.runAllTicks();
```
etoledom: Interesting!
Can we add a comment saying why/when is this needed? 🤔

Tug: Added one [here](https://github.com/WordPress/gutenberg/pull/17228/files#diff-198fbd355f0b071c0a3bbb8a014e2ec9R36) for the same reason. It seems that when `EDIT_ENTITY_RECORD` is dispatched using [this generator function](https://github.com/WordPress/gutenberg/blob/master/packages/core-data/src/actions.js#L169) it's done so asynchronously, meaning the promise does not resolve immediately even during tests. This is related to the changes made on core data entities so I figured it was safe to update our tests for now. I'll try to find a good explanation for this and add something 👍

Issue #1161, opened 2019-06-19
Links: “Open in a new tab” setting doesn’t look like it saves.


thehenrybyrd:
**Expected:**
When I set the “Open in a new tab” setting for a link, I expect it to remain toggled on.

**Actual:**
It does not remain toggled on in all circumstances, even though the target=“_blank” remains in tact.

This happens in Paragraph

**To Reproduce:**
1. Create a new post
2. Add a paragraph block, type some text.
3. Add a link and set it to “Open in a new tab”.
4. Open link settings again and see that “Open in a new tab” is toggled off (don’t touch the toggle, just note). 
5. Tap into another block.
6. Tap back into the block with the link, and carefully place your cursor mid-word within the link.
7. Open link settings again and see that “Open in a new tab” is toggled on.
8. Tap to close link settings. Note that the word is highlighted, but the “link” icon is not.
9. If you tap link settings again, you’ll see that once again “Open in a new tab” is toggled off.

This makes it very difficult to know for sure whether I have in fact set the link to “Open in a new tab”. 

**Video:**

![linksettings](https://user-images.githubusercontent.com/15107387/59776236-1cb9f780-92bb-11e9-926b-121c49ac24d7.gif)

**However, the target=“_blank” remains in tact despite it not appearing consistently.**
This has to be tested in a web browser, since currently preview posts in iOS opens them in the in-app browser. When we tap the link, the device browser opens, so there’s not a simple way to tell whether it was set to open in a new tab or not without browser tools.
In a browser, we can easily Inspect Element, or simply test the link when previewing.

**To Reproduce:**
1. Open WordPress.com in a web browser and navigate to Blog Posts.
2. Open the Post you’ve been working on (doesn’t matter if it’s Draft or Published).
3. Check the link. Is “Open in a new tab” toggled on? Check again. How about now?
4. Preview or View the post. 
5. Check the link, either with Inspect Element or by simply tapping it, and note it opens in a new tab.

<img width="594" alt="Screen Shot 2019-06-19 at 5 12 39 PM" src="https://user-images.githubusercontent.com/15107387/59776285-39562f80-92bb-11e9-9d76-e52bcab8187f.png">

<img width="876" alt="Screen Shot 2019-06-19 at 5 54 46 PM" src="https://user-images.githubusercontent.com/15107387/59776374-5e4aa280-92bb-11e9-92ec-5bf533e77882.png">

[...]

PR #1374, opened 2019-09-20
Only report selection changes if text view is still first responder.

### LINKED ISSUE #1368: [iOS] Can't add link to selected text
### Expected behavior

I expect that if I select existing text in the block editor, I can add a link to that text.

### Actual behavior

If I select text in the block editor and try to add a link to it, the link is added after the text I had selected. (Tested in WPiOS and WPAndroid, and I only observe the issue in WPiOS.)

1. Add a paragraph block.
2. Enter text in the block.
3. Select some of the entered text.
4. Tap the link button.
5. Note that there is no Link Text prefilled. Enter a URL (and optionally, link text).
6. Close the editor and observe that the URL or linked text is added after the text you selected in step 3.

Tested on WPiOS 13.2.1.2 TestFlight beta on iPhone XS, iOS 13.1 public beta.

SergioEstevao:
Fixes #1368 

This PR makes sure that no selection changes are propagated after the aztec text view is no longer the first responder. 
This way when the Link Modal interface is called and the URL input is activated (becomes the first responder) the selection change on the original Aztec content is not propagated to the link UI.

To test:
 - Open the demo app
 - Add or edit a text block, for example paragraph
 - Write some text
 - Select some of the text
 - Tap on the link button on the toolbar
 - Make sure the link text is pre-filled with the selected text

Update release notes:

### DIFF
--- a/react-native-aztec/ios/RNTAztecView/RCTAztecView.swift
+++ b/react-native-aztec/ios/RNTAztecView/RCTAztecView.swift
@@ -571,6 +571,9 @@ class RCTAztecView: Aztec.TextView {
 extension RCTAztecView: UITextViewDelegate {
 
     func textViewDidChangeSelection(_ textView: UITextView) {
+        guard isFirstResponder else {
+            return
+        }
         guard isInsertingDictationResult == false else {
             return
         }
--- a/react-native-aztec/ios/RNTAztecView/RCTAztecView.swift
+++ b/react-native-aztec/ios/RNTAztecView/RCTAztecView.swift
@@ -570,11 +570,8 @@ class RCTAztecView: Aztec.TextView {
 // MARK: UITextView Delegate Methods
 extension RCTAztecView: UITextViewDelegate {
 
-    func textViewDidChangeSelection(_ textView: UITextView) {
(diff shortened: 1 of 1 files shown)


react-native-aztec/ios/RNTAztecView/RCTAztecView.swift
```diff
@@ -571,6 +571,9 @@ class RCTAztecView: Aztec.TextView {
 extension RCTAztecView: UITextViewDelegate {
 
     func textViewDidChangeSelection(_ textView: UITextView) {
+        guard isFirstResponder else {
```
etoledom: Maybe we could add this statement to the previously existing `guard`?
Not sure if we get an advantage by having them separated.

rachelmcr:
@etoledom That weirdness you mentioned looks like the issue reported in https://github.com/wordpress-mobile/gutenberg-mobile/issues/1161.

etoledom:
> @etoledom That weirdness you mentioned looks like the issue reported in #1161.

@rachelmcr you are right! Looks like it's the same issue. Since it's already a separated ticket, let's 🚢 this one ✅

PR #1375, opened 2019-09-20
Don't allow interactions with URL in Aztec text blocks.

### LINKED ISSUE #1373: Tapping on links on GB editor makes it navigate away from the editor
__Expected behaviour__

Tapping on text with a link assigned shouldn't open Safari and navigate to that link

__Actual behaviour__

On tapping on text with a link Safari get's open to the URL associated to the link

This only happens on iOS versions 13 and above.

SergioEstevao:
Fixes #1373 

This PR adds a new delegate method to avoid that interaction with URL attributes happens.
From the documentation it looks this delegate existed since iOS 10, but only with iOS 13 it became a problem. Maybe by default now if no delegate method is implemented the system assumes a `true` answer?

To test:
 - Start the demo app
 - Tap on any link that exists on the content
 - Make sure that the app is not switched to Safari and tries to open the link there.

Update release notes:


SergioEstevao:
@etoledom I'm able to force touch and drag the cursor on the links, are you able to do that with your solution above?

koke:
I think I'm missing something here, but wouldn't it be enough to implement that delegate method to always `return false`?

SergioEstevao:
@koke that was my first solution but the problem then is that the link text will be completely non interactive, if you tap on it, it will not move the cursor to it...

This is a bad bug from iOS 13...

SergioEstevao:
@etoledom I fixed the bug you detected for link removal, can you give it another spin?

koke:
Then I must be missing something because I tried without the selection workaround and I'm seeing the same interactions: [video](https://cloudup.com/cnZ5PzMUI_f)

etoledom:
@koke could you check if the delegate method is being called?

And what iOS version are you testing on? Maybe this bug has been already fixed on the latest iOS 13.1 release. 🤔

SergioEstevao:
What version of IOS you are using? We were seeing it on 13 and 13.1 . Maybe Apple fixed it on 13.1 before they release it.

koke:
I was testing with Xcode 11.1 GM and iOS 13.1

etoledom:
Just updated my testing device to the latest 13.1 ("Seed 4" from the beta program) and now it is behaving as @koke described.

The delegate override is still needed but just returning false is enough to make it behave as it did before. 👍 

There's also an iOS v13.1.1 🤔

koke:
Also related, I came across this PR because I was investigating this other issue and it caught my attention https://github.com/wordpress-mobile/WordPress-iOS/issues/12568

etoledom:
> @etoledom I fixed the bug you detected for link removal, can you give it another spin?

Working good now! 🎉 

Let's update the PR to make it work properly with the released iOS 13.1 version 👍

koke:
👍 Checked the numbers and it seems 13.0 is currently at 6% and trending down, so by the time this gets released, it could be pretty low.

Screen Shot 2019-09-30 at 11 10 29

marecar3:
Hey @SergioEstevao,
Please, can we mention this fix in RELEASE-NOTES?

etoledom:
We should probably merge https://github.com/wordpress-mobile/gutenberg-mobile/pull/1393 first?

Issue #1313, opened 2019-08-22
Media & Text - Media picker buttons are not functional


pinarol:
We need to make Media picker buttons functional.

<img width="331" alt="Screen Shot 2019-08-22 at 16 13 50" src="https://user-images.githubusercontent.com/5032900/63532529-8cdf3680-c513-11e9-9e9d-4af026647155.png">

Support for these items:

1st iteration:
- [ ] Adding image from WordPress media library

2nd iteration:
- [ ] Uploading image from device library
- [ ] Uploading image by capturing

3rd iteration:
- [ ] Adding video from WordPress media library
- [ ] Uploading video from device library
- [ ] Uploading video by capturing

Please refer to already existing image/video blocks for similar functionality.

pinarol:
This work is dependent on this PR which is very close to merge https://github.com/WordPress/gutenberg/pull/16305
But feel free to also checkout the PR branch to start working

geriux:
Hey @pinarol !

So I've finished reviewing the task, I saw the image functionality is already implemented but not fully, here's a summary of the main issue:

### 1st interaction
	
#### Adding image from WordPress media library

Base implementation is there but some are missing:
* Image upload progress / with place holder
* Edit button in toolbar

### 2nd interaction

#### Uploading image from device library

Base implementation is there but some are missing
* Image upload progress / with place holder
* Edit button in toolbar

#### Uploading image by capturing

Base implementation is there but some are missing
* Image upload progress / with place holder
* Edit button in toolbar

### 3rd iteration

#### Adding video from WordPress media library

*Not implemented yet*

#### Uploading video from device library

*Not implemented yet*

####  Uploading video by capturing

*Not implemented yet*

---

After reviewing it I’d like to double check a few things:

Do we want to add the Media Upload progress and the edit button in the toolbar to keep it consistent as the Video / Image blocks?

For those, I’d reuse the MediaUploadProgress component but it's within the image block (the video block uses it as well) so I think it’d be best to extract it out of the image block since it would be used from different components (Image, Video, and MediaText). Any thoughts on this?

To add the video functionality we can as well reuse part of the code of the Video block but it’d also affect if we want to have the upload progress as well and if we extract the previously mentioned component (Media Upload Progress) or not.

Let me know when you can, thanks!

pinarol:
hi @geriux 👋 Thanks for the analysis! I'll share my opinions here but defer the conclusion to @koke 

[...]

PR #1378, opened 2019-09-23
Media & Text - Media picker buttons functionality

geriux:
Gutenberg PR https://github.com/WordPress/gutenberg/pull/17537

Fixes https://github.com/wordpress-mobile/gutenberg-mobile/issues/1313

To test check https://github.com/WordPress/gutenberg/pull/17537 for a full description and screenshots.

It also includes a small change in `RNReactNativeGutenbergBridge.swift` to pass the media type.

#### Parent apps PRs

**WPiOS**: https://github.com/wordpress-mobile/WordPress-iOS/pull/12581
**WPAndroid**: https://github.com/wordpress-mobile/WordPress-Android/pull/10567 (it has the changes from the multiple media PR) but there aren't any new native changes, just the updated gutenberg reference

---

Update release notes:

### DIFF
--- a/react-native-gutenberg-bridge/ios/GutenbergBridgeDelegate.swift
+++ b/react-native-gutenberg-bridge/ios/GutenbergBridgeDelegate.swift
@@ -1,4 +1,4 @@
-public typealias MediaPickerDidPickMediaCallback = (_ id: Int32?, _ url: String?, _ mediaType: String?) -> Void
+public typealias MediaPickerDidPickMediaCallback = (_ media: [(Int32?,String?,String?)]?) -> Void
 
 public enum MediaPickerSource: String {
     case mediaLibrary = "SITE_MEDIA_LIBRARY"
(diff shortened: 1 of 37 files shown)


koke:
I tried to upload a video and after the upload finished, I got this warning:

> Failed to get size for image: https://sandbox.koke.me/wp-content/uploads/2019/09/img_1624.mov

And the video part in media-text only shows the placeholder.

geriux:
> I tried to upload a video and after the upload finished, I got this warning:
> 
> > Failed to get size for image: https://sandbox.koke.me/wp-content/uploads/2019/09/img_1624.mov
> 
> And the video part in media-text only shows the placeholder.

Oh no, that's weird, do you know if I have to do something for the change of `ios/RNReactNativeGutenbergBridge.swift` to work? Did you test it locally? If so, you would have to rebuild because of the native change.

koke:
> If so, you would have to rebuild because of the native change

🤦‍♂ My bad, I didn't realize there were native changes and was testing with the wrong binaries

koke:
I got a weird issue... after testing the video upload, I removed the media-text block, inserted a new one and tried to upload an image. The media-text block will upload correctly but then try to show the image as a video and fail. When I closed the editor I got a crash:

```
Unhandled JS Exception: Cannot read property 'props' of null, stack:
<unknown>@177709:33
<unknown>@40386:21
_callTimer@40303:9
Object.callTimers@40510:9
MessageQueue.__callFunction@15816:44
<unknown>@15558:17
MessageQueue.__guard@15770:13
MessageQueue.callFunctionReturnFlushedQueue@15557:14
<unknown>@80:58
```

I haven't been able to reproduce the crash again 😞

koke:
> <unknown>@177709:33

Looking at the generated JS, the crash seems to be in [KeyboardAwareFlatList](https://github.com/WordPress/gutenberg/blob/a56daebec30a3740116e7e495d9776ef4c3d51a6/packages/components/src/mobile/keyboard-aware-flat-list/index.ios.js#L33), so I would assume it's a coincidence unrelated to these changes, but let me know if you think you changed something that might be causing that.

geriux:
> > @177709:33
> 
> Looking at the generated JS, the crash seems to be in [KeyboardAwareFlatList](https://github.com/WordPress/gutenberg/blob/a56daebec30a3740116e7e495d9776ef4c3d51a6/packages/components/src/mobile/keyboard-aware-flat-list/index.ios.js#L33), so I would assume it's a coincidence unrelated to these changes, but let me know if you think you changed something that might be causing that.

Not really, but I will double check just in case

etoledom:
Milestone moved to [1.15](https://github.com/wordpress-mobile/gutenberg-mobile/milestone/24)

react-native-gutenberg-bridge/ios/GutenbergBridgeDelegate.swift
```diff
@@ -1,4 +1,4 @@
-public typealias MediaPickerDidPickMediaCallback = (_ media: [(Int32?,String?)]?) -> Void
+public typealias MediaPickerDidPickMediaCallback = (_ media: [(Int32?,String?,String?)]?) -> Void
```
etoledom: These callback parameter list is kind of growing a lot 🤔 . Maybe we could think about using a struct?
Having too many unnamed parameters in the public API doesn't sound like a good practice.
cc @koke @pinarol

koke: That sounds like a good idea to me 😄

PR #1371, opened 2019-09-20
Feat: Navigation Upward in InnerBlocks

jbinda:
Fixes #1315 

Please also refer to:
Merge it first gutenberg-mobile PR [Add FloatingToolbar component](https://github.com/wordpress-mobile/gutenberg-mobile/pull/1362)
[Related gutenberg PR](https://github.com/WordPress/gutenberg/pull/17496)
[Related gutenberg-mobile issue](https://github.com/wordpress-mobile/gutenberg-mobile/issues/1315)

It presents:
`Icon` on `Floating Toolbar` component witch allow to select parent of selected block. It is visible only when some of the `Inner Block` is selected

Update release notes:

- [ ] If there are user facing changes, I have added an item to `RELEASE-NOTES.txt`.

PR #1362, opened 2019-09-10
Add FloatingToolbar component

lukewalczak:
Fixes #1317 

To test:

Tested on both platforms
https://github.com/WordPress/gutenberg/pull/17399

Update release notes:

- [ ] If there are user facing changes, I have added an item to `RELEASE-NOTES.txt`.

PR #1379, opened 2019-09-24
Feat: Navigation Down in InnerBlocks

### LINKED ISSUE #1314: Inner Blocks - Navigating Down in the hierarchy
User should be able to navigate downward in the hierarchy.

Please take a look at the [prototype](https://href.li/?https://www.figma.com/proto/SptxS8feaNYtWUzoeNM4qI/Nested-Blocks-i2?node-id=2%3A9167&viewport=-1433%2C80%2C1&scaling=min-zoom) to get a sense of how it should behave.

**Block Focus/Visual Cues**

**Transparency**

You’ll notice below that the selected parent and all of its children are fully visible – everything outside of that is semi-transparent (in this example, 20% opacity). This helps to draw the eye to the current selection, without losing focus of the wider picture.

5544ea2e-3a70-4281-85fc-130d3e3a0be3

**Selection Indication**

8dac2ea8-6785-43b1-8747-1172d0fd9ea5

A selected block that has children gets a strong focus border (blue) and each child on the next level gets a dotted outline that hints at what’s inside.

jbinda:
Fixes #1314 

Please also refer to:
Merge it first gutenberg-mobile PR [Add FloatingToolbar component](https://github.com/wordpress-mobile/gutenberg-mobile/pull/1362) and [Feat: Navigation Upward in InnerBlocks](https://github.com/wordpress-mobile/gutenberg-mobile/pull/1371)
[Related gutenberg PR](https://github.com/WordPress/gutenberg/pull/17547)
[Related gutenberg-mobile issue](https://github.com/wordpress-mobile/gutenberg-mobile/issues/1314)

It presents:
Ability to select all parents in cascade way (according to the way how selecting in groups works on web version)

Update release notes:

- [x] If there are user facing changes, I have added an item to RELEASE-NOTES.txt.


pinarol:
@jbinda Just curious how is it acting when we have paragraph & heading blocks down in the hierarchy and directly tap on them?

pinarol:
@iamthomasbishop this needs a design review, could you be able to take a look at the screen capture here https://github.com/WordPress/gutenberg/pull/17547

iamthomasbishop:
It’s really great to see the hierarchy navigation in practice, great work so far. I have a few requests, some of which are relevant to other in-progress PRs.

* Padding’s/margins jumping around a bit make navigating the hierarchy a bit clunky/jumpy. More on that [here](https://github.com/wordpress-mobile/gutenberg-mobile/issues/1314#issuecomment-535214118)
* The dotted border style (on child blocks) is a bit stronger than the proposed style. Let’s dim the color to match the prototype
* Another thing I’ve already mentioned but worth noting again – it would be beneficial to show the icon of the selected block on the quick toolbar. This would add clarity to what is selected.

I’ll take another look tomorrow, but these are the things that jumped out at me first.

jbinda:
@pinarol paragraph and heading block gets immediately selected after click whatever level it's is.

I know why but didn't figured out the solution yet. I described it more here (#1314) but in few words the RichText editor takes the focus and I need to lock it somehow (take advantages of modify callback on JS side does't work because the text carret still show up. I assume it is something inside the Aztec native code that catch and handle click/focus). Take a look at below demo:

heading

Do you know any prop that can lock the Aztec editor ?

pinarol:
> Do you know any prop that can lock the Aztec editor ?

I don't think we have such a thing. Did you have chance to look into web implementation? We might be able to get inspiration from there.

I think we should look for something generic that can be applicable to all blocks, like for example a putting a transparent view on block.native.js level to prevent taps and removing it when the block is really focused. Not %100 sure how well it'd work but just an idea.

lukewalczak:
That PR also contains the logic for `FloatingToolbar` functionality from [PR](https://github.com/WordPress/gutenberg/pull/17811). More details with the PR ⬆️ .

PR #1383, opened 2019-09-25
Fix when multiple media selection adds only one image or video block

### LINKED ISSUE #1346: [Android] Multiple media selection adds only one image/video block
Current behavior:

- Add an image/video block
- Add multiple images/videos from WordPress media library
- Only 1 image/video block is added

Expected behavior:

- Add an image/video block
- Add multiple images/videos from WordPress media library
- Multiple image/video blocks are added with the selected images/videos

Root cause analysis

It looks like we are having a side effect of a [recently merged PR](https://github.com/WordPress/gutenberg/pull/16677/files). The code that subscribes for mediaAppend events are moved and since inserter menu is not rendered after we pick the media it won't be able to listen. We can try moving this subscription back to its old place as a solution, or anywhere more suitable which won't be unmounted along the way.

This is only observable via WordPress Android app because WordPress iOS app doesn't let multiple selection on WordPress media library currently.

marecar3:
Fixes #1346  . Additional details and testing instructions are in the description of the related [gutenberg PR](https://github.com/WordPress/gutenberg/pull/17579).

Update release notes:


hypest:
👋 @marecar3 and @dratwas , it's not clear to me and unless you two have already coordinated, can you both make sure the works in this PR and #1331 are compatible? They seem to touch similar functionality so, just wanted to make sure. Thanks!

marecar3:
Hey @hypest, thanks for the ping.

As I can see, It seems that this [PR](https://github.com/wordpress-mobile/gutenberg-mobile/pull/1331) is introducing multiple select on a native level and this PR is fixing existing support (for multiple select) on JS side.

We can wait for @dratwas to confirm.

cc: @pinarol @mkevins

dratwas:
Hey, in my PR I added the possibility to select how we will handle the multiple select. If multiple is set to false - we will use the old implementation and few blocks with image/video should be added by sending event through the bridge. If multiple is set to true, all selected media will be passed to the JS callback instead of sending them in event one by one.

RELEASE-NOTES.txt
```diff
@@ -1,3 +1,8 @@
+1.15.0
+------
+* Fix issue when multiple media selection adds only one image or video block
+* Fix issue when force Touch app shortcut doesn't work properly selecting "New Photo Post"
```
etoledom: Should we specify the platform?
The first one is Android only and the second one is iOS only.

etoledom:
Milestone moved to [1.15](https://github.com/wordpress-mobile/gutenberg-mobile/milestone/24)

PR #1392, opened 2019-09-30
Release/1.14.0

marecar3:
Merging release/1.14.0 branch into master

Remaining work
- [ ] Merge [gutenberg side](https://github.com/WordPress/gutenberg/pull/17676)
- [x] Test and approve [WPiOS PR](https://github.com/wordpress-mobile/WordPress-iOS/pull/12574)
- [x] Test and approve [WPAndroid PR](https://github.com/wordpress-mobile/WordPress-Android/pull/10543)
- [ ] Merge this PR
- [x] Tag Gutenberg
- [ ] Tag head of master branch for 1.14.0 release
- [ ] Update iOS PR to reference gutenberg-mobile release tag and merge iOS PR
- [ ] Update Android PR to reference tagged commit and merge Android PR
- [ ] Open PR from master to develop
- [ ]  🎉

To test:

Test native apps against this PR

- [WPiOS PR](https://github.com/wordpress-mobile/WordPress-iOS/pull/12574)
- [WPAndroid PR](https://github.com/wordpress-mobile/WordPress-Android/pull/10543)

Update release notes:

- [ ] If there are user facing changes, I have added an item to RELEASE-NOTES.txt.


hypest:
Thanks for wrangling this release cut @marecar3 !

> there is nothing to merge because 

It may be true that there's nothing to merge today that the branches are cut but, since we plan to let this PR settle for a few days, it's possible that we might add things in the days before merging. If we add commits, the text might be confusing so, may I suggest revising it by either removing it or adding something about the plan?

>  it was branched off of rnmobile/master

I might have missed something so, I wonder if there's a reason for not branching off the last known-good Gutenberg `master` instead. I think [a recent PR](https://github.com/wordpress-mobile/gutenberg-mobile/pull/1366) has brought the gutenberg-mobile develop back to targeting Gutenberg master.

> Open PR from master to develop (won't be necessary unless we end up adding changes to this PR)

I think we always need to merge back to gutenberg-mobile develop so, for example, the bumped version number is merged too.

> Submodule gutenberg updated from aeaf0a to 29ad4d

Noticed that this points to Gutenberg master but to a commit that is newer than any tested on gutenberg-mobile master. Effectively, this brings in new Gutenberg code that hasn't been tested yet, right? If so, I think it's better to avoid introducing new code in a release cut and instead, point to the last known-good Gutenberg commit. Does that make sense?

marecar3:
Thanks @hypest for the good 👁 review.
Text in the description wasn't correct as I used some template which I found interesting (from earlier releases). Sorry about that.

I agree about the reference. I have pointed to the latest well know gb ref.

Also, I have fixed WPAndroid and WPiOS description templates.

mchowning:
:wave: @marecar3 ! Do we need to update the release notes with this note from @SergioEstevao 's fix?

```
 * Fix a link editing issue, where trying to add a empty link at the start of another link would remove the existing link.
```

It sounds like that fix applies to both platforms.

marecar3:
Thank you @etoledom for the great support!!!

PR #1397, opened 2019-10-02
Introduce grouping in the block settings inspector

### LINKED ISSUE #1384: Introduce grouping in the Block Settings Inspector
We are about to complete [refactoring Block Settings](https://github.com/wordpress-mobile/gutenberg-mobile/pull/1341) in a cross platform manner. Next step is introducing the grouping behavior similar to web. We had some discussions around the issue but I think we don't have a clear conclusion yet. Inviting @iamthomasbishop into conversation at this point.

From technical side of view, considering that this component will be cross platform, it'd be beneficial to keep things similar to web which displays settings inside a single page as a tree structure. But we can always explore other options, I just wanted to mention the technical aspect of it.

lukewalczak:
Fixes #1384 
Ref to gutenberg: https://github.com/WordPress/gutenberg/pull/17703

To test:

1. Add an image.
2. Press and icon 
3. Check the `BlockSettings` inspector

Update release notes:

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,6 @@
 [submodule "gutenberg"]
 	path = gutenberg
-	url = ../../WordPress/gutenberg.git	
+	url = ../../WordPress/gutenberg.git
 [submodule "react-native-aztec"]
 	path = react-native-aztec-old-submodule
 	url = ../react-native-aztec.git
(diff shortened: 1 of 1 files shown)


iamthomasbishop:
I gave a thorough design review [over here](https://github.com/WordPress/gutenberg/pull/17703#issuecomment-537573314). Let me know if I missed anything or when another review is needed!

PR #1405, opened 2019-10-04
[FIX] Fix button component

dratwas:
Fixes https://github.com/WordPress/gutenberg/issues/17768
There is the wrong assumption that the child is always an array. https://github.com/WordPress/gutenberg/blob/master/packages/components/src/button/index.native.js#L96
When we pass only one component like View to the button it will be not rendered

Gutenberg PR - https://github.com/WordPress/gutenberg/pull/17766
To test:
Pass only one component as a child to Button and check if it is rendered correctly

Update release notes:

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,6 @@
 [submodule "gutenberg"]
 	path = gutenberg
-	url = ../../WordPress/gutenberg.git
+	url = https://github.com/callstack/gutenberg.git
 [submodule "react-native-aztec"]
 	path = react-native-aztec-old-submodule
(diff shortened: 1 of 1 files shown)


etoledom:
Moved this PR to the milestone `1.16`.

If this need to ship on v1.15, please feel free to change back the milestone and reference the release branch.

dratwas:
I think it is not needed at all. Since the PR on `gutenber` was merged some time ago.

pinarol:
Closing this as gutenberg ref is already updated with another PR

PR #1426, opened 2019-10-09
Update to iOS11 and Swift 5.

SergioEstevao:
This PR just updates the GB projects to have a minimum requirements of iOS 11 and to use Swift 5.

To test:
 - Check if project builds
 - Smoke test the demo app to see if all is correct.

Update release notes:

### DIFF
--- a/Gutenberg.podspec
+++ b/Gutenberg.podspec
@@ -11,12 +11,12 @@ Pod::Spec.new do |s|
   s.homepage     = 'https://github.com/wordpress-mobile/gutenberg-mobile'
   s.license      = package['license']
   s.authors          = 'Automattic'
-  s.platform     = :ios, '10.0'
+  s.platform     = :ios, '11.0'
   s.source       = { :git => 'https://github.com/wordpress-mobile/gutenberg-mobile.git' }
   s.source_files = 'react-native-gutenberg-bridge/ios/*.{h,m,swift}'
   s.requires_arc = true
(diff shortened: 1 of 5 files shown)


koke:
This is being reverted in #1426. When I tried to update WPiOS to the new develop, I found two build issues.

The first is that this PR removed an `#available(iOS 11.0, *)` check in the aztec wrapper, but RNTAztecView.podspec still has a target of iOS 10. That one has an easy fix: just update `s.platform` on that podspec.

The second one, I got some errors building Gutenberg.framework about libxml2 not finding a header file. I investigated a bit but I couldn't make sense of the error:

```
<module-includes>:1:9: note: in file included from <module-includes>:1:
#import "libxml/HTMLparser.h"
        ^
/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS13.1.sdk/usr/include/libxml2/libxml/HTMLparser.h:15:10: error: 'libxml/xmlversion.h' file not found
#include <libxml/xmlversion.h>
         ^
<unknown>:0: error: could not build Objective-C module 'libxml2'
```

etoledom:
> I got some errors building Gutenberg.framework about libxml2 not finding a header file. I investigated a bit but I couldn't make sense of the error:

@koke - I think we should also bump the `s.swift_version    = '5.0'` on RNAztec, that should fix the libxml2 error.

PR #1431, opened 2019-10-10
Revert "Update to iOS11 and Swift 5."

koke:
Reverts wordpress-mobile/gutenberg-mobile#1426

That PR was causing problems with the integration of gutenberg-mobile into WordPress for iOS. Reverting until we find a solution to those issues


etoledom:
This revert might not be needed if https://github.com/wordpress-mobile/gutenberg-mobile/pull/1434 works.

PR #1435, opened 2019-10-11
Import new location for base styles

koke:
In https://github.com/WordPress/gutenberg/pull/17883 the base SCSS styles are moving to a separate package. This updates our SASS transformer so it knows to look for them in the new path.

This can be merged before https://github.com/WordPress/gutenberg/pull/17883 since I didn't remove the old location.

To test:

Update the gutenberg submodule to https://github.com/WordPress/gutenberg/pull/17883 and see if everything runs correctly

Update release notes:

### DIFF
--- a/sass-transformer.js
+++ b/sass-transformer.js
@@ -67,6 +67,7 @@ if ( reactNativeMinorVersion >= 59 ) {
 const autoImportIncludePaths = [
 	path.join( path.dirname( __filename ), 'src' ),
 	path.join( path.dirname( __filename ), 'gutenberg/assets/stylesheets' ),
+	path.join( path.dirname( __filename ), 'gutenberg/packages/base-styles' ),
 ];
 const autoImportAssets = [
 	'_colors.scss',


sass-transformer.js
```diff
@@ -67,6 +67,7 @@ if ( reactNativeMinorVersion >= 59 ) {
 const autoImportIncludePaths = [
 	path.join( path.dirname( __filename ), 'src' ),
 	path.join( path.dirname( __filename ), 'gutenberg/assets/stylesheets' ),
```
gziolo: This line can be removed.
```suggestion
```

koke: It can, after https://github.com/WordPress/gutenberg/pull/17883 is merged 😉 
I left that so we don't have to coordinate merging both PRs. Once the base-styles are merged, we can remove that line in another PR.

hypest: I think it's a good strategy in this case to keep both the "old" and the new paths so we smoothly transition to the new one. I'm approving this PR, but let us know if you have some more feedback here @gziolo and we can address it in the follow up PR.

PR #1436, opened 2019-10-11
Add mobile Spacer component

### LINKED ISSUE #1428: Add Spacer block - 1st iteration
We need to add mobile equivalent of Spacer block, it will look same with web but for the 1st iteration we won't be adding a draggable handle to resize the canvas. The size of the canvas will be determined by block settings.

cc @iamthomasbishop

lukewalczak:
Fixes #1428 
Ref to gutenberg PR: https://github.com/WordPress/gutenberg/pull/17896

To test: Instruction attached in gutenberg PR.

Update release notes:


ios/gutenberg.xcodeproj/project.pbxproj
```diff
@@ -903,7 +903,7 @@
 						TestTargetID = 13B07F861A680F5B00A75B9A;
 					};
 					13B07F861A680F5B00A75B9A = {
-						DevelopmentTeam = PZYM8XX95Q;
+						DevelopmentTeam = PJL7DA2X5F;
```
pinarol: Is this for testing on device? Let's not forget to revert this before merge

src/initial-html.js
```diff
@@ -4,212 +4,15 @@
  */
```
pinarol: Same for this, just dropping a comment to not to forget reverting this file. Let's add the new block somewhere appropriate in the demo content though.

```
<!-- wp:spacer -->
 <div style="height:100px" aria-hidden="true" class="wp-block-spacer"></div>		 <div style="height:100px" aria-hidden="true" class="wp-block-spacer"></div>
 <!-- /wp:spacer -->		 <!-- /wp:spacer -->
		
```

pinarol:
Could you add a sample spacer block into the demo content?(initial-html.js)

etoledom:
> Could you add a sample spacer block into the demo content?(initial-html.js)

@pinarol - There are two already. I guess they used to be hidden as “unsupported”.

pinarol:
No need then, thanks for the heads up @etoledom :) let’s merge this anyway. gutenberg ref is prolly updated by another pr already but if we close this without merging then it can be misleading for the weekly reports.

hypest:
👋, my impression is that the Spacer block is currently behind the __DEV__ flag so, not visible to the end/production user. Can you confirm @lukewalczak and if that's the case, let's also update the PR description to note that fact, thanks!

lukewalczak:
@hypest You're right, `Spacer` component is behind the `DEV` [flag](https://github.com/WordPress/gutenberg/blob/master/packages/block-library/src/index.native.js#L148). I've updated the description above 👍

Issue #1441, opened 2019-10-11
Media & Text icon shows wrong color in dark mode


koke:
![Simulator Screen Shot - iPhone 11 - 2019-10-11 at 15 40 10](https://user-images.githubusercontent.com/8739/66656009-6f8a3700-ec3d-11e9-8a15-de619a2be112.png)

etoledom:
Fixed via https://github.com/wordpress-mobile/gutenberg-mobile/pull/1443

PR #1443, opened 2019-10-14
Fix icons for Media & Text in dark mode

### LINKED ISSUE #1411: [Exploration] Make @wordpress/components/Button cross-platform using styled-system
This is the second iteration on https://github.com/wordpress-mobile/gutenberg-mobile/issues/1386

If preferred this work can be based on the existing experiment branch: https://github.com/WordPress/gutenberg/pull/17614

We want to refactor Button component in a way that uses styled-system similar to what we did for `Box` here in [this PR](https://github.com/WordPress/gutenberg/pull/17614).
gutenberg/packages/components/src/button/index.js

The aim is being able to test that we can apply its styles using props. But we should be careful about keeping it backward compatible. What does it mean to keep it backward compatible? Well, currently [themes](https://wordpress.org/themes/) rely on tag names as well as class names to be able to override existing styles we use. So we shouldn't cause unwanted regressions on themes. In order this to happen we should keep classes applied as the same way. Fortunately this is looks easy, [styled-components allow it](https://href.li/?https://www.styled-components.com/docs/faqs#can-i-use-css-frameworks) and we already demonstrated how to inject the already existing css into it [here](https://github.com/WordPress/gutenberg/pull/17614).

1st iteration(only effects web version):
- Allow passing style via props to Button on gutenberg repo 
- Refactor a few places Button is used and pass the style via props, 3 sample cases would be fine.
- Remove unused css from .scss files but keep the class names applied to Button as themes might be using those
- Other Button instances should keep on working the same

2nd iteration:
Another PR to demonstrate that the Button API works on mobile too. Currently the props on mobile and web are pretty different. At this point choose some props that mobile doesn't support, like isDefault, isLink, add mobile side implementation in button/index.native.js for them. Currently there's no use case on mobile using these props so you can create a new test block(or add them to an existing simple block) just to demonstrate them using the example app.

koke:
Fixes #1411 

To test: See https://github.com/WordPress/gutenberg/pull/17936

Update release notes:

### DIFF
--- a/src/initial-html.js
+++ b/src/initial-html.js
@@ -8,6 +8,12 @@ export default `
 <figure class="wp-block-image"><img alt=""/></figure>
 <!-- /wp:image -->
 
+<!-- wp:media-text -->
+<div class="wp-block-media-text is-stacked-on-mobile"><figure class="wp-block-media-text__media"></figure><div class="wp-block-media-text__content"><!-- wp:paragraph {"placeholder":"Content…","fontSize":"large"} -->
+<p class="has-large-font-size"></p>
+<!-- /wp:paragraph --></div></div>
+<!-- /wp:media-text -->
+
 <!-- wp:image -->
 <figure class="wp-block-image"><img src="https://cldup.com/cXyG__fTLN.jpg" alt=""/></figure>
 <!-- /wp:image -->


pinarol:
> Fixes #1411

The link to the fixed issue doesn't look right

etoledom:
Changed the issue link for future reference to #1441
👍

PR #1453, opened 2019-10-16
[try] Image Size networking

etoledom:
This PR implements apiFetch to get the media object, and set the corresponding image URL when the image size is changed on an image block.

The wpcom specific code is located in this repo. And the needed information (authentication headers and site id/slug) are provided via initial props from the parent app.

This can only be tested on WPiOS for now, and only works on WPCom sites, as a first step.

gutenberg PR: https://github.com/WordPress/gutenberg/pull/17979
WPiOS PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/12706

To test:
- Follow the instructions on https://github.com/wordpress-mobile/WordPress-iOS/pull/12706

Update release notes:

### DIFF
--- a/react-native-gutenberg-bridge/ios/Gutenberg.swift
+++ b/react-native-gutenberg-bridge/ios/Gutenberg.swift
@@ -63,6 +63,14 @@ public class Gutenberg: NSObject {
         if let translations = dataSource.gutenbergTranslations() {
             initialProps["translations"] = translations
         }
+
+        if let slug = dataSource.siteSlug {
+            initialProps["siteSlug"] = slug
+        }
+
+        if let extraHeaders = dataSource.extraHTTPHeaders {
+            initialProps["extraHTTPHeaders"] = extraHeaders
+        }
         
         return initialProps
     }
--- a/src/api-fetch-setup.js
+++ b/src/api-fetch-setup.js
@@ -1,0 +1,55 @@
+/**
+ * External dependencies
+ *
+ * @format
+ */
+
+/**
+ * External dependencies
+ */
+import { fetchRequest } from 'react-native-gutenberg-bridge';
+
+/**
+ * WordPress dependencies
+ */
+import apiFetch from '@wordpress/api-fetch';
+
+const fetchHandler = ( nextOptions, extraHeaders ) => {
+	const { url, ...remainingOptions } = nextOptions;
+
+	headers = { ...apiFetch.DEFAULT_HEADERS, ...extraHeaders };
+
+	const options = {
+		headers,
+		...remainingOptions,
+	};
+	const responsePromise = fetch( url, options );
+
+	const parseResponse = ( response ) => {
+		return response.json();
+	};
+
+	return responsePromise.then( parseResponse ).catch( ( error ) => {
+		return error;
+	} );
+};
+
+const wpcomPathMappingMiddleware = ( options, next, siteSlug ) => {
+	// wp/v2 namespace mapping
+	//
+	// Path rewrite example:
+	// 		/wp/v2/types/post →
+	//		/wp/v2/sites/example.wordpress.com/types/post
+	if ( /\/wp\/v2\//.test( options.path ) ) {
+		const path = options.path.replace( '/wp/v2/', `/wp/v2/sites/${ siteSlug }/` );
+
+		return next( { ...options, path } );
+	}
+	return next( options );
+};
+
+export default ( setupApiFetch = ( siteSlug = '', headers = {} ) => {
+	apiFetch.use( apiFetch.createRootURLMiddleware( 'https://public-api.wordpress.com/' ) );
+	apiFetch.use( ( options, next ) => wpcomPathMappingMiddleware( options, next, siteSlug ) );
+	apiFetch.setFetchHandler( ( options ) => fetchHandler( options, headers ) );
+} );
--- a/src/api-fetch-setup.js
+++ b/src/api-fetch-setup.js
@@ -5,40 +5,21 @@
  */
 
 /**
- * External dependencies
- */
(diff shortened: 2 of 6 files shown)


etoledom:
Even though it's a draft, please feel free to comment on this approach 👍

react-native-gutenberg-bridge/ios/Gutenberg.swift
```diff
@@ -63,6 +63,14 @@ public class Gutenberg: NSObject {
         if let translations = dataSource.gutenbergTranslations() {
             initialProps["translations"] = translations
         }
+
+        if let slug = dataSource.siteSlug {
+            initialProps["siteSlug"] = slug
+        }
+
+        if let extraHeaders = dataSource.extraHTTPHeaders {
+            initialProps["extraHTTPHeaders"] = extraHeaders
```
koke: I would try to keep auth tokens away from initial props since react native seems to aggressively log those, and we shouldn’t log tokens

src/api-fetch-setup.js
```diff
@@ -0,0 +1,55 @@
+/**
+ * External dependencies
+ *
+ * @format
+ */
+
+/**
+ * External dependencies
+ */
+import { fetchRequest } from 'react-native-gutenberg-bridge';
```
koke: Is this from an older attempt or I’m missing where it’s used?

src/api-fetch-setup.js
```diff
@@ -0,0 +1,55 @@
...
+
+		return next( { ...options, path } );
+	}
+	return next( options );
+};
+
+export default ( setupApiFetch = ( siteSlug = '', headers = {} ) => {
+	apiFetch.use( apiFetch.createRootURLMiddleware( 'https://public-api.wordpress.com/' ) );
+	apiFetch.use( ( options, next ) => wpcomPathMappingMiddleware( options, next, siteSlug ) );
+	apiFetch.setFetchHandler( ( options ) => fetchHandler( options, headers ) );
```
koke: You don't need a custom fetch handler to inject headers, I got this one working in our meetup:
https://github.com/WordPress/gutenberg/commit/109182dfcea4600d9dfc18cd2a5b77849e004595

src/api-fetch-setup.js
```diff
@@ -0,0 +1,55 @@
...
+		const path = options.path.replace( '/wp/v2/', `/wp/v2/sites/${ siteSlug }/` );
+
+		return next( { ...options, path } );
+	}
+	return next( options );
+};
+
+export default ( setupApiFetch = ( siteSlug = '', headers = {} ) => {
+	apiFetch.use( apiFetch.createRootURLMiddleware( 'https://public-api.wordpress.com/' ) );
+	apiFetch.use( ( options, next ) => wpcomPathMappingMiddleware( options, next, siteSlug ) );
```
koke: Even if we don't have .org auth ready, we should make these conditional. The app should provide the root URL, and depending on whether that's `*.wordpress.com` or a site URL, the path mapping (and `public-api` root url) should be added automatically.

This also means that `siteSlug` might be inferred from the passed root URL, as  it's only needed for WordPress.com requests.

react-native-gutenberg-bridge/ios/Gutenberg.swift
etoledom: Yes that makes sense. This was the easier way to make it work.

And also we will need to refresh it when the user switches the blog from the editor, avoiding fetching it on every network call.

src/api-fetch-setup.js
etoledom: Yes, this was part of the tests making network from the parent apps. I plan to post a PR with this other approach.

src/api-fetch-setup.js
etoledom: Sounds good!
Implemented a simpler approach, since we are passing a generic "extra headers" prop, not necessarily containing auth related fields.

koke:
Taking this out of the line comment because it's important enough:

> ...we are passing a generic "extra headers" prop, not necessarily containing auth related fields

I'd rather deal with auth separately and explicitly. I know this might seem premature but [I've been scarred before](https://techcrunch.com/2019/04/02/wordpress-bug-account-tokens/).

If you look at [the documentation for api-fetch](https://github.com/WordPress/gutenberg/tree/2c11c83c262c4043425d583e2eb038f77a4a6f61/packages/api-fetch), it supports passing an absolute `url`. I haven't seen any use case yet in core where `apiFetch` is called with anything other than a relative path, but it's a real possibility.

Core doesn't have to worry about that, because it uses cookie-based authentication, and browsers will do the worrying about where to send each cookie. My middleware implementation was also too naive, and didn't check where it was sending the request before injecting a token 🤦‍♂ 

I think we have two options:

- Using a custom fetch handler: we know that at this point nothing else should modify the request, and we can decide there if the header needs to be added based on host name. However, we would be forking the default handler and losing any future improvements.
- Using a middleware, but instead of providing just a token, we pass a pair of token and a hostname/pattern where it's ok to send that token. I prefer this solution, but for extra safety we should investigate ways to add "priorities" to middleware and ensure this runs after every other middleware has had a chance to modify the request.

I think we can easily build the first today, but I'd love for us to invest in the second for a longer term solution.

etoledom:
> I'd rather deal with auth separately and explicitly. I know this might seem premature but I've been scarred before.

That makes total sense. Will take a look at it.

This is one of the reasons I also like the option of doing the networking bits from the parent apps: https://github.com/wordpress-mobile/gutenberg-mobile/pull/1455

But I'm not in favor of any option in particular, so I'm happy to continue improving this approach too.

etoledom:
Closing this in favor of https://github.com/wordpress-mobile/WordPress-iOS/pull/12714

PR #1455, opened 2019-10-17
[try] Image Size with networking on native client

etoledom:
Alternative to https://github.com/wordpress-mobile/gutenberg-mobile/pull/1453

This PR pass the networking request intent to the client app, and awaits for a response.

As a proof of concept, this PR implements the minimum necessary to make Image Size works.
In the future we might need to pass the request method and parameters.

**Related PRs:**
* gutenberg: https://github.com/WordPress/gutenberg/pull/17979 (same as https://github.com/wordpress-mobile/gutenberg-mobile/pull/1453)
* WPiOS: https://github.com/wordpress-mobile/WordPress-iOS/pull/12714
* WPAndroid: https://github.com/wordpress-mobile/WordPress-Android/pull/10779
* FluxC: [merged](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1424)

**To test on iOS** refer to WPiOS PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/12714
**To test on Android** refer to WPAndroid PR: https://github.com/wordpress-mobile/WordPress-Android/pull/10779

Update release notes:


etoledom:
@mchowning FYI - Just rebased all relevant branches to keep them up to date.

src/api-fetch-setup.js
```diff
@@ -0,0 +1,31 @@
...
+import apiFetch from '@wordpress/api-fetch';
+
+const fetchHandler = ( options ) => {
+	const responsePromise = fetchRequest( options.path );
+
+	const parseResponse = ( response ) => {
+		return response;
+	};
+
+	return responsePromise.then( parseResponse ).catch( ( error ) => {
```
koke: Do we need the `then`/`error` if they're just returning the passed value?

mchowning:
Hey @etoledom ! I merged the code from `release/1.17` into this branch and updated the PR to target that release branch.

src/api-fetch-setup.js
mchowning: I've updated `parseResponse` so it no longer just returns the passed value like it used to. I made this change because I have Android returning a json string because that's much simpler to do. If there's an issue with that I can certainly update it to return a json object like iOS.

With respect to the `catch` clause, I believe adding that does change the behavior so that now the resulting promise essentially "resolves" with the error (i.e., the resulting promise will pass the error in any subsequent `then` calls). Seems like this probably is _not_ desired behavior. What do you think @etoledom ?

etoledom:
So far it looks good to me ✅ 

I have found a few issues that I think are beyond this (Networking specific) PR:

- [WPiOS]: When selecting different a different image size, the image's size doesn't change visually. This is due to the [`Image.getSize`](https://github.com/WordPress/gutenberg/blob/ff6afdad160f73b88eba3b427ad2fd15c87fc9a0/packages/block-editor/src/components/media-upload-progress/image-size.native.js#L55) call fails to return the correct size on urls with the form `...?w=100` (from what I tested), returning the full size all the time. Then this value is sent to the native side which uses Photon API to get the image with the given size.

- [WPiOS (Probably Android too)] When the image size request fails, Gutenberg won't ever request it again for the same image. This might be a problem for users with bad connectivity.

- The default Image Size slug is Large, but the actual image URL is the full size one. On web, a new added image has set the proper Large image URL.

src/test/api-fetch-setup.test.js
```diff
@@ -0,0 +1,32 @@
...
+	'wp/v2/media/54/',
+	'wp/v2/media/',
+	'wp/v2/media?context=edit&_locale=user',
+];
+
+const unsupportedPaths = [
+	'wp/v1/media/', // made up example
+];
+
+describe( 'isPathSupported', () => {
```
SergioEstevao: Love the tests for this.

Issue #720, opened 2019-03-08
Pexels and Giphy search not available form image block


hlashbrooke:
I'm using the alpha-155 version of the Android app, but I doubt this is specific to any platform or version.

When setting a featured image on mobile you have the option to search the Pexels free photo library for an image (which is great), but that option isn't available when adding an image with the image block. I'm not sure if this is intentional or not.

Steps to reproduce:
1. Add an image block
2. Click the 'choose image' link
3. Notice the lack of an option for using the Pexels library

koke:
As we do this, let's make sure we also bring in Giphy and the "Other apps" file picker on iOS

frosty:
+1 from me – I was writing a post yesterday and missed the Other Apps option. I had to import the file in the Media Library section, and then hop back into the editor to pull it into the post.

marecar3:
Added @etoledom on this one as we need iOS support when it's available.

hypest:
Reopening since the iOS side of the implementation is still pending.

designsimply:
From a 4-star review "Need help 4" by Buddah216 – Nov 5, 2019 Version 13.5

> Can we no longer access GIPHY’s through WordPress? I’m working on my blog post and noticed, that I no longer can add GIPHY’s.

PR #1469, opened 2019-10-18
720-Add support for giphy and pexel images

marecar3:
Added support for giphy and pexels images

Fixes https://github.com/wordpress-mobile/gutenberg-mobile/issues/720

WPAndroid PR: https://github.com/wordpress-mobile/WordPress-Android/pull/10628
Gutenberg PR: https://github.com/WordPress/gutenberg/pull/18026

ezgif com-video-to-gif (8)

**Note: works only on Android (@etoledom  will implement iOS part of it)**

To test:

1. Open WPAndroid application
2. Create a new post or open existing
3. Create a new image block
4. Open picker
5. Click on `Choose from Free Photo Library`
6. Confirm and see that it will be shown in the new post
7. Open picker
8. Click on `Choose from Giphy`
9. Confirm and see that it will be shown in the new post

giphy_pexels

**@iamthomasbishop I have two questions regarding UI/UX:**
1. Should we show some icons for Giphy and Pexels options in the picker?
2. Are you ok with strings which were reused from WPAndroid or you think we should put some new strings for gb-mobile?

giphy_centered

@iamthomasbishop also about this one. There are some giphy images that can be smaller than a size of the component. I have centered them, but do we want some background on unused space?

Update release notes:

### DIFF
--- a/react-native-gutenberg-bridge/android/build.gradle
+++ b/react-native-gutenberg-bridge/android/build.gradle
@@ -117,6 +117,9 @@ dependencies {
     implementation "org.webkit:android-jsc:r241213"
     implementation project(':react-native-aztec')
 
+    // For animated GIF support
+    implementation 'com.facebook.fresco:animated-gif:2.0.0'
+
     if (rootProject.ext.buildGutenbergFromSource) {
         println "using gutenberg from source"
         implementation project(':react-native-svg')
--- a/react-native-gutenberg-bridge/android/src/main/java/org/wordpress/mobile/WPAndroidGlue/WPAndroidGlueCode.java
+++ b/react-native-gutenberg-bridge/android/src/main/java/org/wordpress/mobile/WPAndroidGlue/WPAndroidGlueCode.java
@@ -38,7 +38,6 @@ import org.wordpress.mobile.ReactNativeAztec.ReactAztecPackage;
 import org.wordpress.mobile.ReactNativeGutenbergBridge.GutenbergBridgeJS2Parent;
 import org.wordpress.mobile.ReactNativeGutenbergBridge.GutenbergBridgeJS2Parent.MediaSelectedCallback;
 import org.wordpress.mobile.ReactNativeGutenbergBridge.GutenbergBridgeJS2Parent.MediaUploadCallback;
-import org.wordpress.mobile.ReactNativeGutenbergBridge.GutenbergBridgeJS2Parent.OtherMediaOptionsReceivedCallback;
 import org.wordpress.mobile.ReactNativeGutenbergBridge.GutenbergBridgeJS2Parent.RNMedia;
 import org.wordpress.mobile.ReactNativeGutenbergBridge.RNReactNativeGutenbergBridgePackage;
 
--- a/react-native-gutenberg-bridge/android/src/main/java/org/wordpress/mobile/WPAndroidGlue/WPAndroidGlueCode.java
+++ b/react-native-gutenberg-bridge/android/src/main/java/org/wordpress/mobile/WPAndroidGlue/WPAndroidGlueCode.java
@@ -35,7 +35,6 @@ import com.horcrux.svg.SvgPackage;
 import org.wordpress.android.util.AppLog;
 import org.wordpress.mobile.ReactNativeAztec.ReactAztecPackage;
 import org.wordpress.mobile.ReactNativeGutenbergBridge.GutenbergBridgeJS2Parent;
-import org.wordpress.mobile.ReactNativeGutenbergBridge.GutenbergBridgeJS2Parent.MediaSelectedCallback;
 import org.wordpress.mobile.ReactNativeGutenbergBridge.GutenbergBridgeJS2Parent.MediaUploadCallback;
 import org.wordpress.mobile.ReactNativeGutenbergBridge.GutenbergBridgeJS2Parent.RNMedia;
 import org.wordpress.mobile.ReactNativeGutenbergBridge.RNReactNativeGutenbergBridgePackage;
@@ -59,7 +58,6 @@ public class WPAndroidGlueCode {
     private ReactInstanceManager mReactInstanceManager;
     private ReactContext mReactContext;
     private RNReactNativeGutenbergBridgePackage mRnReactNativeGutenbergBridgePackage;
-    private MediaSelectedCallback mPendingMediaSelectedCallback;
     private MediaUploadCallback mPendingMediaUploadCallback;
     private boolean mMediaPickedByUserOnBlock;
 
@@ -78,9 +76,6 @@ public class WPAndroidGlueCode {
     private boolean mShouldUpdateContent;
     private CountDownLatch mGetContentCountDownLatch;
     private WeakReference<View> mLastFocusedView = null;
-
-    public static final String MEDIA_SOURCE_STOCK_MEDIA = "MEDIA_SOURCE_STOCK_MEDIA";
-    public static final String MEDIA_SOURCE_GIPHY_MEDIA = "MEDIA_SOURCE_GIPHY_MEDIA";
 
     private static final String PROP_NAME_INITIAL_DATA = "initialData";
     private static final String PROP_NAME_INITIAL_TITLE = "initialTitle";
@@ -120,8 +115,7 @@ public class WPAndroidGlueCode {
         void onCancelUploadForMediaClicked(int mediaId);
         void onCancelUploadForMediaDueToDeletedBlock(int mediaId);
         ArrayList<MediaOption> onGetOtherMediaImageOptions();
-        void onOtherMediaStockMediaButtonClicked(boolean allowMultipleSelection);
-        void onOtherMediaGiphyMediaButtonClicked(boolean allowMultipleSelection);
+        void onOtherMediaButtonClicked(String mediaSource, boolean allowMultipleSelection);
     }
 
     public interface OnReattachQueryListener {
@@ -159,9 +153,9 @@ public class WPAndroidGlueCode {
             }
 
             @Override
-            public void requestMediaPickFromMediaLibrary(MediaSelectedCallback mediaSelectedCallback, Boolean allowMultipleSelection, MediaType mediaType) {
+            public void requestMediaPickFromMediaLibrary(MediaUploadCallback mediaSelectedCallback, Boolean allowMultipleSelection, MediaType mediaType) {
                 mMediaPickedByUserOnBlock = true;
-                mPendingMediaSelectedCallback = mediaSelectedCallback;
+                mPendingMediaUploadCallback = mediaSelectedCallback;
                 if (mediaType == MediaType.IMAGE) {
                     mOnMediaLibraryButtonListener.onMediaLibraryImageButtonClicked(allowMultipleSelection);
                 } else if (mediaType == MediaType.VIDEO) {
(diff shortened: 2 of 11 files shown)


iamthomasbishop:
@marecar3 thanks for the ping! Happy to see that the implementation we did for Giphy Aztec is working here :)

> Should we show some icons for Giphy and Pexels options in the picker?

We could always use the same icon we're using for the first option in that list (dashicons "image" icon, I believe) for the first iteration. I think we might eventually want to iterate on the bottom sheet used here for selection, but for now we can use that.

> Are you ok with strings which were reused from WPAndroid or you think we should put some new strings for gb-mobile?

Assuming you mean the text label strings used on the bottom sheet, I think that's fine.

> There are some giphy images that can be smaller than a size of the component. I have centered them, but do we want some background on unused space?

Let's look at what the web does and match that as a starting point in terms of alignment, and I don't know if I'd apply a background – assuming that area around the image is still tappable to select the block, that's probably fine.

marecar3:
Hey @iamthomasbishop, thanks for the help 👍 

This is how it looks with icons, is that ok with you?

giphy_pexels

react-native-gutenberg-bridge/android/src/main/java/org/wordpress/mobile/WPAndroidGlue/WPAndroidGlueCode.java
```diff
@@ -79,6 +79,9 @@
     private CountDownLatch mGetContentCountDownLatch;
     private WeakReference<View> mLastFocusedView = null;
 
+    public static final String MEDIA_SOURCE_STOCK_MEDIA = "MEDIA_SOURCE_STOCK_MEDIA";
+    public static final String MEDIA_SOURCE_GIPHY_MEDIA = "MEDIA_SOURCE_GIPHY_MEDIA";
```
etoledom: I wouldn't mention "stock media" or "giphy" at the bridge level. I would say that they are all "extra options".
Ideally we would be able to add more media sources on the client app (like the iOS-only `Other apps` option) without the need to modify the Bridge.

iamthomasbishop:
@marecar3 The sheet is looking pretty good, but I have one suggestion that just jumped to my mind. We should probably use a Giphy logo from their [design guidelines](https://developers.giphy.com/docs/sdk#design-guidelines). I grabbed one of the logos from their Sketch asset library and made an SVG, which you can [download here](https://cloudup.com/cOYhpSFSo1X). For a preview, it looks like this (32x32, but we will scale down to match the size of our other icons, which IIRC is 24x24):

image

react-native-gutenberg-bridge/android/src/main/java/org/wordpress/mobile/WPAndroidGlue/WPAndroidGlueCode.java
marecar3: Hey @etoledom that was my initial solution but I had a problem with this part of the code: https://github.com/wordpress-mobile/gutenberg-mobile/pull/1469/files#diff-37882e44592184b021430fb0aae5d21cR129-R135

For media stock, we don't have information about the upload process
which means that we need to listen for `MediaSelectedCallback` ( and not  `MediaUploadCallback` )
and for Giphy we have information about the upload process so we need to use `MediaUploadCallback`

Some ugly solution would be to put prefix `upload` in the key so that we know which callback we need to assign to the media source.

Do you have something on your mind in this case?

etoledom: I see, that makes sense.

Just an idea: Do you think it would be possible to use a single progress-based callback (`MediaUploadCallback`?), and send the progress as completed from the beginning in the Stock Photos case?

This might be a simpler workaround 🤔

marecar3: Hey @etoledom I managed to only use one callback for all cases (`MediaUploadCallback`)
which gives us the opportunity to hide 3rd party options from RN layer.

Let me know when you do another iteration of review. Thanks.

marecar3:
> @marecar3 The sheet is looking pretty good, but I have one suggestion that just jumped to my mind. We should probably use a Giphy logo from their [design guidelines](https://developers.giphy.com/docs/sdk#design-guidelines). I grabbed one of the logos from their Sketch asset library and made an SVG, which you can [download here](https://cloudup.com/cOYhpSFSo1X). For a preview, it looks like this (32x32, but we will scale down to match the size of our other icons, which IIRC is 24x24):
> 
> image

Thanks for the suggestion, I think it's a good one but with the current implementation, it's not so easy to set custom icons for Other resources. Let's go with default icons for this iteration and I will create a separate ticket that will bring that feature, WDYT?

iamthomasbishop:
Works for me! This would be an enhancement, not a blocker 👍

react-native-gutenberg-bridge/index.js
```diff
@@ -89,4 +89,12 @@ export function requestImageUploadCancel( mediaId ) {
 	return RNReactNativeGutenbergBridge.requestImageUploadCancel( mediaId );
 }
 
+export function getOtherMediaOptions( filter, callback ) {
+	return RNReactNativeGutenbergBridge.getOtherMediaOptions( filter, callback );
+}
+
+export function requesOtherMediaPickFrom( mediaSource, multiple, callback ) {
```
mchowning: Minor typo here I believe: request is missing the T?

Also, just noting that I think you could shorten these a bit to just be: 

```
export const getOtherMediaOptions = RNReactNativeGutenbergBridge.getOtherMediaOptions;
export const requestOtherMediaPickFrom = RNReactNativeGutenbergBridge.requesOtherMediaPickFrom;
``` 

I personally prefer avoiding explicitly passing the arguments, but I don't think it's wrong the way you have it either.

react-native-gutenberg-bridge/index.js
```diff
@@ -89,4 +89,12 @@ export function requestImageUploadCancel( mediaId ) {
 	return RNReactNativeGutenbergBridge.requestImageUploadCancel( mediaId );
 }
 
+export function getOtherMediaOptions( filter, callback ) {
```
koke: I don't want to block on this to get this feature out, but why are these "other"? Is there something intrinsically different in requesting an image from the device library or pexels?

I understand we might need to pass the available options through the bridge because we want to show them in the existing bottom sheet (although I think a native one is in the works), but other than that, I think the apps should define which options are available more generally.

marecar3: That's a good question @koke. 

Main idea is to provide 3rd party media sources from native applications so that in the future we can just change native apps WPAndroid or WPiOS and everything should work (without changing JS side or bridge).

I agree that it would be easier to control which options are available from one place (JS) and not from two places (WPAndroid and WPiOS), but we can see in the future if this current solution has some big downsides and if's true, we can change approach.

@etoledom can also share his opinion on this one.

koke: > I agree that it would be easier to control which options are available from one place (JS) and not from two places (WPAndroid and WPiOS)

I actually think it makes more sense to define those in the apps, and avoid gutenberg making any assumptions about the available options.

As an example, the demo app still offers device library, camera, and wp media library, but only the device library works as expected. The camera option does nothing in the iOS simulator (it has no camera), and the media library option inserts a sample image/video.

It would be better if gutenberg just asked about the available options to present them in the picker, and then request the selected one. In the demo app, we could have some options like `Choose from device`, `Add sample image`, and `Add sample video`, instead of pretending to support options that won't work.

I'm not sure how much effort this would be, but if we are already requesting the "others" through the bridge, it doesn't feel like it would be a hard thing to request all the options.

marecar3: I missed the point at first but now it's clear.
Yeah, we can do that, it does make sense, then we can easily mock resources for the demo apps also :)

etoledom: > I actually think it makes more sense to define those in the apps, and avoid gutenberg making any assumptions about the available options.

Actually, the only option on gutenberg web (stand alone) is to add an image from a link. The WordPress Media Picker is some kind of "add on" iirc.

etoledom: I'd suggest to implement the bridge with this new options as it is now, and on a second step, migrate the "older" options to be handled on this new way.

Hopefully we keep the testing steps and possible issues lower on this PR this way.

iamthomasbishop:
@marecar3 I had a chance to review this, and overall it feels great. The only thing that I was going to suggest is that we focus the search input right away when launching into the FPL/Giphy search.

marecar3:
Hey @iamthomasbishop, it seems that Aztec is working the same, so maybe we can raise a separate ticket, for a fix which will be presented in both Aztec and Gutenberg?

marecar3:
@etoledom @hypest Giphy is disabled so we now only have Pexels support.

koke:
Tested on media-text and it's working nicely 👌

iamthomasbishop:
>  it seems that Aztec is working the same, so maybe we can raise a separate ticket, for a fix which will be presented in both Aztec and Gutenberg?

Odd, I could've sworn we had this same discussion when building for Aztec and made it select the input first 🤔 I'll create separate issues on WPAndroid and WPiOS today.

Tug:
> > Are you ok with strings which were reused from WPAndroid or you think we should put some new strings for gb-mobile?
> 
> Assuming you mean the text label strings used on the bottom sheet, I think that's fine.

@iamthomasbishop I had a comment about that related to i18n https://github.com/wordpress-mobile/WordPress-Android/pull/10628#discussion_r339527642

iamthomasbishop:
@Tug good catch! Added [a comment](https://github.com/wordpress-mobile/WordPress-Android/pull/10628#issuecomment-546968342) w/ thoughts.

react-native-gutenberg-bridge/android/build.gradle
```diff
@@ -117,6 +117,9 @@ dependencies {
     implementation "org.webkit:android-jsc:r241213"
     implementation project(':react-native-aztec')
 
+    // For animated GIF support
+    implementation 'com.facebook.fresco:animated-gif:2.0.0'
```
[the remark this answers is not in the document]
marecar3: Good question @cameronvoell!
I was thinking to remove it but then I realize that we might need it as the user could already have `gif` in his Wordpress Library and it would be cool that he can see animated `gif` while editing the Post.

react-native-gutenberg-bridge/ios/RNReactNativeGutenbergBridge.swift
```diff
@@ -42,6 +42,16 @@ public class RNReactNativeGutenbergBridge: RCTEventEmitter {
             })
         }
     }
+    
+    @objc
+    func requestOtherMediaPickFrom(_ source: String, allowMultipleSelection: Bool, callback: @escaping RCTResponseSenderBlock) {
+        //TODO implement me
```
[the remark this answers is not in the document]
marecar3: Great questions @cameronvoell !
Yes, that's correct. The idea is to merge JS and Android part and then @etoledom will work on iOS side of it.

react-native-gutenberg-bridge/android/build.gradle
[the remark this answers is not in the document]
cameronvoell: Just tested this, very cool!

PR #1483, opened 2019-10-23
Add simple media library mock for demo app

mkevins:
### Description

This PR adds a very simple mock to the demo app for Android, which returns 1 image if `allowMultipleSelection` is false, and 2 if true. It may be useful for testing some behaviors without requiring a build of the parent app (WordPress-Android).

It also provides for a "fake" uploading media flow, with two pre-selected images, taking 3 and 7 seconds to upload (sending progress update events as they go).

#### To test:

* Open the demo app and insert an image block
* Tap `ADD IMAGE` button
* Tap `WordPress Media Library` in the BottomSheet

#### Expected:

It should insert the familiar mountain image from cldup.com.

---

* Open the demo app and insert an image block
* Tap `ADD IMAGE` button
* Tap `Choose from device` in the BottomSheet

#### Expected:

A pumpkin and a galaxy image should appear to start "uploading", and fail to upload.

#### Additional comments

I've been using this locally to iterate more quickly, and think it might be useful in other scenarios. In my case, I've only used images, but I expect other media types would be easy to add, if necessary.

Update release notes:

(there are none)


hypest:
Any learnings we can pick out of this PR? Can we close otherwise?

mkevins:
We can close this one.. this was something I was using locally which was helpful while integrating uploads for Gallery block. The idea(s) here would need adaptation in the new structure, so probably best to revisit with a new PR.

PR #1489, opened 2019-10-23
Add content alignment options to paragraph block

### LINKED ISSUE #1266: Add content alignment options in paragraphs
Add support for left/center/right alignment of text in paragraph blocks.

Aztec-Android supports it, Aztec-iOS doesn't.

mchowning:
Fixes #1266 . Further details available in the related [gutenberg PR](https://github.com/WordPress/gutenberg/pull/17577). Newer Gutenberg PR that targets the 1.17 release: https://github.com/WordPress/gutenberg/pull/18433

Because this includes an update to the AztecEditor-iOS version, merging this also needs to be coordinated with merging that update into WPiOS: https://github.com/wordpress-mobile/WordPress-iOS/pull/12929

### Release Notes

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -1,19 +1,17 @@
 1.17.0
 ------
-* [iOS] Add support for Preformatted block.
 * Include block title in Unsupported block's UI
 * Show new-block-indicator when no blocks at all and when at the last block
 * Use existing links in the clipboard to prefill url field when inserting new link.
 * Media & Text block alignment options
+* Add alignment controls for paragraph blocks
 * [iOS] Fix issue where the keyboard would not capitalize sentences correctly on some cases.
 * [iOS] Support for Pexels image library
+* [iOS] Add support for Preformatted block.
 
 1.16.1
 ------
 * [iOS] Fix tap on links bug that reappear on iOS 13.2
-1.16.1
-------
-* Add alignment controls for paragraph blocks
 
 1.16.0
 ------
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -10,6 +10,9 @@
 1.16.1
 ------
 * [iOS] Fix tap on links bug that reappear on iOS 13.2
+1.16.1
+------
+* Add alignment controls for paragraph blocks
 
 1.16.0
 ------
--- a/react-native-aztec/ios/Cartfile
+++ b/react-native-aztec/ios/Cartfile
@@ -1,1 +1,1 @@
-github "wordpress-mobile/AztecEditor-iOS" "1.11.0"
+github "wordpress-mobile/AztecEditor-iOS" "1.12.0"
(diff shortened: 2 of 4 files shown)


SergioEstevao:
This is working pretty good but I found two edge cases were it's failing:

1. If you have a paragraph block selected with empty text and the cursor on it, if you change the alignment the place holder label moves but the cursor stays on the old alignment and if you type the old alignment is kept

Simulator Screen Shot - iPhone 11 - 2019-10-24 at 11 44 30

2. When selecting right alignment the placeholder text is cropped, see that the last dot of the ellipsis is not visible on the placeholder
Simulator Screen Shot - iPhone 11 - 2019-10-24 at 11 42 58

mchowning:
Thanks for the review @SergioEstevao !

>1. If you have a paragraph block selected with empty text and the cursor on it, if you change the alignment the place holder label moves but the cursor stays on the old alignment and if you type the old alignment is kept

This doesn't seem to be happening for me. 🤔  I thought it might be a device/os-specific issue, but I've tried on a few different devices (physical iPhone Xr w/iOS 13, and simulators with iOS 12 & 11). What device and os are you using? Any unusual steps to reproduce?

>2. When selecting right alignment the placeholder text is cropped, see that the last dot of the ellipsis is not visible on the placeholder

Good catch! I didn't notice the missing third period, but I see it now that you've pointed it out. I'll look into it.

mchowning:
>2. When selecting right alignment the placeholder text is cropped, see that the last dot of the ellipsis is not visible on the placeholder

I've pushed a fix for this in the [latest commit](https://github.com/wordpress-mobile/gutenberg-mobile/pull/1489/commits/ad222d05bb14646fc82feff5e95c82980152c762). Would be great if @etoledom  could give that commit a look since it is tweaking [some code he wrote](https://github.com/wordpress-mobile/gutenberg-mobile/pull/591) for handling RTL. 

For that reason, any testing of my changes should be sure to check that RTL is handled appropriately (I tested RTL with Run Options -> Application Language: Right-To-Left Pseudolanguage). In particular, keep an eye out that when an empty paragraph block is selected the cursor is appropriately positioned relative to the placeholder text.

SergioEstevao:
@mchowning here is a video of the first bug I mentioned:



AlignBug

mchowning:
Thanks for the video @SergioEstevao !

Here are the steps to reproduce I'm seeing and some notes. I've spent some time digging into this and I am stumped so far. I'm not that familiar with Aztec-iOS, so please let me know if you have any suggestions on what I might want to investigate @SergioEstevao . 

As an alternative, what would you think about opening a new issue for this and merging this as-is since it seems like this probably will not occur very often, and when it does occur it is not too difficult to recover alignment by just updating the alignment once text has been added?

 Step # | Step | Info
--- | --- | ---
1a | (Alternative a) Load a post with a paragraph block that already has text | The text can be either aligned or unaligned. The bug will not occur if you load a post with an empty paragraph block.
1b | (Alternative b) Create a new block, add text to it, and then deselect and reselect that block. | Focus must leave the block in question, if you remain focused on the block the whole time then the bug will not appear.
2 | Delete all the text from the paragraph block | The bug appears regardless of whether you change the alignment before deleting all the text in the block
3 | Update the alignment of the empty block | The block must be empty, if you delete all the text and add any text back, then the bug does not appear.

SergioEstevao:
Remember to update the release notes to say this change is landing on release 1.17.0

Tug:
@mchowning Looks like the gb PR has been merge, is there anything else remaining on this PR, to have it merged?

PR #1494, opened 2019-10-24
Fix requestImport call to send the correct arguments.

SergioEstevao:
Beside fixing the request for import to send the correct parameters in
the callback, we are also improving the interface by replacing a tupple
for a structure to make parameters more explicit.

This can also be tested on the main app using this PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/12773

To test:
 - Start the demo app
 - On the image blocks check that the different sources work correctly
 - Try to copy one image from the photos app and paste inside a text block, check that an image block is added and you see an image upload simulation happening.

Update release notes:

### DIFF
--- a/react-native-gutenberg-bridge/ios/GutenbergBridgeDelegate.swift
+++ b/react-native-gutenberg-bridge/ios/GutenbergBridgeDelegate.swift
@@ -1,7 +1,7 @@
 public struct MediaInfo {
-    let id: Int32?
-    let url: String?
-    let type: String?
+    public let id: Int32?
+    public let url: String?
+    public let type: String?
 
     public init(id: Int32?, url: String?, type: String?) {
         self.id = id
--- a/react-native-gutenberg-bridge/ios/GutenbergBridgeDelegate.swift
+++ b/react-native-gutenberg-bridge/ios/GutenbergBridgeDelegate.swift
@@ -1,4 +1,18 @@
-public typealias MediaPickerDidPickMediaCallback = (_ media: [(Int32?,String?,String?)]?) -> Void
+public struct MediaInfo {
+    let id: Int32?
+    let url: String?
+    let type: String?
+
+    public init(id: Int32?, url: String?, type: String?) {
+        self.id = id
+        self.url = url
+        self.type = type
+    }
+}
+
+public typealias MediaPickerDidPickMediaCallback = (_ media: [MediaInfo]?) -> Void
+
+public typealias MediaImportCallback = (_ media: MediaInfo?) -> Void
 
 public enum MediaPickerSource: String {
     case mediaLibrary = "SITE_MEDIA_LIBRARY"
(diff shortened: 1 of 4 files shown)


react-native-gutenberg-bridge/ios/RNReactNativeGutenbergBridge.swift
```diff
@@ -27,17 +27,22 @@ public class RNReactNativeGutenbergBridge: RCTEventEmitter {
...
+
+                let jsFormattedMedia = mediaToReturn.map { mediaInfo in
+                    return [mediaDictKeys.IDKey: mediaInfo.id as Any,
+                            mediaDictKeys.URLKey: mediaInfo.url as Any,
+                            mediaDictKeys.TypeKey: mediaInfo.type as Any]
+                }
+                if allowMultipleSelection {
+                    callback([jsFormattedMedia])
+                } else {
+                    callback(jsFormattedMedia)
```
etoledom: It might be good to merge this and send the full array. If multi select is false, the callback should bring just one media item. Maybe we can enforce that via the callback signature, and keep the bridge simpler 🤔 

Again, thinking about the future!

SergioEstevao: We don't have control on the callback interface it's defined by RN library.
It's always expecting an array where each item is an argument of the JS function.
What we can do is to cleanup the callback functions on the other side to always expect an array of media.

react-native-gutenberg-bridge/ios/GutenbergBridgeDelegate.swift
```diff
@@ -1,4 +1,18 @@
...
+    public let type: String?
+
+    public init(id: Int32?, url: String?, type: String?) {
+        self.id = id
+        self.url = url
+        self.type = type
+    }
+}
+
+public typealias MediaPickerDidPickMediaCallback = (_ media: [MediaInfo]?) -> Void
```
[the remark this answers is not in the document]
SergioEstevao: I thought that too, but we will need to check the code on the JS side and see what are the implications of sending nil or not.

Can we do this in the follow up PR?

react-native-gutenberg-bridge/ios/RNReactNativeGutenbergBridge.swift
```diff
@@ -27,17 +27,22 @@ public class RNReactNativeGutenbergBridge: RCTEventEmitter {
...
-                    }
-                    callback([[mediaDictKeys.IDKey: mediaID, mediaDictKeys.URLKey: mediaURL, mediaDictKeys.TypeKey: mediaType]])
+                    mediaToReturn = Array(media.prefix(1))
+                }
+
+                let jsFormattedMedia = mediaToReturn.map { mediaInfo in
+                    return [mediaDictKeys.IDKey: mediaInfo.id as Any,
+                            mediaDictKeys.URLKey: mediaInfo.url as Any,
+                            mediaDictKeys.TypeKey: mediaInfo.type as Any]
+                }
```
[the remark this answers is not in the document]
SergioEstevao: I followed your suggestion and added a method to do the conversion.

SergioEstevao:
@etoledom ready for another look.

react-native-gutenberg-bridge/ios/RNReactNativeGutenbergBridge.swift
etoledom: Let's do that! On another PR someday in the future 😬
It might be good to create a ticket with all these little cleanup tasks to keep track of them.

Issue #872, opened 2019-04-15
JSApplicationIllegalArgumentException in RCTAztecView


jtreanor:
This crash was first seen in 12.1 of WPAndroid. It has happened 48 times so far:

```
Fatal Exception: com.facebook.react.bridge.JSApplicationIllegalArgumentException: Error while updating property 'text' of a view managed by: RCTAztecView
       at com.facebook.react.uimanager.ViewManagersPropertyCache$PropSetter.updateViewProp(ViewManagersPropertyCache.java:95)
       at com.facebook.react.uimanager.ViewManagerPropertyUpdater$FallbackViewManagerSetter.setProperty(ViewManagerPropertyUpdater.java:132)
       at com.facebook.react.uimanager.ViewManagerPropertyUpdater.updateProps(ViewManagerPropertyUpdater.java:51)
       at com.facebook.react.uimanager.ViewManager.updateProperties(ViewManager.java:37)
       at com.facebook.react.uimanager.NativeViewHierarchyManager.updateProperties(NativeViewHierarchyManager.java:136)
       at com.facebook.react.uimanager.UIViewOperationQueue$UpdatePropertiesOperation.execute(UIViewOperationQueue.java:95)
       at com.facebook.react.uimanager.UIViewOperationQueue$1.run(UIViewOperationQueue.java:917)
       at com.facebook.react.uimanager.UIViewOperationQueue.flushPendingBatches(UIViewOperationQueue.java:1025)
       at com.facebook.react.uimanager.UIViewOperationQueue.access$2600(UIViewOperationQueue.java:46)
       at com.facebook.react.uimanager.UIViewOperationQueue$DispatchUIFrameCallback.doFrameGuarded(UIViewOperationQueue.java:1085)
       at com.facebook.react.uimanager.GuardedFrameCallback.doFrame(GuardedFrameCallback.java:29)
       at com.facebook.react.modules.core.ReactChoreographer$ReactChoreographerDispatcher.doFrame(ReactChoreographer.java:166)
       at com.facebook.react.modules.core.ChoreographerCompat$FrameCallback$1.doFrame(ChoreographerCompat.java:84)
       at android.view.Choreographer$CallbackRecord.run(Choreographer.java:874)
       at android.view.Choreographer.doCallbacks(Choreographer.java:688)
       at android.view.Choreographer.doFrame(Choreographer.java:620)
[...]

PR #1507, opened 2019-10-28
Fix Aztec Android "SetSpan" ends beyond length on merging blocks

daniloercoli:
This PR does fix a sub-case of #872 and wordpress-mobile/WordPress-Android#9832 where the editor crashes on merging 2 blocks when the first block does have BR tag(s) at the end of its content.

GB side PR: https://github.com/WordPress/gutenberg/pull/18138

To test:
**Note, we should test this PR with tons of input since the changes in Aztec may break other part of the GB-Editor.**
- Create a post on Gutenberg web with a para block that ends with one or more empty line(s) (Shift+Enter)
Screenshot 2019-10-23 at 19 55 08
- Open this new post in GB mobile
- Tap on the `Plus` symbol to add a new block below the para
- Put the focus into this new block and taps on Backspace to merge this new block with the one above.
- It should not crash, and the 2 blocks merged, with the caret at the end of the block

Also try many other testing scenarios.

cc @SergioEstevao 

Update release notes:

- [ x] If there are user facing changes, I have added an item to `RELEASE-NOTES.txt`.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,32 @@
+---
+name: Bug report
+about: Create a report to help us improve
+title: ''
+labels: ''
+assignees: ''
+
+---
+
+**Describe the bug**
+A clear and concise description of what the bug is.
+
+**To Reproduce**
+Steps to reproduce the behavior:
+1. Go to '...'
+2. Click on '....'
+3. Scroll down to '....'
+4. See error
+
+**Expected behavior**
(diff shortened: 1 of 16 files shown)


SergioEstevao:
While this PR solves some problems after this was merge I'm seeing the following issues:
 - On list, quote blocks and image captions write some text, add one single space to the end and press Enter. See the new line being inserted but the cursor staying up.
 - On the new preformatted list blocks this is even worst because inside the <pre> elements space are to be kept and this PR removes any extra space or empty lines that are inserted so when trying to insert a new line inside a pre we get a red screen.

Issue #939, opened 2019-04-29
Fix our i18n process


Tug:
Internationalization is broken since Gutenberg moved away from [`babel-plugin-makepot`](https://github.com/WordPress/gutenberg/tree/master/packages/babel-plugin-makepot). 

Indeed, build files for each packages in gutenberg is now used as input for glotpress to generate pot files. This means our assumption that mobile native strings in the gutenberg repository will be picked up automatically was wrong.

We need to fix this and either:
1. have gutenberg include our strings 
2. include those in our `gutenberg-mobile.pot`.

Solution 2. is probably more realistic as it will induce less friction with the WordPress core and Gutenberg projects but will bring us one step back in terms of automating translations for gutenberg-mobile.

See related discussions:
https://make.wordpress.org/core/2018/11/09/new-javascript-i18n-support-in-wordpress/
https://make.wordpress.org/core/2018/05/01/javascript-internationalization-the-missing-pieces/
https://wordpress.slack.com/archives/C04KRK0KA/p1525791742000720
https://meta.trac.wordpress.org/ticket/3876

swissspidy:
Let me know if I can help with the i18n part in Gutenberg, WordPress core, or WordPress.org.

hypest:
Not sure if you're actively working on this one @Tug , if so, better to move the ticket to the "In Progress" column in the project board, otherwise let's un-assign the ticket. Thanks!

Tug:
👍 Unassigned

PR #1520, opened 2019-10-30
Generate i18n files for WPAndroid and WPiOS to use

Tug:
Fixes https://github.com/wordpress-mobile/gutenberg-mobile/issues/939

This PR adds 2 scripts to generate translation files for Wordpress-Android (as a `strings.xml` file) and WordPress-iOS (as a .swift file containing static strings). Those files will be generated and committed in the `bundle` folder as:
- `bundle/android/strings.xml` for android
- `bundle/ios/GutenbergNativeTranslations.swift` for iOS

This process will require a script on each app repository so those 2 files are read and included in the glotpress input file and thus the gutenberg mobile translations will be available in glotpress: 
- For Android: https://github.com/wordpress-mobile/WordPress-Android/pull/10719
- For iOS: https://github.com/wordpress-mobile/WordPress-iOS/pull/12811

Note that only the strings that are in `.(native|ios|android).js` files are considered and those that are already part of gutenberg glotpress project (web) are excluded as well.


hypest:
👋

I know this is still WIP so, feel free to ignore this question: `gutenberg-native.strings` seems largely identical to the stings.xml so, I wonder if one or the other will get removed from the PR in favor of a single file.

Tug:
Hi @hypest, thanks for having a look!
Indeed `gutenberg-native.strings` is a file that needed to be removed. It's done now :)

bundle/ios/GutenbergNativeTranslations.swift
```diff
@@ -0,0 +1,81 @@
...
+let string72 = NSLocalizedString("Link text", comment: "")
+let string73 = NSLocalizedString("Add link text", comment: "")
+let string74 = NSLocalizedString("Translate", comment: "")
+let string75 = NSLocalizedString("hello", comment: "")
+let string76 = NSLocalizedString("%d banana", comment: "")
+let string76Plural = NSLocalizedString("%d bananas", comment: "")
+let string77 = NSLocalizedString("hello %s", comment: "")
+let string78 = NSLocalizedString("cheeseburger", comment: "")
+let string79 = NSLocalizedString("%d cat", comment: "")
+let string79Plural = NSLocalizedString("%d cats", comment: "")
```
Tug: Is there a way to have those grouped so glotpress knows `string79Plural` is the plural version of `string79`? Not sure who to ping to answer this 🤔 Any idea @hypest ?

bin/po2android.js
```diff
@@ -0,0 +1,72 @@
...
+	} ).filter( Boolean );
+	return `<?xml version="1.0" encoding="utf-8"?>\n<resources xmlns:tools="http://schemas.android.com/tools">\n${ androidResources.join( '' ) }</resources>\n`;
+}
+
+if ( require.main === module ) {
+	if ( process.stdin.isTTY ) {
+		const potFileName = process.argv[ 2 ];
+		const destination = process.argv[ 3 ];
+		const potFileContent = fs.readFileSync( potFileName );
+		const swiftOutput = po2Android( potFileContent, process.argv[ 3 ] );
```
mchowning: Nitpick, but perhaps "xmlOutput" here instead. 😄

package.json
```diff
@@ -68,7 +68,7 @@
     "prebundle": "yarn i18n-cache:force",
     "prebundle:android": "yarn patch-metro-no-file-watch",
     "postbundle:android": "yarn un-patch-metro-no-file-watch",
-    "bundle": "yarn bundle:android && yarn bundle:ios",
+    "bundle": "yarn bundle:android && yarn bundle:ios && yarn clean:pot && yarn genstrings",
```
mchowning: If there is no "gutenberg*.pot` file, then `yarn clean:pot` fails and prevents `yarn genstrings` from executing.

mchowning:
With these changes running `yarn bundle` will build gutenberg, which runs npm install inside gutenberg, and the the node_modules directory is generated inside gutenberg and remains after this script is run. For me locally, having the node_modules directory inside gutenberg causes the mobile gutenberg tests (`yarn test`) to fail. Maybe this is something unique to my local setup, but if it's not, we might want to update this script to clean up after itself so that someone doesn't run `yarn bundle` and then later not realize why all their tests are failing.

Of course, this probably won't be an issue once we move to the monorepo.

package.json
Tug: Weird it's not failing for me when there is no pot file, I think this might be because of the wildcard?
I can add `-f` if it fixes it for you 👍

Tug:
Thank you for reviewing @mchowning 
I addressed the issues you mentioned, could you have another look :)?

bundle/ios/GutenbergNativeTranslations.swift
koke: We don't have support for plurals yet 😞 https://github.com/wordpress-mobile/WordPress-iOS/issues/6327

bin/po2swift.js
```diff
@@ -0,0 +1,48 @@
...
+}
+
+if ( require.main === module ) {
+	if ( process.stdin.isTTY ) {
+		const potFileName = process.argv[ 2 ];
+		const destination = process.argv[ 3 ];
+		const potFileContent = fs.readFileSync( potFileName );
+		const swiftOutput = po2Swift( potFileContent, process.argv[ 3 ] );
+		fs.writeFileSync( destination, swiftOutput );
+	} else {
```
koke: Can you explain what the non-TTY variant is for? I'm not that familiar with node on the cli, and I don't understand why there are two variants

package.json
koke: Maybe worth moving this to `prebundle` with the rest of the i18n work

bin/po2swift.js
Tug: This is so the script can work with pipes: `cat gutenberg.pot | node bin/po2swift.js`.
I started with this approach because it was easier to test, and I wasn't sure which one I was going to use in package.json. It's not needed anymore so we could remove it but I figured it would be a nice addition

bundle/ios/GutenbergNativeTranslations.swift
Tug: I see. So I guess it's not worth sending those as extra strings for translations as they won't be picked up by glotpress as the plural version of other strings 🤔 Yeah in that case we should probably remove those

Tug:
@koke thanks for reviewing :)
Indeed they won't work. I'll remove the plurals for iOS and check the ci failing test on android 👍

PR #1535, opened 2019-11-04
Signals the need to relayout view after HTML change.

### LINKED ISSUE #1534: Multiline blocks/richtext are not updating size when pressing Enter
**Describe the bug**

When editing on a richtext component that is multiline (list block, quote block, image captions) if you press enter the content is updated but the layout/size of the block insn't. Only when you write new text the layout is updated.

**To Reproduce**
1. Open the demo the app
2. Add a list block
3. Write some test
4. Press Enter
5. See the error

**Expected behavior**

The block should extend size and relayout.

**Screenshots**

 - Device: iPhoneX
 - OS: iOS 13.2
 - Version any version running on iOS 13.2 (1.15 and 1.16)

SergioEstevao:
Fixes #1534 

To test:
 - Start the demo app
 - Add a new list
 - Check that when adding new items to the list the block sizes correctly
 - Do the same checks with quotes and image captions

Update release notes:

### DIFF
--- a/react-native-aztec/ios/RNTAztecView/RCTAztecView.swift
+++ b/react-native-aztec/ios/RNTAztecView/RCTAztecView.swift
@@ -403,7 +403,7 @@ class RCTAztecView: Aztec.TextView {
             let end = selection["end"]  as? NSNumber {
             setSelection(start: start, end: end)
         }
-        // This signal the RN/JS system that the component needs to the relayout
+        // This signal the RN/JS system that the component needs to relayout
         setNeedsLayout()
     }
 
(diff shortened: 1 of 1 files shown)


react-native-aztec/ios/RNTAztecView/RCTAztecView.swift
```diff
@@ -403,6 +403,8 @@ class RCTAztecView: Aztec.TextView {
             let end = selection["end"]  as? NSNumber {
             setSelection(start: start, end: end)
         }
+        // This signal the RN/JS system that the component needs to the relayout
```
etoledom: Small typo (I think)
```suggestion
        // This signals the RN/JS system that the component needs to relayout
```

PR #1536, opened 2019-11-04
Prepare hotfix release 1.15.3

SergioEstevao:
This a PR to prepare the release 1.15.3

This release only changes the iOS bridge code to fix a bug that arise in release 13.2 of iOS and it uses the same web gutenberg code than release 1.15.2

To test:
 - Smoke test the demo app and the test the related WPiOS PR.

Update release notes:


etoledom:
I think this should be targeting `master`, right?

PR #1537, opened 2019-11-04
Update GB reference to support copy of url in clipboard.

### LINKED ISSUE #680: Proposal: Autofill link URL field with link in clipboard
I'd like to propose that when adding a link, Gutenberg autofill the link URL field with the clipboard contents when the clipboard contains a URL. (This is a feature of Aztec that I really appreciate.)

The user flow is to open the browser or other app, copy a URL, and insert it as a link in the editor. Autofilling the link URL field means the user does not have to deal with long tap and paste actions to insert the URL — it "magically" appears in the link sheet when they create the link.

SergioEstevao:
Fixes #680 

GB PR: https://github.com/WordPress/gutenberg/pull/18279

To test:
 - Start the demo app
 - Switch to Safari and copy a link from the url bar
 - Switch to GB demo app
 - Select a text block
 - Select a bit of text
 - Tap on the link button
 - Check that the URL that you copied from Safari is added to the URL field in the link modal
 - Edit an already existing link and check that the link in the clipboard does not override the one already existing.

Update release notes:

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -1,3 +1,7 @@
+1.17.0
+------
+* Use existing links in the clipboard to prefill url field when inserting new link.
+
 1.16.0
 ------
 * [Android] Add support for pexels images


SergioEstevao:
@hypest PR update with release notes, do you mind give it another look.

Issue #1265, opened 2019-08-06
Choose image from "Other Apps" on iOS


hypest:
To choose an image in the Image block, let's also support choosing from "Other Apps" on iOS, like Aztec does.

etoledom:
Closed via https://github.com/wordpress-mobile/WordPress-iOS/pull/12883

PR #1539, opened 2019-11-05
[iOS] Implement client-defined media options.

etoledom:
Fixes https://github.com/wordpress-mobile/gutenberg-mobile/issues/720

This PR implements the iOS Bridge code needed for Stock Photos (Pexels) and https://github.com/wordpress-mobile/gutenberg-mobile/issues/1265

`WPiOS` PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/12858

pexels

**Note:**
- The "Free Photo Library" is the same as used on Aztec.
- I decided to make the Free Photos Library multi-selectable, even though we haven't activated this feature on iOS yet. I can keep it single selectable if we decide that this is better.

- There are some few small changes on the bridge media picker/upload methods, so it would be good to test all options on all media related blocks. (cc @pinarol)

- I noticed that we could improve some things in the JS side, but opted to keep this PR focused on adopting the current implementation to avoid the need of Android changes.

**To test:**
- Test on WPiOS running: https://github.com/wordpress-mobile/WordPress-iOS/pull/12858

Image Block:
- Add an Image block.
- Check that the "Free Photo Library" option is there.
- Open the Free Photo Library.
- Choose one (or many) images.
- Check that all images selected are added.
- Check that you can select only images in all other options.

Video Block:
- Add a video block.
- Check that "Free Photo Library" is NOT in the options.
- Check that you can only select videos in all other options.

Media&Text:
- Add an Media&Text block.
- Check that the "Free Photo Library" option is there.
- Open the Free Photo Library.
- Choose one (or many) images. (Multi-select is also possible on Android)
- Check that all images selected are added. (Extra images will be added as Image Block)
- Check that you can select images and videos in all other options.

cc @marecar3 

Update release notes:

### DIFF
--- a/react-native-gutenberg-bridge/ios/GutenbergBridgeDelegate.swift
+++ b/react-native-gutenberg-bridge/ios/GutenbergBridgeDelegate.swift
@@ -17,9 +17,9 @@ public typealias MediaImportCallback = (_ media: MediaInfo?) -> Void
 /// Label and Type are not relevant since they are delcared on the JS side.
 /// Hopefully soon, this will need to be declared on the client side.
 extension Gutenberg.MediaSource {
-    public static let mediaLibrary = Gutenberg.MediaSource(id: "SITE_MEDIA_LIBRARY", label: "", type: .image)
-    public static let deviceLibrary = Gutenberg.MediaSource(id: "DEVICE_MEDIA_LIBRARY", label: "", type: .image)
-    public static let deviceCamera = Gutenberg.MediaSource(id: "DEVICE_CAMERA", label: "", type: .image)
+    public static let mediaLibrary = Gutenberg.MediaSource(id: "SITE_MEDIA_LIBRARY", label: "", types: [.image, .video])
+    public static let deviceLibrary = Gutenberg.MediaSource(id: "DEVICE_MEDIA_LIBRARY", label: "", types: [.image, .video])
+    public static let deviceCamera = Gutenberg.MediaSource(id: "DEVICE_CAMERA", label: "", types: [.image, .video])
 
     static var registeredInternalSources: [Gutenberg.MediaSource] {
         return [
@@ -28,13 +28,6 @@ extension Gutenberg.MediaSource {
             .mediaLibrary,
         ]
     }
-}
-
-public enum MediaFilter: String {
-    case image
-    case video
-    case audio
-    case other
 }
 
 /// Ref. https://github.com/facebook/react-native/blob/master/Libraries/polyfills/console.js#L376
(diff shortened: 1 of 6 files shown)


react-native-gutenberg-bridge/ios/GutenbergBridgeDelegate.swift
```diff
@@ -11,20 +11,23 @@ public struct MediaInfo {
...
-    case mediaLibrary = "SITE_MEDIA_LIBRARY"
-    case deviceLibrary = "DEVICE_MEDIA_LIBRARY"
-    case deviceCamera = "DEVICE_CAMERA"
-}
-
-public enum MediaFilter: String {
-    case image
-    case video
-    case audio
-    case other
```
etoledom: Merged `MediaType` with `MediaFilter` since we filters by `Types` of media. It seemed like redundant.

react-native-gutenberg-bridge/ios/GutenbergBridgeDelegate.swift
```diff
@@ -77,7 +80,7 @@ public protocol GutenbergBridgeDelegate: class {
     ///     - source: the source from where the picker will get the media
     ///     - callback: A callback block to be called with an array of upload mediaIdentifiers and a placeholder images file url, use nil on both parameters to signal that the action was canceled.
     ///
-    func gutenbergDidRequestMedia(from source: MediaPickerSource, filter: [MediaFilter]?, allowMultipleSelection: Bool, with callback: @escaping MediaPickerDidPickMediaCallback)
+    func gutenbergDidRequestMedia(from source: Gutenberg.MediaSource, filter: [Gutenberg.MediaType], allowMultipleSelection: Bool, with callback: @escaping MediaPickerDidPickMediaCallback)
```
etoledom: Made `filter` non-optional, since a `nil` and `empty` both means no filters (use all types).

react-native-gutenberg-bridge/ios/RNReactNativeGutenbergBridge.swift
```diff
@@ -45,15 +43,25 @@ public class RNReactNativeGutenbergBridge: RCTEventEmitter {
             })
         }
     }
-    
+
     @objc
     func requestOtherMediaPickFrom(_ source: String, allowMultipleSelection: Bool, callback: @escaping RCTResponseSenderBlock) {
-        //TODO implement me
+        requestMediaPickFrom(source, filter: nil, allowMultipleSelection: allowMultipleSelection, callback: callback)
     }
```
etoledom: This event (`requestOtherMediaPickFrom`) is probably not needed? We could reuse `requestMediaPickFrom`.

etoledom:
@marecar3 - I just found a small issue on the Android implementation.

We should not show the Free Media Library option on Self Hosted site, since this is a WPCom/Jetpack-only feature.

I tested on a Self Hosted site and the images fail to upload, but curiously, the images show.

etoledom:
cc @iamthomasbishop

marecar3:
> @marecar3 - I just found a small issue on the Android implementation.
> 
> We should not show the Free Media Library option on Self Hosted site, since this is a WPCom/Jetpack-only feature.
> 
> I tested on a Self Hosted site and the images fail to upload, but curiously, the images show.

Here is the quick fix PR for that case: https://github.com/wordpress-mobile/WordPress-Android/pull/10743

iamthomasbishop:
@etoledom great catch! and thanks for the quick fix @marecar3! 😄

react-native-gutenberg-bridge/ios/RNReactNativeGutenbergBridge.swift
etoledom: Made this mentioned change here: https://github.com/WordPress/gutenberg/pull/18303

etoledom:
Thank you!

PR #1548, opened 2019-11-06
Update to latest version of GB master.

SergioEstevao:
Update develop branch to have the latest version of GB master.

Includes #1507 too, for which the GB ref bump was missed somehow.
Also includes https://github.com/WordPress/gutenberg/pull/17937, with which we hope we address https://github.com/wordpress-mobile/WordPress-iOS/issues/11950.

To test:
 - Just smoke test the demo app to see if all is running.

Update release notes:


hypest:
👋 @SergioEstevao , do you perhaps have a scenario that we can trigger https://github.com/wordpress-mobile/WordPress-iOS/issues/11950#issuecomment-541019293? It'd be good to test it while in this PR since it includes the GB side fix (https://github.com/WordPress/gutenberg/pull/17937).

SergioEstevao:
So the one I detected was in my preformatted PR where a block used a RichText component and didn't set the 'identifier' prop to a correct value. When you tried to merge a block that didn't have the prop set you would get an exception. 

But if all blocks have the identifier set properly I don't think you will be able to reproduce any longer.
Unless you remove the identifier in one of the blocks explicitly, for example in the paragraph block.

PR #1552, opened 2019-11-06
Prevents selection after ZWSP character

### LINKED ISSUE #1530: Media & Text keyboard is set to lowercase
When you create a new post and insert a media-text block, the keyboard is set to lowercase instead of capitalizing the text.

IMG_43EC293B38BB-1

_Originally posted by @koke in https://github.com/wordpress-mobile/WordPress-iOS/pull/12783#issuecomment-547483495_

koke:
The original issue was that when tapping on an empty RichText, the selection was going to the end of the text by default. As we use the ZWSP character as a placeholder, the selection was after that character, and the keyboard appeared in lowercase mode.

I have a fix that corrects this, but I'm not 100% sure if this is the best place to fix it.

Fixes #1530

To test:

Some scenarios to reproduce the issue:

- Insert a media-text block and tap on the empty paragraph
- On a blank new post, tap on the placeholder, then tap again
- On a blank new post, dismiss the keyboard, then tap on the post title

Update release notes:


react-native-aztec/ios/RNTAztecView/RCTAztecView.swift
```diff
@@ -603,6 +603,15 @@ class RCTAztecView: Aztec.TextView {
         let caretData = packCaretDataForRN()
         onSelectionChange(caretData)
     }
+
+    // MARK: - Selection
+    private func correctSelectionAfterZWSP() {
+        guard selectedTextRange?.start == endOfDocument,
+            text == String(.zeroWidthSpace) else {
```
SergioEstevao: Do you think we could associate the check with this configuration in common init: https://github.com/wordpress-mobile/gutenberg-mobile/pull/1552/files#diff-ad68d1bf3bf68556db4be6e6215573adL136

Just in case the character changes in the future?

koke: Good catch, pushed a change to make it more generic

PR #1553, opened 2019-11-06
Release 1.16.1

SergioEstevao:
Prepares release 1.16.1

WP-iOS PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/12897
WP-Android: https://github.com/wordpress-mobile/WordPress-Android/pull/10750
WP-GB PR: Not needed because there are no gutenberg side changes.

To test:
 - Smoke test all functionality for a new release
 - Pay special attention to the multiline text blocks and see if the update their size when empty line are inserted or deleted.

Update release notes:

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -1,3 +1,8 @@
+1.16.0
+------
+* [Android] Add support for pexels images
+* Add left, center, and right image alignment controls
+
 1.15.2
 ------
 * Fix issue when copy/paste photos from other apps, was not inserting an image on the post.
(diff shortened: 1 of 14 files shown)


mchowning:
Looks like we may need to update these bundles again since the merge from master @SergioEstevao ?

SergioEstevao:
@mchowning the bundles don't need update because after the merge we are still keeping the JS side of 1.6.0 and that was already generated on the PR.

etoledom:
> the bundles don't need update because after the merge we are still keeping the JS side of 1.6.0 and that was already generated on the PR.

I also did a local bundle update and it was clean 👍

PR #1560, opened 2019-11-08
Remove extra padding on text blocks in Android 

### LINKED ISSUE #992: Remove extra inner padding on text blocks
Since we last solved these issues looks like there might have been a regression in one of the last releases. There is some extra padding applied inside the body of text blocks (paragraph, heading, etc). It's slightly different between platforms:

## iOS

There appears to be a slight bit of additional **left/right** padding. It's hard to tell if there's also extra padding on top/bottom edges of the blocks, so we should double check. Same thing applied to the title (only on iOS).

57C57374-7663-41BD-A8BE-C42B558C371A

## Android

It looks like extra **top/bottom** padding, but left/right looks aligned properly.

(Note: title block looks to have proper padding.)

681D53FD-67D9-4486-AC30-6F1AD68F289F
060C26AF-EDB5-4240-80BF-26C73DBED080

## Note

All blocks have inner padding of **12 top/bottom, 16 left/right**, so the total vertical spacing between blocks is always 24.

maxme:
Fixes #992: Remove extra padding on text blocks.

[Related gutenberg PR](https://github.com/WordPress/gutenberg/pull/18759)

⚠️ In testing this we need make sure and run `yarn start:reset`. Otherwise the metro bundler will not reflect the css changes in this PR.

Update release notes:

### DIFF
--- a/react-native-aztec/android/src/main/java/org/wordpress/mobile/ReactNativeAztec/ReactAztecTextShadowNode.java
+++ b/react-native-aztec/android/src/main/java/org/wordpress/mobile/ReactNativeAztec/ReactAztecTextShadowNode.java
@@ -10,6 +10,13 @@ public class ReactAztecTextShadowNode extends ReactTextInputShadowNode {
     private @Nullable ReadableMap mTextMap = null;
     private @Nullable Integer mColor = null;
 
+    @Override
+    public void setDefaultPadding(int spacingType, float padding) {
+        // Do nothing in order to avoid adding unwanted default padding to views
+        // (https://github.com/wordpress-mobile/gutenberg-mobile/issues/992) due to
+        // https://github.com/facebook/react-native/blob/6ebd3b046e5b71130281f1a7dbe7220eff95d74a/ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactTextInputShadowNode.java#L70
+    }
+
     @ReactProp(name = PROP_TEXT)
     public void setText(@Nullable ReadableMap inputMap) {
         mTextMap = inputMap;
--- a/react-native-aztec/android/src/main/java/org/wordpress/mobile/ReactNativeAztec/ReactAztecTextShadowNode.java
+++ b/react-native-aztec/android/src/main/java/org/wordpress/mobile/ReactNativeAztec/ReactAztecTextShadowNode.java
@@ -1,17 +1,18 @@
 package org.wordpress.mobile.ReactNativeAztec;
 
+import android.widget.EditText;
+
 import androidx.annotation.Nullable;
 
 import com.facebook.react.bridge.ReadableMap;
+import com.facebook.react.uimanager.ThemedReactContext;
 import com.facebook.react.uimanager.annotations.ReactProp;
 
 public class ReactAztecTextShadowNode extends ReactTextInputShadowNodeFork {
 
     @Override
-    public void setDefaultPadding(int spacingType, float padding) {
-        // Do nothing in order to avoid adding unwanted default padding to views
-        // (https://github.com/wordpress-mobile/gutenberg-mobile/issues/992) due to
-        // https://github.com/facebook/react-native/blob/6ebd3b046e5b71130281f1a7dbe7220eff95d74a/ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactTextInputShadowNode.java#L70
+    protected EditText createDummyEditText(ThemedReactContext themedContext) {
+        return new EditText(themedContext, null, 0);
     }
 
     @ReactProp(name = PROP_TEXT)
--- a/react-native-aztec/android/src/main/java/org/wordpress/mobile/ReactNativeAztec/ReactAztecTextShadowNode.java
+++ b/react-native-aztec/android/src/main/java/org/wordpress/mobile/ReactNativeAztec/ReactAztecTextShadowNode.java
@@ -4,9 +4,8 @@ import androidx.annotation.Nullable;
 
 import com.facebook.react.bridge.ReadableMap;
 import com.facebook.react.uimanager.annotations.ReactProp;
-import com.facebook.react.views.textinput.ReactTextInputShadowNode;
 
-public class ReactAztecTextShadowNode extends ReactTextInputShadowNode {
+public class ReactAztecTextShadowNode extends ReactTextInputShadowNodeFork {
 
     @Override
     public void setDefaultPadding(int spacingType, float padding) {
--- a/react-native-aztec/android/src/main/java/org/wordpress/mobile/ReactNativeAztec/ReactAztecTextShadowNode.java
+++ b/react-native-aztec/android/src/main/java/org/wordpress/mobile/ReactNativeAztec/ReactAztecTextShadowNode.java
@@ -7,8 +7,6 @@ import com.facebook.react.uimanager.annotations.ReactProp;
 import com.facebook.react.views.textinput.ReactTextInputShadowNode;
 
 public class ReactAztecTextShadowNode extends ReactTextInputShadowNode {
-    private @Nullable ReadableMap mTextMap = null;
-    private @Nullable Integer mColor = null;
 
     @Override
     public void setDefaultPadding(int spacingType, float padding) {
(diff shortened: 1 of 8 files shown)


SergioEstevao:
Simulator Screen Shot - iPhone 11 Pro Max - 2019-11-19 at 11 58 48

I updated the margin/paddings on iOS too.

maxme:
Note: setting `heading_vertical_padding` to `0dp` in f972a75  do the following change:

default heading_vertical_padding | heading_vertical_padding == 0
-|-
Screenshot 2019-11-20 at 16 41 39 | Screenshot 2019-11-20 at 16 41 45

maxme:
Note we need to release a new Aztec version before merging that one.

mchowning:
:wave: @maxme  I added two commits. The first one (c9050f6) removes unnecessary padding and the second one (7d2f7b7) is just cleaning up some css variables that we don't need anymore (so it is not strictly necessary). Cleaning up the css variables did require changes on the gutenberg side ([gutenberg side PR](https://github.com/WordPress/gutenberg/pull/18759)).

It appears that the reason we were still seeing excess padding is because that padding is applied due to the default background that is applied to `EditText` components.  This background varies significantly  between different devices, so that is why we were seeing different results on different devices (it is also the reason you needed to add the `setPadding(0, 0, 0, 0) code). While on that subject, it is worth noting that even with your changes I was still seeing extra padding on both paragraphs and headings (see my "Before" screenshot which was before my changes, but included your changes).

In addition, React Native was internally using this padding by setting a `defaultPadding` value that was equal to the padding of a new `EditText` view (which RN was internally creating and storing inside a `ShadowNode`). This `defaultPadding` was then being applied to the view when the `ShadowNode` was being initialized. Interestingly, it seems that if _either_ (1) the explicit setting of 0 padding on the view with `aztecText.setPadding(0, 0, 0, 0)` or (2) the disabling of the default padding results in the view getting padding.

The reason I also removed a few css variables is because previously we had these set differently on the two platforms in order to accommodate for the extra padding that was coming from Aztec views.

Before | After
--- | ---
Screen Shot 2019-11-26 at 12 22 46 PM | Screen Shot 2019-11-26 at 12 21 12 PM

mchowning:
⚠️ Because this involves css changes, in testing this we need to run `yarn start:reset` otherwise the metro bundler will not reflect the new css (in my experience at least).

react-native-aztec/android/src/main/java/org/wordpress/mobile/ReactNativeAztec/ReactAztecTextShadowNode.java
```diff
@@ -10,6 +10,13 @@
     private @Nullable ReadableMap mTextMap = null;
     private @Nullable Integer mColor = null;
 
+    @Override
+    public void setDefaultPadding(int spacingType, float padding) {
+        // Do nothing in order to avoid adding unwanted default padding to views
+        // (https://github.com/wordpress-mobile/gutenberg-mobile/issues/992) due to
+        // https://github.com/facebook/react-native/blob/6ebd3b046e5b71130281f1a7dbe7220eff95d74a/ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactTextInputShadowNode.java#L70
+    }
+
```
mchowning: Another approach that works equally well is to explicitly set padding to 0 on the `ShadowNode` when it is constructed instead of preventing the default padding from being set like I am doing:

```
    public ReactAztecTextShadowNode() {
        super();
        setPadding(Spacing.START, 0);
        setPadding(Spacing.TOP, 0);
        setPadding(Spacing.END, 0);
        setPadding(Spacing.BOTTOM,0);
    }
```

I do not have a strong preference between these two approaches. It is nice how ignoring the default padding does not have us setting a padding of 0 both here and in the `ReactAztecManager`.

maxme:
Looks good to me.

Before we can merge that one:
1. merge the gutenberg PR first.
2. create a new Aztec release for Android (and then update the dependency here)

mchowning:
:wave: @maxme  !

Wanted to give you a heads up that I rebased these changes onto develop. The branch had gotten so out-of-date that adding another merge commit was getting a bit messy.

mchowning:
I have updated this branch with the latest code from `develop`. In addition, I included a fix for https://github.com/wordpress-mobile/WordPress-Android/issues/10505 ([relevant comment](https://github.com/wordpress-mobile/WordPress-Android/issues/10505#issuecomment-567174703)), which I believe is the most frequently occurring crash in the Android app. 

Note that this fix will _not_ fix the crash with respect to React Native `TextInput` components because those components are still using the unpatched ShadowNode. I believe that fixing the crash for `RichText` should be sufficient since we use that for blocks much more than `TextInput`. This means though, that if you add a bunch of `TextInput` based components (such as the code block, which is still dev-only at this time), it is possible to create the crash even with this fix. My intention is to monitor the crash reports after this fix is released and determine at that point whether we need a more invasive fix (i.e. creating our own `TextInput` fork) because I suspect that this relatively minor fix will take care of the problem for us since it addresses our far more common usage of`RichText`.

The only way to recreate the crash is to create a post with a lot of paragraph blocks (500-1000 of them is what I use), and then you quickly scroll through the post for a while. I usually get a crash in a minute or two. If there's no crash after a minute or so I'll kill the app and restart it to try again. To give you an idea of how I recreate the crash, here is [a video of my very scientific method for causing the crash](https://cloudup.com/cK4Y6p_2Uae).

I know this PR has been up for quite a while now, but I think it is in good shape to review and 🤞 get merged. cc: @cameronvoell , @marecar3 , @SergioEstevao

marecar3:
Hey @mchowning if this PR fixed the crash on Android, should we maybe update release notes in this PR, WDYT? thanks.

react-native-aztec/android/src/main/java/org/wordpress/mobile/ReactNativeAztec/ReactTextInputShadowNodeFork.java
```diff
@@ -0,0 +1,240 @@
...
+import com.facebook.yoga.YogaMeasureMode;
+import com.facebook.yoga.YogaMeasureOutput;
+import com.facebook.yoga.YogaNode;
+
+/**
+ * This is a fork from {@link com.facebook.react.views.textinput.ReactTextInputShadowNode} for the purpose
+ * of customizing that class so that the construction of the dummy {@link EditText} instance
+ * can be overridden (see {@link ReactTextInputShadowNodeFork#createDummyEditText(ThemedReactContext)}).
+ */
+public class ReactTextInputShadowNodeFork extends ReactBaseTextShadowNode
```
marecar3: Is there some plan on how can we follow improvements in `ReactAztecTextShadowNode` or `ReactBaseTextShadowNode` which will be potentially presented in the new version of RN?

mchowning:
>Hey @mchowning if this PR fixes the crash on Android, should we maybe update release notes in this PR, WDYT?

I think you're absolutely right @marecar3 ! I've added release notes addressing both the padding fix and the crash fix.

react-native-aztec/android/src/main/java/org/wordpress/mobile/ReactNativeAztec/ReactTextInputShadowNodeFork.java
mchowning: Not really. Ideally, we'll be able to get React Native itself updated to allow overriding the construction of the `EditText` instance [like this](https://github.com/wordpress-mobile/gutenberg-mobile/pull/1560/commits/048f07042603cf9579098eb250989f18dd2592b4#diff-50887f88a59f4879f48ea11a48b84e14R74-R95). Otherwise (and until then) we will just have to manually pull in any updates.

mchowning: Got this fix [merged into React Native](https://github.com/facebook/react-native/commit/a5b5d1a805a9c54d325763b432be1cf2c8811dc9), so once we start using a version of React Native that includes that change, we won't need to have this `ReactTextInputShadowNodeFork` anymore! 😄

PR #1563, opened 2019-11-10
Add Android native handling of api calls

mchowning:
Just looking to add the changes required for Android to your branch @etoledom . I realize this won't make a ton of sense without the WPAndroid changes for context (I'll get that up soon), but definitely let me know if you have any thoughts.

### DIFF
--- a/react-native-gutenberg-bridge/android/build.gradle
+++ b/react-native-gutenberg-bridge/android/build.gradle
@@ -91,6 +91,11 @@ android {
     lintOptions {
         abortOnError false
     }
+
+    compileOptions {
+        sourceCompatibility JavaVersion.VERSION_1_8
+        targetCompatibility JavaVersion.VERSION_1_8
+    }
 }
 
 repositories {
--- a/src/api-fetch-setup.js
+++ b/src/api-fetch-setup.js
@@ -1,8 +1,4 @@
-/**
- * External dependencies
- *
- * @format
- */
+/** @format */
 
 /**
  * External dependencies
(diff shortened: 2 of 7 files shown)


src/api-fetch-setup.js
```diff
@@ -18,6 +14,9 @@ const fetchHandler = ( options ) => {
 	const responsePromise = fetchRequest( options.path );
 
 	const parseResponse = ( response ) => {
+		if ( typeof response === 'string' ) {
+			response = JSON.parse( response );
```
mchowning: For Android, it is a fair bit easier to just send a string across the bridge instead of a json object. Let me know if there's a reason we should prefer an object though. I can definitely do it, it just requires a fair bit of code something along [these lines](https://github.com/AppsFlyerSDK/react-native-appsflyer/blob/master/android/src/main/java/com/appsflyer/reactnative/RNUtil.java).

react-native-gutenberg-bridge/android/build.gradle
```diff
@@ -91,6 +91,11 @@ android {
     lintOptions {
         abortOnError false
     }
+
+    compileOptions {
+        sourceCompatibility JavaVersion.VERSION_1_8
+        targetCompatibility JavaVersion.VERSION_1_8
```
mchowning: Needed in order to use method references and lambdas.

mchowning:
I'm going to go ahead and rebase this onto your PR @etoledom so that we can get your gutenberg-mobile PR opened up for review with the work from both platforms, but when you get back definitely let me know if you see any issues with what I've done here.

PR #1589, opened 2019-11-18
Automate gutenberg translations followup

Tug:
This PR aims to improve our i18n flow. There are 3 main improvements we need to take care of:
- [x] Fix plurals on android. GlotPress supports `<string-array>` tags for plurals instead of `<plurals>`
- [x] Find a way to keep the same ref/name for each string so not all strings are updated when we refresh the translation files.
- [x] Exclude strings from tests files in gutenberg
 
### Testing Instructions
- Make sure the gutenberg submodule is updated
- Run `yarn genstrings`
- Make sure there are no changes to string files in bundles (`git diff`)

Update release notes:


bin/po2android.js
```diff
@@ -1,12 +1,19 @@
...
 
 const gettextParser = require( 'gettext-parser' ),
-	fs = require( 'fs' );
+	fs = require( 'fs' ),
+	crypto = require( 'crypto' );
 
 const indent = '    ';
 
+/**
+ * Encode a raw string into an Android-compatible value
```
hypest: Should we also encode quotes and double-quotes?

Tug: Yep, `JSON.stringify` is going to take care of that

Tug: Single quotes don't need to be encoded though as they'll be contained in a double quote string

Issue #1587, opened 2019-11-18
Make Spacer block publicly available


pinarol:
We completed Spacer block implementation, let's remove DEV flag to make it available in production.

To test:
- Generate JS bundles
- Open PRs on WPiOS & WPAndroid
- Open the WordPress app and open editor with metro off
- Confirm Spacer bock is insertable and make some smoke tests on it

pinarol:
Done

PR #1594, opened 2019-11-19
Enable Spacer block in production

geriux:
This enables the `Spacer` block in production

Fixes https://github.com/wordpress-mobile/gutenberg-mobile/issues/1587

`Gutenberg` PR -> https://github.com/WordPress/gutenberg/pull/18605

To test:

`WordPress iOS` -> https://github.com/wordpress-mobile/WordPress-iOS/pull/12970
`WordPress Android` -> https://github.com/wordpress-mobile/WordPress-Android/pull/10817

Update release notes:

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -1,6 +1,7 @@
 1.18.0
 ------
 * [iOS] Added native fullscreen preview when clicking image from Image Block
+* New block: Spacer
 
 1.17.0
 ------


marecar3:
Hey @geriux, not sure if I am missing some context but usually the practice is that we update bundle files through the release process. 
The release process for the v`1.18.0` started yesterday so you can be sure that spacer will be part of the new bundle when we merge it to the master.

geriux:
> Hey @geriux, not sure if I am missing some context but usually the practice is that we update bundle files through the release process.
> 
> The release process for the v`1.18.0` started yesterday so you can be sure that, if we merge this PR, spacer will be part of the new release bundle.

Yup, I updated the bundles so reviewers were able to test the `Spacer` block in the main apps before merging. It also was a good test as part of my onboarding to know how to generate bundles, etc =)

maxme:
I'm not sure if there is a process for this, but I'll drop the bundle changes from the PR and keep the gutenberg ref change only.

maxme:
@geriux the e2e tests seem to fail - can you check it's related to your changes?

pinarol:
> I'll drop the bundle changes from the PR and keep the gutenberg ref change only

Sounds good 👍bundles are usually generated only on release branches, this one was needed just for testing purposes because we needed to verify Spacer block appears in the block picker while metro is off.

geriux:
> @geriux the e2e tests seem to fail - can you check it's related to your changes?

It shouldn't be but I'll check, thanks!

Issue #1463, opened 2019-10-18
Create feature flag for template picker


koke:
There should be a launch option to decide whether the template picker will be visible. We only want this for new pages, so the apps need to be able to specify when the picker will be available

koke:
I think the feature flag might be easily set up as a [Gutenberg experiment](https://github.com/WordPress/gutenberg/blob/master/lib/experiments-page.php), and we can with on that experiment via editor settings from our own initialization.

Only showing this on pages will require some extra changes, since we are currently mocking the post data and Gutenberg can't tell if we are editing a page or a post. I still need to get myself more familiar with editor initialization and the current bridge communication to make a decision on how to implement this.

koke:
All the PRs were merged, so this one was ready to close

PR #1606, opened 2019-11-21
Add post type to editor

koke:
To be able to show the Page Templates selector only on pages, we need to differentiate between posts and pages. Currently we are hardcoding a `postType` of `post` in all cases.

This PR adds the necessary structure for the apps to inject that post type into Gutenberg, but the changes to the apps to actually do that will come in future PRs.

I have changed the bridge code in a way that would be compatible with the existing app implementations (it would just default to `post`), so these should be fine to merge before we make changes to the apps.

Part of #1463. The proper feature flag is coming up as part of #1576, and we'll need PRs in both apps to finish that issue.

To test:

From WPAndroid, hardcode the "page" postType, or try with the approach in https://github.com/wordpress-mobile/WordPress-Android/pull/10775:

```diff
diff --git a/libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/GutenbergContainerFragment.java b/libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/GutenbergContainerFragment.java
index 2d874b7105..1549e391cf 100644
--- a/libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/GutenbergContainerFragment.java
+++ b/libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/GutenbergContainerFragment.java
@@ -79,4 +79,5 @@ public class GutenbergContainerFragment extends Fragment {
                 BuildConfig.DEBUG,
                 BuildConfig.BUILD_GUTENBERG_FROM_SOURCE,
+                "page",
                 isNewPost,
                 localeString,
```

From WPiOS: https://github.com/wordpress-mobile/WordPress-iOS/commit/c7885da6a3b5fe2baaf47845bf7a48f959389883

Then check with the JS debugger that the editor is initialized with the right post type.

Update release notes:


pinarol:
small note I had to change
```
 func gutenbergPostType() -> String? {
         return post is Page ? "page" : "post"
     }
```

into 

```
 func gutenbergPostType() -> String {
         return post is Page ? "page" : "post"
 }
```

on WPiOS side, it was just giving a warning but the method was not getting called.

PR #1608, opened 2019-11-22
Update tests for Xcode 11/iOS 13

JavonDavis:
Fix CI tests for Xcode 11 on the server and locally.

To test:
 - Run tests locally
 - For iOS tests to run locally make sure you have Xcode 11.2 installed locally, and the iPhone 11 simulator installed to.
 - Check the tests on the server.

Update release notes:


__device-tests__/helpers/utils.js
```diff
@@ -92,9 +92,11 @@ const setupDriver = async () => {
...
 	} else {
-		desiredCaps = _.clone( ios12 );
+		desiredCaps = _.clone( ios );
 		if ( isLocalEnvironment() ) {
 			desiredCaps.app = path.resolve( localIOSAppPath );
+			delete desiredCaps.platformVersion;
+			desiredCaps.deviceName = 'iPhone 11';
 		} else {
 			desiredCaps.app = `sauce-storage:Gutenberg-${ safeBranchName }.app.zip`; // App should be preloaded to sauce storage, this can also be a URL
 		}
```
mchowning: **Not suggesting we need to make this change in this PR**, but as a future improvement, what would you think about moving as much of this "logic" as we can into separate objects in caps (maybe `iosCi` and `iosLocal`), and just having the logic here select the appropriate caps file (and possibly resolving the app path for local runs)? Seems like it would be nice to keep as much of this "logic" in one place as possible. I'm imagining something like:

```
ios = {
	browserName: '',
	platformName: 'iOS',
	os: 'iOS',
	deviceOrientation: 'portrait',
	automationName: 'XCUITest',
	appiumVersion: '1.15.0', // SauceLabs requires appiumVersion to be specified.
	app: undefined, // will be set later, locally this is relative to root of project
};

exports.iosCi = {
	...ios,
	deviceName: 'iPhone 11 Simulator',
	platformVersion: '13.0',
};

exports.iosLocal = {
	...ios,
	deviceName: 'iPhone 11',
};
```

__device-tests__/gutenberg-editor-block-insertion.test.js
```diff
@@ -100,6 +102,7 @@ describe( 'Gutenberg Editor tests for Block insertion', () => {
 			await editorPage.dismissKeyboard();
 		}
 
+		await swipeDown( driver );
```
hypest: @SergioEstevao , do you happen to know why do we need this operation here?

SergioEstevao: This was @JavonDavis work, so I'm not sure about the rational of this.

__device-tests__/helpers/utils.js
hypest: I like that suggestion, adding perhaps some comment about why we don't set the `platformVersion` for the local run. That said, and like Sérgio commented over Slack, this can probably be done in a separate PR so to unblock this one. @SergioEstevao , can you open a relevant ticket? Thanks!

Issue #1264, opened 2019-08-06
Preformatted block


hypest:
Support the Gutenberg "Preformatted" block.

SergioEstevao:
@iamthomasbishop what font do you want to use for preformatted blocks/html?

Normally we use a monospace font, and in iOS the one available is `Courier`

@hypest is this font available on Android too?

iamthomasbishop:
@SergioEstevao What are we using on Aztec/HTML mode? We can use the same font we're using there for consistency. Another option (which I _think_ is available, per [this list](https://developer.apple.com/fonts/system-fonts/)) could be Menlo.

iamthomasbishop:
It's hard to tell what the default is on Android, so that could change my answer above 😄 I _think_ Droid Sans Mono was the default at some point but not sure if that's still the case. [Roboto Mono](https://fonts.google.com/specimen/Roboto+Mono) is also nice.

SergioEstevao:
@iamthomasbishop we were using Courier in Aztec, but Menlo looks so much better that I'm going to change in both. What size do you think we should use?

Do you mind to post here a little mockup? and also tell me what colour you want to use for the background of the Pre block.

iamthomasbishop:
> we were using Courier in Aztec, but Menlo looks so much better that I'm going to change in both.

👍 Awesome.

As I understand it, `pre` should use the same font styling as `code` block, minus syntax highlighting and additional block settings (code language, show line numbers, etc). So let's match the code block styling that we documented here: https://github.com/wordpress-mobile/gutenberg-mobile/issues/368

Our font-size can match what I mentioned in the issue mentioned above, 14px. If that's a bit too small, we can bump up to our base paragraph sizing of 16px.

PR #1615, opened 2019-11-25
Activate preformat block on android platform

marecar3:
Fixes [Active preformat block on the Android platform](https://github.com/wordpress-mobile/gutenberg-mobile/issues/1264)

AztecEditor-Android PR: https://github.com/wordpress-mobile/AztecEditor-Android/pull/869
Gutenberg PR: https://github.com/WordPress/gutenberg/pull/18777
WPAndroid PR: https://github.com/wordpress-mobile/WordPress-Android/pull/10868

To test:
 - Open the demo app
 - Add a Preformatted block to the content
 - Write around on it
 - Try to merge it with other blocks

Special case:

1. Create a draft Post on Gutenberg Web
2. Insert bellow content
```
<!-- wp:preformatted -->
<pre class="wp-block-preformatted">just a text<br>line2<br></pre>
<!-- /wp:preformatted -->
```
3. Open mobile application
4. Open drafted Post
5. Place the caret on the end of the Post
6. Go back (leave Editor)
7. Open draft Post again

Result: There shouldn't be a red screen: https://github.com/wordpress-mobile/AztecEditor-Android/pull/869#issuecomment-552864686

Update release notes:


SergioEstevao:
@mchowning bug number 2 is a [know bug](https://github.com/WordPress/gutenberg/issues/18476), and it was happening in iOS and the web too, it's now fixed in GB master by this [PR](https://github.com/WordPress/gutenberg/pull/18656)

marecar3:
Changed target branch to `develop` as we don't want to hurry with this feature (it still has some active issues)

hypest:
> 5. Observe the cursor briefly jump, that a new line is added to the block, but the cursor remains at its current position.

This is probably caused by https://github.com/WordPress/gutenberg/blob/bb0529f59746e67dae47077bf91833e42a9081c5/packages/rich-text/src/component/index.native.js#L650. Aztec is trimming spaces in other cases but in the `<pre>` case it will not so, the JS code need to cater for that case too. @marecar3 can you have a look there?

> 6. Press Enter again
> 7. Observe that the previously added new line is removed and the cursor remains unmoved

I wasn't able to replicate that 🤔

marecar3:
> > 1. Observe the cursor briefly jump, that a new line is added to the block, but the cursor remains at its current position.
> 
> This is probably caused by https://github.com/WordPress/gutenberg/blob/bb0529f59746e67dae47077bf91833e42a9081c5/packages/rich-text/src/component/index.native.js#L650. Aztec is trimming spaces in other cases but in the `<pre>` case it will not so, the JS code need to cater for that case too. @marecar3 can you have a look there?

Thanks a lot, @hypest, you opened my eyes, will let you know when I fix it.

> > 1. Press Enter again
> > 2. Observe that the previously added new line is removed and the cursor remains unmoved
> 
> I wasn't able to replicate that 🤔

Hey @hypest, I have fixed that issue.

hypest:
>  it's now fixed in GB master by this PR

Already [commented in the Gutenberg PR](https://github.com/WordPress/gutenberg/pull/18777#issuecomment-559494281) so, when that PR gets rebased the fix will be incorporated.

marecar3:
> > 1. Observe the cursor briefly jump, that a new line is added to the block, but the cursor remains at its current position.
> 
> This is probably caused by https://github.com/WordPress/gutenberg/blob/bb0529f59746e67dae47077bf91833e42a9081c5/packages/rich-text/src/component/index.native.js#L650. Aztec is trimming spaces in other cases but in the `<pre>` case it will not so, the JS code need to cater for that case too. @marecar3 can you have a look there?
> 
> > 1. Press Enter again
> > 2. Observe that the previously added new line is removed and the cursor remains unmoved
> 
> I wasn't able to replicate that 🤔

Hey @hypest, I have fixed the problem with this change: https://github.com/WordPress/gutenberg/pull/18777/commits/2faec829bc9c157d7042b47a4499dcf5ff32156e

The main concern with that change is that I needed to remove `pre` and also `br` as regex match will be true if I just remove `pre`. Not sure if it's a safe to remove `br`?

marecar3:
> 👋 @marecar3 ! This is looking pretty good. There are just a two issues I'm seeing with newline handling on Android.
> 
> Also, I think we need to update the WPAndroid branch to reference the latest gutenberg and Aztec code.
> 
> ## 1. Adding newline when last character on a line is a space causes problems with consecutive empty lines
> Steps:
> 
> 1. Add preformat block
> 2. Enter text on line _that ends with a space_
> 3. Press Enter and observe a new line as expected
> 4. Press Enter again
> 5. _Observe the cursor briefly jump, that a new line is added to the block, but the cursor remains at its current position._
> 6. Press Enter again
> 7. _Observe that the previously added new line is _removed_ and the cursor remains unmoved_
> 
> preformat-blocks-space-newline mp4
> 
> ## 2. Merging blocks with newlines loses newlines
> Steps:
> 
> 1. Add preformat block
> 2. Add content spanning multiple lines
> 3. Create another preformat block
> 4. (Optional) Add content to the new block
> 5. Merge the new block back into the first block by pressing backspace at the beginning of the second block
> 6. _Observe that all newlines in the two blocks appear to have been replaced with spaces_
> 
> preformat-lose-newlines-on-merge mp4

Hey @mchowning, you can do another round of testing, thanks! :)

hypest:
> Hey @hypest, I have fixed the problem with this change: WordPress/gutenberg@2faec82
>
> The main concern with that change is that I needed to remove pre and also br as regex match will be true if I just remove pre. Not sure if it's a safe to remove br?
I can't remember the case where we have html block with only br, but maybe I am wrong?

Without diving too much into this yet, I think that the `willTrimSpaces` pattern should match what Aztec will do when it will trim spaces. But, since in the PRE block we tell Aztec to not trim anything at all, it means that changing the regex is not exactly relevant. Instead, how about finding a way to shortcircuit willTrimSpaces() to return `false` if we know we are in the PRE case?

marecar3:
> > Hey @hypest, I have fixed the problem with this change: [WordPress/gutenberg@2faec82](https://github.com/WordPress/gutenberg/commit/2faec82)
> > The main concern with that change is that I needed to remove pre and also br as regex match will be true if I just remove pre. Not sure if it's a safe to remove br?
> > I can't remember the case where we have html block with only br, but maybe I am wrong?
> 
> Without diving too much into this yet, I think that the `willTrimSpaces` pattern should match what Aztec will do when it will trim spaces. But, since in the PRE block we tell Aztec to not trim anything at all, it means that changing the regex is not exactly relevant. Instead, how about finding a way to shortcircuit willTrimSpaces() to return `false` if we know we are in the PRE case?

Hey @hypest, I have implemented another approach which you can find there: https://github.com/WordPress/gutenberg/pull/18777/commits/60dc8388f7f17977f86be35fbcb8be3eced28f80
let me know if you have some doubt about it, thanks.

react-native-aztec/android/src/main/java/org/wordpress/mobile/ReactNativeAztec/ReactAztecText.java
```diff
@@ -23,11 +24,13 @@
 import com.facebook.react.views.textinput.ReactTextInputLocalData;
 import com.facebook.react.views.textinput.ScrollWatcher;
 
+import org.jetbrains.annotations.NotNull;
```
hypest: Hmm, any particular reason for using the JetBrains `NotNull` instead of AndroidX's `NonNull`?

Asking since I'd find it kinda more "natural" to to use the AndroidX's annotations unless they don't suit us well.

react-native-aztec/android/src/main/java/org/wordpress/mobile/ReactNativeAztec/ReactAztecText.java
```diff
@@ -23,11 +24,13 @@
...
 import org.wordpress.aztec.AztecText;
 import org.wordpress.aztec.AztecTextFormat;
 import org.wordpress.aztec.ITextFormat;
 import org.wordpress.aztec.plugins.IAztecPlugin;
 import org.wordpress.aztec.plugins.IToolbarButton;
+import org.wordpress.aztec.watchers.EndOfBufferMarkerAdder;
```
hypest: This is an unused import at this stage, right? Can we remove it or is it an indication that some code that uses is got lost?

react-native-aztec/android/src/main/java/org/wordpress/mobile/ReactNativeAztec/ReactAztecText.java
```diff
@@ -525,11 +547,19 @@ public void onTextChanged(CharSequence s, int start, int before, int count) {
...
-            if (!mIsSettingTextFromJS && mListeners != null) {
+            if (shouldRedirectTextChangeCalls()) {
                 for (TextWatcher listener : mListeners) {
                     listener.afterTextChanged(s);
                 }
             }
         }
     }
+
+    private boolean shouldRedirectTextChangeCalls() {
```
hypest: Nit: what do you think about naming this method as `shouldDelegateTextChangeCalls` to match the class' name?

react-native-aztec/android/src/main/java/org/wordpress/mobile/ReactNativeAztec/ReactAztecText.java
```diff
@@ -525,11 +547,19 @@ public void onTextChanged(CharSequence s, int start, int before, int count) {
...
+        return (!mIsSettingTextFromJS || isPreTag()) && mListeners != null;
```
hypest: The conditional has now started to be harder to read/follow. Can we split it up into documented individual ones that return early? WDYT?

For example:
```
        if (mListeners == null) {
            // No listeners so, no-one to delegate the calls to
            return false;
        }
```

react-native-aztec/android/src/main/java/org/wordpress/mobile/ReactNativeAztec/ReactAztecText.java
marecar3: Yup, I agree.

react-native-aztec/android/src/main/java/org/wordpress/mobile/ReactNativeAztec/ReactAztecText.java
marecar3: Yup, removed it.

react-native-aztec/android/src/main/java/org/wordpress/mobile/ReactNativeAztec/ReactAztecText.java
marecar3: Good suggestion. I have split condition into multiple individuals.

marecar3:
Hey @hypest, I think that I have addressed all issues so this one is ready for another iteration of review.

react-native-aztec/android/build.gradle
```diff
@@ -12,7 +12,7 @@ buildscript {
         wordpressUtilsVersion = '1.22'
         espressoVersion = '3.0.1'
 
-        aztecVersion = '03dd2332c1deb7c26dac82a7638121ca7b4ca0fd'
+        aztecVersion = '754086b517b1da25869102482048086ef928fb5b'
```
hypest: I notice that this commit includes 3 changes that are not actually related (the Aztec red bump, the `org.jetbrains.annotations.NotNull` change and the expansion of the conditional statement. No harm done but, I'd probably suggest keeping unrelated changes to separate commits. That helps with code review but most importantly, better documentation of the code changes. Our future selves usually like that isolation when investigating issues :)

react-native-aztec/android/src/main/java/org/wordpress/mobile/ReactNativeAztec/ReactAztecText.java
```diff
@@ -171,7 +170,7 @@ public boolean onTextContextMenuItem(int id) {
     }
 
     @Override
-    public float getPreformatBackgroundAlpha(@NotNull TypedArray styles) {
+    public float getPreformatBackgroundAlpha(@Nullable TypedArray styles) {
```
hypest: Looking at the [corresponding Aztec code here](https://github.com/wordpress-mobile/AztecEditor-Android/pull/869/files#diff-7ca71cd1f009ab8438e0901205001b14R715) it looks like the parameter is not nullable (no `?` in the parameter type). Perhaps `@NonNull` is the appropriate annotation here?

react-native-aztec/android/src/main/java/org/wordpress/mobile/ReactNativeAztec/ReactAztecText.java
hypest: Left an additional comment on the particular change: https://github.com/wordpress-mobile/gutenberg-mobile/pull/1615/commits/a665093164086ad2f7ce6bed83e349da7cccfd46#r354960106

react-native-aztec/android/src/main/java/org/wordpress/mobile/ReactNativeAztec/ReactAztecText.java
marecar3: Hey @hypest, yes you were right in the first place. 
For some reason, I mixed those two annotations.
It's fixed now. Sorry about misunderstanding.

react-native-aztec/android/build.gradle
marecar3: Good to know that :)

marecar3:
> Made another pass Marko and only one comment needs attention, the one about the annotation.

👋 Thanks @hypest, I have fixed the comment which points to fix annotation.
You can do another iteration of testing.

hypest:
Hmm, the Android device tests consistently fail on the `rotate device and continue adding blocks` test. I've tried re-triggering them 3 times already 🤔

react-native-aztec/android/src/main/java/org/wordpress/mobile/ReactNativeAztec/ReactAztecText.java
```diff
@@ -525,11 +546,29 @@ public void onTextChanged(CharSequence s, int start, int before, int count) {
...
+        if (isPreTag()) {
+            // If tag is pre tag we want to delegate calls in every case
+            return true;
+        }
+
+        return !mIsSettingTextFromJS;
+    }
+
+    private boolean isPreTag() {
+        return !TextUtils.isEmpty(mTagName) && mTagName.equals(PRE_TAG);
```
mchowning: This is just a nice-to-have, but I think this could be shortened to be a bit more clear by just reversing the equals check so you get a null and empty check for free since `PRE_TAG` is a constant:

```
return PRE_TAG.equals(mTagName);
```

marecar3: Good point. Fixed.

PR #1616, opened 2019-11-25
Button Block

### LINKED ISSUE #1513: Button block - 1st iteration
We need to add support for Button block on the mobile side. It is highly used in pages and it would help us add variety on our starter page templates.

Let discuss which items would make a good first iteration, here are my thoughts but we can drop/add items:

- [ ] 1. Render the block on mobile with given color, border radius, `Link` option. No settings, no URLFinder for now.

Screen Shot 2019-10-29 at 18 52 00

The tricky part here will be the gradient colors, we'll need to add a new library to handle this. I am actually willing to leave gradients for a later iteration but not sure how to display a button that was created on web and already using gradient colors. If there's a good idea that can cover this we can leave it to the 2nd iteration.

Screen Shot 2019-10-29 at 18 52 11

 - [ ] 2. Add Border radius settings:

Screen Shot 2019-10-29 at 18 55 34

- [ ] 3. Add style settings(Needs design for mobile):

Screen Shot 2019-10-29 at 18 58 06

- [ ] 4. Add link settings

Screen Shot 2019-10-29 at 18 56 37

- [ ] 5. Add color settings(Needs design for mobile)(Custom color link can allow user change the hex value directly instead of displaying a color picker for this 1st iteration)

Screen Shot 2019-10-29 at 19 00 34

lukewalczak:
Fixes #1513 

Ref to gutenberg PR:https://github.com/WordPress/gutenberg/pull/18823
Ref to WordPress-Android: https://github.com/wordpress-mobile/WordPress-Android/pull/10902
Ref to WordPress-iOS:https://github.com/wordpress-mobile/WordPress-iOS/pull/13060

This PR introduces 1st iteration of mobile `Button` block component.

To achieve gradient background color on mobile `react-native-linear-gradient` comes in handy.

The first iteration includes:
* Rendering `Button` with:
	* gradient background color ✅ 
	* custom background color chosen from picker ✅ 
	* custom text color chosen from picker ✅ 
* Options to change border radius, add link, open url in new tab ✅ 
* Expanding and shrinking `RichText` ✅ 

Blockers:
* cannot render button with background color chosen from color palette ❌
* cannot render button with text color chosen from color palette ❌
~~* issue with expanding and shrinking `RichText` on iOS ❌~~
~~* issue with shrinking  `RichText` on Android ❌~~
~~* issue with text color on WordPress-iOS ❌~~

To test:

**-> Gradient**
* open draft post on web
* add Button
* choose the gradient background color
* save post
* open draft post on mobile
* check if a gradient is correct

**-> Custom background color**
* open draft post on web
* add Button
* choose the custom background color from picker
* save post
* open draft post on mobile
* check if a background color is correct

**-> Custom text color**
* open draft post on web
* add Button
* choose the custom text color from picker
* save post
* open draft post on mobile
* check if a text color is correct

**-> Border radius**
* open draft post on mobile
* add Button
* open settings
* move slider to change border radius value

**-> Add link**
* open draft post on mobile
* add Button
* open settings
* pass a url
* switch on open in new tab
* open a preview mode
* click Button and check if it opens a new window with passed url

Screenshot /  gifs:

**ANDROID**

regular | linear gradient
--- | ---
Screenshot 2019-12-02 at 10 35 50 | Screenshot 2019-12-02 at 10 45 54

**IOS**

regular | linear gradient
--- | ---
Screenshot 2019-12-02 at 11 05 07 | Screenshot 2019-12-02 at 10 58 11

**IOS Dark Mode**

regular | linear gradient
--- | ---
Screenshot 2019-12-02 at 11 12 28 | Screenshot 2019-12-02 at 11 12 43

**Expanding and shrinking RichText**

android | ios
--- | ---
android_rich_text | ios_rich_text

**Settings**

ios | android
--- | ---
Screenshot 2019-12-02 at 11 33 20 | Screenshot 2019-12-02 at 11 05 49

Update release notes:


iamthomasbishop:
@lukewalczak Based on screenshots above, this is looking really nice! I also noticed the platform differences for padding, so I think we should fix that ASAP. 

In terms of colors, I think as a starting point this is reasonable. If we aren’t going to support changing the text/bg Coles in the inspector, I think it might be worth implementing a little hint that we are building this functionality. It could be a similar flow to unsupported blocks, where we show a `(?)` icon as an entry point to an informational bottom sheet. Something like this:

44E6F944-9E78-4EAB-8A54-E34EB6F18861

One more question: what does the border radius of the  outline (outset blue border) look like with larger radii? Curious how that border radius looks relative to the button radius itself.

lukewalczak:
@iamthomasbishop Currently, the value is a total sum of `button radius + spacing between button and outline + outline border width`. Let me know, what should be the value if the current one is incorrect.

iamthomasbishop:
@lukewalczak I think it looks right. I was mostly curious how it looks as you change the border-radius up/down, but from the looks of it, it should scale nicely.

dratwas:
> - issue with text color on WordPress-iOS ❌

This issue exists because of this change
https://github.com/wordpress-mobile/AztecEditor-iOS/commit/152307508dcdd1d4cb5eeba730ee5ffaf4eb0854#diff-8d8015f3f021c0252c1569e4965dce80R245

 It doesn't care if `textColor` is set or not just use defaults one

It landed just 4h ago in `gutenberg-mobile` so I'm able to reproduce it in our example as well.

lukewalczak:
Can confirm that after merge `develop` into PR issue:
> issue with text color on WordPress-iOS ❌

doesn't exist.

light mode | dark mode 
--- | ---
Screenshot 2019-12-06 at 21 57 34 | Screenshot 2019-12-06 at 21 57 54

react-native-aztec/ios/RNTAztecView/RCTAztecView.swift
```diff
@@ -172,14 +174,35 @@ class RCTAztecView: Aztec.TextView {
...
+        let container = NSTextContainer(size: containerSize)
+        let storage = NSTextStorage(attributedString: attributedText)
+        let layoutManager = NSLayoutManager()
+        layoutManager.addTextContainer(container)
+        storage.addLayoutManager(layoutManager)
+        let rawSize =  layoutManager.usedRect(for: container).size
+        return rawSize
+    }
+    
+    private func measureWidth() {
```
pinarol: can we rename this method since it is not just measuring, it is also changing things. `adjustWidth` can be a better candidate.

react-native-aztec/ios/RNTAztecView/RCTAztecView.swift
```diff
@@ -172,14 +174,35 @@ class RCTAztecView: Aztec.TextView {
...
+        if(maxWidth > 0 && minWidth > 0) {
+            let newWidth = textSize(for: maxWidth).width
```
pinarol: I was wondering if `sizeThatFits:` can also work here which can save us from introducing the new method: `textSize`

dratwas: I remember that at some point of work I changed it to that function but i think that the `sizeThatFits:`  could work in the current implementation. Will try it and let you know :)

lukewalczak: Confirmed it works properly with `sizeThatFits`

react-native-aztec/ios/RNTAztecView/RCTAztecView.swift
lukewalczak: Renamed

android/app/src/main/java/com/gutenberg/MainApplication.java
```diff
@@ -126,13 +127,15 @@ public boolean getUseDeveloperSupport() {
...
                 return Arrays.asList(
-                        new MainReactPackage(),
-                        new ReactSliderPackage(),
-                        new ReactVideoPackage(),
-                        new SvgPackage(),
-                        new ReactAztecPackage(),
-                        new RNRecyclerviewListPackage(),
-                        mRnReactNativeGutenbergBridgePackage);
+                    new MainReactPackage(),
+            new LinearGradientPackage(),
```
pinarol: Indentation doesn't look right.

bundle/ios/GutenbergNativeTranslations.swift
```diff
@@ -1,11 +1,14 @@
 import Foundation
 
 private func dummy() {
-    _ = NSLocalizedString("'%s' isn't yet supported on WordPress for Android", comment: "translators: %s: Name of the block")
-    _ = NSLocalizedString("'%s' isn't yet supported on WordPress for iOS", comment: "")
+    _ = NSLocalizedString("%1$s. Current value is %2$s", comment: "translators: accessibility text. Inform about current value. %1$s: Control label %2$s: Current value.")
```
pinarol: @koke do we still need to make additions to this file?

koke: I think that's still the case, @Tug can you confirm?

react-native-aztec/ios/RNTAztecView/RCTAztecView.swift
```diff
@@ -172,14 +174,25 @@ class RCTAztecView: Aztec.TextView {
...
+    // MARK: - View height and width: Match to the content
 
     override func layoutSubviews() {
         super.layoutSubviews()
+        adjustWidth()
         fixLabelPositionForRTLLayout()
         updateContentSizeInRN()
     }
 
+    private func adjustWidth() {
```
koke: It seems very strange to me that we would need this. Can you explain why we need to adjust the width manually and it's not enough with flexbox?

bundle/ios/GutenbergNativeTranslations.swift
Tug: Yes, but those should only be updated when we make a release though. Not sure why we're updating `bundle/` in this PR?

lukewalczak: I was publishing `bundle` to create a testing build in WordPress-iOS for Thomas. Will revert these changes.

react-native-aztec/ios/RNTAztecView/RCTAztecView.swift
lukewalczak: @dratwas could you please explain that?

dratwas: I couldn't make it work with `flexbox` only. I think it is because on initial render we have width set and if the text is bigger than that width it does the break line instead of expanding width. Additional thing is that the RN parent view doesn't know that the textView wants to expand until we send that information. However, if you have any idea how we could handle that in a different way, I would be more than grateful :)

koke: Just a quick update, I did some testing with a sample app and `TextInput` and what we want here seems well supported. I'm still not sure what exactly is preventing that from working in Gutenberg, but it should be possible in theory.

RPReplay_Final1578911172



```jsx
export default function App() {
  const [value, setValue] = React.useState( 'Button label' );
  return (
    <View style={ styles.container }>
      <TextInput
        style={ styles.input }
        value={ value }
        multiline={ true }
        onChangeText={ text => setValue( text ) }
      />
    </View>
  );
}

const styles = StyleSheet.create({
  container: {
    flex: 1,
    justifyContent: 'center',
    alignItems: 'center',
    alignContent: 'center',
  },
  input: {
    color: '#fff',
    backgroundColor: '#00f',
    fontSize: 24,
  }
});
```

PR #1621, opened 2019-11-27
Use plural method for multiple media selection handling

mkevins:
This PR aims to consolidate various media selection handling by refactoring methods that utilize both singular and plural paradigms to only use plural methods where possible.

We should preserve the behavior that occurs when multiple images are selected via the media library intent (multiple selection is possible for the image block even though the `allowMultipleSelection` flag is false).

Related WordPress-Android PR: https://github.com/wordpress-mobile/WordPress-Android/pull/10857

**To test:**

---

**Image block:**

When multiple items are selected via the media library in the image block, the additional items are appended to the editor as new image blocks.

---

**Gallery block:**

Related PR: https://github.com/wordpress-mobile/WordPress-Android/pull/10703

Checkout related Gallery PR and build with source.

* In a post, add a gallery block
* Tap "ADD MEDIA"
* Tap "WordPress Media Library"
* Select multiple images

**Expected behavior:**

The selected images should be added to the gallery, and _not_ added as new image blocks.

---

Update release notes:


react-native-gutenberg-bridge/android/src/main/java/org/wordpress/mobile/WPAndroidGlue/WPAndroidGlueCode.java
```diff
@@ -579,33 +586,42 @@ public void toggleEditorMode(boolean htmlModeEnabled) {
...
+//       if (isMediaUploadCallbackRegistered() && mMediaPickedByUserOnBlock) {
+//           String mediaType = getMediaType(isVideo);
+//           mMediaPickedByUserOnBlock = false;
+//           List<RNMedia> mediaList = new ArrayList<>();
+//           mediaList.add(new Media(mediaId, mediaUri, mediaType));
+//           mPendingMediaUploadCallback.onUploadMediaFileSelected(mediaList);
+//       } else {
+//           // we can assume we're being passed a new image from share intent as there was no selectMedia callback
+//           sendOrDeferAppendMediaSignal(mediaId, mediaUri, isVideo);
+//       }
```
dratwas: Is it intentional to leave commented code here? :)

react-native-gutenberg-bridge/android/src/main/java/org/wordpress/mobile/WPAndroidGlue/WPAndroidGlueCode.java
```diff
@@ -579,33 +586,42 @@ public void toggleEditorMode(boolean htmlModeEnabled) {
...
+                }
+            } else {
+                rnMediaList.addAll(mediaList);
+                mPendingMediaUploadCallback.onUploadMediaFileSelected(rnMediaList);
             }
-            mPendingMediaUploadCallback.onUploadMediaFileSelected(rnMediaList);
-            mPendingMediaUploadCallback = null;
         }
+
+        mAppendsMultipleSelectedToSiblingBlocks = false;
```
dratwas: Not sure if it is needed since we already set it to the correct value each time we call `requestMediaPickFromMediaLibrary`

malinajirka: Really good question ;)! I believe `appendUploadMediaFiles` can be invoked even as a result for `requestMediaPickFromDeviceLibrary`, `requestMediaPickerFromDeviceCamera` etc. 

Imagine the following scenario 
1. _requestMediaPickFrom**MediaLibrary**_ is invoked and `mAppendsMultipleSelectedToSiblingBlocks` is set to `true`
2. `appendUploadMediaFiles` is invoked and we don't re-set `mAppendsMultipleSelectedToSiblingBlocks` back to `false`
3. _requestMediaPickerFrom**DeviceCamera**_ is invoked
4. `appendUploadMediaFiles` is invoked and `mAppendsMultipleSelectedToSiblingBlocks` is `true` even though it should be `false`

I like your thinking and I think we might want to consider removing this line as you suggested and explicitly setting `mAppendsMultipleSelectedToSiblingBlocks` in all the other `requestMediaPickerXYZ` as preparation for the future. What do you think about it @mkevins ? Is there a change we'll ever enable multi-select for the other sources?

react-native-gutenberg-bridge/android/src/main/java/org/wordpress/mobile/WPAndroidGlue/WPAndroidGlueCode.java
mkevins: Indeed this is no longer needed :smiley: . Since the dependent tests have been refactored, I have removed the method chain.

react-native-gutenberg-bridge/android/src/main/java/org/wordpress/mobile/WPAndroidGlue/WPAndroidGlueCode.java
mkevins: Very good question! For the image block, the only time multiple selection is possible is via the WordPress Media Library. For this case, the special logic to append blocks has been implemented. In all other cases, I think it should currently be false. IMO this value should only be true for that special case: image block, media library selected.

> Is there a chance we'll ever enable multi-select for the other sources?

I'm not sure about that. I've set the flag to false explicitly in the other cases as well, and that can be updated later if the need arises.

react-native-gutenberg-bridge/android/src/main/java/org/wordpress/mobile/WPAndroidGlue/Media.java
```diff
@@ -21,6 +23,21 @@ public Media(int id, String url) {
         this.mType = "";
     }
 
+    public static Media createRNMediaUsingMimeType(final int id, final String url, @NonNull final String mimeType) {
+        String type;
+
+        if (mimeType.startsWith("image")) {
+            type =  "image";
```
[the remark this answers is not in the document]
mkevins: Sounds good :+1: . Updated!

react-native-gutenberg-bridge/android/src/main/java/org/wordpress/mobile/WPAndroidGlue/WPAndroidGlueCode.java
```diff
@@ -163,6 +168,8 @@ public void responseHtml(String title, String html, boolean changed) {
             @Override
             public void requestMediaPickFromMediaLibrary(MediaUploadCallback mediaSelectedCallback, Boolean allowMultipleSelection, MediaType mediaType) {
                 mMediaPickedByUserOnBlock = true;
+                // flag to append as siblings when allowMultipleSelection = false is not respected
```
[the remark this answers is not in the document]
mkevins: Sure! I've added it as a Javadoc so its purpose is easily accessible wherever it's used.

react-native-gutenberg-bridge/android/src/main/java/org/wordpress/mobile/WPAndroidGlue/WPAndroidGlueCode.java
```diff
@@ -579,33 +586,42 @@ public void toggleEditorMode(boolean htmlModeEnabled) {
...
     }
 
     public void appendUploadMediaFiles(ArrayList<Media> mediaList) {
         if (isMediaUploadCallbackRegistered() && mMediaPickedByUserOnBlock) {
             mMediaPickedByUserOnBlock = false;
             List<RNMedia> rnMediaList = new ArrayList<>();
-            for (Media media : mediaList) {
-                rnMediaList.add(new Media(media.getId(), media.getUrl(), media.getType()));
+
+            if (mAppendsMultipleSelectedToSiblingBlocks && 1 < mediaList.size()) {
```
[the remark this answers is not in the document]
mkevins: Good suggestion! I've added a clarifying comment, and also a note about when in the future this may no longer be needed (after image block <-> gallery block transforms are implemented on mobile).

react-native-gutenberg-bridge/android/src/main/java/org/wordpress/mobile/WPAndroidGlue/WPAndroidGlueCode.java
malinajirka: Do we need to add it also to `requestMediaPickFrom(..)`  or is it used for a different scenario?

mkevins: Right! Nice catch! Updated.

mkevins:
@malinajirka thanks for pairing / collaborating on this :+1:

  @dratwas :wave: thanks for reviewing. I've set the `mAppendsMultipleSelectedToSiblingBlocks` flag explicitly to false in the other media selection paths. I've also left the set-to-false assignment after the for each loop to guard, in case a future media selection intent is added without the explicit assignment (so the default will always be to respect the `allowMultipleSelection` flag). This is ready for another pass.

mkevins:
I had to make [one additional change](https://github.com/wordpress-mobile/gutenberg-mobile/pull/1621/commits/3bcdcf63c05684f40830f14f546e6b9184f877f4) to make the enum - mimeType truncation work (the comparison / truncation should be all lowercase).
I tested with Media Library (selecting single and multiple images), and also with uploading from device for the image block. For the uploading flow, I also tested leaving the editor while the upload is in progress, and opening the post after the upload is finished.

PR #1630, opened 2019-11-27
Improve code block style

### LINKED ISSUE #368: Improve style of code block
There's not a lot of difference with the paragraph block, except for the monospace font, and some wrong margins:

img_419a975c4e0c-1

The web version for comparison:

img_36271d76b0eb-1

SergioEstevao:
Fixes #368 

This PR improves the style of the code block by:
 - Change the font to Menlo size 14pt
 - Add paddings has indicated in the ticket
 - Adds a border like in the web block

CodeDarkLight

Gutenberg PR: https://github.com/WordPress/gutenberg/pull/18785

To test:
 - Open the demo app
 - Create a new code block
 - Edit the block and see if all looks good

Update release notes:

### DIFF
--- a/src/_native.ios.scss
+++ b/src/_native.ios.scss
@@ -1,7 +1,7 @@
 /** @format */
 
 // Fonts
-$default-monospace-font: courier;
+$default-monospace-font: menlo;
 $default-regular-font: 'Noto Serif';
 $title-block-padding-top: 12;
 $title-block-padding-bottom: 12;


iamthomasbishop:
👋 Nice work, @SergioEstevao, I have updated the block blueprint - here is the latest:

image

* For the body background, we can use the same as the `pre` block
* No border
* Not sure what you've got applied here, but padding within the content area can be set to: 12px on top/bottom, 16px on left/right

Looks good to me other than those things! 👍

SergioEstevao:
@iamthomasbishop , thanks for the feedback, one question though if I apply your suggestion what will make it different from the pre block visually? 

After you write some text they will look exactly the same in the editor, no?

iamthomasbishop:
>  if I apply your suggestion what will make it different from the pre block visually?

The main diff is the background fill color. In your example the background looks white, and the block content area should have a dim color fill (one of the light grays in light mode, a dark gray in dark mode) to match the pre block style.

> After you write some text they will look exactly the same in the editor, no? 

In terms of the contents of the block, static and selected states look the same, just like other text-based blocks.

SergioEstevao:
@koke design is updated to follow @iamthomasbishop  recomendations, can you please give it another look?

SergioEstevao:
@koke I didn't change the font in Android, that still specifies monospace, that I'm assuming defaults to Roboto Mono.

@iamthomasbishop do we want Menlo in Android to?

iamthomasbishop:
@SergioEstevao I was assuming we'd use Menlo on iOS (considering it's either the default or at least an _available_ font on iOS) and fall back to the default Roboto Mono on Android. Are we able to use Menlo on Android as well?

SergioEstevao:
@iamthomasbishop just confirmed with @maxme that for Android there is only one mono font by default and it's Roboto Mono. So we will keep that in Android and use Menlo in iOS.

So all is ok with PR and I will merge it :).

iamthomasbishop:
Sounds good! :shipit:

PR #1637, opened 2019-11-28
Do no escape strings with double quotes in android strings.xml

Tug:
Wrapping strings in double quotes such as
```
<string name="my_string">"\n    This is my string.\nHello"</string>
```
Will import the string in glotpress as `"\n    This is my string.\nHello"` instead of `\n    This is my string.\nHello`.
In order not to update glotpress and translate.wordpress.org which would be costly, let's revert to using simple escaping of xml entities and not use `""`. In theory we should only lose the ability to have leading and trailing whitespace characters in a string, which is not something that happens frequently. So the resulting sting in glotpress will be `This is my string.\nHello`.

To make sure the created xml is valid we run `xmllint` automatically once the npm script is complete.

### To test:
- Check that XML looks valid in `bundle/android/strings.xml`
- Optional: copy-paste all the strings from `bundle/android/strings.xml` into `WordPress/src/main/res/values/strings.xml` in WPAndroid and build and run the app with those strings. You can boot the block editor in debug mode to see `translations` being printed in the console
- Optional: Try to edit `gutenberg-android.pot`, adding special characters to it such as `<`, `>`, `"`, `\n`..., then run `./bin/po2android.js gutenberg-android.pot bundle/android/strings.xml`, it should create a valid xml file. Validate file with `yarn genstrings:android:check`

Update release notes:


hypest:
> Try to edit `gutenberg-android.pot`

👋 @Tug , where is that file?

Tug:
@hypest It's at the root of the project, you should have it after running `yarn genstrings` once

hypest:
Tried out manually adding ` \ ` (with the spaces) and ` \\ ` in the .pot file and what I see in the output strings.xml is that the single backslash is just gone, while the double-backslash is preserved as is. Should I have seen them be replaced with ` \\ ` and ` \\\\ ` instead?

Tug:
Actually the POT format supports character escaping using backslash so `\` in a pot file is actually a special character that needs to be escaped to `\\` in order to register a `\`.

So the POT parser will have removed those when the string is given to `escapeResourceXML`. This function will simply add them back before we dump the string in the XML/

So basically: (POT => XML)
`\ ` => ` `
`\n` => `\n`
`\\` => `\\`

hypest:
I see. The PR looks good to me @Tug , let's bring it up to date with develop to fix the conflicts and I'll do last review pass afterwards.

Tug:
@hypest updated!

hypest:
Chances are, the failing Android device tests are due to a recent issue (https://github.com/wordpress-mobile/gutenberg-mobile/issues/1703), which is now fixed. Can you update again feom develop @Tug ? Thanks!

Tug:
I merged in `develop` but e2e tests are still failing. The error seems unrelated to my changes though

PR #1649, opened 2019-12-04
[e2e] improve e2e iOS tests

dratwas:
I think we have a few redundant actions in `test:e2e:ios:local`, they increase the time of build steps.

the flow of script to run local e2e tests for iOS:
`"test:e2e:ios:local": "yarn test:e2e:build-app:ios && yarn test:e2e:install-app:ios && TEST_RN_PLATFORM=ios yarn device-tests:local"`

- `yarn test:e2e:build-app:ios does` :
1. Build the app by the `react-native run-ios` script with `--no-packager` flag and `--configuration Release` (it builds JS bundle and copy it to the app in Build Phases script)
2. Build the JS bundle by `react-native bundle` one more time and copy it to the app (it is already there from step 1)
3. Run `cd ./ios/build/gutenberg/Build/Products/Release-iphonesimulator && zip -r $WORK_DIR/ios/Gutenberg.app.zip gutenberg.app` it is needed only for CI to upload the app to the sauce labs (this scripts from package.json is not used in ci config https://github.com/wordpress-mobile/gutenberg-mobile/blob/develop/.circleci/config.yml#L137)

- `yarn test:e2e:install-app:ios`  runs only `yarn test:e2e:build-app:ios` under the hood, so basically runs steps 1, 2, 3 one more time.

- `TEST_RN_PLATFORM=ios yarn device-tests:local` runs appium tests

I think we could make it much faster by removing points 2 and 3 from `yarn test:e2e:build-app:ios`  and remove the `yarn test:e2e:install-app:ios`

Since we use caching of app builds on CI but we want to build JS bundle each time, I also set the flag `SKIP_BUNDLING` in CI config to skip creating of JS bundle during the build step. JS bundle is created in the next step

To test:
1. Local tests
- remove `ios/build` directory (just to be sure)
- run `yarn test:e2e:ios:local` 
- the app should be built and all tests should pass

2. Green CI

Update release notes:


SergioEstevao:
@dratwas thanks for this! Do you mind to update with the latest develop? We just merged some changes that upgrade the version of Appium and some configuration and it will good to make sure these changes play well together.

dratwas:
Hi @SergioEstevao sorry for the delay. Today I checked all my PRs and found this one. I updated it with the develop branch and should be still valid :) Could you please review it one more time?

SergioEstevao:
@dratwas Because we now are using CocodPods for the build process should we add: `yarn preios` to `yarn test:e2e:build-app:ios` to make sure we have the right pods and workspace setup?

dratwas:
Yes! You're right it should be run before. Let me add it :) 

Btw. WDYT about changing the `preios` to 
```
cd ios && (bundle check --path=vendor/bundle > /dev/null || bundle install) && bundle exec pod install --repo-update
```

to use the `--repo-update` flag instead of running `bundle exec pod repo update`

PR #1680, opened 2019-12-13
Master to develop post release v1.19.0

cameronvoell:
Bringing updates from [release v.19.0 ](https://github.com/wordpress-mobile/gutenberg-mobile/pull/1667)back into develop. Gutenberg reference updated to latest release commit on master branch. 

PR submission checklist:


hypest:
👋 Cameron! At this point, and judging from the commit history, it looks to me that the JS bundles are not corresponding to the code so, I'd suggest creating a new set of them to merge. Other than this, I think the PR is ready to merge and only contains trivial diff compared to `develop`. WDYT?

cameronvoell:
> 👋 Cameron! At this point, and judging from the commit history, it looks to me that the JS bundles are not corresponding to the code so, I'd suggest creating a new set of them to merge. Other than this, I think the PR is ready to merge and only contains trivial diff compared to `develop`. WDYT?

Yea I agree since we're bringing in bundle updates they should correspond with the code in the PR. Agreed we're now ready to merge .

Issue #1662, opened 2019-12-09
Testing: Some UI tests are flaky


rachelmcr:
Some of our UI tests have flaky results, including on CI, leading to unreliable test results. A couple possible solutions we've discussed:

- Improve reliability of the tests themselves, to reduce flakiness.
- Rerun the tests on CI automatically and only report repeated failures.

Internal ref: p9ugOq-Nl-p2/#comment-1589

rachelmcr:
As reported in https://github.com/wordpress-mobile/gutenberg-mobile/pull/1615#issuecomment-562705259 the `rotate device and continue adding blocks` test has some flaky failures. An example of the test result on one of those failed test runs:

```
Error: expect(received).toBe(expected) // Object.is equality

- Expected
+ Received

  <!-- wp:paragraph -->
  <p>the finer continuum interprets the polynomial rabbit. when can the geology runs? an astronomer runs. should a communist consent?</p>
  <!-- /wp:paragraph -->
  
  <!-- wp:paragraph -->
- <p>the finer continuum interprets the polynomial rabbit. when can the geology runs? an astronomer runs. should a communist consent?</p>
+ <p>the finer continuum interprets the polynomial rabbit. when can the geology runs? an astronomer runs. should a communist cons</p>
  <!-- /wp:paragraph -->
    at EditorPage.toBe (/home/circleci/project/__device-tests__/pages/editor-page.js:124:32)
    at tryCatch (/home/circleci/project/node_modules/@babel/runtime/node_modules/regenerator-runtime/runtime.js:45:40)
    at Generator.invoke [as _invoke] (/home/circleci/project/node_modules/@babel/runtime/node_modules/regenerator-runtime/runtime.js:271:22)
    at Generator.prototype.<computed> [as next] (/home/circleci/project/node_modules/@babel/runtime/node_modules/regenerator-runtime/runtime.js:97:21)
    at tryCatch (/home/circleci/project/node_modules/@babel/runtime/node_modules/regenerator-runtime/runtime.js:45:40)
    at invoke (/home/circleci/project/node_modules/@babel/runtime/node_modules/regenerator-runtime/runtime.js:135:20)
    at /home/circleci/project/node_modules/@babel/runtime/node_modules/regenerator-runtime/runtime.js:145:13
    at tryCallOne (/home/circleci/project/node_modules/promise/lib/core.js:37:12)
    at /home/circleci/project/node_modules/promise/lib/core.js:123:15
[...]

PR #1700, opened 2019-12-19
Fix image preview with small image size selected

etoledom:
This PR fixes an issue that happens when the user selects a thumbnail (or any smaller image size), and then press on it so see it full screen (this happens on self-hosted sites).

`Gutenberg` side PR: https://github.com/WordPress/gutenberg/pull/19247
`WPiOS` side PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/13145

Expected: The user sees the full sized image.
Result: The user sees the smaller, low quality version of the image.

Also on this PR:
- Sending both the current image url (sized) and original url (full size). This is not currently used for image preview, but will be used for Image Editing. Since this is iOS only (so far), the Android version of it will still receive only one parameter, but still fixing the original issue. 

**Note:** I'm planning/thinking about renaming the `requestImageFullscreenPreview` method action to something more generic, like `requestImageAction`, with an action parameter that could be `preview`, `edit`, etc... so we can reuse it for various future actions.

small_preview

To test:
- Open Gutenberg on WPiOS, on a self-hosted site.
- Add a new image.
- Change its size to thumb.
- Tap on the image to show the image preview.
- Check that the image shown is the original full size image.

PR submission checklist:


react-native-gutenberg-bridge/index.js
```diff
@@ -101,8 +101,11 @@ export function getOtherMediaOptions( filter, callback ) {
 	return RNReactNativeGutenbergBridge.getOtherMediaOptions( filter, callback );
 }
 
-export function requestImageFullscreenPreview( mediaUrl ) {
-	return RNReactNativeGutenbergBridge.requestImageFullscreenPreview( mediaUrl );
+export function requestImageFullscreenPreview( currentImageUrl, originalImageUrl ) {
+	if ( isIOS ) {
+		return RNReactNativeGutenbergBridge.requestImageFullscreenPreview( currentImageUrl, originalImageUrl );
+	}
+	return RNReactNativeGutenbergBridge.requestImageFullscreenPreview( originalImageUrl || currentImageUrl );
```
cameronvoell: tested, fix looks good on Android as well 👍. I was initially surprised this change didn't require android changes as well, but this works great.

cameronvoell:
> Note: I'm planning/thinking about renaming the requestImageFullscreenPreview method action to something more generic, like requestImageAction, with an action parameter that could be preview, edit, etc... so we can reuse it for various future actions.

If you'd like to coordinate PRs to update this, I can do the WP Android one, let me know.

react-native-gutenberg-bridge/index.js
etoledom: On android works by sending the big image url even if we are showing a smaller version. No native changes needed.

We wouldn't need changes on iOS either, but these changes aim to help on the image editing work.

On the future, we will need these changes on Android too, since the plan is to implement image editing on Android too, but still it would be cool to use the same API on Android soon. 👍

etoledom:
The failing test on Android on Device is a known [CI flakiness issue](https://github.com/wordpress-mobile/gutenberg-mobile/issues/1662#issuecomment-576325584).

Merging directly.

etoledom:
> > Note: I'm planning/thinking about renaming the requestImageFullscreenPreview method action to something more generic, like requestImageAction, with an action parameter that could be preview, edit, etc... so we can reuse it for various future actions.
> 
> @etoledom If you'd like to coordinate PRs to update this, I can do the WP Android one, let me know.

Thank you! We will need that at some point soon 👍

Issue #1696, opened 2019-12-18
Focus loop on Rich Text


pinarol:
**Describe the bug**
I am using the html snippet here:

```
<!-- wp:group -->
<div class="wp-block-group"><div class="wp-block-group__inner-container"><!-- wp:button -->
<div class="wp-block-button"><a class="wp-block-button__link">Button</a></div>
<!-- /wp:button -->

<!-- wp:image {"id":984,"sizeSlug":"large"} -->
<figure class="wp-block-image size-large"><img src="https://pinarspremiumtestsite123443.files.wordpress.com/2019/12/img_0111.jpg?w=1024" alt="" class="wp-image-984"/></figure>
<!-- /wp:image -->

<!-- wp:paragraph -->
<p>Faded from</p>
<!-- /wp:paragraph -->

<!-- wp:group -->
<div class="wp-block-group"><div class="wp-block-group__inner-container"></div></div>
<!-- /wp:group -->

<!-- wp:group -->
<div class="wp-block-group"><div class="wp-block-group__inner-container"><!-- wp:button -->
<div class="wp-block-button"><a class="wp-block-button__link">Button 2</a></div>
<!-- /wp:button --></div></div>
<!-- /wp:group --></div></div>
<!-- /wp:group -->

<!-- wp:media-text {"mediaId":990,"mediaType":"image"} -->
<div class="wp-block-media-text is-stacked-on-mobile"><figure class="wp-block-media-text__media"><img src="https://pinarspremiumtestsite123443.files.wordpress.com/2019/12/img_0111-1.jpg" alt="" class="wp-image-990"/></figure><div class="wp-block-media-text__content"><!-- wp:paragraph -->
<p></p>
<!-- /wp:paragraph --></div></div>
<!-- /wp:media-text -->

<!-- wp:group -->
<div class="wp-block-group"><div class="wp-block-group__inner-container"></div></div>
<!-- /wp:group -->

<!-- wp:video -->
<figure class="wp-block-video"></figure>
<!-- /wp:video -->

<!-- wp:image {"id":987,"sizeSlug":"large"} -->
<figure class="wp-block-image size-large"><img src="https://pinarspremiumtestsite123443.files.wordpress.com/2019/12/img_0005-1-1.jpg?w=1024" alt="" class="wp-image-987"/></figure>
<!-- /wp:image -->

```
--->>> Make sure autocorrect is on

Select one of the sub-blocks of Group block
Add an image block in it, add some image
[...]

PR #1698, opened 2019-12-19
Fix focus loop when autocompleting text on iOS

koke:
See https://github.com/WordPress/gutenberg/pull/19234
Fixes #1696 

PR submission checklist:

- [x] I have considered adding unit tests where possible.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #1702, opened 2019-12-19
[FIX] Rich Text loop when autocompleting text on iOS

dratwas:
Fixes #
Gutenberg PR: https://github.com/WordPress/gutenberg/pull/19240
Fixes: https://github.com/wordpress-mobile/gutenberg-mobile/issues/1696

In this PR I created a local variable for focus state instead of using `TextInputState`, thanks to that we do not call `focus()` in `onFocus` which was a bit weird. I also pass the latest native value to `onBlur` event to update JS value if that is not correct.

PR submission checklist:

### DIFF
--- a/react-native-aztec/src/AztecView.js
+++ b/react-native-aztec/src/AztecView.js
@@ -1,6 +1,7 @@
 import PropTypes from 'prop-types';
 import React from 'react';
 import ReactNative, {requireNativeComponent, TextViewPropTypes, UIManager, ColorPropType, TouchableWithoutFeedback, Platform} from 'react-native';
+import TextInputState from 'react-native/lib/TextInputState';
 
 const AztecManager = UIManager.getViewManagerConfig('RCTAztecView');
 
@@ -34,6 +35,14 @@ class AztecView extends React.Component {
     ...TextViewPropTypes, // include the default view properties
   }
 
+  componentDidMount() {
+    TextInputState.registerInput(ReactNative.findNodeHandle(this))
+  }
+
+  componentWillUnmount() {
+    TextInputState.unregisterInput(ReactNative.findNodeHandle(this))
+  }
+
   dispatch(command, params) {
     params = params || [];
     UIManager.dispatchViewManagerCommand(
@@ -57,6 +66,10 @@ class AztecView extends React.Component {
   }
 
   _onEnter = (event) => {
+    if (!this.isFocused()) {	
+      return;	
+    }
+
     if (!this.props.onEnter) {
       return;
     }
@@ -87,18 +100,16 @@ class AztecView extends React.Component {
   }
 
   _onFocus = (event) => {
-    this.focused = true;
     const { onFocus } = this.props;
-
     if (onFocus) {
       onFocus(event);
     }
   }
   
   _onBlur = (event) => {
-    this.focused = false;
     this.selectionEndCaretY = null;
     const { onBlur } = this.props;
+    TextInputState.blurTextInput(ReactNative.findNodeHandle(this));
 
     if (onBlur) {
       onBlur(event);
@@ -106,13 +117,13 @@ class AztecView extends React.Component {
   }
 
   _onSelectionChange = (event) => {
-    if (this.focused && this.props.onSelectionChange) {
+    if (this.props.onSelectionChange) {
       const { selectionStart, selectionEnd, text } = event.nativeEvent;
       const { onSelectionChange } = this.props;
       onSelectionChange( selectionStart, selectionEnd, text, event );
     }
 
-    if (this.focused && this.props.onCaretVerticalPositionChange && 
+    if (this.props.onCaretVerticalPositionChange && 
       this.selectionEndCaretY != event.nativeEvent.selectionEndCaretY) {
         const caretY = event.nativeEvent.selectionEndCaretY;
         this.props.onCaretVerticalPositionChange( event.target, caretY, this.selectionEndCaretY );
@@ -121,31 +132,23 @@ class AztecView extends React.Component {
   }
 
   blur = () => {
-    if (this.focused) {
-      if (Platform.OS === "ios") {
-        UIManager.blur(ReactNative.findNodeHandle(this))
-      } else {
-        this.dispatch(UIManager.getViewManagerConfig('AndroidTextInput').Commands.blurTextInput)
-      }
-    }
-    this.focused = false;
+    TextInputState.blurTextInput(ReactNative.findNodeHandle(this));
   }
 
   focus = () => {
-    if (!this.focused) {
-      if (Platform.OS === "ios") {
-        UIManager.focus(ReactNative.findNodeHandle(this))
-      } else {
-        this.dispatch(UIManager.getViewManagerConfig('AndroidTextInput').Commands.focusTextInput)
-      }
-    }
-    this.focused = true;
+    TextInputState.focusTextInput(ReactNative.findNodeHandle(this));
   }
 
-  isFocused = () => this.focused
+  isFocused = () => {
+    const focusedField = TextInputState.currentlyFocusedField();	
+    return focusedField && ( focusedField === ReactNative.findNodeHandle(this) );
+  }
 
   _onPress = (event) => {
-		this._onFocus(event); // Check if there are listeners set on the focus event
+    if( !this.isFocused() ) {
+      this.focus();
+      this._onFocus(event); // Check if there are listeners set on the focus event
+    }
   }
 
   _onAztecFocus = (event) => {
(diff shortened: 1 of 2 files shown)


mchowning:
Do we have two PRs fixing the same issue? cc: @koke 

https://github.com/wordpress-mobile/gutenberg-mobile/pull/1698

pinarol:
Yes we do @mchowning , we are trying both approaches :)

react-native-aztec/src/AztecView.js
```diff
@@ -91,59 +87,65 @@ class AztecView extends React.Component {
...
+      } else {
+        this.dispatch(UIManager.getViewManagerConfig('AndroidTextInput').Commands.blurTextInput)
+      }
+    }
+    this.focused = false;
   }
 
   focus = () => {
-    TextInputState.focusTextInput(ReactNative.findNodeHandle(this));
+    if (!this.focused) {
```
pinarol: Could you help me understand why we need to change here? how this is giving a different result than the check TextInputState uses? 

```
function focusTextInput(textFieldID: ?number) {
  if (currentlyFocusedID !== textFieldID && textFieldID !== null) {
  ...//focus
  }
}
```

react-native-aztec/src/AztecView.js
```diff
@@ -91,59 +87,65 @@ class AztecView extends React.Component {
...
+      onBlur(event);
     }
-
-    const { onBlur } = this.props;
-    onBlur(event);
   }
 
   _onSelectionChange = (event) => {
-    if ( this.props.onSelectionChange ) {
+    if (this.focused && this.props.onSelectionChange) {
```
pinarol: Aren't we able to do the same thing with the old function?

isFocused = () => {		 
     const focusedField = TextInputState.currentlyFocusedField();		
     return focusedField && ( focusedField === ReactNative.findNodeHandle(this) );		
   }

react-native-aztec/src/AztecView.js
```diff
@@ -91,59 +87,65 @@ class AztecView extends React.Component {
...
 
-  isFocused = () => {
-    const focusedField = TextInputState.currentlyFocusedField();
-    return focusedField && ( focusedField === ReactNative.findNodeHandle(this) );
-  }
+  isFocused = () => this.focused
 
   _onPress = (event) => {
-	if ( ! this.isFocused() ) {
-		this.focus(event); // Call to move the focus in RN way (TextInputState)
```
pinarol: I am a bit afraid of removing this code since it is [really old](https://github.com/wordpress-mobile/react-native-aztec/pull/92). I haven't had the chance to test this with Android yet but this seems like a fix for a possible keyboard dismissing issue on Android.

pinarol:
I realized that this PR is breaking auto-scroll on iOS, because keyboard-aware-scroll-view trusts TextInputState to find out the currently focused TextInput and then it adjusts the scroll position accordingly. 

This is PR is making the global TextInputState not trustable anymore so it can have lots of other side effects as well. 

So how about we just add onBlur fix onto @koke’s [PR](https://github.com/WordPress/gutenberg/pull/19234)?

dratwas:
Yeah, right we need the `TextInputState`. However, I still think we should change the way we use it.

The paint point is that there is no method `setCurrentlyFocusedID` and we need to call `TextInputState.focusTextInput` in `_onPress` instead.
https://github.com/wordpress-mobile/gutenberg-mobile/blob/develop/react-native-aztec/src/AztecView.js#L144
The `TextInputState.focusTextInput`  sets `currentlyFocusedID` for us but also calls `UIManager.focus` which tries to focus our input.
https://github.com/facebook/react-native/blob/master/Libraries/Components/TextInput/TextInputState.js#L52
We do the same for `_onBlur`. We call `TextInputState.blurTextInput` which clears the `currentlyFocusedID` but also tries to blur the input.
https://github.com/facebook/react-native/blob/master/Libraries/Components/TextInput/TextInputState.js#L73

My point is I think there is no reason to call `focus` in `_onFocus` callback and `blur` in `_onBlur` callback. We should only set the correct ID since we already know that the input is focused/blurred.

Let's imagine this kind of scenario:
- Focus input and write something as we did with autocorrection (with different length)
- Click on "Move up" button - `blur()` is called
- `onSelectionChange()` is called because suggestion has a different length than original text - `focus()` is called
- `_onBlur()` is called - remember that in the meantime the `focus()` was called and input is focused and in the current implementation we call `blur()` in `_onBlur()`
- `_onFocus()` is called - like above - the `blur()` was called in `_onBlur()` so input is blurred and we call `focus()` in `_onFocus()`
- _onBlur() is called - input is focused - we call one more time `blur()`
-`_onFocus()` is called - input is blurred - we call `focus()`

And the loop starts. It's because we call `blur()` in` _onBlur()` and `focus()` in `onFocus()`  instead of setting the correct `currentlyFocusedID` w/o focusing/bluring in `TextInputState`.


I added this function to `TextInputState` and used it just to check if it will fix these issues and seems like yes.
```
function setCurrentlyFocusedFieldID(textFieldID: number) {
  if (textFieldID != currentlyFocusedID) {
    currentlyFocusedID = textFieldID;
  }
}
```

Good news is that there are actually functions `blurField` and `focusField` on the `master` branch of `react-native` and they do what we need :slightly_smiling_face: https://github.com/facebook/react-native/blob/master/Libraries/Components/TextInput/TextInputState.js#L31-L41

koke:
Those new methods in react-native are nice and will likely make this more robust, but I think that today's problem lies in here:

> `onSelectionChange()` is called because suggestion has a different length than original text - `focus()` is called

A change on selection shouldn't trigger a change in focus. This kind of makes sense on the web since selection is global across the whole document, but on native, selection is specific to each view, so that's making things confusing:

- The selection changes in the text view because of the autocorrection
- `onSelectionChange()` is called on `RichText` because the selection has changed
- Normally, `RichText` would propagate that to the store with `this.props.onSelectionChange`, which is the other kind of selection global to the document. The store then sees the new selection and understands that the selection is back on the text view, which ends up in it requesting focus again.

We still want to receive the first `onSelectionChange()` to keep the internal state of `RichText` updated, but we don't want to update the global selection if that's not the currently selected view.

react-native-aztec/src/AztecView.js
dratwas: I already reverted those changes

dratwas:
Make sense. I reverted those changes related to the `TextInputState` and added `registerInput` and `unregisterInput`. I also cherry-picked @koke's changes in `gutenberg` PR https://github.com/WordPress/gutenberg/pull/19240

react-native-aztec/src/AztecView.js
```diff
@@ -34,6 +34,14 @@ class AztecView extends React.Component {
     ...TextViewPropTypes, // include the default view properties
   }
 
+  componentDidMount() {
+    TextInputState.registerInput(ReactNative.findNodeHandle(this))
```
pinarol: Could you explain a bit how this registerInput/unregisterInput works and how do they contribute to the solution?

dratwas: TBH I don't think that they are needed in that solution but I added it because I realized that we don't to that at this moment. They add or remove `fieldID` to/from `Set` of inputs in `TextInputState` 
https://github.com/facebook/react-native/blob/master/Libraries/Components/TextInput/TextInputState.js#L85-L95

W/o this the `false` will be returned when `isTextInput` is called.
https://github.com/facebook/react-native/blob/master/Libraries/Components/TextInput/TextInputState.js#L93-L95

pinarol:
the new solution looks a lot better but I come across with keyboard close button issue. It is not working in certain scenarios:

- create a new post
- write something
- tap keyboard close button

keyboard-close3

- open a post with media text
- type something on text area
- delete the text 
- tap delete button once again so the paragraph block is also deleted
- focus on text area again
- tap keyboard close button

keyboard-close

pinarol:
But the issues related to focus loop on auto correct is fixed.

I tested both cases:

text from autocorrection has the same length as written text:

- type something that auto-correct wants to correct
- make sure suggested text has **same** length
- tap move down
- see that auto-correct made the change
- switch to html mode
- verify text is matching ✅ 

text from autocorrection has a different length as written text

- type something that auto-correct wants to correct
- make sure suggested text has **different** length
- tap the image
- see that auto-correct made the change
- verify there's no focus loop ✅

react-native-aztec/src/AztecView.js
pinarol: This sounds really nice but I am a bit afraid about consequences. We don't know who is calling `isTextInput` and what could be done differently if it turns true this time. So I'd avoid adding it unless we have enough context. I am just trying to be extra cautious since this has been a pretty fragile component so far.

dratwas: Ok, I understand your worries so I will remove it then :)

pinarol:
The issue about keyboard close button is fixed! 🎉

Issue #1705, opened 2019-12-20
[Android] Crash when the video library picks up newer RN version


hypest:
**Describe the bug**
Demo app crashes on Android dues to the `react-native-video` lib picking up a newer-than-supported version of react-native from our maven mirror.

Stacktrace:

```
2019-12-20 14:16:41.637 23174-23174/com.gutenberg E/AndroidRuntime: FATAL EXCEPTION: main
    Process: com.gutenberg, PID: 23174
    java.lang.NoSuchMethodError: No interface method pushMap(Lcom/facebook/react/bridge/ReadableMap;)V in class Lcom/facebook/react/bridge/WritableArray; or its super classes (declaration of 'com.facebook.react.bridge.WritableArray' appears in /data/app/com.gutenberg-OJuYadh646qOe814qf9Ing==/base.apk!classes34.dex)
        at com.brentvatne.exoplayer.ReactExoplayerView.getAudioTrackInfo(ReactExoplayerView.java:657)
        at com.brentvatne.exoplayer.ReactExoplayerView.videoLoaded(ReactExoplayerView.java:634)
        at com.brentvatne.exoplayer.ReactExoplayerView.onPlayerStateChanged(ReactExoplayerView.java:602)
        at com.google.android.exoplayer2.ExoPlayerImpl$PlaybackInfoUpdate.notifyListeners(ExoPlayerImpl.java:803)
        at com.google.android.exoplayer2.ExoPlayerImpl.updatePlaybackInfo(ExoPlayerImpl.java:717)
        at com.google.android.exoplayer2.ExoPlayerImpl.handlePlaybackInfo(ExoPlayerImpl.java:648)
        at com.google.android.exoplayer2.ExoPlayerImpl.handleEvent(ExoPlayerImpl.java:593)
        at com.google.android.exoplayer2.ExoPlayerImpl$1.handleMessage(ExoPlayerImpl.java:125)
        at android.os.Handler.dispatchMessage(Handler.java:106)
        at android.os.Looper.loop(Looper.java:193)
        at android.app.ActivityThread.main(ActivityThread.java:6718)
        at java.lang.reflect.Method.invoke(Native Method)
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:493)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:858)
```

PR #1706, opened 2019-12-20
Libs forks need to use node_modules when building locally

### LINKED ISSUE #1703: RNSVGPath Crash on Start of Example App
**Describe the bug**
Red banner on startup in the example app on Android. 

**To Reproduce**
1. Start the example gutenberg-mobile Example App. 
2. Notice the red banner mentioning RNSCGPath, seen in the screenshot below.

**Screenshots**
image

**Additional context**
This error was showing up on branches/commits that were previously working, and is likely related to a change in the build environment/workflow. This error is showing on Android only.

hypest:
Fixes #1703, #1705 

On Android, some of the npm libraries we're forking are using the Maven mirror of React Native both when building them as standalone libs (on JitPack) and when built locally as normal npm deps. That wouldn't be a problem per se but, the RN version referenced in the local builds is the "latest' ([see this for example](https://github.com/wordpress-mobile/react-native-svg/blob/f155fbdd309f5e28bdbb326b819c5bf5af6f4cc9/android/build.gradle#L49)) so, when a newer version becomes available in the maven mirror it gets picked up.

We recently made available in the maven mirror some RN versions that are unfortunately incompatible with the versions expected by the forks, leading to crashes.

The versions used in local builds need to come from the local node_modules folder, and this PR brings in updates to the forks to do just that.

Related PRs:
`react-native-svg`:  https://github.com/wordpress-mobile/react-native-svg/pull/8
`react-native-video`:  https://github.com/wordpress-mobile/react-native-video/pull/2
`react-native-slider`: https://github.com/wordpress-mobile/react-native-slider/pull/1

To test:

1. Demo app should work as normal, including showing icons and videos

2. Pointing WPAndroid's gutenberg-mobile submodule to this PR should also work for all build modes and variants

PR submission checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -139,7 +139,7 @@
   "dependencies": {
     "@babel/runtime": "^7.3.1",
     "@react-native-community/cli": "^1.5.2",
-    "@react-native-community/slider": "git+https://github.com/wordpress-mobile/react-native-slider.git#e20f268e61327321a5bbd66afc5238388d65f2a4",
+    "@react-native-community/slider": "git+https://github.com/wordpress-mobile/react-native-slider.git#1ac7533fad0eb7f091d1ea501010b113efa2e5f8",
     "classnames": "^2.2.5",
     "dom-react": "^2.2.1",
     "domutils": "^1.7.0",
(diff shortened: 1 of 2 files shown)


hypest:
This PR got merged before the various fork PRs got merged so, will create a new PR to point to merged versions.

Issue #1675, opened 2019-12-12
Dark Mode - Android integration


maxme:
We want to support dark mode for Android like it was done recently for iOS.

The work is in progress in this wpandroid branch:
https://github.com/wordpress-mobile/WordPress-Android/tree/feature/material-theme - cc @khaykov 

Some color name and value changes might be required to align with the new color system - cc @iamthomasbishop 

Related iOS Dark mode PR: https://github.com/wordpress-mobile/gutenberg-mobile/pull/1309

marecar3:
Thanks, @maxme!

PR #1717, opened 2019-12-21
Feature/1675 dark mode android integration

marecar3:
Fixes https://github.com/wordpress-mobile/gutenberg-mobile/issues/1675

Gutenberg PR: https://github.com/WordPress/gutenberg/pull/19293
WordPress-Android PR: https://github.com/wordpress-mobile/WordPress-Android/pull/11004

ezgif com-video-to-gif (13)

To test:

On devices with API <=28:
1. Open App Settings from Me screen.
2. Notice the App Theme preference.
3. The default value should be set to `Set by Battery Saver`.
4. From the preference, dialog select `Dark` and confirm that the app switches to Dark Theme.
5. Open the Gutenberg Editor and confirm that it has Dark Theme
6. Select `Light` and confirm that the app switches to Light Theme.
7. Open the Gutenberg Editor and confirm that it has Light Theme
8. Select `Set by Battery Saver` and enable Battery Saver on the device.
9. Confirm that Dark Theme is used when Battery Saver is ON.

On devices with API 29+:
1. Open App Settings from Me screen.
2. Notice the App Theme preference.
3. The default value should be set to `System Default`.
4. From the preference, dialog select `Dark` and confirm that the app switches to Dark Theme.
5. Open the Gutenberg Editor and confirm that it has Dark Theme
6. Select `Light` and confirm that the app switches to Light Theme.
7. Open the Gutenberg Editor and confirm that it has Light Theme
8. Select `System Default` and close the app.
9. Enable Dark Theme in Display settings of the device.
10. Launch the app again and confirm that the dark theme is applied.

Note for API 29+: At this moment we aren't able to get the app theme to switch automatically when the Dark Theme is toggled. For now, you need to restart the app for new theme to take effect. @khaykov has created a separate issue for this at https://github.com/wordpress-mobile/WordPress-Android/issues/10748

PR submission checklist:

- [ ] replace "future release" in release notes with some version


marecar3:
Early feedback from @iamthomasbishop :

- [ ] Android-only: change primary surface background to #121212
- [ ] Link color should be blue 30 (double-check iOS for proper value)
- [ ] When editor is loading, background is white (haven’t seen this in iOS)
- [ ] Bottom sheet icons don't always display as same/correct color (sometimes blueish gray, sometimes neutral. See gallery appender/image sources sheet for example)
- [ ] Slider drag handle is using too dark of a blue (should be same as process fill)

maxme:
>  Link color should be blue 30 (double-check iOS for proper value)

I noticed the iOS / Android color difference on links:

iOS | Android
-|-
Simulator Screen Shot - iPhone 11 - 2019-12-31 at 12 06 02|Screenshot_1577790361

react-native-gutenberg-bridge/android/src/main/java/org/wordpress/mobile/WPAndroidGlue/WPAndroidGlueCode.java
```diff
@@ -305,6 +309,13 @@ public void performRequest(String pathFromJS, Consumer<String> onSuccess, Consum
             public void requestImageFullscreenPreview(String mediaUrl) {
                 mOnImageFullscreenPreviewListener.onImageFullscreenPreviewClicked(mediaUrl);
             }
+
+            @Override
+            public void getPreferredColorScheme(
+                    PreferredColorSchemeReceivedCallback preferredColorSchemeReceivedCallback) {
+                String preferredColorScheme = mOnPreferredColorSchemeListener.getPreferredColorScheme();
+                preferredColorSchemeReceivedCallback.onPreferredColorSchemeReceived(preferredColorScheme);
```
mchowning: I'm curious if we could avoid using a callback here and instead just return the color scheme `String` value directly? I feel like that helps simplify the code. [Something like this](https://github.com/wordpress-mobile/gutenberg-mobile/commit/089181305e88cd4bb782c5d4249e654406a426f0) is what I had in mind.

react-native-gutenberg-bridge/android/src/main/java/org/wordpress/mobile/WPAndroidGlue/WPAndroidGlueCode.java
```diff
@@ -149,6 +149,10 @@ public boolean isContentChanged() {
         void onEditorAutosave();
     }
 
+    public interface OnPreferredColorSchemeListener {
+        String getPreferredColorScheme();
+    }
```
mchowning: Not a big deal, but "Listener" doesn't quite seem right here to me. I think of a "Listener" object as something that has events/updates pushed to it and then responds to those events, not something that provides those updates to others on demand like this interface does. 

To put it another way, I would expect a `PreferredColorSchemeListener` interface would be something like this:

```
   public interface PreferredColorSchemeListener {
        void onColorSchemeChanged(String colorScheme);
    }
``` 

Renaming this interface from `OnPreferredColorSchemeListener` to something like `PreferredColorSchemeSource` or `PreferredColorSchemeProvider` would seems a bit more accurate to me (I'm not remotely invested in either of those suggestions though).

etoledom:
@marecar3 - Could you please update the branches? Thank you!

RELEASE-NOTES.txt
```diff
@@ -1,3 +1,7 @@
+1.21.0
+------
+* [Android] Dark Mode
+
```
etoledom: We should probably target v1.23.0 by now 👍

marecar3: Hey @etoledom, thanks for the feedback.
I just put: "future release" instead of version as we aren't sure when the WPAndroid part will be done. Also, I will update 


> > Link color should be blue 30 (double-check iOS for proper value)
> 
> I noticed the iOS / Android color difference on links:
> 
> iOS	Android
> Simulator Screen Shot - iPhone 11 - 2019-12-31 at 12 06 02	Screenshot_1577790361

This should be fixed.

react-native-gutenberg-bridge/android/src/main/java/org/wordpress/mobile/WPAndroidGlue/WPAndroidGlueCode.java
marecar3: Hey @mchowning 👋 
It seems that I changed a lot of code here so if mind doing another iteration of review?
Thanks a lot!

react-native-gutenberg-bridge/android/src/main/java/org/wordpress/mobile/WPAndroidGlue/WPAndroidGlueCode.java
marecar3: Same here. As I have changed a lot of code we would need another iteration of code review.

marecar3:
> > Link color should be blue 30 (double-check iOS for proper value)
> 
> I noticed the iOS / Android color difference on links:
> 
> iOS	Android
> Simulator Screen Shot - iPhone 11 - 2019-12-31 at 12 06 02	Screenshot_1577790361

Thanks @maxme, It should be fixed now.

react-native-gutenberg-bridge/android/src/main/java/org/wordpress/mobile/ReactNativeGutenbergBridge/RNReactNativeGutenbergBridgePackage.java
```diff
@@ -12,19 +12,23 @@
 public class RNReactNativeGutenbergBridgePackage implements ReactPackage {
     private GutenbergBridgeJS2Parent mGutenbergBridgeJS2Parent;
     private RNReactNativeGutenbergBridgeModule mRNReactNativeGutenbergBridgeModule;
+    private boolean mIsDarkMode;
```
mchowning: Would be nice if we could marm `mIsDarkMode` and `mGutenbergBridgeJS2Parent` as `final`. Would make it clear that `mRNReactNativeGutenbergBridgeModule` is the only thing that can/should change. I do not feel strongly about this change though, feel free to leave as-is if you prefer.

mchowning:
This is looking good @marecar3 !

Noticed that on API<29 devices switching battery mode on/off while within the editor would update the dark mode for the app (i.e. the toolbar on the top would update), but not for the editor. This seems like a bit of an edge case, and I wouldn't mind proceeding with this PR as is and opening a new issue for that (like [what we did with the similar issue on API 29+ devices](https://github.com/wordpress-mobile/WordPress-Android/issues/10748)).

Starting editor with battery Saver OFF | Starting editor with battery saver ON
--- | ---
drak_mode_start_light | dark_mode_switching_api26.

I also left [a comment about a couple of issues I saw on the WPAndroid PR](https://github.com/wordpress-mobile/WordPress-Android/pull/11004#issuecomment-596778555), but those issues are almost certainly not related to your changes.

react-native-gutenberg-bridge/android/src/main/java/org/wordpress/mobile/ReactNativeGutenbergBridge/RNReactNativeGutenbergBridgePackage.java
marecar3: Thanks for the good advice.
Will update them!

marecar3:
Hey @mchowning!

Thanks for the good review! I have found where is the problem and now it should be fixed.

Please, can you do another iteration of testing? Thanks!

RELEASE-NOTES.txt
```diff
@@ -5,10 +5,10 @@
 1.24.0
 ------
 * New block: Latest Posts
-* [Android] Can now scroll post when in landscape orientation with the soft keyboard displayed
```
marecar3: This shouldn't be part of release notes as it was removed: https://github.com/wordpress-mobile/gutenberg-mobile/pull/1989/files#diff-a3d054590c32de57d87be233927e755dL3
But for some reason, it was reverted from revert.

marecar3:
> Looks good! The issue with the editor not immediately switching themes when battery mode is changed is now fixed. Great work @marecar3 !
> 
> I did notice that the WPAndroid PR already got merged. Should we open a new WPAndroid PR once this gets merged so that we can update that to use the merged hash? (might not be necessary since the WPAndroid PR is only targetting a feature branch)

Hey @mchowning, thanks for asking!
@khaykov has opened a new PR https://github.com/wordpress-mobile/WordPress-Android/pull/11469 where we will use a merged hash.

PR #1719, opened 2019-12-23
Scroll down to reveal the Trash buttons

hypest:
Tries to fix some of the flakiness of the e2e tests by scrolling the content to reveal the Trash button.

Another detail is with the `should be able to insert block into post` test, removed the special treatment (removal) of one of blocks on Android since now with the scroll-to-reveal-trash-icon the dedicated removal is not needed.

Tested locally and the Appium tests are green on my Pixel 2 XL, Android 9.

To test:

1. CI should be green

2. Locally running `yarn test:e2e:android:local` and `yarn  test:e2e:ios:local` should be green

PR submission checklist:

### DIFF
--- a/__device-tests__/gutenberg-editor-block-insertion.test.js
+++ b/__device-tests__/gutenberg-editor-block-insertion.test.js
@@ -12,6 +12,7 @@ import {
 	stopDriver,
 	isAndroid,
 	swipeDown,
+	swipeUp,
 	clickMiddleOfElement,
 } from './helpers/utils';
 import testData from './helpers/test-data';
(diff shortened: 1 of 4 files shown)


hypest:
Triggered the CI tests twice ans got green on e6693af. Have now triggered it a third time to see how consistent it is.

hypest:
I've increased the delay before the rotation, in an attempt to have the text setting before it settle. At least, that's my assumption there. I'll trigger the tests a few times to see if that improved reliability or not.

hypest:
Now trying 9f20d23 and e4860a0.

Run No1:
Run No2:
Run No3:

hypest:
FWIW, some else is also experiencing issues with the adb input text command not delivering all the chars: `And on some devices I experienced that a long text (ca. 200 characters) written with adb shell input text "<longText>" was partly lost - even the bigger part at the end! It looked to me depending on the manufacturer.` from https://stackoverflow.com/a/54680722

hypest:
Closing as it has become stale for long time. We can circle back and reopen or update if needed.

PR #1733, opened 2020-01-05
Implement toggle bold and italic shortcuts

### LINKED ISSUE #861: Keyboard shortcuts
Gutenberg should support at least the basic keyboard shortcuts for rich text formatting.

kmdupr33:
Fixes #861 

This pr pairs with another pr that contains [changes to the gutenberg repo](https://github.com/WordPress/gutenberg/pull/19408). It follows the proposal I laid out [here](https://href.li/?https://hogwartsp2.wordpress.com/2019/12/21/shortcuts-implementation-proposal/).

This only contains work for the italics and bold shortcuts because the undo and redo shortcuts are [already working](https://href.li/?https://docs.google.com/spreadsheets/d/1AwWfi2ZmrFRtYt7igFCuhfRubkZbqX5xtkWSwvbnOGU/edit#gid=71534297).

* The main listeners for shortcuts are in the `GutenbergViewController` and `MainActivity` that host the React native app. On iOS, this enables us to leverage some of the native shortcut discoverability features that are on iOS as well as making shortcuts "global" for free. If we agree this is a reasonable choice on iOS, I'll need to add localized strings for shortcut discoverability (see checkbox below). For consistency and for making it easy to make shortcuts global, I also implemented the shortcut listener in the `MainActivity` on android.
* On iOS, since the `GutenbergViewController` is already pretty long, I've created a `ShortcutsCoordinator` in the same spirit as the `Media*Coordinators`. In order for the shortcuts coordinator to do non-trivial work in handling shortcuts, I needed to modifier the responder chain so that it has a chance to handle shortcut events. On Android, the ideal home for this functionality was less obvious, and since the `MainActivity` is pretty lean, I threw some of the shortcut code there for now. I'm totally open to feedback if anyone has any ideas where that code should live.
* A key assumption I made here is that keyboard shortcuts for us are always global and don't have target elements, which is unlike web's implementation. I figured this was kosher, but if I'm wrong about this, I'll need to rethink some things on the js side and on the native side. 

To test:

If you want to test the bold shortcut, its as simple as clicking a textview and pressing modifier-b.

If you want to test italics, you'll have to run the apps in release mode so that the developer menu is disabled and the toggle inspector shortcut doesn't interfere with the toggle italics shortcut. 

PR submission checklist:

### DIFF
--- a/android/app/src/main/java/com/gutenberg/MainActivity.java
+++ b/android/app/src/main/java/com/gutenberg/MainActivity.java
@@ -1,6 +1,13 @@
 package com.gutenberg;
 
+import android.os.Bundle;
+import android.view.KeyEvent;
+
 import com.facebook.react.ReactActivity;
+import com.facebook.react.ReactApplication;
+
+import org.wordpress.mobile.ReactNativeGutenbergBridge.RNReactNativeGutenbergBridgeModule;
+import org.wordpress.mobile.ReactNativeGutenbergBridge.RNReactNativeGutenbergBridgeModule.Shortcut;
 
 public class MainActivity extends ReactActivity {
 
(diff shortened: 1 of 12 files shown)


android/app/src/main/java/com/gutenberg/MainActivity.java
```diff
@@ -12,4 +19,33 @@
...
+    @Override
+    protected void onCreate(Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+        ((ReactApplication) getApplication()).getReactNativeHost()
+                .getReactInstanceManager()
+                .addReactInstanceEventListener(context -> module = context.getNativeModule(RNReactNativeGutenbergBridgeModule.class));
+    }
+
+    @Override
+    public boolean onKeyUp(int keyCode, KeyEvent event) {
```
Tug: I wonder if we will need to add a debouncer here. Just trying to toggle bold formatting makes it toggle twice or more for me (tested on an iOS Simulator).
Also and more worrying is when I keep the modifier key (so command on macos) pressed it keeps repeating the key over and over. so for `cmd+b` it keeps on toggling the selected text in bold, and for `cmd+v` it keeps on pasting the copied text over and over until I release the `cmd` key.
I'll look into it a bit, but maybe you have an idea @kmdupr33 ?

Tug: Just tested in `develop` and it looks like a regression, at least on iOS

kmdupr33: Interesting. Looking into this now.

kmdupr33: Ok. This is resolved 👍

hypest:
Closing this as not updated in a while.

Issue #1747, opened 2020-01-06
Cherry pick "max depth" fix on our react-native fork


pinarol:
Let's first check if [this PR](https://github.com/wordpress-mobile/gutenberg-mobile/pull/1450) is close to merge, if yes, let's wait for it and do this afterwards.

We need [this fix](https://github.com/facebook/react-native/pull/26986) to be able to ship Group, Column blocks. It will be available with React Native 0.62 which is in the RC phase currently but there's no clarity around when it'll be shipped(discussions going on [here](https://github.com/react-native-community/releases/issues/145)). So we can cherry pick that fix on our fork and unblock Group and Column.

PR #1450, opened 2019-10-15
Update React Native to 0.61.5

jtreanor:
This is a draft of the first steps for upgrading the project to React Native 0.61.5

I have focussed on verifying that the CocoaPods patch for RN 0.61.x works correctly (https://github.com/facebook/react-native/issues/25349). Here is what I have done so far:

- Updated RN and associated packages in `package.json`.
- Made a few small changes to get it to bundle successfully.
- Updated `bin/generate-podspecs.sh` to work with the new RN version.
- Generated the new podspecs and deleted obsolete ones.

Sister PRs:
* WPAndroid PR: https://github.com/wordpress-mobile/WordPress-Android/pull/10997
~https://github.com/wordpress-mobile/WordPress-Android/pull/10978~
* WP-iOS corresponding PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/13116
* Gutenberg PR: 
~https://github.com/WordPress/gutenberg/pull/19167~ https://github.com/WordPress/gutenberg/pull/19369
* `react-native-video` fork PR: https://github.com/wordpress-mobile/react-native-video/pull/4
* `react-native-svg` fork PR: https://github.com/wordpress-mobile/react-native-svg/pull/9

~As you can see from the failing CI, there is quite a bit more to do. However, it already successfully builds in WPiOS (see https://github.com/wordpress-mobile/WordPress-iOS/pull/12699).~

~The biggest outstanding piece of work that I encountered is that `gutenberg.xcodeproj` will have to be completely overhauled to use CocoaPods. It may be helpful to follow the upgrade helper [here](https://react-native-community.github.io/upgrade-helper/?from=0.59.8&to=0.61.2).~

Update release notes:

- [x] If there are user facing changes, I have added an item to `RELEASE-NOTES.txt`.

PR #1762, opened 2020-01-08
Use patch-package to fix max-depth issue in react-native (temporary)

dratwas:
Fixes https://github.com/wordpress-mobile/gutenberg-mobile/issues/1747

There is an issue on iOS with nested views.

>Searching for the ancestor view while measuring a shadow view layout relative to the ancestor is limited to 30. I couldn't find the reason nor commit where it was introduced (because I have no access to internal facebook commits).
I think we should keep looking for an ancestor view until the shadow view has the superview property.
https://github.com/facebook/react-native/pull/26986

This is fixed here https://github.com/facebook/react-native/pull/26986
It's merged to master and will land in 0.62 version of `react-native`, however it is a blocker for us since we can not nest more than 3 group blocks because of that. I know there is a PR with updating `react-native` to 0.61 but it will not solve that issue since the fix is not there yet. We could cherry-pick this commit to our fork but as I can see in this PR https://github.com/wordpress-mobile/gutenberg-mobile/pull/1450 we will not use our fork anymore.

That's why I created this PR - to handle this on `gutenberg-mobile` repo level with `patch-package`.

The mechanism is very simple. I created a patch that is in the `patches` directory and contains my changes. The patch is applied in the `prepare` hook, so it doesn't require any additional steps. 
So the flow looks like:
- install dependencies
- apply all patches from `patches` directory (it's done in `prepare` hook) 

More info about patch-package - https://callstack.com/blog/say-goodbye-to-old-fashioned-forks-thanks-to-the-patch-package/

To test:
- run `yarn install`
- run `yarn ios`
- add at least 4 nested group blocks
- the app shouldn't crash

PR submission checklist:


pinarol:
Aha, didn't know we aren't going to use our fork anymore. I haven't tested this yet but can't think of any reasons to not to try. cc @koke 

I have a question. Are we going to need to update our patch file name "react-native+0.60.0-patched.patch" as "react-native+0.61.2-patched.patch" ? If so, I suggest doing that in the [existing PR](https://github.com/wordpress-mobile/gutenberg-mobile/pull/1450) once we merge this one.

dratwas:
>I have a question. Are we going to need to update our patch file name "react-native+0.60.0-patched.patch" as "react-native+0.61.2-patched.patch" ? If so, I suggest doing that in the existing PR once we merge this one.

Yes we would need to create a new patch

koke:
> Aha, didn't know we aren't going to use our fork anymore. I haven't tested this yet but can't think of any reasons to not to try. cc @koke

I haven't tested this, but it looks like a good approach to me 👌

pinarol:
It looks like there are still things to do for [0.61](https://github.com/wordpress-mobile/gutenberg-mobile/pull/1450). Since applying the patch is simple we can get this merged without needing to wait more. And we can re-apply the patch after that 0.61 is merged. Or directly into that PR. @koke wdyt? would you be able to review this one if that's ok for you too?

package.json
```diff
@@ -39,6 +39,8 @@
     "jest-junit": "^6.3.0",
     "metro-react-native-babel-preset": "^0.52.0",
     "node-fetch": "^2.3.0",
+    "patch-package": "^6.2.0",
+    "postinstall-prepare": "^1.0.1",
```
koke: Do you have more insights into why this is needed? The repo has [an explanation](https://github.com/ds300/patch-package#why-use-postinstall-postinstall-with-yarn) but I still don't understand why we need to run `prepare` after removing a dependency.

pinarol:
That crash was fixed on the release branch with [this PR](https://github.com/WordPress/gutenberg/pull/19457).

package.json
dratwas: `node_modules` are regenerated when I run `yarn remove xyz` and the patch will not be applied w/o `postinstall-prepare` package.

To test it, please checkout to this branch, run `yarn remove postinstall-prepare` and check `node_modules/react-native/React/Views/RCTShadowView.m:159` file.  You should have the original version w/o the patch. When you add`postinstall-prepare` and run i.e `yarn remove flow-bin` the patch is applied and this is the expected result.

I think it is necessary since `yarn remove` is a regular command that is used and we want to have `react-native` patched after that.

koke: That makes sense, thanks. It seems odd that yarn would do that 😞

pinarol:
Piotr 👋 Could you update the pr’s branch so the [fix](https://github.com/WordPress/gutenberg/pull/19457) is included? Thanks!

dratwas:
@pinarol, @koke Done :)

jbinda:
>I noticed something that I don't think are caused by this PR. I added several levels of nested groups, and each level adds a bit of vertical padding to the placeholder:

@koke I think the issue is in the bordering logic. Adding condition which eliminates top margin in second child of selected item should fix that behaviour.

pinarol:
> I think the issue is in the bordering logic. Adding condition which eliminates top margin in second child of selected item should fix that behaviour.

@jbinda could you open an issue for that and add to callstack board? Thanks in advance.

jbinda:
> @jbinda could you open an issue for that and add to callstack board? Thanks in advance.

Sure

PR #1782, opened 2020-01-17
Add media edit icon to image block

### LINKED ISSUE #1583: Add overflow menu button to image/video/gallery image overlays
We've often struggled to handle cells on image, video, and gallery images that have multiple possible actions – such as `replace image/video`, `remove` (on gallery), and soon `edit image` with the coming media-editing capabilities.

This has been proposed inside of other block/feature tickets, but I wanted to create a ticket specific to the menu so that we can attack it holistically, across various block types.

## Proposal

We should add a button that when tapped shows a menu or sheet with relevant options. I'd like to propose that the icon is an ellipsis (`•••`) icon, which would sit on a dark button background that uses the same styling as `floatingToolbar` and gallery images.

Here's a sketch of the flow:

image

leandroalonso:
Fixes #1583

`Gutenberg` PR -> https://github.com/WordPress/gutenberg/pull/19723

`WPiOS` related PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/13291
`WPAndroid` related PR: https://github.com/wordpress-mobile/WordPress-Android/pull/11175

This PR adds a button over Image blocks that shows two options:

1. Edit: shows the Media Editor so the user can edit the image
2. Replace: to replace the media

As for now, the button is enabled to show only in iOS, since the feature is in the first stages of development in Android.

### To test

In order for this PR to work you'll have to checkout my fork-branch (I don't have permissions on the main `gutenberg` repo). You can [use a tool like hub to make that easier](https://hub.github.com/). Eg.: `hub pr checkout 19723`

Then, for both Android and iOS (to enable in Android just change `showMediaEditorButton` to `true`:

1. Check that the button appears only for Images
2. Check that the button appears only for images that are uploaded (and not being uploaded)
3. Tap the button and checks that a Picker appears

PR submission checklist:

- [ ] I have considered adding unit tests where possible. — Refer to Gutenberg's parent PR. I had some issues using `shallow`.


leandroalonso:
@iamthomasbishop @mbshakti I implemented the button based on the last version that was shared:

Screen Shot 2019-12-12 at 7 07 57 PM

The button is slightly bigger because I wanted to respect the minimum 44/44pt touch area. Let me know if that's a problem.

Also, I wasn't able to implement the "Delete" option. It seems to me that this is a little bit complicated, but I need help from @etoledom here. Maybe it is just my lack of experience in React talking.

leandroalonso:
@hypest we talked about not including a conditional platform to show this button.

However, we just start Android and it's way behind iOS. I've added the conditional and I can promise I'll watch that code and remove it once Android is ready. 🤞 

Are you ok with that? Let me know if you have any other suggestions. :)

iamthomasbishop:
Looking pretty solid! I one request: can we change the “delete” label to “remove”? I think this makes more sense because you’re not actually deleting the image, rather removing from the document.

leandroalonso:
@iamthomasbishop the image in the PR description is actually what was implemented.

Quoting my last comment:

> Also, I wasn't able to implement the "Delete" option. It seems to me that this is a little bit complicated, but I need help from @etoledom here. Maybe it is just my lack of experience in React talking.

But yeah, we can definitely change the label if we're going to implement that. :)

iamthomasbishop:
Ah sorry, I missed the earlier context 😬 Thank you!

mchowning:
@iamthomasbishop , is there a plan to consolidate all of these "edit" buttons. With this change we have three 😱 ? Perhaps the intention is for this button to replace the swap image button in the toolbar since they seem to duplicate each other? Even if that's true, I think it will still be a bit confusing to figure out which of the remaining two remaining "edit" buttons are used for a particular "edit". Just my $.02. 😄 

Screen Shot 2020-01-17 at 2 39 40 PM

iamthomasbishop:
> is there a plan to consolidate all of these "edit" buttons. With this change we have three 😱 ? 

@mchowning My thought is that the most obvious place to put a "replace" entry point is on the image itself. The image in the toolbar is not at all obvious, so I'd like to do one of the following:

1. Change the icon to a text label of `Replace` (as core recently changed to, IIRC). This would require us to change other media blocks that offer replace functionality.
2. Remove it once we have this menu button on the image itself.

It's worth noting that the button in the toolbar is always visible, whereas there is always a chance that the others aren't. Having a button in the toolbar is generally valuable real estate for primary actions because of this. The replace action specifically is just not a particularly obvious representation/icon, imo. 😄

The other one is the options in the settings sheet. I'm not as worried about this because it's "hidden" UI until the user needs it, thus less of a conflict.

etoledom:
Moved milestone to v1.22.
Please feel free to re-target if you consider this PR must be shipped together with v1.21.

ios/gutenberg/GutenbergViewController.swift
```diff
@@ -185,6 +185,10 @@ extension GutenbergViewController: GutenbergBridgeDelegate {
     func gutenbergDidRequestImagePreview(with fullSizeUrl: URL, thumbUrl: URL?) {
         print("Gutenberg requested fullscreen image preview for " + fullSizeUrl.absoluteString)
     }
+
+    func gutenbergDidRequestMediaEditor(with mediaUrl: URL, callback: @escaping MediaPickerDidPickMediaCallback) {
+        print("Gutenberg requested media editor for " + mediaUrl.absoluteString)
```
etoledom: It would be nice to test the callback too, just to be sure.

We can return a random media url (for example `https://cldup.com/_rSwtEeDGD.jpg` taken from https://frontenberg.tomjn.com). In this way, if the proper block is updated with the given URL, we know everything is working 👍

react-native-gutenberg-bridge/ios/GutenbergBridgeDelegate.swift
```diff
@@ -20,12 +20,14 @@ extension Gutenberg.MediaSource {
...
     public static let deviceLibrary = Gutenberg.MediaSource(id: "DEVICE_MEDIA_LIBRARY", label: "", types: [.image, .video])
     public static let deviceCamera = Gutenberg.MediaSource(id: "DEVICE_CAMERA", label: "", types: [.image, .video])
+    public static let mediaEditor = Gutenberg.MediaSource(id: "MEDIA_EDITOR", label: "", types: [.image])
 
     static var registeredInternalSources: [Gutenberg.MediaSource] {
         return [
             .deviceCamera,
             .deviceLibrary,
             .mediaLibrary,
+            .mediaEditor,
```
etoledom: This is meant to be used to select an image on an empty image block. Not sure if we want the Edit image option here.

Btw, I just noticed this is not working properly, and the internal sources are still defined in JS.

react-native-gutenberg-bridge/ios/GutenbergBridgeDelegate.swift
```diff
@@ -133,7 +135,11 @@ public protocol GutenbergBridgeDelegate: class {
 
     /// Tells the delegate to display a fullscreen image from a given URL
     ///
-    func gutenbergDidRequestImagePreview(with fullSizeUrl: URL, thumbUrl: URL?)
+    func gutenbergDidRequestImagePreview(with mediaUrl: URL, thumbUrl: URL?)
+
+    /// Tells the delegate to display the media editor from a given URL
+    ///
+    func gutenbergDidRequestMediaEditor(with mediaUrl: URL, callback: @escaping MediaPickerDidPickMediaCallback)
```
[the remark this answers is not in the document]
leandroalonso: Since the "Edit" option always appears, I guess that shouldn't be optional. I will open a WPiOS PR for this. Thanks for the heads up!

react-native-gutenberg-bridge/ios/GutenbergBridgeDelegate.swift
leandroalonso: Removed in 4535a1bea1d809b19e482b60f0d0cbf9d801426a

react-native-gutenberg-bridge/index.js
```diff
@@ -14,8 +14,11 @@ export const mediaSources = {
 	deviceLibrary: 'DEVICE_MEDIA_LIBRARY',
 	deviceCamera: 'DEVICE_CAMERA',
 	siteMediaLibrary: 'SITE_MEDIA_LIBRARY',
+	mediaEditor: 'MEDIA_EDITOR',
 };
 
+export const showMediaEditorButton = isIOS;
```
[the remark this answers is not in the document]
leandroalonso: > Since the feature is not ready yet, it might be good to add also a __DEV__ check too.

Added in 5ca3c27a04d57c375ed565cc4ab971fbed44beb8

> It would be also great if we ask this to the parent app, so we avoid branching platforms and WPAndroid would say false directly. But this is not really necessary at this stage. Just an idea 👍

I thought about doing that! I gave up because:

* This will mean more code (in the bridge and in the apps)
* Once we have the feature on both platforms, we will remove all this conditional code. Less code to remove = better.

Does that sound reasonable to you?

ios/gutenberg/GutenbergViewController.swift
leandroalonso: Do you mean a UI Test? There's any similar test example you can point me to?

react-native-gutenberg-bridge/index.js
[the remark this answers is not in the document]
etoledom: Works for me 👍 

> Once we have the feature on both platforms, we will remove all this conditional code. Less code to remove = better.

Is good to think about Gutenberg as a library for others to implement too, and other library users might not want to have that edit button there. In the other hand, right now 3rd parties are not priority for the mobile version, so let's keep it as it is.

react-native-gutenberg-bridge/ios/GutenbergBridgeDelegate.swift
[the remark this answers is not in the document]
etoledom: Continuing with the idea on https://github.com/wordpress-mobile/gutenberg-mobile/pull/1782#discussion_r370140525, the implementation of this delegate method might be the signal to show/hide this button. But again, not needed at this stage! 😁

Let's do as you mentioned. ✅

ios/gutenberg/GutenbergViewController.swift
etoledom: This is a good example: https://github.com/wordpress-mobile/gutenberg-mobile/blob/40f4c19eaa2147cc8165332af68776f338471f92/ios/gutenberg/GutenbergViewController.swift#L71:L71

The idea would be, by selecting that option in the demo app, to see something happening.
On the real app we would edit the image and upload the edited version of it. Here in the demo app we can simulate a new edited version by giving back this new url. In this way we ensure that the method and callback work properly.

Does it make sense? Maybe I'm confused 😅

leandroalonso: Done in 2257b1c3d889555be7ade0a9418047ee7e40dd9a

leandroalonso:
@etoledom I guess I addressed all the necessary changes. Let me know if there's anything else.

As for the PR in WPiOS, I'll create one that will both implement this method and also provide the Media Editor (it wouldn't make sense to have just an option that does nothing).

react-native-gutenberg-bridge/ios/RNReactNativeGutenbergBridge.swift
```diff
@@ -155,6 +155,31 @@ public class RNReactNativeGutenbergBridge: RCTEventEmitter {
...
+        }
+
+        DispatchQueue.main.async {
+            self.delegate?.gutenbergDidRequestMediaEditor(with: url) { media in
+                guard let media = media else {
+                    callback(nil)
+                    return
+                }
+
+                let mediaToReturn = media
```
etoledom: Is this necessary? Maybe we could use`media` directly

react-native-gutenberg-bridge/android/src/main/java/org/wordpress/mobile/WPAndroidGlue/WPAndroidGlueCode.java
```diff
@@ -378,7 +390,8 @@ public void attachToContainer(ViewGroup viewGroup, OnMediaLibraryButtonListener
                                   OnEditorAutosaveListener onEditorAutosaveListener,
                                   OnAuthHeaderRequestedListener onAuthHeaderRequestedListener,
                                   RequestExecutor fetchExecutor,
-                                  OnImageFullscreenPreviewListener onImageFullscreenPreviewListener) {
+                                  OnImageFullscreenPreviewListener onImageFullscreenPreviewListener,
+                                  OnMediaEditorListener onMediaEditorListener) {
```
mchowning: We also need a PR to WPAndroid that adds this new parameter where [`GutenbergContainerFragment` calls this method](https://github.com/wordpress-mobile/WordPress-Android/blob/df83d585f5a70442bf718b46e4236b064a3f0d0d/libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/GutenbergContainerFragment.java#L57). Otherwise, merging this will break building Gutenberg from source on WPAndroid (i.e., building with `wp.BUILD_GUTENBERG_FROM_SOURCE=true` added to WPAndroid's `gradle.properties` file).

mchowning:
:wave: @leandroalonso ! Nice work on this! 

We have recently merged a fix for our e2e tests that _should_ make them more stable. So you may want to merge/rebase the latest from `develop` into your PR to increase the likelihood of getting green builds on CI. 😄

react-native-gutenberg-bridge/android/src/main/java/org/wordpress/mobile/WPAndroidGlue/WPAndroidGlueCode.java
leandroalonso: Done in https://github.com/wordpress-mobile/WordPress-Android/pull/11175

PR #1795, opened 2020-01-21
Enforce consistent RN version across javascript and Android projects

mchowning:
This PR enforces that we use the same react native version in our javascript and Android projects.

### Testing
Verify that:
1. demo app builds
2. react-native-gutenberg-bridge android project can be built directly (i.e., `react-native-gutenberg-bridge/android/gradlew assembleDebug`)
3. react-native-aztec android project can be built directly (i.e., `react-native-aztec/android/gradlew assembleDebug`)
4. Jitpack can build with these changes. In other words, WPAndroid can be built with this checked out in `gutenberg-mobile` and `BUILD_GUTENBERG_FROM_SOURCE` set to `false`

PR submission checklist:

### DIFF
--- a/android/shared_config.gradle
+++ b/android/shared_config.gradle
@@ -1,0 +1,11 @@
+import groovy.json.JsonSlurper
+
+/*
+ * This function returns the relevant React Native version for Android as parsed from
+ * the provided package.json file.
+ */
+ext.getRnVersion = { String pathToPackageJson, String section ->
+    def packageSlurper = new JsonSlurper()
+    def packageJson = packageSlurper.parse(file(pathToPackageJson))
+    return packageJson.get(section).get('react-native')
+}
--- a/src/test/project.test.js
+++ b/src/test/project.test.js
@@ -1,0 +1,24 @@
+/** @format */
+
+/**
+ * External dependencies
+ */
+import fs from 'fs';
+import _ from 'lodash';
+
+describe( 'RN version from root is same in', () => {
+	const rootRNVersion = parseRNVersionFromPackageJson( '.', 'dependencies' );
+	expect( rootRNVersion ).toBeDefined();
+
+	[ 'react-native-aztec', 'react-native-gutenberg-bridge' ].forEach( ( path ) => {
+		it( path, () => {
+			const otherRNVersion = parseRNVersionFromPackageJson( path, 'peerDependencies' );
+			expect( otherRNVersion ).toEqual( rootRNVersion );
+		} );
+	} );
+} );
+
+function parseRNVersionFromPackageJson( packageJsonPath, section ) {
+	const packageJson = JSON.parse( fs.readFileSync( `${ packageJsonPath }/package.json` ) );
+	return _.get( packageJson, [ section, 'react-native' ] );
+}
--- a/react-native-aztec/android/build.gradle
+++ b/react-native-aztec/android/build.gradle
@@ -31,9 +31,6 @@ apply plugin: 'com.android.library'
 apply plugin: 'kotlin-android'
 apply plugin: 'com.github.dcendents.android-maven'
 
-// import the `readReactNativeVersion()` function
-apply from: 'https://gist.githubusercontent.com/hypest/742448b9588b3a0aa580a5e80ae95bdf/raw/8eb62d40ee7a5104d2fcaeff21ce6f29bd93b054/readReactNativeVersion.gradle'
-
 group='com.github.wordpress-mobile.gutenberg-mobile'
 
 // fallback flag value for when lib is compiled individually (e.g. via jitpack)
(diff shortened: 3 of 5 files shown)


android/shared_config.gradle
```diff
@@ -0,0 +1,27 @@
+import groovy.json.JsonSlurper
+
+/*
+ * This function returns the relevant React Native version. Before doing so,
+ * it performs a check to insure that the javascript version for React Native
```
hypest: > javascript version for React Native

I assume that by "javascript" version you mean the version defined in the package.json, right? How about naming that "the package.json version of RN"? I don't think mentioning "JavaScript" is relevant here anyway, the package.json is a npm/Node construct anyway.

If we rename here, we should also rename the variable names in the code below too.

mchowning: 👍 Makes sense to me. Updated.

android/shared_config.gradle
```diff
@@ -0,0 +1,27 @@
...
+ * has not changed. It does this so that anytime the javascript version of React
+ * Native is changed, this method will fail the build until this method is also
+ * updated.
+ */
+ext.getRnVersion = { String pathToPackageJson, String section ->
+
+    def packageSlurper = new JsonSlurper()
+    def packageJson = packageSlurper.parse(file(pathToPackageJson))
+    def actualJavascriptRnVersion = packageJson.get(section).get('react-native')
```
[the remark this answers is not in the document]
mchowning: Oh, but whoever wrote that original code was so hacky with it! 😝 (and I _definitely_ didn't just copy that code out when I wrote this method 😁)

That change definitely does read better. Updated.

android/shared_config.gradle
```diff
@@ -0,0 +1,28 @@
...
+ * it performs a check to insure that the version for React Native in package.json
+ * has not changed. It does this so that anytime the version of React Native in
+ * package.json is changed, this method will fail the build until this method is also
+ * updated to return the new React Native version applicable for Android.
+ */
+ext.getRnVersion = { String pathToPackageJson, String section ->
+    def actualPackageJsonRnVersion = parseRnVersion(pathToPackageJson, section)
+    def expectedPackageJsonRnVersion = 'jtreanor/react-native#v0.60.0-patched'
+    if (actualPackageJsonRnVersion == expectedPackageJsonRnVersion) {
+        return '0.60.0-patched'
```
hypest: I think it feels weird to have to look in this non-standard file, inside the demo app android folder to check for what version the expected one is. The native parts of the demo app don't play any role when building the bridge module, for example, and that's what a WPAndroid develop would probably look first to understand the dependencies.

In a sense, what I'm still trying to understant is whether the `gutenberg-mobile/package.json` can or cannot be the (single) source of truth for which RN version should be picked up 🤔.

mchowning: >I think it feels weird to have to look in this non-standard file, inside the demo app android folder to check for what version the expected one is. The native parts of the demo app don't play any role when building the bridge module, for example, and that's what a WPAndroid develop would probably look first to understand the dependencies.

My thinking was that it was an Android-specific file, so I put it in `android/`. I see your point though, perhaps this would be better (a) down one level at the root; or (b) in `react-native-gutenberg-bridge/android`. WDYT?

>In a sense, what I'm still trying to understand is whether the gutenberg-mobile/package.json can or cannot be the (single) source of truth for which RN version should be picked up 🤔.

In principle, yes. The reason I went with this approach was that _before we updated to RN 0.61.5_, the version numbers between the package.json and Android did not align well (package.json was vXXX, whereas Android was just XXX). This meant that to parse the version from the package.json would have involved some one-off fragile arbitrary parsing logic (i.e., check if there's a `v` after `#` and if there is drop the `v`) that I didn't want to use and then have lingering there forever. I put the check that throws an exception if the package.json version changed with the idea that it would force us to update this function to something more reasonable when we changed the version in package.json.

So... we've updated the RN version, which means I should probably get on updating this function 😄, and it does look like we should be able to now directly parse the version from the package.json file easily.

src/test/project.test.js
```diff
@@ -0,0 +1,24 @@
...
+	const rootRNVersion = parseRNVersionFromPackageJson( '.', 'dependencies' );
+	expect( rootRNVersion ).toBeDefined();
+
+	[ 'react-native-aztec', 'react-native-gutenberg-bridge' ].forEach( ( path ) => {
+		it( path, () => {
+			const otherRNVersion = parseRNVersionFromPackageJson( path, 'peerDependencies' );
+			expect( otherRNVersion ).toEqual( rootRNVersion );
+		} );
+	} );
+} );
```
mchowning: :wave: @hypest! I'm curious if you have any thoughts about how I'm using a jest test to enforce that the RN versions are consistent across the different package.json files. This feels rather hacky to me, but I assume we would always want these to be consistent, and I couldn't think of a better way to handle it. WDYT?

android/shared_config.gradle
mchowning: Made the changes. This is now actually back to being more similar to what you had set up. As I see it, the two main differences are:

1. We are storing the function to parse the react-native version in a gradle file inside the repository instead of in a github gist; and
2. There are jest tests verifying that the react-native version between the different `package.json` files are consistent.

mchowning:
Closed this because the PRs been sitting for quite awhile, and it feels a bit weird to me that I'm using unit tests to enforce consistent versions in our dependencies. If we see problems from these versions getting out of sync again though we may want to reconsider this change.

PR #1800, opened 2020-01-22
Fix local media id collisions

### LINKED ISSUE #1610: Mobile gallery block - Upload options
This aims to track issues related to uploading media from the Gallery block. It is part of this issue: https://github.com/wordpress-mobile/gutenberg-mobile/issues/1416.

## Test Builds

Builds are open for testing on these draft PRs:
 * `WordPress-Android`: https://github.com/wordpress-mobile/WordPress-Android/pull/11234#issuecomment-582252774
* `WordPress-iOS`: https://github.com/wordpress-mobile/WordPress-iOS/pull/13374#issuecomment-582251935

### Media uploading and ids

In the current state, the `MediaPlaceholder` component utilizes the `addToGallery` prop described here: https://github.com/WordPress/gutenberg/pull/18262 to remove duplicates via the `id` prop whenever new media is selected. This is because the collection of image elements is currently keyed on `id` or `url` if `id` is not defined.

When a media item is local (i.e. not finished uploading, in the case of adding media from device), it may have a temporary local id. This can result in "collisions" (i.e. the local id of a newly added and currently uploading media item being equal to the "server id" of an already uploaded item within the same gallery).

One way to address this is to negate the local media id over the bridge (both directions), this way, the media models / database on WordPress-Android will not require updating, and the Gutenberg (JavaScript) side will not see collisions. I'd love to know if there are alternative approaches to solving this as well.

**Update:** The collision id issue has been confirmed with [steps here](https://github.com/wordpress-mobile/WordPress-Android/issues/11239#issue-560256478).

mkevins:
Fixes https://github.com/wordpress-mobile/WordPress-Android/issues/11239

Related PR:

`WordPress-Android`: https://github.com/wordpress-mobile/WordPress-Android/pull/11125

This PR aims to prevent local media ids from colliding with remote ids during ongoing uploads. The issue is described in more detail here: https://github.com/wordpress-mobile/gutenberg-mobile/issues/1610#issue-527049243. The approach used in this PR is to negate the local media id over the bridge so local ids are represented as negative integers to the JavaScript code. All interactions over the bridge involving these local ids (primarily media upload events) will negate the local id. In this way, the media models and database logic used within WordPress-Android can remain unaffected, and the React Native (JavaScript) logic will avoid id collisions.

#### Variable names changes:

In several places we are using names like Media_Upload_Callback. This is not always correct, since we use the same callback for media selection from the WordPress Media Library (which does not involve any uploading). I made some variable name changes to improve the clarity while working with this code.

To test:
tbd

PR submission checklist:


SergioEstevao:
@mkevins is this PR still active, are you still planning to work on this? It will probably need to be update to the monorepo structure.

mkevins:
> @mkevins is this PR still active, are you still planning to work on this? It will probably need to be update to the monorepo structure.

Not really active at the moment. I believe the possibility still exists to have id collisions, but much of this PR will have conflicts, since the interfaces have already been refactored a bit. I'll close this for now, and revisit the strategy if necessary at a later point. Thanks for the reminder!

PR #1820, opened 2020-01-24
Rename iOS module name of sample app from gutenberg to GutenbergDemo.

SergioEstevao:
With the update to React 0.61.5 the iOS demo project is now built using CocoaPods.

This creates a scenario where sometimes Xcode build system gets confused between the pod library Gutenberg (the Gutenberg Bridge podspec) and the binary of the demo project itself: gutenberg.

You will see Xcode sometimes complaining that we are importing `Gutenberg` instead of `gutenberg`, and after getting on this state you will probably need to clean all build folders and restart the project.

So the solution I propose here is to rename the module name of the `gutenberg.xcodeproj` from `gutenberg` to `GutenbergDemo` to avoid these conflicts going forward.

To test:
 - Check if projects build
 - Check that tests build and pass

PR submission checklist:


hypest:
👋 @SergioEstevao , as a follow up, do you think that we could/should also rename the local files that are still named as "gutenberg.xxx"? For example, all the files/folders in https://github.com/wordpress-mobile/gutenberg-mobile/tree/develop/ios are still named/prefixed `gutenberg`.

PR #1828, opened 2020-01-28
Remove DEV flag from Media Editor button

leandroalonso:
Remove the `__DEV__` flag so the Media Editor button will be available for iOS in production too.

PR submission checklist:


leandroalonso:
@etoledom The Media Editing is already merged in WPiOS `develop`. Do you think we can move on with this PR?

PR #1835, opened 2020-01-29
Make inserter to show options on long-press to add before/after

### LINKED ISSUE #1433: Proposal: Long-press on inserter to show options for "add above" and "add below"
Long-pressing on the inserter button (`(+)` on the quick toolbar) should show options to `Add Block Before` and `Add Block After` the currently selected block. If there is no selection these options could be `Add to beginning` and `Add to end`. On iOS, this could be an `ActionSheet` and Android a `BottomSheet`.

Example:

IMG_0645

Note: I think we could expand on this pattern in the future, but I've found myself wanting to do add blocks before the selected one quite often, especially now that we have InnerBlocks.

We might also want to add these options to the `•••` menu on the inline toolbar when we get a chance, similar to web.

ceyhun:
Fixes #1433 

To test: see Gutenberg PR: https://github.com/WordPress/gutenberg/pull/18791

PR submission checklist:


designsimply:
Looks lovely! 😍

Screenshot_20200221-160053
<sup>Tested with WPAndroid 14.2-rc-2 on Pixel 3 Android 10.</sup>

Issue #1802, opened 2020-01-22
Refine selected block borders


iamthomasbishop:
While working on InnerBlocks, we decided to try a new style for borders to highlight the difference between nested and non-nested blocks. I must say, after playing w/ blocks that use this method, it feels right for all blocks, so I'm proposing that we extend to use this across the board.

Essentially what we did was move the left/right borders in towards the center so that there were slight _optical margins_ on the left/right sides of the block. The two approaches look like this in practice (both are in use at the moment):

<img width="1116" alt="image" src="https://user-images.githubusercontent.com/1200432/72917688-e5f10780-3d09-11ea-9d54-0bd80c526a93.png">

Note that all blocks (other than special cases w/ columns, etc) span 100% of the content container, but we used a mix of margin and padding to define the border so that there is an optical inset. So instead of using 12px/16px as the padding on all blocks (and no margin), we use something like 12px/12px, and then there is a 1px border and 3px margins on left/right of the block (don't quote me on these exact values because I can't recall what we're using in the code). This might help to illustrate the difference:

<img width="971" alt="image" src="https://user-images.githubusercontent.com/1200432/72917864-39fbec00-3d0a-11ea-8c7e-e3c1c0aab64c.png">

There are a few wins we get by doing this:

* Above all, it's much clearer which block is selected
* 12/12 padding inside blocks feels more balanced
* Right now, we have a couple of different styles for (parent) block borders depending on device size — this would consolidate into a single style
* As we start getting into more complex layout and nesting, this should make the UI feel more natural, because elements won't feel as much like they're "bumping" against each other

// cc @lukewalczak @jbinda @pinarol @hypest — any thoughts?

jbinda:
>So instead of using 12px/16px as the padding on all blocks (and no margin), we use something like 12px/12px, and then there is a 1px border and 3px margins on left/right of the block (don't quote me on these exact values because I can't recall what we're using in the code).

That's the correct values and in the code it looks almost the same. In the `_variables.scss` file there is a definition of border connected values (see below):

<image src="https://user-images.githubusercontent.com/21242757/72971494-10a29500-3dca-11ea-9a94-558a93923a7a.png" width="300" />

Base on that each block receive proper class with already calculated value of margin. The cool thing about that are:
1. Code simplicity and readability
2. It's easy to play with those value and change how the bordering behaviour changes

[...]

PR #1842, opened 2020-01-30
Avoid fullscreen text entry UI in landscape mode

### LINKED ISSUE #1829: Android landscape mode - soft keyboard done/next button adds new line/block
## Summary
On Android, pressing "Next" or "Done" soft keyboard button when editing paragraph block in landscape orientation adds a new paragraph block instead of closing the keyboard.

android_landscape_done_button mp4

## To Reproduce
1. Create a paragraph block
2. With the phone in landscape mode, place the cursor in the middle of the block's text
3. Tap the soft keyboard's Done/Next button **(I'm seeing the "Next" button on the demo app, but WPAndroid is properly showing "Done")**
4. Observe that only the text after the cursor remains
5. Close the keyboard
6. Observe that a new paragraph block has been added.

## Expected Behavior
I think we want this the done button to dismiss the keyboard without creating a new block, but it would be good to get @iamthomasbishop 's thoughts.

## Smartphone
- 1 
  - Device: Pixel
  - OS: Android 10
   - Version: commit 7963eeed

- 2
   - Device: Pixel 3
   - OS: Android 10
   - Version: alpha-206 (latest from beta channel on Jan. 28, 2020)

mchowning:
Fixes #1829 

## Summary

On Android when we are editing a paragraph block in landscape mode and you press the done button, an "Enter" keypress is registered. Therfore, the block is split at the point of the cursor. @iamthomasbishop indicated that the "Done" button should just close the keyboard, but that he would prefer not have the "full screen" editing type view when in landscape mode. Conveniently enough, the preferred solution is actually easier to implement. 😄 I would like to get @iamthomasbishop 's sign-off on this first though to make sure he's ok with how the view is mostly taken up by the keyboard.

### But What Does It Look Like?!?!?

I'm glad you asked! 

Before | After
--- | ---
Screen Shot 2020-01-30 at 11 42 06 AM | Screen Shot 2020-01-30 at 11 39 35 AM
android_landscape_no_extract mp4 | android_landscape_no_extract mp4

## To Test

1. Open post with paragraph block
2. Select paragraph block so that keyboard displays
3. Switch the phone to landscape mode
4. Observe that the phone looks like the "After" picture ^above^ (i.e., there isn't a "Done" button).

## PR Submission Checklist

### DIFF
--- a/react-native-aztec/android/src/main/java/org/wordpress/mobile/ReactNativeAztec/ReactAztecText.java
+++ b/react-native-aztec/android/src/main/java/org/wordpress/mobile/ReactNativeAztec/ReactAztecText.java
@@ -18,6 +18,7 @@ import android.text.InputType;
 import android.text.Spannable;
 import android.text.TextUtils;
 import android.text.TextWatcher;
+import android.view.inputmethod.EditorInfo;
 import android.view.inputmethod.InputMethodManager;
 import android.widget.TextView;
 
@@ -138,6 +139,7 @@ public class ReactAztecText extends AztecText {
                 ReactAztecText.this.propagateSelectionChanges(selStart, selEnd);
             }
         });
+        this.setImeOptions(getImeOptions() | EditorInfo.IME_FLAG_NO_EXTRACT_UI);
         this.setInputType(InputType.TYPE_CLASS_TEXT | InputType.TYPE_TEXT_FLAG_CAP_SENTENCES | InputType.TYPE_TEXT_FLAG_MULTI_LINE);
     }
 
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -1,3 +1,7 @@
+1.24.0
+------
+* [Android] Can now scroll post when in landscape orientation with the soft keyboard displayed
+
 1.23.0
 ------
 * New block: Group


iamthomasbishop:
This feels much more natural, imo. One tiny thing (which may be solved by [this proposal](https://github.com/wordpress-mobile/gutenberg-mobile/issues/1802) if we work on it) is that there seems to be a momentary "stutter" in adjusting the border styling upon rotation in your last gif above, where it attempts to switch from top/bottom-only borders (in portrait) to full borders w/ rounded corner (in landscape).

iamthomasbishop:
One other thing, and we might want to tackle it separately. Do we want to automatically hide the keyboard upon rotation? I'm not 100% sure, just think it'd make the effects more visible.

mchowning:
Thanks for the comments @iamthomasbishop !

> One tiny thing (which may be solved by this proposal if we work on it) is that there seems to be a momentary "stutter" in adjusting the border styling upon rotation in your last gif above, where it attempts to switch from top/bottom-only borders (in portrait) to full borders w/ rounded corner (in landscape).

This stutter is pretty noticeable on Android whereas I don't really notice it iOS. I suspect it may be tricky to address, so IMO it would be better to address it separately.

> One other thing, and we might want to tackle it separately. Do we want to automatically hide the keyboard upon rotation? I'm not 100% sure, just think it'd make the effects more visible.

We could certainly do that. A few questions... Would you want this on both platforms or just on Android? If both, on iOS currently I believe that dismissing the keyboard generally also removes focus from the block, so would you want to lose focus on rotation as well on iOS, or do you not have a strong preference? Also, do you have a strong preference on whether we also dismissed the keyboard when rotating from landscape to portrait? FWIW, making the keyboard dismiss on Android will be pretty easy. At this point, I'm not sure how difficult that would be on iOS.

One interesting and unexpected (to me at least) side effect of dismissing the keyboard on Android is that dismissing the keyboard on rotation creates enough "movement" on the screen during rotation that the stutter you mentioned is actually much less noticeable (to my eyes at least).

iamthomasbishop:
> This stutter is pretty noticeable on Android whereas I don't really notice it iOS. I suspect it may be tricky to address, so IMO it would be better to address it separately.

Sounds good 👍 

> Would you want this on both platforms or just on Android? If both, on iOS currently I believe that dismissing the keyboard generally also removes focus from the block, so would you want to lose focus on rotation as well on iOS, or do you not have a strong preference?

That's a good point. After thinking about this again, I think we should probably just simplify the whole situation and leave it as it currently is (don't remove focus/hide keyboard). While hiding the keyboard would certainly clear space, removing the focus might feel like a bug in practice (and require the user to re-select the block).

react-native-aztec/android/src/main/java/org/wordpress/mobile/ReactNativeAztec/ReactAztecText.java
```diff
@@ -138,6 +139,7 @@ public void onSelectionChanged(int selStart, int selEnd) {
                 ReactAztecText.this.propagateSelectionChanges(selStart, selEnd);
             }
         });
+        this.setImeOptions(getImeOptions() | EditorInfo.IME_FLAG_NO_EXTRACT_UI);
```
cameronvoell: This is a great improvement! Read through IME options [here](https://developer.android.com/reference/android/view/inputmethod/EditorInfo#IME_FLAG_NO_EXTRACT_UI) and agree IME_FLAG_NO_EXTRACT_UI looks like the right choice.

mchowning:
Hey @iamthomasbishop ! I was looking at this in the WPAndroid app, and I think we may want to revert this change. With the app toolbar virtually no space is left above the keyboard for the post, which results is a pretty poor experience. WDYT?

landscape_tight_android mp4

iamthomasbishop:
Agreed that's not great, although tbh landscape w/ any sort of text editing is quite restrictive considering the height of the keyboard. In the long run, I think we might want to consider a more flexible/collapsing app bar — which would help a lot — but for the time being, we can probably revert.

PR #1861, opened 2020-02-06
Remove react-native-aztec-old-submodule

ceyhun:
`react-native-aztec` was moved into this repo a while ago and it's submodule was removed: https://github.com/wordpress-mobile/gutenberg-mobile/pull/465 
But this submodule was re-introduced again because of JitPack problems: https://github.com/wordpress-mobile/gutenberg-mobile/pull/486
Seems like JitPack is not an issue anymore as the CI is green in this PR so we can remove this submodule now.

### DIFF
--- a/react-native-gutenberg-bridge/android/build.gradle
+++ b/react-native-gutenberg-bridge/android/build.gradle
@@ -157,10 +157,6 @@ if (isJitPack) {
     def assetsFolder = 'src/main/assets'
     def aarFolder = 'src/main/assets'
 
-    task removeReactNativeAztecOldSubmodule(type: Delete) {
-        delete '../../react-native-aztec-old-submodule'
-    }
-
     task buildJSBundle(type: YarnTask) {
         args = ['bundle:android']
     }
@@ -214,6 +210,5 @@ if (isJitPack) {
         backupHermesReleaseAAR.dependsOn(copyJSBundle)
         copyJSBundle.dependsOn(buildJSBundle)
         buildJSBundle.dependsOn(yarn_install, ensureAssetsDirectory)
-        ensureAssetsDirectory.dependsOn(removeReactNativeAztecOldSubmodule)
     }
 }
--- a/react-native-gutenberg-bridge/android/build.gradle
+++ b/react-native-gutenberg-bridge/android/build.gradle
@@ -157,6 +157,10 @@ if (isJitPack) {
     def assetsFolder = 'src/main/assets'
     def aarFolder = 'src/main/assets'
 
+    task removeReactNativeAztecOldSubmodule(type: Delete) {
+        delete '../../react-native-aztec-old-submodule'
+    }
+
     task buildJSBundle(type: YarnTask) {
         args = ['bundle:android']
     }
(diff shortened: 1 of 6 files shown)


Tug:
Changes looks good to me. Could you maybe open a PR on WPAndroid that targets this branch to make sure ewverything is ok now?

ceyhun:
[Android PR](https://github.com/wordpress-mobile/WordPress-Android/pull/11289) builds failed with a JitPack error:
 
https://jitpack.io/com/github/wordpress-mobile/gutenberg-mobile/fc7493eb81/build.log

Tried rebasing but got the same error:
https://jitpack.io/com/github/wordpress-mobile/gutenberg-mobile/c972d0d89a/build.log

@hypest maybe you have an idea?

hypest:
Looking at the JitPack log:
> jest-haste-map: Haste module naming collision: react-native-aztec
>  The following files share their name; please adjust your hasteImpl:
>    * <rootDir>/react-native-aztec-old-submodule/package.json
>    * <rootDir>/react-native-aztec/package.json

Looks like the "old-submodule" is still there. Not fully sure yet if this is essentially the same error that we got back in the day and that's why we decided to keep the old-submodule around 🤔 .

hypest:
Or, perhaps we have to add a step in JitPack's config to `rm -Rf` the submodule which is probably checked out from the initial clone (before JitPack then switches the to branch under test). Does that make sense?

ceyhun:
> Not fully sure yet if this is essentially the same error that we got back in the day and that's why we decided to keep the old-submodule around 🤔 .

I think it was a different error before https://github.com/jitpack/jitpack.io/issues/3707 mentioned here https://github.com/wordpress-mobile/gutenberg-mobile/pull/486. 

> Or, perhaps we have to add a step in JitPack's config to `rm -Rf` the submodule folder which is probably checked out from the initial clone (before JitPack then switches the to branch under test). Does that make sense?

Let me try that 👍

react-native-gutenberg-bridge/android/build.gradle
```diff
@@ -210,5 +214,6 @@ if (isJitPack) {
         backupHermesReleaseAAR.dependsOn(copyJSBundle)
         copyJSBundle.dependsOn(buildJSBundle)
         buildJSBundle.dependsOn(yarn_install, ensureAssetsDirectory)
+        ensureAssetsDirectory.dependsOn(removeReactNativeAztecOldSubmodule)
```
hypest: 👋 @ceyhun , deleting during the gradle build should do the trick as well, but I was thinking that perhaps we don't make this part of the build process and instead do the cleaning in the JitPack config file, [here for example](https://github.com/wordpress-mobile/gutenberg-mobile/blob/6352c210a79dbcf7e9285971761e8eaf94100a01/jitpack.yml#L2). The "cleanup" is mainly a git checkout issue, not a build issue so, I'd expect it not to be involved in the gradle steps. WDYT?

ceyhun: Yeah that might be better. Let me try that too 👍

ceyhun:
It worked perfectly 🎉 Android build is now green as well: https://github.com/wordpress-mobile/WordPress-Android/pull/11289

PR #1869, opened 2020-02-07
Delete react-native-aztec's example project

ceyhun:
Seems like it's not used anymore and old dependencies are getting a lot of [automated PRs](https://github.com/wordpress-mobile/gutenberg-mobile/pulls?q=is%3Aopen+is%3Apr+react-native-aztec%2Fexample+label%3Adependencies) (the Swift and react-native versions are pretty outdated as well), so instead of updating dependencies maybe we can delete this folder. 

- [ ] Will add a git tag to a commit before the removal to easily search for it in the future (to reinstate and update)


ceyhun:
Added git tag to the commit before removal of react-native-aztec example project:

https://github.com/wordpress-mobile/gutenberg-mobile/tree/react-native-aztec-example-project

PR #1874, opened 2020-02-07
Schedule to request focus instead of asking immediately

### LINKED ISSUE #1870: Scrolling jump to top when typing a new paragraph block
**Describe the bug**
When hitting Enter to create a new paragraph block, the whole content seems to jump to the top and scroll back down to the new block that got created.

See this GIF (slowed down to make it more noticable, notice the momentary jump to the top where an image block is there):

screencapture-1581083047041 2020-02-07 15_58_53

**Expected behavior**
The content should not jump to the top it and scroll back down again.

**To Reproduce**
1. Open the demo app and add an image block at the top (to make the effect more noticeable)
3. Start adding paragraph bloc. Add more than a few so the view will need to scroll to show the whole content
4. Type "Enter" on the last paragraph block to create a new one
5. Notice that the whole view jumps to the top for a split-second and scrolls back down to the end on the newly created block

 - Device: Pixel 2 XL
 - OS: Android 10
 - App Version Happening at least on the 1.22 PR, [this commit](2657a54053f35b8dc8a67427b8a9b5717c4b88d7).

hypest:
Fixes #1870 

My current theory is that the new View is not yet in the proper position in the layout and requesting focus to it causes a scroll to the top of the ScrollView, since the coordinates of the view are not yet updated in respect to its parent. I can be wrong, haven't fully reached the bottom of this issue.

### To test A:

1. Open the demo app and add an image block at the top (to make the effect more noticeable)
2. Start adding paragraph block. Add more than a few so the view will need to scroll to show the whole content
3. Type "Enter" on the last paragraph block to create a new one
4. There shouldn't be a "whole list" jump to the top and back down again

### To test B:

There's a prior known issue with scrolling to new block added doesn't work so, this PR is not introducing that bug: https://github.com/wordpress-mobile/gutenberg-mobile/issues/1881

### PR submission checklist:

- [x] I have considered adding unit tests where possible. I haven't added any because we don't currently have a way to detect this scrolling jump in automated tests.

### DIFF
--- a/react-native-aztec/android/src/main/java/org/wordpress/mobile/ReactNativeAztec/ReactAztecManager.java
+++ b/react-native-aztec/android/src/main/java/org/wordpress/mobile/ReactNativeAztec/ReactAztecManager.java
@@ -7,6 +7,8 @@ import android.os.Build;
 import androidx.annotation.Nullable;
 import androidx.core.graphics.ColorUtils;
 
+import android.os.Handler;
+import android.os.Looper;
 import android.text.Editable;
 import android.text.Layout;
 import android.text.TextUtils;
@@ -488,7 +490,13 @@ public class ReactAztecManager extends BaseViewManager<ReactAztecText, LayoutSha
     public void receiveCommand(final ReactAztecText parent, int commandType, @Nullable ReadableArray args) {
         Assertions.assertNotNull(parent);
         if (commandType == mFocusTextInputCommandCode) {
-            parent.requestFocusFromJS();
+            // schedule a request to focus in the next layout, to fix https://github.com/wordpress-mobile/gutenberg-mobile/issues/1870
+            new Handler(Looper.getMainLooper()).post(new Runnable() {
+                @Override
+                public void run() {
+                    parent.requestFocusFromJS();
+                }
+            });
             return;
         } else if (commandType == mBlurTextInputCommandCode) {
             parent.clearFocusFromJS();


cameronvoell:
I confirmed that this fixed the list scrolling to the top and back to the bottom when hitting enter to create a new block (testing steps) when I tested on Android in the example app. 

I still saw some jumpy looking animations when I opened the block inserter and am not sure if that is already a known issue. Will confirm, and post a video start of day tomorrow if helpful.

mchowning:
This fix also appears to address the "scroll to the top when hitting enter to create a new paragraph block" issue in my testing.

hypest:
Cameron and Matt have already given a first look at this PR and seems to fix the issue but, added you @marecar3 as a reviewer too to have a look too. Let me know if you can think of some scrolling cases that might be broken by this PR. Thanks!

marecar3:
Hey @hypest, nice work here!
Good investigation!

It's working much better now, but it seems that scroll still has some jumpy behavior.

ezgif com-video-to-gif (14)

hypest:
Right, I think that the jumping remaining is an older "issue", where the virtual keyboard actually gets hidden momentarily while the focus is first lost from the current block and then refocusing on the new block. The fix in this PR doesn't try to address that part :(

If it's not a new issue (would appreciate if you try v1.21.0 to check if it's there too), what do you think about addressing it in a different PR @marecar3?

Tug:
Tested on gb-mobile `develop` and gb `master` by cherry picking this commit and it indeed fixes the worst scroll issues (scroll loops and crashes) but as noticed by others it still jumps a couple times and end up not at the optimal place in the end in particular for group blocks it seems

hypest:
> but as noticed by others it still jumps a couple times and end up not at the optimal place in the end in particular for group blocks it seems

👋 @Tug , let me ask, is that behavior happening on an emulator and/or physical device?

PR #1920, opened 2020-02-17
Pullquote block

### LINKED ISSUE #1854: Pullquote block
Add support for pullquote block

ceyhun:
Add support for Pullquote block behind dev flag. Issues blocking the release of this block are here: #1854 

Gutenberg PR: https://github.com/WordPress/gutenberg/pull/20952

To test: See Gutenberg PR

PR submission checklist:


marecar3:
Hey @mchowning 👋 

As I have mentioned on Slack, @ceyhun has a problem with text alignment in `pullquote` block and I tried to find some nice/good fix, but couldn't find an easy one.

Let me try to explain where is the problem and what I have tried in order to fix it.

**Problem:** 

Probably alignment isn't working because `<p>` is wrapped with `<div>` and here is the picture of all spans in example text.

Screen Shot 2020-02-25 at 2 33 49 PM

Even if we have `ParagraphSpan` it seems that the problem is caused by `HiddenHtmlBlock`.

**Solution:**

The easiest solution is to call method: `      view.toggleFormatting(AztecTextFormat.FORMAT_ALIGN_CENTER);`
but there is a problem how to determinate when to call the method as:
 - we don't have a specific block type / tag
 - method call can only happen after the text is set 
 - we need to know the type of alignment (to store it, but it will be ugly?)

Maybe I am missing something so if you have some ideas which can work, please be free to share, thanks!

mchowning:
> Even if we have ParagraphSpan it seems that the problem is caused by HiddenHtmlBlock.

I bet you're right. Since `HiddenHtmlSpan` implements `IAztecAlignmentSpan` it's going to always ignore the `View`'s gravity. I wonder if it would be possible to make it so that there is a `HiddenHtmlSpan` class that does not extend `IAztecAlignmentSpan`, which we could then use to allow the `gravity` to take effect. Similar to [what I did with `ParagraphSpan`](https://github.com/wordpress-mobile/AztecEditor-Android/pull/851)? My suspicion is that will not be an easy change since it looks like there were specific needs that led to [adding alignment to the `HiddenHtmlSpan`](https://github.com/wordpress-mobile/AztecEditor-Android/pull/624).

>The easiest solution is to call method: view.toggleFormatting(AztecTextFormat.FORMAT_ALIGN_CENTER);
but there is a problem how to determinate when to call the method...

Am I understanding that all of your problems would go away if just used `view.toggleFormatting(...)` for all alignments? I know I ran into a lot of issues trying to use that for paragraphs (I outlined some of those issues in my [PR description](https://github.com/wordpress-mobile/AztecEditor-Android/pull/851)), but we can always reassess that decision if we need to.

mchowning:
I have opened an [AztecEditor-Android  PR](https://github.com/wordpress-mobile/AztecEditor-Android/pull/899) that I believe will address the alignment issue. That does not solve the problem of the placeholder not showing though, so that still needs some investigation. As [you noted](https://github.com/wordpress-mobile/gutenberg-mobile/issues/1970), the quote block also has the same problem with the placeholder, so I don't think that issue should block this PR.

SergioEstevao:
You need to update this PR with the latest changes in develop, after the merge if the CI tests pass feel free to merge.

ceyhun:
Closing as gutenberg PR is merged and gutenberg ref already updated by another merged PR.

PR #1925, opened 2020-02-18
e2e test update ios simulators versions

SergioEstevao:
This PR just updates the simulators being used for e2e tests and refactor the code a bit to make clear what simulators are used in CI and the one used locally.

To test:
 - Run the iOS e2e tests locally: `yarn test:e2e:ios:local`
 - Check if all is ok.

PR submission checklist:


__device-tests__/helpers/caps.js
```diff
@@ -1,15 +1,27 @@
-exports.ios = {
+const ios = {
 	browserName: '',
 	platformName: 'iOS',
-	platformVersion: '13.0',
-	deviceName: 'iPhone 11 Simulator',
+	platformVersion: '13.3',
```
dratwas: I think that the `platformVersion` and `deviceName` could be removed from here since the `ios` is not exported. They are declared in `iosLocal `and `iosServer`.

Issue #1746, opened 2020-01-06
Add Latest Posts block - 1st iteration


pinarol:
Would that be possible to come up with a simple 1st iteration for this? @iamthomasbishop This time I am talking about something that can also ship to prod. What I have in mind is not showing block previews but just allowing the user to change the most important block settings, wdyt?

iamthomasbishop:
Are we talking about the "Latest Posts" block that is in core? I know there was some discussion around which block will end up being used in Core (there are a few that are on the radar), so I want to make sure our target is aligning with core.

For reference, the current Latest Posts block looks like this:

<img width="973" alt="image" src="https://user-images.githubusercontent.com/1200432/71851975-59a1dc00-309d-11ea-80f2-c294c0087ff7.png">

pinarol:
> Are we talking about the "Latest Posts" block that is in core?

Yes, this issue is about the core one.

lukewalczak:
Hi @iamthomasbishop , can you upload some design for that block?

chipsnyder:
Hey, @iamthomasbishop I'm getting ready to start looking into the support for this. Do we have any designs ready for it yet?

chipsnyder:
Chatting with @iamthomasbishop, we discussed the following:

For the canvas, we won't focus on adding any block specific editing to the canvas but instead have a basic UI with the standard controls for moving the block up and down.

For the Block settings, we'll add support for:
- Post Content
  - Excerpt / Full Post (Only shown if Post Content is toggled on)
  - Maximum Number of words (Only shown if Post Content is toggled on AND Excerpt is selected)
- Display Post Date
- Order By
- Category
- Number of Items

For "Post Content," I'll start by investigating some flows around that to see if there are any blockers around the conditional rows or if the flow feels weird.

PR #1931, opened 2020-02-18
Latest Posts Block v1 Support in Mobile

chipsnyder:
**Fixes:** https://github.com/wordpress-mobile/gutenberg-mobile/issues/1746
**Related PRs:**
- `gutenberg` https://github.com/WordPress/gutenberg/pull/20301
- For testing:
    - `WordPress-iOS` https://github.com/wordpress-mobile/WordPress-iOS/pull/13495
    - `WordPress-Android` https://github.com/wordpress-mobile/WordPress-Android/pull/11338

## To Test

These tests were validated via WPiOS and WPAndroid 
Categories make a network request to fetch the available categories, so they were tested against a Wordpress.com site as well as a self-hosted site.

### Add New Block 
1) Select an existing Page or create a new one
2) Add a Latest Posts Block

**Expect:** block to be added

3) Tap on the block or the gear menu

**Expect** the settings to show
**Validate** the block is configured based on the settings in the table above

### Modify Block
This should be tested by modifying in mobile and validating on the web and by modifying on the web and testing on mobile.

1) Select a Latest Post Block and bring up the settings
2) Modify Settings

**Conditional items:** 
- Tapping on the toggle for "Show Post Content" should add a row for "Show" with the options of "Excerpt" or "Full Post" depending on the configuration of the block.
- Changing "Show" to "Full Post" should hide "Max number of words in excerpt" 
- Changing "Show" to "Excerpt" should show "Max number of words in excerpt" with the last selected value

3) Close the settings
4) Save as a draft or publish
5) Leave the screen

**Validate** by checking the page on the other platform and by going into the same page/post that all settings match how you configured them.

### Categories known behaviors
Categories are fetched in `componentDidMount` which results in a flow where multiple Latest-Posts Blocks can show different data. 
1) Open a page
2) Add a Latest-Post Block
**- Take Note of the Categories -** 
3) Modify the available categories. If on the web, do this in a new tab.
**- Note the first block does not reflect the changes in the categories -** 
4) Add a new Latest-Post Block
**- Note the second block does reflect the changes in the categories -** 

This behavior is consistent on the web and mobile. Leaving the page and coming back will sync the categories.

## Screenshots 
latestPost

## PR submission checklist:


iamthomasbishop:
@chipsnyder Great work thus far! Per our discussion in Slack on Friday, I have some feedback. I see you've already made some changes and I've pulled down the latest build and I'll update my notes as a result 😄 I have only tested in iOS thus far, but I am assuming most of these issues also exist on Android. So I will give Android a good review on Monday. 

## Relevant Issues

* After seeing the Post Content Settings part in action again, it feels a little backwards to me to _toggle off_ an action to show a conditional cell below it. So would it make more sense to change the second toggle to be `Only show excerpt` or similar? (example: https://cloudup.com/cdJj9w-cE-C). It might also help to put a footnote below it that explains what's going on. If the default actually _is_ excerpt, then we can flip this (toggle on by default, and footnote might not be as relevant as a result).

* In dark mode, the `Customize` button text color should be the same as the main link color in dark mode (`$blue-30`)

* Separators should extend to the 16px keylines (example: https://cloudup.com/cIuFHpgqcFd)

* We can remove the last separator in each table section (see previous example)

* I think `Excerpt length` might be more succinct as a label instead of `Max number of words in excerpt`, but it's a very minor thing

## Unrelated issues

* After interacting with an input sometimes you can't scroll back upward on the bottom sheet

chipsnyder:
> So would it make more sense to change the second toggle to be Only show excerpt or similar? (example: https://cloudup.com/cdJj9w-cE-C). It might also help to put a footnote below it that explains what's going on. If the default actually is excerpt, then we can flip this (toggle on by default, and footnote might not be as relevant as a result).

The default is `excerpt` so changing that toggle as you suggest is good with me.


> In dark mode, the Customize button text color should be the same as the main link color in dark mode ($blue-30)

👍 

> Separators should extend to the 16px keylines (example: https://cloudup.com/cIuFHpgqcFd)
We can remove the last separator in each table section (see previous example)

Oops, those changes must not have made it in the build I'll double-check that they're included. 

> I think Excerpt length might be more succinct as a label instead of Max number of words in excerpt, but it's a very minor thing

Yeah, I wasn't the biggest fan of the original wording either. 


These should all be straight forward changes. I'll get something together for you later today.

iamthomasbishop:
> Oops, those changes must not have made it in the build I'll double-check that they're included.

I just deleted and re-installed the [test build](https://github.com/wordpress-mobile/WordPress-iOS/pull/13495) and I'm seeing them now — not sure if this is an updated build or I just missed them in my second review All good now! 😄

> These should all be straight forward changes. I'll get something together for you later today. Thanks for taking a look at the updated tweaks :)

Awesome, thank you! 🙏

PR #1934, opened 2020-02-19
Contact Info jetpack native blocks behind DEV flag

Tug:
This is a work in progress

See Jetpack PR: https://github.com/Automattic/jetpack/pull/15389

PR submission checklist:


cameronvoell:
Issues remaining:

- [ ] Save a block and return, it should still show as a contact block
- [ ] Add a contact info on the web, edit on mobile, it should show
- [ ] Fix weird focus problem where you can type in address/email/phone without the inner lock showing as selected

src/jetpack-editor-setup.js
```diff
@@ -19,7 +19,7 @@ const supportedJetpackBlocks = {
 	},
 };
 
-const setJetpackData = async ( {
```
cameronvoell: Without removing this, we would start parsing html to blocks before the jetpack blocks got registered. I'm not sure if this changes our existing thoughts on when we should check for jetpack permission. Perhaps it should be done on the app side. But it will at least have to block the parsing of HTML to blocks otherwise existing jetpack blocks will show up as unknown blocks.

Tug: Agreed block parsing (and thus loading the editor) should be done afterwards. We can delay the loading of the editor for a bit but it wouldn't be a great UX if it happened each time. It would be nice if we could get jetpack data from the parent app, or at least cache here it somehow.

src/index.js
```diff
@@ -55,6 +56,7 @@ export class RootComponent extends React.Component {
 		super( props );
 		setupLocale( props.locale, props.translations );
 		setupApiFetch();
+		setupJetpackEditor( props.jetpackState || { blogId: 1, isJetpackActive: true } );
```
Tug, line 59: We'll have to make this async when we actually fetch the data from the site, but for now let's keep it that way 👍

src/jetpack-editor-setup.js
```diff
@@ -0,0 +1,47 @@
...
+};
+
+export default ( jetpackState ) => {
+	if ( ! jetpackState.isJetpackActive ) {
+		return;
+	}
+
+	const jetpackData = setJetpackData( jetpackState );
+
+	require( '../jetpack/extensions/editor' );
```
Tug, line 44: we need to add a __DEV__ flag here

cameronvoell:
@iamthomasbishop As we mentioned in the chat this feature is initially being added behind a DEV flag while we iterate on some of the API requirements. In the mean time, it can be tested in it's current state on iOS using AppCenter build number: 27246 ([test PR with link here](https://github.com/wordpress-mobile/WordPress-iOS/pull/13918#issuecomment-614964417)). 

I created this issue for tracking design review and api requirements for enabling this block in production: https://github.com/wordpress-mobile/gutenberg-mobile/issues/2166

src/index.js
cameronvoell: Sounds good. Just going to note here that if we make an async call to check if we're enabling jetpack blocks, then we need to block the action of parsing raw html to blocks until after that function returns, otherwise jetpack blocks will show up as not enabled, even if api response says they should be enabled.

src/jetpack-editor-setup.js
```diff
@@ -0,0 +1,49 @@
...
+	const jetpackEditorInitialState = {
+		available_blocks: availableBlocks,
+		jetpack: {
+			is_active: isJetpackActive,
+		},
+		siteFragment,
+		tracksUserData: userData,
+		wpcomBlogId: blogId,
+	};
+	global.window[ JETPACK_DATA_PATH ] = jetpackEditorInitialState;
```
cameronvoell, line 33: Just want to ask you @Tug about this line since it's hard to understand with the API calls temporarily disabled. Are we utilizing `global.window` here so jetpack code can access `jetpackEditorInitialState` when initializing, similar to how they're doing it on the web side in [jetpack/extensions/editor.js](https://github.com/Automattic/jetpack/blob/master/extensions/editor.js)?

maxme:
I was wondering if the bundle went bigger with the dependency to Jetpack (and sub-dependencies). Can we do a sanity check and also verify if the loading times?

cameronvoell:
> I was wondering if the bundle went bigger with the dependency to Jetpack (and sub-dependencies). Can we do a sanity check

@maxme I just performed yarn bundle on gutenberg-mobile develop before and after this PR is merged and I didn't see any increase in the size of the bundle directory.

Before jetpack PR (commit ca78090f8386677259cdfd3727054374a70760ab):
5.0M	bundle/ios/assets
 24M	bundle/ios
5.0M	bundle/android/raw
 24M	bundle/android
 48M	bundle

After jetpack PR (commit fd39dd9d2f31418933d48529c4a77d8747f23937)
5.0M	bundle/ios/assets
 24M	bundle/ios
5.0M	bundle/android/raw
 24M	bundle/android
 48M	bundle

> also verify if the loading times?
I can check this too, but figure it's worth clarifying first since I think you might have had a typo. If I'm understanding you correctly, a robust check of bundle load times would be to test how long it takes to launch the gutenberg editor in WPiOS and WPAndroid before and after the jetpack code is added. Let me know if that's what you meant, or it was something else.

cameronvoell:
See differences in App.js files as well, before and after the update. Looks like it should be negligible. 

  | Before Jetpack | After Jetpack | differences
-- | -- | -- | --
android App.js | 8523434 | 8523915 | -481
android App.js.map | 11217927 | 11231035 | -13108
  |   |   |  
ios App.js | 8541370 | 8541748 | -378
ios App.js.map | 11269901 | 11283125 | -13224

maxme:
Thanks @cameronvoell - By `also verify if the loading times?` I meant  `Can we also check if loading times went up?` ;) - but let's forget that since the bundle size didn't change much.

PR #1938, opened 2020-02-20
Issue/1916 fix align issues

### LINKED ISSUE #1916: [GlobalStep] iOS – Alignment is not displayed after pressing "return"
### Description
When typing into a Paragraph Block, selecting a Text Alignment causes it to be displayed with the selected alignment but pressing the keyboard’s “return” key causes the preview of that block while on the Block Editor to display the text aligned to the left despite the setting being properly saved and the post accurately published with the selected text alignment setting.

### Reproduction Rate
4/4 100%

### Expected behavior
The preview of the block should retain the selected text alignment.

### Actual behavior
The preview of the text alignment defaults to the left.

1. Install WordPress 14.2
2. Log in to an account.
3. Press “Add new”.
4. Tap the “Start writing...” block.
5. Select the “Align Right” setting.
6. Type some content.
7. Press “return”.
##### Tested on the following
iPhone SE (12.0.1)

#### Please see the attached video for more information
[iOSAlignment.zip](https://github.com/wordpress-mobile/gutenberg-mobile/files/4205187/iOSAlignment.zip)

### Submitted by:
Luis Pimenta

SergioEstevao:
Fixes #1916 

This PR fixes the set alignment property on RCTAzteView to avoid conflict condition with the alignment property on the base TextView Class.

To test:
 - Open the demo app
 - Add a paragraph block
 - Write some text
 - Set the alignment to Right or Center
 - Check if the alignment is set
 - Tap Enter
 - Check that the alignment stays set on the previous block
 - Switch to HTML
 - Check that the HTML has the alignment attributes
 - Go back to visual and check if the alignment is applied
 - Now tap Enter in the middle of the text of the block
 - Check that both block keep the alignment.

PR submission checklist:

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -8,6 +8,7 @@
 * [Android] Fix issue preventing correct placeholder image from displaying during image upload
 * [iOS] Fix diplay of large numbers on ordered lists
 * Fix issue where adding emojis to the post title add strong HTML elements to the title of the post
+* [iOS] Fix issue where alignment of paragraph blocks were not being respected whem splitting the paragraph or reading from the original Post content.
 
 1.22.0
 ------
--- a/react-native-aztec/ios/RNTAztecView/RCTAztecView.swift
+++ b/react-native-aztec/ios/RNTAztecView/RCTAztecView.swift
@@ -37,10 +37,14 @@ class RCTAztecView: Aztec.TextView {
     }
 
     override var textAlignment: NSTextAlignment {
-        didSet {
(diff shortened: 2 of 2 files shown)


SergioEstevao:
@mchowning I didn't fully tracked down the issue, but it looked like when we were using `didSet` we were calling `super.setAlignment` and that was causing the value being set on the base `UITextView` and then I think the `didSet` on `TextView` didn't track down properly the oldValue difference.

In either case, I don't think we were doing the correct step of calling a `super.setTextAlignment` on a `didSet`.

Issue #1274, opened 2019-08-08
Autoscroll regressions 


etoledom:
1. On older phones (no X family), autoscroll won't show the inner toolbar:

![Autoscroll-old](https://user-images.githubusercontent.com/9772967/62698412-bf524500-b9e5-11e9-86c2-90fbaebc11c3.gif)

2. On list block, autoscroll won't scroll unless the caret is completely out of view. On previews version it used to scroll right after some part of it goes out of sight.

![Autoscroll-lists](https://user-images.githubusercontent.com/9772967/62698491-ee68b680-b9e5-11e9-9597-11aa0b85e769.gif)

3. With the phone on landscape, the inner toolbar is not completely visible:

![Autoscroll-landscape](https://user-images.githubusercontent.com/9772967/62698560-17894700-b9e6-11e9-9928-11396fa2e769.gif)

cc @Tug

Issue #891, opened 2019-04-18
Autoscroll doesn't work properly when the software keyboard is hidden.


etoledom:
This can happen when a hardware keyboard is attached, but I have been able to test it just with the laptop keyboard and the simulator.

When Gutenberg is presented without the software keyboard present, autoscroll won't work.

![doesnt-work](https://user-images.githubusercontent.com/9772967/56377594-00420600-620b-11e9-9152-7c1dbae0abbe.gif)

When the software keyboard is set to hidden while Gutenberg is present (cmd+k on the simulator), the list block gets a jumpy behavior on the ListBlock, scrolling down on Enter, and scrolling back up when typing a letter. This doesn't seem to happen on the paragraph block.

This might be the case of connecting a hardware keyboard to a real device.

![jumpy](https://user-images.githubusercontent.com/9772967/56377657-1bad1100-620b-11e9-9f81-35f0099fa838.gif)

PR #1939, opened 2020-02-20
[iOS] Fix autoscroll behaviour.

### LINKED ISSUE #1275: iOS autoscroll not working when typing goes under the keyboard
iOS autoscroll not working when typing goes under the keyboard for about 2 lines, this looks related to the removed props in [this PR](https://github.com/WordPress/gutenberg/pull/16677), the underlying code probably doesn't understand the caret going under the keyboard as it used to before.

auto-scroll

Old version:
auto-scroll-old

To repro this bug: put the caret in the middle of the text and type

cc @Tug

### LINKED ISSUE #1276: Tech debt: Refactoring of the magic number used in the calculation of extraScrollHeight
In this [line](https://github.com/WordPress/gutenberg/blob/67b2328cfc2c0fef6c0ab54f69e1c5aa808f2fa3/packages/block-editor/src/components/block-list/index.native.js#L104) there's a "+ 10" in the calculation of extraScrollHeight. Let's analyze what this refers to and try to replace it with a real style value, if this is not applicable let's try to give it a name to make it more clear.

First introduced in PR https://github.com/WordPress/gutenberg/pull/16677

etoledom:
This PR tries to fix wrong autoscroll behaviour on iOS.

Fixes https://github.com/wordpress-mobile/WordPress-iOS/issues/13224
Fixes https://github.com/wordpress-mobile/gutenberg-mobile/issues/1274
Fixes #1276
Fixes #1275
Fixes https://github.com/wordpress-mobile/gutenberg-mobile/issues/891

**Note:** Aztec changes on Podfile are unrelated. It might be a missing `yarn preios` on a previous PR.

`gutenberg` PR: https://github.com/WordPress/gutenberg/pull/20334
`rn-keyboard-aware-scroll-view` PR: https://github.com/wordpress-mobile/react-native-keyboard-aware-scroll-view/pull/9

### Enhancements:
- The position to scroll to is now properly calculated (to show the inner toolbar).
- The `is caret under keyboard` calculation used to kick the autoscroll on typing now considers the extra height (the bars). In this way, if the bottom part of the inner toolbar goes behind the global toolbar, it will autoscroll.
- Autoscrolling on lines that are not the last one, now will show a bit more space below the line. This is just to give more "air" to it. (cc @iamthomasbishop to know your opinion about it).

All these changes are demonstrated on the following gifs:

autoscroll-iphone
autoscroll-ipad

### To test:
- Follow the examples on the provided Gifs.
- Test with longer texts too.
- Test on other RichText powered blocks.

**NOTE:** For the external keyboard test is necessary to test with a real device.

@pinarol - Any other edge case that you consider we should test?

PR submission checklist:


SergioEstevao:
@etoledom The link for the rn-keyboard-aware-scroll-view PR is incorrect.

etoledom:
> @etoledom The link for the rn-keyboard-aware-scroll-view PR is incorrect.

Fixed, thank you!

pinarol:
> @pinarol - Any other edge case that you consider we should test?

On the top of my head: 

- devices with/without safe area
- switching between landscape&portrait
- text inputs other than rich text
- merging/splitting blocks
- rich text inside group block

pinarol:
> The is caret under keyboard calculation used to kick the autoscroll on typing now considers the extra height (the bars). In this way, if the bottom part of the inner toolbar goes behind the global toolbar, it will autoscroll.

This can be related to removed parameters during the first set of inner blocks feature:

Screen Shot 2020-02-21 at 20 33 43

659b32723f8e7c4e401f97705a4e7a133ca663da

There are already some props we can use for this, like: inputAccessoryViewHeight, extraBottomInset. So before introducing new fixes we could benefit from trying these.

pinarol:
Thanks! it is so much better! I tested this with following:

- [x] In a line other than the last line: typing till the caret is in the next line and checking if caret gets under the keyboard
- [x] switching between landscape&portrait
- [x] text inputs other than rich text(but keep in mind there’s no scroll to caret for them)
- [x] merging/splitting blocks
- [x] rich text inside group block
- [ ] Devices with/without safe area > Failed ❗️

It looks like the current mechanism acts differently on devices without safearea, and the extra caret space isn't working:

non-iphonex

I think this is not too big of a deal on user's end but it is worth investigating because we made it so far and it'd be good not to leave a mystery bug for the future 🤔

etoledom:
Thank you for testing @pinarol !

This last issue should be fixed by https://github.com/wordpress-mobile/react-native-keyboard-aware-scroll-view/pull/10

I updated this PR to include that change, so it can be tested after pulling and `yarn install`

etoledom:
Thanks for testing @pinarol 🎉 
Added the extra issues this PR fixes to the description.
5 issues going away :yay:

PR #1936, opened 2020-02-19
Add logging to Page Template Picker

chipsnyder:
**Fixes:** https://github.com/wordpress-mobile/gutenberg-mobile/issues/1883
**Related PRs:**
- `gutenberg` https://github.com/WordPress/gutenberg/pull/20317
- `WordPress-iOS` https://github.com/wordpress-mobile/WordPress-iOS/pull/13510
- `WordPress-iOS-Shared` https://github.com/wordpress-mobile/WordPress-iOS-Shared/pull/250
- `gutenberg-mobile Android` https://github.com/wordpress-mobile/gutenberg-mobile/pull/1945
- `WordPress-Android` https://github.com/wordpress-mobile/WordPress-Android/pull/11362
---
### Create a new page using a template 
1) Open the WordPress iOS app 
2) Navigate to pages and create a new page
3) Select one of the starter page templates.
4) Select "Apply"
5) Edit the page if desired 
6) Then save or publish and exit

**Expect** 
- The event `editor_session_end` should have an additional property of `template` with the name of your selected template
- A new event titled `editor_session_template_preview` should be generated with a property of `template` with the name of your selected template
- A new event titled `editor_session_template_apply` should be generated with a property of `template` with the name of your selected template

### Start a page stop and continue
1) Open the WordPress iOS app 
2) Navigate to pages and create a new page
3) Add a title to the page
4) Save as draft 
5) Open the page again
6) Select a template and Apply it
7) Then save or publish and exit

**Expect** 
- The event `editor_session_end` should have an additional property of `template` with the name of your selected template
- A new event titled `editor_session_template_preview` should be generated with a property of `template` with the name of your selected template
- A new event titled `editor_session_template_apply` should be generated with a property of `template` with the name of your selected template

### Create a new page and cancel the template
1) Open the WordPress iOS app 
2) Navigate to pages and create a new page
3) Select one of the start page templates.
4) Select "Close"
5) Edit the page if desired 
6) Then save or publish and exit

**Expect** 
- The event `editor_session_end` should **_NOT_** have an additional property of `template` with the name of your selected template
- A new event titled `editor_session_template_preview` should be generated with a property of `template` with the name of your selected template
- `editor_session_template_apply` should **_NOT_** be generated 

## PR submission checklist:

- [x] I have considered adding unit tests where possible.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #1945, opened 2020-02-21
[Android] Add logging to Page Template Picker

### LINKED ISSUE #1883: Add metrics for SPTs
We should be able to track when a template was applied in an editor session.

I initially thought about using a `@wordpress/data` plugin to hook into store actions, but considering that the feature is mobile only for now, it would be much simpler to just fire an event from the picker directly.

How I imagine this working:

1. We add a method to the bridge to send user actions to the apps (`RNReactNativeGutenbergBridge.logUserEvent( name, properties )`). The picker sends a `page_template_applied` event with the template name in `properties`.
2. When the template picker applies a template, it calls this bridge method.
3. The apps fire a `editor_session_template_apply` event with a `template` property containing the name.
4. The apps keep track of the last applied template in the editor session ([iOS](https://github.com/wordpress-mobile/WordPress-iOS/blob/develop/WordPress/Classes/ViewRelated/Post/PostEditorAnalyticsSession.swift), [Android](https://github.com/wordpress-mobile/WordPress-Android/blob/develop/WordPress/src/main/java/org/wordpress/android/ui/posts/PostEditorAnalyticsSession.java)), and include it as part of the `editor_session_end` event in a `template` property.

mkevins:
Fixes #1883

This is the Android part of this PR: https://github.com/wordpress-mobile/gutenberg-mobile/pull/1936

#### Related PRs:

* `gutenberg` https://github.com/WordPress/gutenberg/pull/20317
* `WordPress-iOS` https://github.com/wordpress-mobile/WordPress-iOS/pull/13510
* `WordPress-Android` https://github.com/wordpress-mobile/WordPress-Android/pull/11362

#### To test:

Test via steps here: https://github.com/wordpress-mobile/WordPress-Android/pull/11362#issue-378287710

PR submission checklist:


react-native-gutenberg-bridge/android/src/main/java/org/wordpress/mobile/WPAndroidGlue/WPAndroidGlueCode.java
```diff
@@ -317,6 +324,16 @@ public void requestMediaEditor(MediaUploadCallback mediaUploadCallback, String m
                 mPendingMediaUploadCallback = mediaUploadCallback;
                 mOnMediaEditorListener.onMediaEditorClicked(mediaUrl);
             }
+
+            @Override
+            public void logUserEvent(GutenbergUserEvent event, ReadableMap eventProperties) {
+                switch (event) {
+                    case PAGE_TEMPLATE_APPLIED:
+                        mOnLogGutenbergUserEventListener.onGutenbergDidLogSessionTemplateEvent(eventProperties.getString(
```
koke: I'm not familiar with conventions on WPAndroid or Kotlin, but I wonder if it'd be less verbose if we had just one callback method and let the app switch through the different cases of the enum?

react-native-gutenberg-bridge/android/src/main/java/org/wordpress/mobile/ReactNativeGutenbergBridge/GutenbergBridgeJS2Parent.java
```diff
@@ -81,6 +84,28 @@ public static MediaType getEnum(String value) {
         }
     }
 
+    enum GutenbergUserEvent {
+        PAGE_TEMPLATE_APPLIED("page_template_applied");
```
chipsnyder: I went through and made some changes thanks to the clarification in https://github.com/wordpress-mobile/gutenberg-mobile/issues/1883#issuecomment-590247803 I renamed this event to `editor_session_template_apply`

react-native-gutenberg-bridge/android/src/main/java/org/wordpress/mobile/ReactNativeGutenbergBridge/GutenbergBridgeJS2Parent.java
```diff
@@ -81,6 +84,28 @@ public static MediaType getEnum(String value) {
...
+        String name;
+
+        GutenbergUserEvent(String name) {
+            this.name = name;
+        }
+
+        public static GutenbergUserEvent getEnum(String eventName) {
+            return MAP.get(eventName);
+        }
+    }
```
mchowning: Do you think it would work to simplify this enum a bit along these lines?

```
enum GutenbergUserEvent { 
  PAGE_TEMPLATE_APPLIED;

  @Nullable 
  public static getEnum(String eventName) {
    switch (eventName) {
      case "page_template_applied": return PAGE_TEMPLATE_APPLIED;
      default: null;
    }
  }
}
```

mkevins: The benefit of using the static map and static initializer is that we only need to add new events in one place. Your example is cleaner, but will require new entries in two places. Not a huge issue, but that's the reason I went with the static map.

react-native-gutenberg-bridge/android/src/main/java/org/wordpress/mobile/ReactNativeGutenbergBridge/GutenbergBridgeJS2Parent.java
mkevins: Thanks, I'll make a similar change. :+1:

react-native-gutenberg-bridge/android/src/main/java/org/wordpress/mobile/WPAndroidGlue/WPAndroidGlueCode.java
mkevins: I agree. Initially, I tried to multiplex the entire "interface chain" to reduce the total usage of forwarding methods, but couldn't use `ReadableMap` from the parent app scope. I was able to partially address this with the `toHashmap` method in this commit: https://github.com/wordpress-mobile/gutenberg-mobile/pull/1945/commits/a7b991d6ae7567918917ec24c54dd526f2f78958.

mkevins:
Thanks for reviewing :+1: 

Merging this into the main feature PR, as the iOS device test error: `error: type 'GutenbergViewController' does not conform to protocol 'GutenbergBridgeDelegate'` will be solved there.

PR #1969, opened 2020-02-28
[iOS] Fix [Obj] character after dictation

etoledom:
This PR fixes an issue where an unwanted character was inserted after dictation.

And also, now dictations on the middle of the string will be positioned correctly (not always at the end).

To test:
- Tap on the Post Title.
- Start dictating with voice.
- When finished, press Enter.
- Check that there is no [UNDEFINED] appended to the post title.
- Check for regressions on #606

PR submission checklist:


etoledom:
Thank you! 🙏

PR #1981, opened 2020-03-04
Add accessibility label to Block List Footer

shiki:
Fixes part of https://github.com/wordpress-mobile/WordPress-iOS/issues/13225. I worked with @jd-alexander on this during the Gutenberg Day. 🙂 

Please see  https://github.com/WordPress/gutenberg/pull/20633 for the changes.

## Findings

As described in the **Unlabeled Element** section of https://github.com/wordpress-mobile/WordPress-iOS/issues/13225, there is an accessible element below the last block. It does not have a label. But it can be activated, which creates a new block. 

## Solution

We thought about making the element inaccessible but it looks like that element is used for _adding a block at the bottom_ while the _Add Block_ button on the toolbar is used for _adding a block next to the currently selected block_. So we decided to add an accessibility label instead.

## Testing

1. Turn VoiceOver/TalkBack on
2. Navigate to the Block List Footer
3. Confirm that VoiceOver/TalkBack speaks "Add block". 

## Reviewing 

We're not sure if this is the correct process for submitting PRs. Please let us know if it should be done in another way. 🙂 

## Submitter Checklist

- [x] If it's feasible, I have added unit tests. 


shiki:
Thank you very much, @etoledom and @Tug. 🙏 I have learned a lot from this experience. I'm sure @jd-alexander can say the same.

jd-alexander:
Yes indeed @shiki Thanks much @etoledom  @Tug 🙏

PR #2009, opened 2020-03-12
Dismiss bottom sheet on return key submit

### LINKED ISSUE #1164: [Bottom Sheets] Tapping "return" should dismiss sheet
### Expected

When I'm editing something on a bottom sheet, I expect tapping "return" to dismiss the sheet.

### Actual

Tapping return dismisses the keyboard, but not the sheet with it.

guarani:
Addresses #1164

To test: See https://github.com/WordPress/gutenberg/pull/20848

PR submission checklist:

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -3,6 +3,7 @@
 * New block: Cover
 * Improve icon on the "Take a Video" media option
 * Removed the dimming effect on unselected blocks
+* Block Editor: When editing link settings, tapping the keyboard return button now closes the settings panel as well as closing the keyboard.
 
 1.24.0
 ------


guarani:
Since using a forked Gutenberg here, this PR needs to be updated before merging.

mchowning:
What do you think about adding a release note for this change?

guarani:
> What do you think about adding a release note for this change?

@mchowning, good question. I'm leaning towards this being too trivial of a change to warrant inclusion in the release notes, but I could be wrong. If we were going to add it, it could be something like this:

_"Block Editor: When editing link settings, tapping the keyboard return button now closes the settings panel as well as closing the keyboard."_

mchowning:
>  I'm leaning towards this being too trivial of a change to warrant inclusion in the release notes

It seems like a user-facing change to me, so I think it should be added to the release notes. Your suggested release note sounds good to me.

PR #2024, opened 2020-03-16
Update Gutenberg ref to use video-alt2 icon

### LINKED ISSUE #1697: Improve icon for the "Take a Video" media option
**Describe the bug**
Currently, the "Take a Photo" and "Take a Video" options when adding media to a Media&Text block use the same icon, which makes it confusing, especially since the copy itself is quite similar between the two options. See screenshot:

**Expected behavior**
It would be better if the icons were different. For the video item, @iamthomasbishop has suggested using Dashicon's `dashicons-video-alt2`.

hypest:
Fixes #1697 

Improves the "Take a Video" option in the choose media bottom sheet by using a different icon than the "camera" one used so far

See screenshot and testing instructions at the Gutenberg PR: https://github.com/WordPress/gutenberg/pull/20927

PR submission checklist:

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -1,3 +1,7 @@
+1.25.0
+------
+* Improve icon on the "Take a Video" media option
+
 1.24.0
 ------
 * New block: Latest Posts


RELEASE-NOTES.txt
```diff
@@ -1,3 +1,7 @@
+1.25.0
+------
+* Improve icon on the "Take a Video" media option
```
etoledom, line 3: This change affected Android only, since on iOS we are not showing images on Media Options.
Please let's prefix these cases with `[Android]` or `[iOS]` as needed 👍

hypest: Sorry I didn't see this earlier. @SergioEstevao is in the process of updating this as part of the 1.25. release process.

PR #2042, opened 2020-03-20
Feat: change dimmed style opacity for InnerBlocks

jbinda:
This PR change the opacity to 1 (so basically it turn off the dimming of unselected block). It's because latest decision about design. I keep the logic because we want to have possibility to bring it back easily and/or adjust the opacity

Please also refer to:
[Related gutenberg PR](https://github.com/WordPress/gutenberg/pull/21042)
[Columns PR](https://github.com/WordPress/gutenberg/pull/19013)
[Buttons PR](https://github.com/WordPress/gutenberg/pull/20191)

After this gets merged we should also update the sanity-test cases for [Group block](https://github.com/wordpress-mobile/test-cases/blob/master/test-cases/gutenberg/group.md#tc007)

To test:
- expect during navigation through nested structure that none of the block gets dimmed

PR submission checklist:


jbinda:
Ci tests passed, however I wonder if I shouldn't add some note in `RELEASE-NOTES.txt` because it affects current production behaviour about dimming ? @pinarol wdyt ?

pinarol:
Good point. It’d be good to add a note @jbinda

RELEASE-NOTES.txt
```diff
@@ -1,6 +1,7 @@
 1.25.0
 ------
 * Improve icon on the "Take a Video" media option
+* Change dimmed style opacity to 1 for unselected block when navigating in nested structure
```
pinarol: Can we make it more easy to understand for external people/users?

```suggestion
* Removed the dimming effect on unselected blocks
```

jbinda:
sure, done !

PR #2059, opened 2020-03-24
Remove reference to nonexistent mocks

guarani:
## Description

We use the `node-libs-react-native` module to provide React Native compatible implementations of Node core modules. This module exports a mapping of module name to module paths and this is configured in our Metro configuration files, `rn-cli.config.js` and `extra-node-modules.config.js` via the `extraNodeModules` parameter.

While looking into why sass changes were not live reloading I came across `extra-node-modules.config.js`. This file mocks some Node modules that are not used, but the mock it points to, `__mocks__/nodejsMock.js`, doesn't exist anymore (deleted in https://github.com/wordpress-mobile/gutenberg-mobile/pull/1215/files). Since `node-libs-react-native` [already sets the path for the `fs` and `net` modules to `null`](https://github.com/parshap/node-libs-react-native/blob/cd8d78a520cd002690a448d6d30b7ad0551ce301/index.js#L12), the PR simply removes our mock for these two modules. `node-libs-react-native` does not mock `canvas`, so to keep behavior the same, this PR sets its path to `null`.

To test:
1. Make sure all CircleCI tests are green

PR submission checklist:


SergioEstevao:
@guarani  I'm moving this to 1.26, feel free to move it back to 1.25 if this is something that can be considered a regression.

pinarol:
I am changing the milestone of this PR as 1.27 because the 1.26 release branch will be cut quite soon. Let me know if you instead wanted to include this to 1.26.

mkevins:
Hi @guarani :wave: :smiley: since this one is still awaiting review, I'll bump the milestone to `1.28`. Please let me know if you wanted to get this into the `1.27` release.

extra-node-modules.config.js
```diff
@@ -1,7 +1,5 @@
 /** @format */
 const enm = require( 'node-libs-react-native' );
-enm.fs = __dirname + '/__mocks__/nodejsMock.js';
-enm.net = __dirname + '/__mocks__/nodejsMock.js';
-enm.canvas = __dirname + '/__mocks__/nodejsMock.js';
+enm.canvas = null;
```
ceyhun, line 3: I think this file is not used anywhere, anymore. Maybe we can delete it?

rn-cli-inside-gb.config.js
```diff
@@ -26,7 +24,6 @@ const alternateRoots = [ path.resolve( __dirname, '../node_modules' ) ].concat(
 );
 
 module.exports = {
-	extraNodeModules: Object.assign( enm, wppackages ),
```
ceyhun, line 29: I'm not sure if `wppackages` is used in this case but I think it's better to only remove `enm` and keep it like `extraNodeModules: wppackages`

extra-node-modules.config.js
guarani: That's true, thanks for pointing that out. I've removed `extra-node-modules.config.js` now.

rn-cli-inside-gb.config.js
guarani: Good catch, I missed that!

guarani:
Thanks for the review @ceyhun! I removed `extra-node-modules.config.js` completely and fixed the accidental deletion of `wppackages`.

guarani:
Thanks for the review, @ceyhun!

PR #2060, opened 2020-03-24
Add tests for Latest-Posts Bock

chipsnyder:
Adds a device test for latest posts and adds the block to the initial html in order to help catch regressions

To test:
The CI run of the device tests should be enough to validate

PR submission checklist:

### DIFF
--- a/__device-tests__/pages/editor-page.js
+++ b/__device-tests__/pages/editor-page.js
@@ -21,6 +21,7 @@ export default class EditorPage {
 	headingBlockName = 'Heading';
 	imageBlockName = 'Image';
 	galleryBlockName = 'Gallery';
+	latestPostsBlockName = 'Latest Posts';
 	unorderedListButtonName = 'Convert to unordered list';
 	orderedListButtonName = 'Convert to ordered list';
 
@@ -431,5 +432,21 @@ export default class EditorPage {
 		const textViewElement = await this.getTextViewForHeadingBlock( block, false );
 		const text = await textViewElement.text();
 		return text.toString();
+	}
+
+	// ============================
+	// Latest-Posts Block functions
+	// ============================
+
+	async addNewLatestPostsBlock() {
+		await this.addNewBlock( this.latestPostsBlockName );
+	}
+
+	async getLatestPostsBlockAtPosition( position: number ) {
+		return this.getBlockAtPosition( position, this.latestPostsBlockName );
+	}
+
+	async removeLatestPostsBlockAtPosition( position: number ) {
+		return await this.removeBlockAtPosition( position, this.latestPostsBlockName );
 	}
 }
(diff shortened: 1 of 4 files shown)


chipsnyder:
Looks like Android is having trouble finding the element. I think because it needs to scroll

chipsnyder:
@geriux This is good to go now

__device-tests__/pages/editor-page.js
```diff
@@ -161,10 +162,27 @@ export default class EditorPage {
...
+		if ( ! isAndroid() ) {
+			return await this.driver.elementByAccessibilityId( blockName );
+		}
+
+		// Checks if the Block Button is available, and if not will scroll to the second half of the available buttons.
+		if ( ! await this.driver.hasElementByAccessibilityId( blockName ) ) {
+			for ( let step = 0; step < 5; step++ ) {
+				await this.driver.pressKeycode( 20 ); // Press the Down arrow to force a scroll.
+			}
+		}
```
geriux: Just wondering if you tried to use the `swipeDown` helpers, like let's say we add an ID to the Bottom Sheet and then swipe down to find the element. What do you think?

chipsnyder: I tried using the [TouchActions](http://appium.io/docs/en/writing-running-appium/touch-actions/#touchaction) `TouchAction().press(...).moveTo(...).release()` for some reason I couldn't get it to recognize the touch within the modal. I even tried using it to Click on an already visible view with no luck. 

I then took a look at the [mobile commands](http://appium.io/docs/en/commands/mobile-command/), but `swipeDown` and `scroll` aren't supported for Android, so I didn't bother trying those. 

I could have been missing something for the touch events, but I would probably need some pair programing there to help figure it out.

chipsnyder: More specifically this was the example I was working from:

```
let action = new wd.TouchAction();
action.press({x: 10, y: 10});
action.moveTo({x: 10, y: 100});
action.release();
await action.perform();
```
http://appium.io/docs/en/commands/interactions/touch/touch-perform/ 

For the coordinates, I replaced the `x` and `y` with the middle of the screen and set the `y` to be a few pixels from the bottom of the scroll to a few pixels from the top

SergioEstevao:
@chipsnyder  I'm moving this to 1.26, feel free to move it back to 1.25 if this is something that can be considered a regression.

__device-tests__/pages/editor-page.js
geriux: Thanks for the detailed explanation! Since I'm also new to appium I wanted to investigate a bit but as you mentioned in your findings we are very limited on Android, tried some things with no luck. This makes me realize we should definitely try out Detox. It's kinda crazy we can't swipe 🤯

chipsnyder: Thanks for taking the time to explore your other ideas! I was actually hoping one of your ideas worked 😅 this is a bit of a hacky approach haha

PR #2062, opened 2020-03-25
Enable "Edit" image option for Android

ashiagr:
Fixes https://github.com/wordpress-mobile/WordPress-Android/issues/11147

This PR sets `showMediaEditorButton` to true so that "Edit" image option becomes available for Android too.

As we already merged Media Editing PR https://github.com/wordpress-mobile/WordPress-Android/pull/11101 to `WPAndroid-develop`, I think we can enable it now.

Let me know if there are any other steps to be followed.

PR submission checklist:


react-native-gutenberg-bridge/index.js
```diff
@@ -22,7 +22,7 @@ export const userEvents = {
 	editorSessionTemplatePreview: 'editor_session_template_preview',
 };
 
-export const showMediaEditorButton = isIOS;
+export const showMediaEditorButton = true;
```
mchowning: If this is just going to always be `true` from now on, is there a reason to not just remove it entirely and update (simplify) the logic where it is being used?

ashiagr: Thanks @mchowning ! I tried to remove references for `showMediaEditorButton` const in this PR
https://github.com/WordPress/gutenberg/pull/21165

I replaced `gutenberg` `submodule` in `gutenberg-mobile` with the one in my `forked-branch` and I'm now seeing edit/ replace options in the center instead of start aligned but I made no style changes. Also CI build fails for the PR. 

@leandroalonso , @mchowning can you please advise?

Also wondering if we need to put any dev check?

device-2020-03-26-184935

leandroalonso: @ashiagr in iOS this component uses the native Action Sheet — so it's working ok.

Probably some style was changed along the way and caused that. @mchowning any idea what's the source file that controls this alignment?

mchowning: We'll need to look into this, it's occurring on `develop` and it's also affecting the select an image action sheet, so we'll need to get this fixed before the next Gutenberg-mobile release. I'd say let's go ahead and get this merged and we can fix the bottom action sheet alignment separately. WDYT?

I already approved the gutenberg PR. 🙂 Once that is merged, we can update the submodule ref here and this will be ready to merge too.

PR #2084, opened 2020-03-30
Disable align controls for the Heading block in Android

SergioEstevao:
GB PR: https://github.com/WordPress/gutenberg/pull/21268

To test:
  - Start the demo in Android
 - Check that on heading block the align toolbar is not visible/available.

PR submission checklist:


mchowning:
Let's update the release notes to indicate that alignment is only being added for headings on iOS. 

I would also tweak the wording of that release note since you've updated the alignment toolbar to not appear like a toolbar anymore (i.e., the user now just seems an expandable button). 🎉

PR #2097, opened 2020-04-01
New template for release PRs

SergioEstevao:
This PR will add a new template for release PRs to make it easier to check all the steps needed and to standardize the release PRs.

This template can then be used by using this link `https://github.com/wordpress-mobile/gutenberg-mobile/compare/master...release/1.25.0?expand=1&template=release_pull_request.md`

Or just use the normal pull request flow in github and then add this to url of the pull request preview page: `?expand=1&template=release_pull_request.md`

Fixes #

To test:

PR submission checklist:

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE/release_pull_request.md
+++ b/.github/PULL_REQUEST_TEMPLATE/release_pull_request.md
@@ -1,0 +1,37 @@
+Release for Gutenberg Mobile v1.XX.Y
+
+## Related PRs
+
+- Gutenberg: https://github.com/WordPress/gutenberg/pull/
+- WPAndroid: https://github.com/wordpress-mobile/WordPress-Android/pull/
+- WPiOS: https://github.com/wordpress-mobile/WordPress-iOS/pull/
+
+- Aztec-iOS: N/A
+- Aztec-Android: N/A
+
+## Extra PRs that Landed After the Release Was Cut
+
+- [ ]  PR 1
+- [ ]  PR 2
+
+## Changes
+
+ - Change 1
+ - Change 2
+
+## Test plan
+
+- Use the main WP apps to test the changes above. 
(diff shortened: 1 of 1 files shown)


.github/PULL_REQUEST_TEMPLATE/release_pull_request.md
```diff
@@ -0,0 +1,37 @@
...
+- Aztec-Android: N/A
+
+## Extra PRs that Landed After the Release Was Cut
+
+- [ ]  PR 1
+- [ ]  PR 2
+
+## Changes
+
+ - Change 1
```
mchowning, line 19: It might be worth adding some guidance here for how to determine the relevant changes in the form of a comment (so it doesn't actually show up in in the rendered content, i.e., `<!-- To determine the changes from the last release you need to... -->`). I like moving as many of those kinds of directions to here as is reasonable because then updates to the directions will go through PR review instead of just being changed by one person.

SergioEstevao: I found that for the changes you can just look at the commits show in the pull request itself instead of doing the terminal git command.

mchowning: Sure, that works too. I just thought it might be nice to provide some guidance regarding what should be included as "Changes".

That does make me think though, is there any risk of us missing changes where
1. change #1 is merged into gutenberg
2. change #2 is merged into gutenberg
3. change #2 is merged into gutenberg-mobile
4. since ^that^ brought in change #1  already, there is no pr explicitly bringing change #1 into Gutenberg-mobile, so we miss it.

I think we're mostly ok because there should still be a release note for change #1 that would need to get merged, but we do forget if we forgot to update the release note we'd miss testing it too. This is more me just thinking out loud and not an issue with this PR.

PR #2099, opened 2020-04-02
[UI tests] Run only canary tests on a PR

Stojdza:
Run a subset of tests instead of a full suite against every PR. This is an experiment and the purpose of it is to reduce waiting time for CircleCI jobs and avoid running full suite against every commit. More info - p9ugOq-106-p2

Things to consider before we go with it:
- What tests/functionalities we want to cover with canary tests?
- Number of [workers](https://github.com/wordpress-mobile/gutenberg-mobile/pull/2099/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R105)? With a smaller number of tests the jobs will wait less time when we have _high traffic_. But, should we reduce number of workers? Right now it's [three](https://github.com/wordpress-mobile/gutenberg-mobile/blob/develop/package.json#L104). 

To test:
Look at the tests in CircleCI jobs. Are they green? 
What about execution time? Are they faster than before? 

Note:
This option should not go to `develop` before we add the option to run the full suite against PR.  

PR submission checklist:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -84,7 +84,7 @@ jobs:
             curl -u "$SAUCE_USERNAME:$SAUCE_ACCESS_KEY" -X POST -H "Content-Type: application/octet-stream" https://saucelabs.com/rest/v1/storage/automattic/Gutenberg-$SAUCE_FILENAME.apk?overwrite=true --data-binary @./android/app/build/outputs/apk/debug/app-debug.apk
       - run:
           name: Run Device Tests
-          command: yarn device-tests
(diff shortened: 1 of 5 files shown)


Stojdza:
UI tests jobs' names should be updated in _Branch protection rules_ once this PR is ready for merge https://github.com/wordpress-mobile/gutenberg-mobile/settings/branches

Stojdza:
Changes from this PR are merged to https://github.com/wordpress-mobile/gutenberg-mobile/pull/2137

PR #2107, opened 2020-04-03
Add a new androidReplacements function to comply with Android Typography lint rules

maxme:
Add a new androidReplacements function to comply with Android Typography lint rules

### DIFF
--- a/bin/po2android.js
+++ b/bin/po2android.js
@@ -34,6 +34,21 @@ function escapeResourceXML( unsafeXMLValue ) {
 }
 
 /**
+ * Android specifics replacements.
+ *
+ * @param {string} XMLValue input to apply replacements.
+ * @return {string} valid string passing Android linter rules.
+ */
+function androidReplacements( XMLValue ) {
+	return XMLValue.replace( /(-|\.\.\.)/gm, function( character ) {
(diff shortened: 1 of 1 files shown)


Tug:
Change looks good 👍 , anything else I should be looking at for the review?

maxme:
> anything else I should be looking at for the review?

This can be tested with current `gutenberg-mobile` develop, before the patch, the following commands will result in a lint error (see this https://github.com/wordpress-mobile/WordPress-Android/pull/11585#issuecomment-608390900), after the patch the same commands will work fine.

```
cd libs/gutenberg-mobile
yarn genstrings
cd ../..
python tools/merge_strings_xml.py
./gradlew :WordPress:lintVanillaRelease
```

Tug:
Don't you mean gutenberg-mobile `master`? On `develop`, `yarn genstrings` is failing because of a tsconfig issue in gutenberg.

maxme:
> yarn genstrings is failing because of a tsconfig issue in gutenberg.
Yes, sorry I forgot about that, I had to apply the patch from https://github.com/WordPress/gutenberg/pull/21324

Thanks for reviewing it!

PR #2141, opened 2020-04-09
Correct android font weight

lukewalczak:
Fixes the issue with no effect on Android when `fontWeight` value is numerical.
More about the issue you can read [here](https://github.com/facebook/react-native/issues/26193)

Going through the comments and potential solutions I've created a function for mapping a font to font-weight called `correctTextFontWeight` which has to be called from the application starting point (`_src/index.js_`)

before | after
--- | ---
Screenshot 2020-04-09 at 12 36 41 | Screenshot 2020-04-09 at 12 28 38

before | after
--- | ---
Screenshot 2020-04-09 at 13 01 29 | Screenshot 2020-04-09 at 12 29 08

To test:

1. Open WP-Android app
2. Open PAGE draft
3. Expect correct font weight (thicker - 600) on template buttons texts

To test: 

1. Open playground app on Android
2. Scroll to unsupported block
3. Expect correct font weight (thicker - 600) on template buttons texts

To test: 

1. Open playground app on iOS
2. Expect app is not crashing

NOTE: I would like to add some unit test if you like that approach.

PR submission checklist:


src/text-font-weight-correct.js
```diff
@@ -0,0 +1,75 @@
...
+const correctTextFontWeight = ( args ) => {
+	const baseText = textRender.call( this, args );
+	const { style } = baseText.props;
+
+	const flatStyle = Array.isArray( style ) ? merge( {}, ...style ) : style;
+	const shouldCorrectFontWeight = flatStyle && flatStyle.fontWeight;
+
+	return cloneElement( baseText, {
+		style: [ flatStyle, shouldCorrectFontWeight && getCorrectFontWeight( flatStyle.fontWeight ) ],
+	} );
```
dratwas, line 68: WDYT about something like this? Thanks to that we don't clone element if we don't need to
```suggestion
	return shouldCorrectFontWeight ? cloneElement( baseText, {
		style: [ flatStyle, getCorrectFontWeight( flatStyle.fontWeight ) ],
	} ) : baseText;
```

PR #2163, opened 2020-04-16
Add @-mention support

### LINKED ISSUE #331: Support @-mentions while editing
It would be nice if we could support @-mentions while editing, to add an explicit mention to a site user.

mchowning:
Fixes #331 by adding an `@`-button for inserting mentions for any `RichText`-based block.

Gutenberg PR: https://github.com/WordPress/gutenberg/pull/21651
WPAndroid PR: https://github.com/wordpress-mobile/WordPress-Android/pull/11693
WPiOS PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/13942

### Testing
1. Open any RichText-based block.
2. Tap the @ button on the far right of the format toolbar
3. Use the Add Mention UI to select a mention to use.

PR submission checklist:


android/app/src/main/java/com/gutenberg/MainApplication.java
```diff
@@ -128,6 +128,9 @@ public void editorDidEmitLog(String message, LogLevel logLevel) {
             @Override
             public void performRequest(String path, Consumer<String> onSuccess, Consumer<Bundle> onError) {}
 
+            @Override
+            public void onAddMention(Consumer<String> onSuccess) {}
```
SergioEstevao, line 132: Could we implement a "stub" version of this that always returns an user called `admin` for example?

mchowning: Done!

iamthomasbishop:
@mchowning @SergioEstevao Great work thus far! I've only had a chance to take Android for a spin (Sergio is helping to get me an iOS test build, so I'll test iOS asap). Based on that quick spin, here is some initial feedback. (I hope this is the right issue to drop feedback on — if not, I can move to another ticket!)

## Initial review notes

- There is sometimes a weird glitch where the UI sort of flashes momentarily when you're exiting the flow. To reproduce this, open the @-mention UI, don't type anything, tap the empty toolbar to hide the UI, then tap on another block
- We should add a separator before the @ button ([example](https://cloudup.com/cimyTmoJKTw))
- The @-mention toolbar seems a bit tall (54px?). It should match the height (44px) and styling of the block toolbar ([example](https://cloudup.com/cxFIVsqmsm8))
- [Android] We can probably get rid of the shadow on mentions UI
- Tapping delete when only `@` is shown in the input should exit the @-mention UI
- I expected typing `@` on the keyboard to trigger the mentions UI
- Similar to the last point, after applying a username, I expected to be able to remove some of the username characters to trigger the UI with that string as the active search. For example: if I type something, and then remove characters I expected it to start searching as I delete characters
- When the @-mention is applied, we might want to style it like a link, or at least make it visually distinct

## Bonus points

This is a bit out of scope, but it would certainly be nice-to-have!

- When the list is open all the way to the app bar and you're scrolled to the top of the list, it'd be nice for a swipe-down gesture to dismiss the UI

SergioEstevao:
### iOS-only

- [ ] Mention toolbar should be styled the same as the block toolbar (top border and same height)
- [ ] User list should be full-width (seems to be
- [ ] Tapping outside (above) the user list should exit the @-mention flow (perform same action as tapping on the (x) in the input)
- [ ] The animation when the UI appears is a bit odd (fades in and expands from the right). It could probably be a subtle fade, and if we apply a expand transition, it can expand upward.
- [ ] The expanded user list only shows 3 users at most, so it might be useful if a fourth item was hanging over the edge as to hint that you can scroll more. We could also make the list container a little taller, but this would be fine if its more obvious that you can scroll.

SergioEstevao:
> We should add a separator before the @ button (example)
ok
> The @-mention toolbar seems a bit tall (54px?). It should match the height (44px) and styling of the block toolbar (example)

ok

> I expected typing @ on the keyboard to trigger the mentions UI

While I understand this is good to have, it will involve a bit more of a technical challenge in terms that we will need to send back and forward from JS to native the keypresses of @

> Similar to the last point, after applying a username, I expected to be able to remove some of the username characters to trigger the UI with that string as the active search. For example: if I type something, and then remove characters I expected it to start searching as I delete characters

This is one is even more challenging because for this to work we will need to send all the  keypresses between JS and Native, and the performance will suffer because of this.
    
>When the @-mention is applied, we might want to style it like a link, or at least make it visually distinct

Do we have this on the web? If not it will need something to be coordinated with our web  colleagues.
   
>Can we add a trailing space to the username when one is applied to the block? I can't think of a scenario that I wouldn't want a space after the username.

Easy to do :)

iamthomasbishop:
@SergioEstevao Thanks for the updates!

> While I understand this is good to have, it will involve a bit more of a technical challenge in terms that we will need to send back and forward from JS to native the keypresses of @

Fair enough. It would definitely be nice to have, but if it's a lot of work, it could be part of the next iteration.

> This is one is even more challenging because for this to work we will need to send all the keypresses between JS and Native, and the performance will suffer because of this.

Same thing as the last one, although I think this is less important than the previous one.

> Do we have this on the web?

No, we don't have this on the web, but this is how it works on most other products and it feels odd that it's not styled as a link considering that's what the front-end output is.

mchowning:
Thanks so much @iamthomasbishop ! Since @SergioEstevao has already responded to the iOS and cross-platform issues (and I agree with everything he said), I'll address the Android-specific issues.

### Android-Only

> There is sometimes a weird glitch where the UI sort of flashes momentarily when you're exiting the flow. To reproduce this, open the @-mention UI, don't type anything, tap the empty toolbar to hide the UI, then tap on another block

I'm noticing that tapping on the @-mentions toolbar is dismissing the available suggestions while keeping the mentions "screen" open so it shows the "No matching users" text. That is certainly something that needs to be fixed, but it sounds like you are describing a different issue. Can you tell me a bit more about what you're seeing because I'm having trouble recreating it?

> We can probably get rid of the shadow on mentions UI

I want to do a bit more digging on this, but it's looking like this will be difficult to change. The suggestions are actually appearing in a pop-up window, which the system applies a shadow to. I'm just disguising the fact that it's a popup window by making it full-width. It feels to me like this is not a must-have if it turns out to be particularly difficult, do you agree?

> Tapping delete when only @ is shown in the input should exit the @-mention UI

That makes sense! 👍

mkevins:
Since it is in draft state, I bumped the milestone to `1.28`. Please let me know if you want this in the `1.27` release.

SergioEstevao:
> We should add a separator before the @ button

Done

iamthomasbishop:
@mchowning I forgot to come back to submit my replies to you on Friday, my bad 🤦 

> issue has been fixed

Awesome!

> shadow

I definitely wouldn’t consider it a blocker, but if we can’t remove the shadow altogether, is there a way we can bump up the z-index of the mention toolbar to be above it? In other words, if the suggestions UI is flush against the toolbar, and the toolbar has a higher z-index, it’d cover the shadow.

If that’s not possible, let’s make the UI not-quite full-width — maybe 8 or 16px margin on each side of the suggestions UI?

I’m going to play with an updated iOS build now and report back if I have any additional feedback.

iamthomasbishop:
@SergioEstevao I found a tiny thing on iOS, and I’m not sure where it’s coming from. The background color of the suggestions is different depending on the width of the screen/app. For example, on iPad full-screen, the background looks black, whereas on split-view iPad or iPhone, it looks like a secondary surface material (dark gray). Can we force all breakpoints to use the dark gray? Here are some screenshots:

**Full-screen iPad**
image

**Split-view iPad**
image

Not a blocker, and this might be better left as a refinement to the native part, but I also noticed the suggestions rows going full-width feels a little odd. I wonder if on wider screens we should either make the container of each suggestion narrower or perhaps stack the username and full name in the rows like Github for iOS does — like this:

image

mchowning:
As soon as jitpack decides to let the WPAndroid build succeed 😞this will be ready for another review @iamthomasbishop . Here is a summary of where things stand with respect to Android:

### Done
- Add a separator before @ button
- Update height of mention toolbar to 44px
- We can probably get rid of the shadow on mentions UI
- Tapping delete when only @ is shown in the input should exit the @-mention UI

### Not Done on Android

> Can we add a trailing space to the username when one is applied to the block? I can't think of a scenario that I wouldn't want a space after the username.

This is working on iOS. At this time I have not made this change on Android, however, because this behavior would be broken currently on Android. The technical issue is that if we think that there is an extra space that the native code might remove we reset the cursor's position. As a result, if the mention ends with a space, the cursor is moved to the beginning of the mention. It seems to me that having the cursor after the just-inserted mention without a space is much less disruptive to the writing flow than having the cursor at the beginning of the mention with a space at the end.

I am working on this, but addressing it will probably require touching code that was added to avoid a number of crashes on Android; I want to work on that change separately from this PR since it looks like it will be high risk and should be reviewed in isolation to make sure I don't introduce any regressions. Also, my thinking is that the lack of a space after the mention is probably not a blocking issue.

> There is sometimes a weird glitch where the UI sort of flashes momentarily when you're exiting the flow. To reproduce this, open the @-mention UI, don't type anything, tap the empty toolbar to hide the UI, then tap on another block

I'm not clear on what glitch you're seeing here, could you describe it a bit more or record a video?

### Future Improvements

These are things that will require a bit of work to implement (discussed [here](https://github.com/wordpress-mobile/gutenberg-mobile/pull/2163#issuecomment-618294496)), and I think we should keep track of them because they would be really good future improvements.

- I expected typing @ on the keyboard to trigger the mentions UI
- When the @-mention is applied, we might want to style it like a link, or at least make it visually distinct
- After applying a username, I expected to be able to remove some of the username characters to trigger the UI with that string as the active search. For example: if I type something, and then remove characters I expected it to start searching as I delete characters
- When the list is open all the way to the app bar and you're scrolled to the top of the list, it'd be nice for a swipe-down gesture to dismiss the UI

iamthomasbishop:
@mchowning Just had a chance to review again.

> It seems to me that having the cursor after the just-inserted mention without a space is much less disruptive

Agreed — this would be really nice to have, but not adding a space is indeed less disruptive. I don't consider it a pure blocker, but it's close because (imo) it feels like a bug in practice. Addressing in a separate PR is fine.

> I'm not clear on what glitch you're seeing here, could you describe it a bit more or record a video?

Sure thing, [here's a video](https://cloudup.com/c_A5SScpLse). Note: it's happening when I tapped the user from the suggestions list.

> Future Improvements

Sounds good 👍

mchowning:
Thanks for the video of the UI glitch @iamthomasbishop ! That wasn't happening on my test device, but I found an emulator that showed the same issue, and it should be resolved now. Would you mind taking a look at [the latest installable build](https://github.com/wordpress-mobile/WordPress-Android/pull/11693#issuecomment-615489470) with your device to confirm that it's fixed for you as well?

ios/gutenberg/GutenbergViewController.swift
```diff
@@ -194,6 +195,10 @@ extension GutenbergViewController: GutenbergBridgeDelegate {
     func gutenbergDidLogUserEvent(_ event: GutenbergUserEvent) {
         print("Gutenberg loged user event")
     }
+
+    func gutenbergDidRequestMention(callback: @escaping (Result<String, NSError>) -> Void) {
+        callback(.success("matt"))
```
cameronvoell, line 200: I'd be in favor of "sergio" here for balance, but I guess the name "matt" get's special treatment around here ;-P

react-native-gutenberg-bridge/android/src/main/java/org/wordpress/mobile/WPAndroidGlue/WPAndroidGlueCode.java
```diff
@@ -352,17 +360,43 @@ private ImagePipelineConfig getImagePipelineConfig(OkHttpClient client) {
...
-                             boolean isNewPost, String localeString, Bundle translations, int colorBackground, boolean isDarkMode) {
-        onCreateView(initContext, htmlModeEnabled, application, isDebug, buildGutenbergFromSource, "post", isNewPost
-        , localeString, translations, colorBackground, isDarkMode);
-    }
-
-    public void onCreateView(Context initContext, boolean htmlModeEnabled,
-                             Application application, boolean isDebug, boolean buildGutenbergFromSource,
-                             String postType, boolean isNewPost, String localeString, Bundle translations,
-                             int colorBackground, boolean isDarkMode) {
+    public void onCreateView(Context initContext,
```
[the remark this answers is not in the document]
mchowning: I had the exact same thought. I was _very_ tempted to just remove it, but ended up deciding not to lump that change in with these changes.

Issue #2168, opened 2020-04-20
Update existing templates using the new blocks


pinarol:
Let's revisit existing templates and update them using new blocks.

iamthomasbishop:
## New blocks to utilize
Here are the newly available blocks that we'll make use of:

* Columns
* Cover
* Latest Posts
* Buttons (shipping next)

## Layouts to update
With those new blocks in mind, here are the layouts/templates that will be updated to utilize these blocks:

* About
* Blog page
* Services

Here are some quick static mockups (Figma) that show roughly what the layouts will now look like (phone size mockup on left, portrait tablet on right), and some notes that explain what will need to change about for each.

### About layout

<a href="https://user-images.githubusercontent.com/1200432/79807057-76ea7580-832f-11ea-82ff-bcb121194738.png"><img src="https://user-images.githubusercontent.com/1200432/79807057-76ea7580-832f-11ea-82ff-bcb121194738.png" width="850"></a>

* Use the Columns block to place each quote in a separate Column

We might want to consider adding a fourth column here to balance things out a bit on tablet size, but that won't make or break the template.

### Blog page layout

<a href="https://user-images.githubusercontent.com/1200432/79807332-27f11000-8330-11ea-9eca-1e955c7cc8d7.png"><img src="https://user-images.githubusercontent.com/1200432/79807332-27f11000-8330-11ea-9eca-1e955c7cc8d7.png" width="876"></a>

* We can now utilize the Cover block. Or, if it's not quite ready for use we can use a static image (which I believe we're already doing as of the last release)
* We can remove the "Latest Posts" heading block, as it is a bit redundant

### Services layout

<a href="https://user-images.githubusercontent.com/1200432/79807619-19572880-8331-11ea-9631-8d8c056d2ce2.png"><img src="https://user-images.githubusercontent.com/1200432/79807619-19572880-8331-11ea-9631-8d8c056d2ce2.png" width="788"></a>

<a href="https://user-images.githubusercontent.com/1200432/79807671-38ee5100-8331-11ea-88d2-4b4ce0d133ea.png"><img src="https://user-images.githubusercontent.com/1200432/79807671-38ee5100-8331-11ea-88d2-4b4ce0d133ea.png" width="780"></a>

* Now can utilize Columns for the 3 sections of content in the middle part of the page
* We might want to add a fourth section to balance things a bit

## Other notes

[...]

PR #2179, opened 2020-04-24
Update existing templates to use new blocks

mkevins:
Fixes https://github.com/wordpress-mobile/gutenberg-mobile/issues/2168

#### Related PR

`gutenberg`: https://github.com/WordPress/gutenberg/pull/21857

### Description

This PR updates the mobile pages templates to use new blocks. It also adds some additional starter content.

**To test:**

1. Start a new page, then repeat the following steps for these templates:
   * About
   * Blog
   * Services
2. Tap page template in picker (on the bottom of the screen)
3. Apply the template
4. Scroll through to ensure the content is using the blocks [described here](https://github.com/wordpress-mobile/gutenberg-mobile/issues/2168#issuecomment-616857093)
5. Ensure additional content [listed here](https://github.com/wordpress-mobile/gutenberg-mobile/issues/2168#issuecomment-619654811) has been added

PR submission checklist:


mkevins:
Hi @iamthomasbishop :wave: :smile: 

Thanks for providing the [additional content described here](https://github.com/wordpress-mobile/gutenberg-mobile/issues/2168#issuecomment-619654811). I've added that content, and updated the templates with all the new blocks.

<details>
<summary>Here's how the About template looks in desktop preview after it is applied</summary>

spt-about-desktop
</details>

**Note**: the button block alignment is a known issue.

I noticed that we are using slightly different images for the second and third images within the services template (focus and strategy) compared to your mockup. I also noticed a vertical alignment issue (which is probably related, since the image heights differ). Lastly, I see a word wrap issue when we use all 4 columns, so maybe we can reduce that down? Wdyt?

<details>
<summary>Services template applied and previewed in desktop mode</summary>

spt-services-desktop
</details>

iamthomasbishop:
Thanks for the update @mkevins ! I had completely forgotten that on mobile we are only showing a max of 2 columns per row when in landscape orientation even though this results in 4 pretty-narrow columns on desktop. Perhaps we shouldn’t add a fourth column, after all. I think it does look a tad awkward on landscape phone and tablet, but perhaps we can adjust the Columns block to allow up to 3 columns per row (not as part of this work, but as a separate iteration on the Columns block). I think this would probably solve most of the concerns you had in your [latest comment](https://github.com/wordpress-mobile/gutenberg-mobile/pull/2179#issuecomment-622284296).

cc @pinarol @lukewalczak @dratwas — I know we had a lengthy discussion while working on the Columns/Column block about how many column blocks per row we allow, but I think 2 is proving to be a bit too constricting. Should we consider increasing the maximum to 3 columns per row?

> I noticed that we are using slightly different images for the second and third images within the services template (focus and strategy) compared to your mockup

I didn’t swap in updated images for the columns in my mockups, only chose new images for each of the new columns I added. So what already exists in code is what we should continue using. 👍

pinarol:
> cc @pinarol @lukewalczak @dratwas — I know we had a lengthy discussion while working on the Columns/Column block about how many column blocks per row we allow, but I think 2 is proving to be a bit too constricting. Should we consider increasing the maximum to 3 columns per row?

I don't have a strong preference actually, but if you think that would be better let's do that as part of [this issue.](https://github.com/wordpress-mobile/gutenberg-mobile/issues/2197). I expect it'd be a straight forward change. We can also consider making the readable content area bigger than 512 in some wide screen devices and fit there even more columns.

mkevins:
Hi @iamthomasbishop :wave: :smiley: 

I've reverted the addtion of the 4th column in the services template, so it is back to having 3 columns.

<details>
<summary>Services template with 3 columns - desktop preview</summary>

spt-services-desktop2
</details>

I've left the 4 columns with the quotes, since they didn't seem to have wrapping issues. Please let me know if you wanted that reverted as well, or if it's ok to leave it as is. In case of the latter, should we still add the `Needs Copy Review` label to the original issue for the quote?

pinarol:
Can we accelerate the review of this one please? @iamthomasbishop @chipsnyder Let's target merging this till EOD Fri (May 8th) to make it to the next release.

iamthomasbishop:
> Services template with 3 columns - desktop preview

@mkevins Ah, shoot. Just realize that first image is taller than the others — is there a way we can crop that image at the bottom to be the same height/width as the other images?

> I've left the 4 columns with the quotes, since they didn't seem to have wrapping issues. 

I would also move this back to 3 columns. It doesn’t look _terrible_, but wider columns might make for more readable text.

mkevins:
Hi @iamthomasbishop :wave: 

I'll remove that 4th quote then. Thanks for the guidance.

> is there a way we can crop that image at the bottom to be the same height/width as the other images?

We can use the width parameter with Pexels to limit the width of the image, but it seems the problem is coming from their different aspect ratios (the first image has a ratio of 4:3, and the other two have ratios ~ 3:2). If you can provide a similar Pexels image for the first one, with a 3:2 aspect, I can replace it.

iamthomasbishop:
@mkevins Here are some [updated images](https://cloudup.com/cYup77YRQeF), all cropped to 3:2 so we can ensure consistent sizing.

mkevins:
> @mkevins Here are some [updated images](https://cloudup.com/cYup77YRQeF), all cropped to 3:2 so we can ensure consistent sizing.

Thanks Thomas! I've updated the images and removed the 4th column in the about template, and everything is lining up nicely now :tada: .

Template|Landscape|Desktop Preview
-|-|-
About | spt-about-landscape | spt-about-desktop2
Services | spt-services-landscape | spt-services-desktop3

pinarol:
Thank you everyone! I went ahead and merged it when CI was green.

mkevins:
Thanks for testing and reviewing, and thanks Pinar for wrangling the merge! :tada:

PR #2206, opened 2020-05-01
Remove willTrimSpaces check on Android

mchowning:
Removes willTrimSpaces check on Android. See PR description and testing steps in https://github.com/WordPress/gutenberg/pull/22006.

PR submission checklist:

### DIFF
--- a/__device-tests__/gutenberg-editor-lists-space.test.js
+++ b/__device-tests__/gutenberg-editor-lists-space.test.js
@@ -1,0 +1,71 @@
+/**
+ * @format
+ * */
+
+/**
+ * Internal dependencies
+ */
+import EditorPage from './pages/editor-page';
+import {
+	setupDriver,
+	isLocalEnvironment,
+	stopDriver,
+	isAndroid,
+} from './helpers/utils';
+
+jasmine.DEFAULT_TIMEOUT_INTERVAL = 1000000;
+
+describe( 'Gutenberg Editor tests for List block (end)', () => {
+	let driver;
+	let editorPage;
+	let allPassed = true;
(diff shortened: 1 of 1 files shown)


mchowning:
I added an e2e test to cover this scenario. I'm a bit torn on whether that's a good idea or not. On one hand these out of bounds issues have been a recurrent issue on Android, but on the other hand it seems like a bit of an edge case and if we added e2e tests for every one of these our test suite would take forever to run (admittedly we're a long ways from having that problem). Interested to hear your thoughts.

mchowning:
:wave: @hypest ! Would appreciate it if you could take a look at this PR since I believe you have the most experience with the `willTrimSpaces` code, and I want to make sure removing it doesn't cause any regressions. 😬

hypest:
Had a look and [left a comment](https://github.com/WordPress/gutenberg/pull/22006#discussion_r420099656) Matt. I think that it's OK to remove the `willTrimSpaces` method if Aztec is not removing them, though some more testing to align make sure that indeed Aztec won't trim other white space characters is needed.

In the meantime, tried [the original issue](https://github.com/wordpress-mobile/gutenberg-mobile/issues/871) and still works OK on this PR.

SergioEstevao:
@mchowning do you still want this to go for 1.28?

mchowning:
> do you still want this to go for 1.28?

No, better to bump it.

mchowning:
Switching this to be a draft for the time being because I think I'll be submitting a larger change shortly that will include this.

mchowning:
Closing this because these changes were already handled in https://github.com/wordpress-mobile/gutenberg-mobile/pull/2279

PR #2211, opened 2020-05-01
Use Kotlin in react-native-gutenberg-bridge

### LINKED ISSUE #2203: Kotlin-fy the Android code for the bridge
Currently, the RN bridge and glue code on Android is written in Java. Nothing's wrong with that but it would be better to migrate to Kotlin for smoother navigation between the WPAndroid and gutenberg-mobile codebases.

mchowning:
Addresses #2203 

Adds Kotlin to react-native-gutenberg-bridge and converts the `Media` object to a Kotlin data class.

### Converted `Media` to a Kotlin Data Class

As a part of this I also converted that class to be immutable since it doesn't appear like we need it to be mutable. In my testing, this didn't cause any problems, and any problems _should_ be caught by the type system, but I wanted to point it out.

### Extra Media Object
We had two `Media` objects:

#### 1. [react-native-gutenberg-bridge/.../WPAndroidGlue/Media](https://github.com/wordpress-mobile/gutenberg-mobile/blob/f4a1a793c7199d16824afc9739d2a1e2b67498d8/react-native-gutenberg-bridge/android/src/main/java/org/wordpress/mobile/WPAndroidGlue/Media.java) 
converted to Kotlin in this PR
#### 2. [android/app/.../Media](https://github.com/wordpress-mobile/gutenberg-mobile/blob/f4a1a793c7199d16824afc9739d2a1e2b67498d8/android/app/src/main/java/com/gutenberg/Media.java)
This was added recently in #1851. I don't think we actually need it though. I removed it and am just relying on the `Media` class in `react-native-gutenberg-bridge` (which is now Kotlin).

### Demo App
Confirm that you can add both an Image and a Video block to the demo app with the "WordPress Media Library" option.

### WPAndroid
Confirm that you can add both Image and Video blocks using each of the source options:
1. From device
2. New Video/Image
3. WordPress Media Library
4. (Image block only) Free Photo Library

I don't expect there to be any issue, but we also need to confirm that Jitpack can build this with Kotlin without any problems. I have not been able to confirm this yet because of Jitpack-Github timeout issues. 😖

## PR submission checklist


react-native-gutenberg-bridge/android/src/main/java/org/wordpress/mobile/WPAndroidGlue/WPAndroidGlueCode.java
```diff
@@ -681,25 +681,24 @@ public void appendUploadMediaFiles(ArrayList<Media> mediaList) {
...
+            if (!TextUtils.isEmpty(media.getUrl()) && media.getId() > 0) {
                 // send signal to JS
-                appendNewMediaBlock(mediaId, mediaUri, mediaType);
+                appendNewMediaBlock(media.getId(), media.getUrl(), media.getType());
             }
         } else {
             // save the URL, we'll add it once Editor is mounted
             synchronized (WPAndroidGlueCode.this) {
-                mMediaToAddAfterMounting.put(mediaId, new Media(mediaId, mediaUri, mediaType, caption));
+                mMediaToAddAfterMounting.put(media.getId(), media);
```
mchowning, line 701: We shouldn't need to create a new object now that the `Media` is an immutable data class.

SergioEstevao:
@mchowning do you update this PR to get included in 1.29 or move it?

Issue #2169, opened 2020-04-20
Add a Tooltip for template selection buttons


pinarol:
We need to add a Tooltip to point to the template selection buttons.

iamthomasbishop:
@geriux Here are some notes on the design component and how it should behave in the context of the editor, as a starting point. Keep in mind that this is a first pass, so we will likely iterate to ensure we're aligning stylistically and behaviorally with tooltips elsewhere.

<a href="https://user-images.githubusercontent.com/1200432/80428126-c2ac9a00-88ae-11ea-8993-648b5fa3bffa.gif"><img src="https://user-images.githubusercontent.com/1200432/80428126-c2ac9a00-88ae-11ea-8993-648b5fa3bffa.gif" width="900"></a>

⬆️ [View the prototype on Figma](https://www.figma.com/proto/2E2IBuDA3eBa0G4dQXkWFg/Starter-Page-Templates-Mobile-MVP?node-id=640%3A30358&viewport=-805%2C269%2C0.8444179892539978&scaling=scale-down)

## Visibility

In the context of SPTs, the tooltip should only be shown once, just as a subtle hint to try out the feature via the layout selection buttons.

The tooltip should appear after a slight delay (details on transitions below), and should be dismissed when the user taps anywhere outside of the tooltip, _with the exception of the inline layout selection buttons_ — the tooltip should stay visible as the user scrolls this portion of the UI. Tapping on the canvas above, to its side, or on the keyboard should dismiss the tooltip.

## Style

You can find detailed style, measurements, etc. on the [design documentation](https://zeroheight.com/226850145/p/19a850-tooltip). Note: we might end up tweaking the background color and shadow styling during implementation, so the documentation and our implementation might not be final.

## Transitions

### Entry

Tooltip fades in and slides up ~4px.

* **Delay**: Tooltip starts entering `~500ms` after the editor initially loads
* **Opacity**: When it starts appearing, it enters from `0%` opacity to `100%`
* **Y-translation**: It starts `-4px` down on the Y axis, relative to its end position (There is `~8px` spacing between tooltip pointer and inline layout selection buttons — [example](https://cloudup.com/cDFc7Siu6yn))
* **Duration**: `~300ms`
* **Easing**: `easeOut`

### Exit

Tooltip fades out quickly in-place.

* **Delay**: No delay
* **Opacity**: `100%` opacity to `0%`
* **Y-translation**: none — the tooltip fades out in-place
[...]

PR #2216, opened 2020-05-05
Tooltip for template selection buttons

geriux:
Fixes https://github.com/wordpress-mobile/gutenberg-mobile/issues/2169

`Gutenberg PR` -> https://github.com/WordPress/gutenberg/pull/21974
`WordPress iOS PR` -> https://github.com/wordpress-mobile/WordPress-iOS/pull/14045
`WordPress Android PR` -> https://github.com/wordpress-mobile/WordPress-Android/pull/11807

**Note:** Bundles were generated to test the main apps builds, they will be removed before merging.

PR submission checklist:


geriux:
Hey @mkevins @chipsnyder 👋 

I've added you both to review these PRs 😃 I was thinking @chipsnyder could review the native iOS side and @mkevins Android's.

For Android, I can't make the build to test yet until we fix the issue with Jitpack, but the code can be reviewed in the meantime.

Thanks!

geriux:
Hey @iamthomasbishop 👋 

Here's the [iOS build](https://github.com/wordpress-mobile/WordPress-iOS/pull/14045#issuecomment-623357480) to test/review the Tooltip. Thanks!

iamthomasbishop:
@geriux I had a chance to test over the weekend and it worked as expected in terms of timing and animation. 

One thing that we discussed earlier is that the colors don't match perfectly with the newish native tooltip component on WPiOS and WPAndroid (which I think is using Color Studio colors), but that I think we should start using system-defined grays on that native component anyway. I just made [a proposal](https://github.com/wordpress-mobile/WordPress-iOS/issues/14136) for that, and in the process of discussing this, @bjtitus mentioned we may have access to the system-defined colors via a new API ([PlatformColor](https://reactnative.dev/docs/next/platformcolor)). If that is indeed the case, let's use a system gray and coordinate to decide which colors we should be using for each platform. In the meantime, we can stick with the color you've applied here.

Note: we'll also need to create a ticket and discuss/make the same decision on WPAndroid but if we do have access via that API, I suggest we utilize it 😄

react-native-gutenberg-bridge/index.js
```diff
@@ -136,4 +136,12 @@ export function logUserEvent( event, properties ) {
 	return RNReactNativeGutenbergBridge.logUserEvent( event, properties );
 }
 
+export function getStarterPageTemplatesTooltipShown( callback ) {
```
chipsnyder, line 139: As more of a question more than a suggestion: 

I see [one other spot](https://github.com/wordpress-mobile/gutenberg-mobile/blob/2ee4845b56201610b58a34065a58be160a4767b4/react-native-gutenberg-bridge/index.js#L116) where `get` is used to request information but mostly it seems like we're using the keyword `request` on most calls. 

Is there a pattern where the Gutenberg Bridge is using one over the other?

mkevins: Fwiw, I had a similar thought while reviewing, but forgot to mention it. Your comment reminded me :smile: . I generally like the semantics of "request" better than "get" when there is a callback involved, instead of the function directly returning the value. Functions over the bridge are an interesting case though, because we _must_ use a callback interface to retrieve data from the parent app, even when the data is guaranteed to be available, and the callback _must_ be invoked. Following this convention strictly, I suppose is a mild forfeit of some expressibility, in a sense. It seems that perhaps there is a pattern emerging where "get" signifies an "almost-sort-of" locality of reference, but I'm not sure whether that is intentional.

geriux: Thanks for the feedback! I agree `request` would make better sense and it'll consistent with the rest of the calls. I'll change it =)

iamthomasbishop:
@mkevins that’s a good question and a subtle one. The behavior on our side is a bit less aggressive and I think it’s okay. What we have here is feeling pretty natural in this context, so I propose that we keep the behavior as-is. This is a good thing to consider in future definitions — whether they’re persistent or passive (like ours is).

react-native-gutenberg-bridge/ios/GutenbergBridgeDelegate.swift
```diff
@@ -160,14 +160,20 @@ public protocol GutenbergBridgeDelegate: class {
...
     /// Tells the delegate that the editor requested a mention
     /// - Parameter callback: Completion handler to be called with an user mention or an error
     func gutenbergDidRequestMention(callback: @escaping (Swift.Result<String, NSError>) -> Void)
+
+    /// Tells the delegate that the editor requested to show the tooltip
+    func gutenbergDidRequestStarterPageTemplatesTooltipShown() -> Bool
+    
+    /// Tells the delegate that the editor requested to set the tooltip's visibility
+    /// - Parameter tooltipShown: Tooltip's visibility value    
+    func gutenbergDidRequestSetStarterPageTemplatesTooltipShown(_ tooltipShown: Bool)
```
etoledom, line 176: Hey @geriux @chipsnyder !
First congrats for merging this feature 🎉 

I just recently came across with these new bridge delegate methods, and I was wondering what was the intention. My first thought was that we were displaying UI on native side, but it seems like it's used only to do persistent storage of user actions (in this case, tooltips already shown to the user). In this case I was wondering if it makes sense to do this persistence internally in the bridge itself, and relieve the client from this responsibility that seems to be quite low level.

Also, it would be interesting to be able to get and set settings in general, in order to avoid new bridge messages for every different setting we might have in the future.

Furthermore, web already has a similar system in place that uses `LocalStorage` as persistence. It's used (between other things) to know if gutenberg should display NUX UI. Maybe we can reuse this system and create this generic bridge method to store a JSON payload in the same way that the browser's `LocalStorage` does.

This is clearly no hi-pri, but I've been thinking in ways to improve the bridge usage anyway.

Please let me know what do you think, or perhaps these ideas wouldn't help this feature in particular.

Thank you!

mkevins: Hi @etoledom :wave: :smiley:  As I understand, AsyncStorage was considered (to mirror the behavior of LocalStorage), however, it was decided that though that would simplify the implementation of persistent flags, it was more desirable to have all app-level persistence in one place.

While reviewing the Android side of this feature, I had a similar thought about how this pattern might "grow". But, since there aren't any other similar flags (yet) there was nothing to DRY up. My thinking is, we should keep an eye out if a pattern emerges where we can propose a more general solution to settings / persistent flag retrieval and reduce the "footprint" of the interface / protocol.

etoledom: Thanks @mkevins for the explanation!

What do you think about the first part? Maybe we still can avoid giving the responsibility to store the flag to the client. Could we do de persistence inside the bridge without the need for WP apps to know about it?

mkevins: > Maybe we still can avoid giving the responsibility to store the flag to the client. Could we do de persistence inside the bridge without the need for WP apps to know about it?

I think this was the case in an earlier implementation:
https://github.com/WordPress/gutenberg/blob/9aaf218b11481ebb4fe227426921181b21d252b4/packages/block-editor/src/components/page-template-picker/picker.native.js#L94

I'm not sure I have all the context on the decision to centralize the persistence into the parent apps. Perhaps it is to avoid "fragmenting" settings screens, but that is purely a guess on my part. This particular flag is not a user-toggleable setting, so it seems moot in this instance, but maybe this is more of a general architecture decision? I'm not sure there is a way to access the AsyncStorage data from the native side, so perhaps that's a concern as well? Maybe @hypest or @geriux  can shed some more light on this.

I found [this](https://github.com/sriraman/react-native-shared-preferences) which looks interesting, but I'm not sure if there is an equivalent iOS module.

geriux: Hey @etoledom 👋 thanks for bringing this up and also thanks @mkevins for joining the conversation.

As @mkevins mentioned when I started this feature my intention was to use AsyncStorage. Since it didn't require much other than adding the library to the main apps (because React Native removed it in favor of lean core).

In the process of that, there were some discussions about maybe avoid using AsyncStorage and use what we currently use for some flags instead. Mainly to store data in the same place, even if it's a flag. So I ended up deciding to do it this way.

For this case though, as @mkevins mentions, it is not a user-togglable setting so having it stored somewhere else would work just fine.

> Also, it would be interesting to be able to get and set settings in general, in order to avoid new bridge messages for every different setting we might have in the future.

This could be an option to avoid adding extra code for these types of settings.

> Furthermore, web already has a similar system in place that uses LocalStorage as persistence. It's used (between other things) to know if gutenberg should display NUX UI. Maybe we can reuse this system and create this generic bridge method to store a JSON payload in the same way that the browser's LocalStorage does.

I guess this could also work, in my experience, AsyncStorage works just fine. Not sure how complicated it would be to develop what you mention but I'd lean more for adding AsyncStorage if we decide to go for this kind of system.

etoledom: Thank you both for your explanations! Everything seems more clear now 🙂 

Let's wait for a good moment to re-think this, probably when we need to store a new similar flag.

PR #2226, opened 2020-05-06
Issue/update aztec ios to 1.19.0

SergioEstevao:
Updates Aztec to version 1.19.0

To test:
 - Smoke test the demo app, especially the text blocks
 - Check the initial content block that has the sup and sub and see if it renders correctly.

PR submission checklist:


SergioEstevao:
> The other is something I didn't noticed in the Aztec PR, and I'm not sure if it's expected or not:
When I try to modify text before the subscript (or superscript) that text will be add as part of the subscript.

That is expected it should work the same way as bold, italic, strikethrough.

Issue #2182, opened 2020-04-24
JitPack build fails with "Git clone took longer than 180 s"


hypest:
It has been reported (in Slack) a few times already that the JitPack build randomly fails with the following error message (in JitPack's log):

```
Git error. Retrying.
Git error. Git clone took longer than 180 s
```

**To Reproduce**
No steps to reproduce, but it looks like it is too frequent at this point. An example is [this WPAndroid PR](https://github.com/wordpress-mobile/WordPress-Android/pull/11594) on which we've failed to make the JitPack build dependency succeed.

hypest:
Looks like the issue is still happening, this time on 30d5f0a, part of the 1.27.0 release cut.

mchowning:
Closing this since we no longer rely on Jitpack to build the entire gutenberg-mobile library. We do still use it to build some of our React Native dependencies, but those are substantially smaller and have not encountered this "Git clone took longer than 180 s" error.

PR #2231, opened 2020-05-06
Enable using the JitPack mode of building from a parent project

hypest:
Addresses #2182

We currently experience high failure rate on JitPack building the gutenberg-mobile artifact for WPAndroid, so this PR tries to enable building the artifact and the JS bundle locally, by reusing the gradle-powered Node/RN build.

This mode is then used by WPAndroid. Related PR: https://github.com/wordpress-mobile/WordPress-Android/pull/11833

To test:
1. CI should be green
2. Integration in the WPAndroid PR should be OK and green too. Nothing special to test out on this side.

PR submission checklist:


mchowning:
Do we still want jitpack code [in `react-native-gutenberg-bridge`](https://github.com/wordpress-mobile/gutenberg-mobile/blob/develop/react-native-gutenberg-bridge/android/build.gradle#L130-L150)?

hypest:
> Do we still want jitpack code in react-native-gutenberg-bridge?

Actually, yes. This PR only removes the gutenberg-mobile bridge library building on Jitpack, but we still have the npm package forks be built on JitPack. Those repos are smaller and don't seem to pose an issue yet.

hypest:
The WPAndroid PR will target a release so, this PR shall also target `master`. Will change it to that. Since this PR is introducing integration changes but no functional changes, I'm not planning to bump the library version.

hypest:
cc @geriux since you're currently wrangling a hotfix release and this PR now targets master. Most probably, the hotfix will need to update after this lands on time.

geriux:
> cc @geriux since you're currently wrangling a hotfix release and this PR now targets master. Most probably, the hotfix will need to update after this lands on time.

Cool! thanks for letting me now 👍

hypest:
There's only one CI job pending on the WPAndroid PR ([on this commit](https://github.com/wordpress-mobile/WordPress-Android/pull/11833/commits/bb1e0a022422b56a161d253ce2d375e5da800d41)), which seems to be waiting for Firebase to queue the tests run, for more than 25mins now. I will start the merge domino since everything else is green at this point. Besides, that job will get re-triggered anyway.

Issue #1744, opened 2020-01-06
Use the REST API endpoint to resolve theme colors


pinarol:
Depends on the issue: https://core.trac.wordpress.org/ticket/48798

Use the recently extended REST API endpoint added [here](https://core.trac.wordpress.org/ticket/48798) to resolve color settings

pinarol:
**Some History**

Currently on mobile side we are not able to render theme colors in the block editor. For example, if you:
- Create a Button block in web 
- Set a theme color from color palette 
- View the same Button block on mobile
You'll see a Gray color. That's because we don't know how to resolve the theme color into real color value, so we are using a fallback color for theme colors.
Same thing is true for Cover block.

With the help of the updated `/themes` endpoint, we'll be able to get a color name-value, gradient name-value mappings.

**`/themes` Endpoint**

https://core.trac.wordpress.org/ticket/48798 was closed in favor of https://core.trac.wordpress.org/ticket/49037 . 49037 extends `/themes` endpoint to return lots of theme information including color and gradient values used in that theme.

The endpoint should be available as of WordPress 5.4 RC, but it is not released yet.

There are some instructions here about how to test it: https://wordpress.org/news/2020/03/wordpress-5-4-release-candidate/

**WPiOS/WPAndroid**

We need to:

- Add service calls to the `/themes` endpoint in parent apps
- Bootstrap the block editor with the new theme colors

I believe we can find out how other types of site data is stored and store theme colors in a similar way.

**RN side**

As far as I understand we need to change "withColors" in a way that also resolves the theme colors.

chipsnyder:
Android side of this has merged now it'll be in 15.2 🚀

PR #2241, opened 2020-05-08
Support Theme Colors and Gradients

chipsnyder:
**Fixes:** https://github.com/wordpress-mobile/gutenberg-mobile/issues/1744

**Related PRs**
`gutenberg` https://github.com/WordPress/gutenberg/pull/22197
`WordPress-Android` https://github.com/wordpress-mobile/WordPress-Android/pull/12041
`WordPress-FluxC-Android` https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1593
`WordPress-iOS` https://github.com/wordpress-mobile/WordPress-iOS/pull/14085

----

## Description

Adds support for theme defined colors and gradients to be injected into the editor from the mobile apps. 

Colors can then be displayed by the same mechanisms that utilize the settings lookup for colors and gradients. This is currently supported on Cover and Button blocks

## How has this been tested?

***Note*** Atomic sites don't seem to support the `wp/v2/themes` API changes yet so this can be tested with a free site or a self-hosted site

To help with testing this theme can be added to a self-hosted site: [twentytwenty-copy.zip](https://github.com/wordpress-mobile/WordPress-iOS/files/4657974/twentytwenty-copy.zip)
### Colors
1.) Select a theme with custom colors (such as TwentyTwenty)
2.) Create a post with blocks that have a custom color set (such as Cover or Button)
3.) Load editor on mobile 

**Expect** to see the custom color on the block

### Gradients
1.) Select a theme with custom gradients or [add gradients to a theme](https://github.com/WordPress/gutenberg/blob/12a6ee49c16dd6f327a8a141b6858197536da800/docs/designers-developers/developers/themes/theme-support.md#block-gradient-presets)
2.) Create a post with blocks that have a custom gradient set (such as Cover or Button)
3.) Load editor on mobile 

**Expect** to see the custom gradient on the block

### No Wifi
1.) Set up blocks for custom color and gradient
2.) View the post on mobile to cache the theme
3.) Turn off wifi to the device
4.) Reload the editor

**Expect** to see the custom colors on the blocks

### No Wifi - Force kill the app
1.) Set up blocks for custom color and gradient
2.) View the post on mobile to cache the theme
3.) Turn off wifi to the device
4.) Stop the app from running in the background
5.) Reload the editor

**Expect** to see the custom colors on the blocks

## Screenshots 
<kbd></kbd>

### PR submission checklist:


chipsnyder:
Hey @dratwas 

I made some changes to update the settings with colors and gradients loaded from themes. When you get time will you take a look and see if this will conflict with anything you're working on for the color picker?

ios/Podfile.lock
```diff
@@ -195,10 +195,10 @@ PODS:
     - React
   - react-native-slider (2.0.7):
     - React
-  - react-native-video (5.0.2):
+  - react-native-video (5.0.1):
```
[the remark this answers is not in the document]
chipsnyder: Good catch. I'm probably behind on develop a bit. I'll update and rerun pod install

src/index.js
```diff
@@ -88,6 +90,8 @@ export class RootComponent extends React.Component {
 				initialHtmlModeEnabled={ initialHtmlModeEnabled }
 				initialTitle={ initialTitle }
 				postType={ postType }
+				colors={ colors }
```
dratwas, line 93: Hmmm, it's not important but I'm wondering if we need to create a consts for `colors` and `gradients`. We could use directly `this.props.colors/gradients`. Or, maybe we could do something like:

```js
const {
  initialHtmlModeEnabled,
  initialTitle = 'Welcome to Gutenberg!',
  postType = 'post',
  colors,
  gradients,
} = this.props;

let initialData = this.props.initialData;
```

chipsnyder: I really only went with the const originally because I wasn't sure if I'd need to default them to something. You're right though it would be cleaner if I just directly used `this.props.colors/gradients`

dratwas: Make sense :) Thanks!

dratwas:
It looks good, how about Android? Will you implement it as a part of this PR? :)

chipsnyder:
> How about Android?

@pinarol and I were thinking it would be best to get iOS and the Gutenberg side merged first to get feedback then do a different PR for Android (currently in progress).

pinarol:
Let's just make sure we are not breaking Android side by merging this.

chipsnyder:
> Let's just make sure we are not breaking Android side by merging this.

Agreed. I had success with this via the demo app and via the main app running metro. However, I would appreciate an extra set of eyes to confirm. I can make a PR build for Android if it'll help.

chipsnyder:
>> How about Android?
>
> @pinarol and I were thinking it would be best to get iOS and the Gutenberg side merged first to get feedback then do a different PR for Android (currently in progress). 

I should have the Android side of this done today or tomorrow, so I'll have an Android build available for testing then.

chipsnyder:
> I should have the Android side of this done today or tomorrow, so I'll have an Android build available for testing then.

I was able to get the Android side up and included it in this PR since it kind of went stale. @dratwas Would you be able to re-review with those changes? 

The [iOS side](https://github.com/wordpress-mobile/WordPress-iOS/pull/14085) was reviewed and tested by @SergioEstevao and I reached out to @mchowning to request his review on the [Android side](https://github.com/wordpress-mobile/WordPress-Android/pull/12041)

react-native-gutenberg-bridge/android/src/main/java/org/wordpress/mobile/ReactNativeGutenbergBridge/GutenbergBridgeJS2Parent.java
```diff
@@ -1,5 +1,10 @@
 package org.wordpress.mobile.ReactNativeGutenbergBridge;
 
+import android.os.Bundle;
+import android.os.Parcelable;
+
+import androidx.annotation.Nullable;
+
```
dratwas, line 7: Hmmm, do we need that?

react-native-gutenberg-bridge/android/src/main/java/org/wordpress/mobile/ReactNativeGutenbergBridge/RNReactNativeGutenbergBridgeModule.java
```diff
@@ -1,5 +1,8 @@
 package org.wordpress.mobile.ReactNativeGutenbergBridge;
 
+import android.os.Bundle;
+import android.os.Parcelable;
```
dratwas, line 4: Not sure if we need import of `Parcelable` here

react-native-gutenberg-bridge/android/src/main/java/org/wordpress/mobile/WPAndroidGlue/WPAndroidGlueCode.java
```diff
@@ -430,12 +436,23 @@ public void onCreateView(Context initContext,
...
+        Serializable colors = editorTheme != null ? editorTheme.getSerializable(PROP_NAME_COLORS) : null;
+        if (colors != null) {
+            initialProps.putSerializable(PROP_NAME_COLORS, colors);
+        }
+
+        Serializable gradients = editorTheme != null ? editorTheme.getSerializable(PROP_NAME_GRADIENTS) : null;
+        if (gradients != null) {
+            initialProps.putSerializable(PROP_NAME_GRADIENTS, gradients);
+        }
+
```
dratwas, line 448: I'm wondering if we would have more props that are from the theme. If yes i think it would be worth to implement a method like `addThemeProp(initialProps, propName)` to not duplicate the code since it looks pretty similar

react-native-gutenberg-bridge/android/src/main/java/org/wordpress/mobile/ReactNativeGutenbergBridge/RNReactNativeGutenbergBridgeModule.java
chipsnyder: You're right. I missed this one too once I changed a signature below

dratwas:
I'm not sure if the update of bundles should be a part of that PR but the code looks good!

Awesome work @chipsnyder 🎉

react-native-gutenberg-bridge/android/src/main/java/org/wordpress/mobile/WPAndroidGlue/WPAndroidGlueCode.java
chipsnyder: That's a good idea. I guess that would partially depend on the other items that are added. For now, though this seems to be the limit of added fields with ongoing discussions on if we'll theme the background or not. I think if we add more here we might want to pursue that idea more.

src/index.js
```diff
@@ -85,10 +83,12 @@ export class RootComponent extends React.Component {
 		return (
 			<Editor
 				initialHtml={ initialData }
-				initialHtmlModeEnabled={ initialHtmlModeEnabled }
+				initialHtmlModeEnabled={ this.props.initialHtmlModeEnabled }
```
mkevins, line 86: Not suggesting a change here, just curious if there is a rationale for why `initialHtmlModeEnabled` was removed from a destructuring while the other three props were added to it?

chipsnyder: I think this came about as just a resolution to a merge conflict. For the other const declarations it was suggested to use this.props.whatever which I had converted to on a previous merge then I think just ran into issues when resolving it later

PR #2264, opened 2020-05-14
Improve avoid-jitpack workflow

mchowning:
Trying to improve the avoid-jitpack workflow, particularly the build speeds.

See [WPAndroid PR](https://github.com/wordpress-mobile/WordPress-Android/pull/11921) for testing instructions

PR submission checklist:

### DIFF
--- a/react-native-gutenberg-bridge/android/build.gradle
+++ b/react-native-gutenberg-bridge/android/build.gradle
@@ -174,9 +174,43 @@ if (buildGutenbergMobileJSBundle) {
     task bundleUpToDateCheck {
         description("Checks if the inputs to the javascript bundle and the bundle itself are unchanged. \
 Only if they are is the isBundleUpToDate flag set to true. That flag is used by other tasks.")
-        inputs.files("$jsRootDir/package.json",
-                     "$jsRootDir/yarn.lock")
-        inputs.dir("$jsRootDir/patches")
+
+        def isRelevantFile = { it.name.endsWithAny('.js', '.css') || it.name == 'package.json' }
+        def inputFiles = {
+            def jsFiles = []
+            def dirs = []
+            file(jsRootDir).eachDir { dir ->
+                if (dir.name != 'node_modules'
+                        && dir.name != 'bundle'
+                        && dir.name != 'gutenberg'
+                        && !dir.name.startsWith('symlinked')) {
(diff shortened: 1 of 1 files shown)


mchowning:
The remaining work on this is to figure out a way to include all of the relevant js, css, scss files (any others?) in the task's input so that if any of them change the bundle is regenerated. This worries me because it seems very fragile and when it fails it will not be obvious that the reuse of a previously generated bundle is the problem. 🤔

mchowning:
> The remaining work on this is to figure out a way to include all of the relevant js, css, scss files (any others?) in the task's input so that if any of them change the bundle is regenerated. This worries me because it seems very fragile...

Pushed an update that I _think_ addresses that. But its a bit hacky and I'm still concerned about it being fragile. Remaining things I think should be addressed:
- [ ] Running the `clean` gradle task from WPAndroid should regenerate the bundle. This seems really important to me because it makes it relatively trivial to recover if a bundle gets out of date for some reason. Currently a clean does not cause the bundle to regenerate.
- [ ] Running a WPAndroid build should not make the gutenberg-mobile submodule "dirty". Currently running a WPAndroid build that generates a js bundle modifies ~30 files in the `libs/gutenberg-mobile/bundle/` directory.

mchowning:
In trying to make clean rerun the task I've realized that for some reason my task is not actually checking the output the way I thought it was. I can delete the bundle and the `bundleUpToDateCheck` task still shows as UP-TO-DATE and builds an app that crashes when the editor is loaded. 😬

mchowning:
Just added fc9fbdb, which moves the bundle to the build folder and also has the `bundleUpToDateCheck` task add an indicator file next to the bundle. I registered that indicator file as an output of the `bundleUpToDateCheck` task, and it seems to detect that it needs to rerun when that is deleted (still don't understand why it doesn't do the same with the actual bundle file). This means that if you run the `clean` task, the bundle and the indicator file will both be deleted, and the next build will properly detect that it needs to regenerate the bundle.

cameronvoell:
> Tested via [wordpress-mobile/WordPress-Android#11921](https://github.com/wordpress-mobile/WordPress-Android/pull/11921). Looks good! :shipit:

Did another round of checks after retargeting and rebase with release branch. Looks good!

PR #2266, opened 2020-05-15
Display radial gradient in block which supporting gradients

lukewalczak:
Fixes missing possibility to display radial-gradients in Buttons or Cover blocks.

Ref to gutenberg: https://github.com/WordPress/gutenberg/pull/22384

To test:

1. Open draft on web
2. Add `Buttons` or `Cover`
3. Set radial gradient background color
4. Save draft
5. Open post on mobile
6. Expect to see radial gradient

PR submission checklist:


lukewalczak:
I think that we can mention in release notes that displaying radial gradient is supported in Buttons and Cover blocks. Wdyt @pinarol?

pinarol:
I think we should mention it but maybe with a single star [*] since this is not the whole feature yet.

PR #2276, opened 2020-05-20
Fix NPE which is caused when getLayout() is null in ReactAztecText

marecar3:
Fixes #: https://github.com/wordpress-mobile/WordPress-Android/issues/11821

To test: Gutenberg Editor should work as expected

PR submission checklist:


react-native-aztec/android/src/main/java/org/wordpress/mobile/ReactNativeAztec/ReactAztecText.java
```diff
@@ -200,6 +201,13 @@ public void requestFocusFromJS() {
             @Override
             public void run() {
                 // let's pinpoint the caret line to ask the system to bring that line into the viewport
+                // we need to make sure that getLayout() isn't null
+                // because it will cause the NPE: https://github.com/wordpress-mobile/WordPress-Android/issues/11821
+                if (getLayout() == null) {
+                    AppLog.e(AppLog.T.EDITOR, "Layout is null");
```
hypest, line 207: Since we bail on purpose, how about logging a warning instead of an error?

Also, I think the `Layout is null` offers too little context to anyone reading the app logs so, what do you think about using something like "Layout is null while trying to scroll inside a ReactAztecText block. Canceling the scroll."?

marecar3: > Since we bail on purpose, how about logging a warning instead of an error?
> 
> Also, I think the `Layout is null` offers too little context to anyone reading the app logs so, what do you think about using something like "Layout is null while trying to scroll inside a ReactAztecText block. Canceling the scroll."?

Thanks for the good feedback!
The log message should be better now.

PR #2299, opened 2020-05-21
Insure bundle file exists when determining up-to-date status

mchowning:
This improves the incremental build detection following the no-jitpack changes. 

The primary motivation for this change is that the previous implementation would incorrectly find that a bundle was up-to-date if the bundle file was not present but the `upToDateIndicatorFile` was present. This could occur if there was a build failure in the `buildJSBundle` task after the `bundleUpToDateCheck` task had already created the `upToDateIndicatorFile`. It could also occur if someone deleted only the bundle file while making no other changes (see test step 9). In those scenarios, gradle would think that the bundle did not need to be rebuilt (because the `upToDateIndicatorFile` existed) even though there was no bundle file.

With these changes we no longer need the `upToDateIndicatorFile` hack. Instead, we insure that we only set the `isBundleUpToDate` flag to `true` if a bundle file exists. We do not need to check that it is the correct bundle file because the `isBundleUpToDate` flag will be flipped back to false if any of the inputs to the bundle have changed. The only way I can see that this could break would be if the bundle file were manually replaced with a different file without changing any of the bundle file's many inputs (basically all the js and css source). Even this scenario can be recovered from by running the `clean` task though.

I think this change is also an improvement because it no longer registers the bundle as an output to the `upToDateCheck` task. Any bundle file that existed at the end of that task would have been the result of a _previous_ build so registering its from-the-previous-build state as an output before the currently-in-progress build was complete would have been misleading if the current build actually generated a new bundle file. With that said, I think the worst problem that could have caused us would be to sometimes unnecessarily regenerate a bundle when it didn't need to.

To test:

1. Use WPAndroid's `develop` branch with this branch checked out
2. Insure that `wp.BUILD_GUTENBERG_FROM_SOURCE` is missing from `gradle.properties` or set to `false`.
3. Run `./gradlew clean installVanillaDebug`
4. Observe that the build takes minutes to complete and the JS bundle is created (`libs/gutenberg-mobile/react-native-gutenberg-bridge/android/build/assets/index.android.bundle`).
5. Run the resulting apk and insure you can open the block editor _without_ the metro server.
6. Reinstall the app by re-running `./gradlew installVanillaDebug` (note we are _not_ running `clean` anymore)
7. Observe that the build is much faster (10-20 seconds)
8. Run the resulting apk and insure you can open the block editor _without_ the metro server.
9. Manually delete the bundle file `libs/gutenberg-mobile/react-native-gutenberg-bridge/android/build/assets/index.android.bundle`
10. Run a third build: `./gradlew installVanillaDebug`
11. Observe that the build takes minutes to complete and the JS bundle is created (`libs/gutenberg-mobile/react-native-gutenberg-bridge/android/build/assets/index.android.bundle`).
12. Run the resulting apk and insure you can open the block editor _without_ the metro server.
13. Run `./gradlew clean`
14. Verify that the JS bundle was deleted (`libs/gutenberg-mobile/react-native-gutenberg-bridge/android/build/assets/index.android.bundle`)
15. Run a fourth build: `./gradlew installVanillaDebug`
16. Run the resulting apk and insure you can open the block editor _without_ the metro server.

PR submission checklist:

### DIFF
--- a/react-native-gutenberg-bridge/android/build.gradle
+++ b/react-native-gutenberg-bridge/android/build.gradle
@@ -180,7 +180,7 @@ if (buildGutenbergMobileJSBundle) {
 
     task bundleUpToDateCheck {
         description("Checks if the inputs to the javascript bundle and the bundle itself are unchanged. \
-Only if they are is the isBundleUpToDate flag set to true. That flag is used by other tasks.")
+If they are changed, the isBundleUpToDate flag is switched to false. That flag is used by other tasks.")
 
         def isRelevantFile = { it.name.endsWithAny('.js', '.css') || it.name == 'package.json' }
         def inputFiles = {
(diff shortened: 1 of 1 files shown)


react-native-gutenberg-bridge/android/build.gradle
```diff
@@ -218,24 +218,18 @@ Only if they are is the isBundleUpToDate flag set to true. That flag is used by
         }
         inputs.files(inputFiles())
 
-        outputs.file("$buildAssetsFolder/$bundleName")
+        // We cannot use the bundle file itself as an output because it does not yet exist when this
+        // task finishes. Nevertheless, we have to declare something as an output because only tasks
+        // with outputs are run incrementally, so we're declaring a file that does not exist as the
+        // output. Since that file never exists, that "output" will always be considered "up-to-date".
```
hypest, line 224: > Since that file never exists, that "output" will always be considered "up-to-date"

Heh, I find that counter-intuitive and probably shows my (low) gradle understanding so I wonder, do you have a link somewhere that explains this behavior?

hypest:
By the way, took the liberty to update the PR description to amend the JS bundle path. I think it was missing the `assets/` subfolder inside the `build` folder. Hope the new path is the accurate one.

hypest:
WPAndroid opened here: https://github.com/wordpress-mobile/WordPress-Android/pull/11994

react-native-gutenberg-bridge/android/build.gradle
mchowning: This was counter-intuitive to me too. This is [the forum post](https://discuss.gradle.org/t/unexpected-up-to-date-task-while-output-file-is-missing/5216/2) that clued me into the behavior:

> For many tasks the outputs are not explicit, so for each task that runs, Gradle records the outputs files in the provided dirs or file paths to scan. If none exist, and the tasks was successful, then the set of outputs for the task is null.

mchowning:
When building the 14.9 release build (which did not have the code from this PR), @oguzkocer ran into an issue where the bundle file was not being generated even though the entire `libs/gutenberg-mobile/react-native-gutenberg-bridge/android/build/assets/` directory was missing. The new [manual check for the bundle file included in this PR](https://github.com/wordpress-mobile/gutenberg-mobile/pull/2299/files#diff-1f6a0664aee669119ee1ded189c57259R230) seems like a pretty foolproof way to insure that cannot happen anymore, but I wanted to leave a note about it here since I'm not really sure how that happened even with the old code. Related discussion: p1590765206023000-slack-mobile-gutenberg

PR #2305, opened 2020-05-22
Delete temporary RN tools before building

### LINKED ISSUE #2302: [Android] Build issue with node: Couldn't follow symbolic link
Under circumstances that are not clear, the WPAndroid build fails with this error:

```
FAILURE: Build failed with an exception.
* What went wrong:
Execution failed for task ':react-native-gutenberg-bridge:nodeSetup'.
> Could not list contents of '/var/folders/d3/7bzlyz453w11k42t26xtbkmc0000gn/T/jsbundle/null/nodejs/node-v12.16.1-darwin-x64/bin/npm'. Couldn't follow symbolic link.
```

A simliar issue can happen with Yarn:

```
Could not list contents of ‘/var/folders/py/vdkc920j69b255sld66fbxt40000gp/T/jsbundle/null/yarn/yarn-v1.22.4/bin/yarnpkg’. Couldn’t follow symbolic link.
```

Looking at the symlinks in question, they point to files that are no longer there. We don't know yet why those files get missing and the tools setup becomes inconsistent.

hypest:
Fixes #2302

Under circumstances that we don't quite understand yet, the gradle-node-plugin managed Node/npm/Yarn tools setup ends up in an inconsistent state and break the WPAndroid build. This PR tries to reset the state of the tools by deleting their folders (living in a temporary system directory) and let gradle-node-plugin re-setup them before building.

To test:

1. After a successful WPAndroid build, delete the `libs/gutenberg-mobile/react-native-gutenberg-bridge/android/build/assets/index.android.bundle` JS bundle and also delete the node folder `$TMPDIR/jsbundle/null/nodejs/node-v12.16.1-darwin-x64/lib/node_modules/npm`.
2. Run `./gradlew assembleVanillaDebug`. It should succeed.
3. Install the apk (or use `installVanillaDebug` in Step 2) and run the app. The block editor should work as normal.

PR submission checklist:

### DIFF
--- a/react-native-gutenberg-bridge/android/build.gradle
+++ b/react-native-gutenberg-bridge/android/build.gradle
@@ -33,6 +33,11 @@ def buildGutenbergMobileJSBundle =
         ? false : (rootProject.ext.has("buildGutenbergMobileJSBundle")
                 && rootProject.ext.buildGutenbergMobileJSBundle)
 
+def tempFolder = "${System.getenv('TMPDIR')}/jsbundle/${System.getenv('VERSION')}"
+def nodeFolder = file("${tempFolder}/nodejs")
+def npmFolder = file("${tempFolder}/npm")
+def yarnFolder = file("${tempFolder}/yarn")
+
 if (buildGutenbergMobileJSBundle) {
     println 'Building the Gutenberg Mobile JS bundle'
 
@@ -55,16 +60,14 @@ if (buildGutenbergMobileJSBundle) {
         // If false, it will try to use globally installed node.
         download = true
 
-        def tmpdir = "${System.getenv('TMPDIR')}/jsbundle/${System.getenv('VERSION')}"
-
         // Set the work directory for unpacking node
-        workDir = file("${tmpdir}/nodejs")
+        workDir = file(nodeFolder)
 
         // Set the work directory for NPM
-        npmWorkDir = file("${tmpdir}/npm")
+        npmWorkDir = file(npmFolder)
 
         // Set the work directory for Yarn
-        yarnWorkDir = file("${tmpdir}/yarn")
+        yarnWorkDir = file(yarnFolder)
 
         // Set the work directory where node_modules should be located
         nodeModulesDir = file("${project.projectDir}/../../")
@@ -290,10 +293,16 @@ If they are changed, the isBundleUpToDate flag is switched to false. That flag i
         delete '../../node_modules'
     }
 
+    task resetExtractedRNTools(type: Delete) {
+        println "Deleting temporary folders with extracted RN tools"
+        delete nodeFolder, npmFolder, yarnFolder
+    }
+
     preBuild.dependsOn(cleanupNodeModulesFolder)
     cleanupNodeModulesFolder.dependsOn(backupHermesDebugAAR)
     backupHermesDebugAAR.dependsOn(backupHermesReleaseAAR)
     backupHermesReleaseAAR.dependsOn(copyJSBundle)
     copyJSBundle.dependsOn(buildJSBundle)
     buildJSBundle.dependsOn(yarn_install)
+    nodeSetup.dependsOn(resetExtractedRNTools)
 }


hypest:
I see the following error happening quite often on my end, and not sure if it's a networking issue or not:

```
error /Users/hypest/proj/a8c/android/tmp/t5/WordPress-Android/libs/gutenberg-mobile/node_modules/appium-chromedriver: Command failed.
Exit code: 1
Command: node install-npm.js
Arguments:
Directory: /Users/hypest/proj/a8c/android/tmp/t5/WordPress-Android/libs/gutenberg-mobile/node_modules/appium-chromedriver
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
Output:
[16:25:05] [Chromedriver Install] Installing Chromedriver version '75.0.3770.90' for platform 'mac' and architecture '64'
[16:25:05] [Chromedriver Install] Opening temp file to write 'chromedriver_mac64' to...
[16:25:05] [Chromedriver Install] Opened temp file '/var/folders/py/vdkc920j69b255sld66fbxt40000gp/T/2020422-24205-1seqp3s.8zg9/chromedriver_mac64.zip'
[16:25:05] [Chromedriver Install] Downloading https://chromedriver.storage.googleapis.com/75.0.3770.90/chromedriver_mac64.zip...
[16:25:08] Error installing Chromedriver: Error: read ECONNRESET
[16:25:08] RequestError: Error: read ECONNRESET
    at new RequestError (/Users/hypest/proj/a8c/android/tmp/t5/WordPress-Android/libs/gutenberg-mobile/node_modules/request-promise-core/lib/errors.js:14:15)
    at Request.plumbing.callback (/Users/hypest/proj/a8c/android/tmp/t5/WordPress-Android/libs/gutenberg-mobile/node_modules/request-promise-core/lib/plumbing.js:87:29)
    at Request.RP$callback [as _callback] (/Users/hypest/proj/a8c/android/tmp/t5/WordPress-Android/libs/gutenberg-mobile/node_modules/request-promise-core/lib/plumbing.js:46:31)
    at self.callback (/Users/hypest/proj/a8c/android/tmp/t5/WordPress-Android/libs/gutenberg-mobile/node_modules/request/request.js:185:22)
    at Request.emit (events.js:311:20)
    at Request.onRequestError (/Users/hypest/proj/a8c/android/tmp/t5/WordPress-Android/libs/gutenberg-mobile/node_modules/request/request.js:881:8)
    at ClientRequest.emit (events.js:311:20)
    at TLSSocket.socketErrorListener (_http_client.js:426:9)
    at TLSSocket.emit (events.js:311:20)
    at emitErrorNT (internal/streams/destroy.js:92:8)
    at emitErrorAndCloseNT (internal/streams/destroy.js:60:3)
    at processTicksAndRejections (internal/process/task_queues.js:84:21)
[16:25:08] Downloading Chromedriver can be skipped by using the '--chromedriver-skip-install' flag or setting the 'APPIUM_SKIP_CHROMEDRIVER_INSTALL' environment variable.

> Task :react-native-gutenberg-bridge:yarn_install FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':react-native-gutenberg-bridge:yarn_install'.
> Process 'command '/var/folders/py/vdkc920j69b255sld66fbxt40000gp/T/jsbundle/null/yarn/yarn-v1.22.4/bin/yarn'' finished with non-zero exit value 1
```

react-native-gutenberg-bridge/android/build.gradle
```diff
@@ -290,10 +293,16 @@ If they are changed, the isBundleUpToDate flag is switched to false. That flag i
         delete '../../node_modules'
     }
 
+    task resetExtractedRNTools(type: Delete) {
+        println "Deleting temporary folders with extracted RN tools"
+        delete nodeFolder, npmFolder, yarnFolder
+    }
+
     preBuild.dependsOn(cleanupNodeModulesFolder)
```
mchowning, line 301: Food for thought not directly related to this PR: I did notice how we're cleaning the node modules folder _after_ we generate the bundle, but we're running `resetExtractedRNTools` _before_ we create the bundle. It might help make the build a bit easier to grok if we did all the "cleaning" tasks at the same time. Could we move `cleanupNodeModulesFolder` to the beginning so that we are not left with an empty node_modules after doing a build with `BUILD_GUTENBERG_FROM_SOURCE` set to false? Put another way, by deleting the `node_modules` at the end, we have to do a `yarn install` after doing a not-from-source in order to complete a not-from source build. If we instead deleted the `node_modules` at the beginning we wouldn't have that issue (and it would also do a better job of insuring we have a clean `node_modules` folder anytime we're doing a not-from-source build).

mchowning:
> I see the following error happening quite often on my end, and not sure if it's a networking issue or not

I'm pretty sure I've never seen that, so perhaps it is a networking issue

react-native-gutenberg-bridge/android/build.gradle
hypest: > Do you think we could move cleanupNodeModulesFolder to the beginning so that we are not left with an empty node_modules after doing a build with BUILD_GUTENBERG_FROM_SOURCE set to false?

When we [introduced](https://github.com/wordpress-mobile/gutenberg-mobile/commit/6d5f623d0e3b7d55c66074c611a65f087da4f8c0) the `cleanupNodeModulesFolder` task, the idea was that it had to be at the end so the (Android) artifact bundling doesn't pick up unwanted/unneeded binaries, and to "cleanup" the project folder.

I guess that leaving the node_modules folder in could be an optimization in case `yarn install` has to be run again but we need to see (1) if unwanted binaries still get picked up and (2) how often will people need it when in not-from-source mode.

You think perhaps that (1) doesn't occur anymore and (2) is already shown that we need node_modules often enough?

mchowning: > we need to see (1) if unwanted binaries still get picked up...

I had not thought about (1) at all, so we would definitely need to verify that. 

> and (2) how often will people need it when in not-from-source mode.

I believe it should speed up the build substantially any time a new js bundle was generated. Just noting that in some initial testing I did last week I think I switched out the `yarn_install` task with the plain `yarn` task to get the speed increases. I didn't have a chance to look into that, so it's just something to keep in mind as a possibility if/when we try this change.

My biggest concern is actually that the node_modules could get into bad shape and _need_ to be deleted. For that reason, if we ended up making this change, I think we should also make sure that running the `clean` task wipes out the node_modules folder. That way it will be easy for a dev to get back to a good state.

All things considered, I don't think this is the most important improvement we need in the no-jitpack builds (some change along the lines of https://github.com/wordpress-mobile/gutenberg-mobile/pull/2329 that gives us a clean submodule folder and caching is much more important imo), but if we do end up sticking with this kind of a build, I think it's definitely worth further investigation.

PR #2306, opened 2020-05-22
Update initial HTML.

SergioEstevao:
Clean up of the initial HTML to better show all supported blocks,
in a specific order and with the latest version of block attributes.

To test:
 - Build the demo app
 - Check that the initial content is ok 

PR submission checklist:

### DIFF
--- a/src/initial-html.js
+++ b/src/initial-html.js
@@ -13,7 +13,7 @@ export default `
 <!-- /wp:heading -->
 
 <!-- wp:paragraph -->
-<p><strong>Bold</strong> <em>Italic</em> <s>Striked</s> Superscript<sup>(1)</sup> Subscript<sub>(2)</sub> <a href="http://www.wordpress.com" target="_blank" rel="noreferrer noopener">Link</a></p>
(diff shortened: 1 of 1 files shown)


src/initial-html.js
```diff
@@ -4,236 +4,229 @@
...
+<h2>Text Blocks</h2>
+<!-- /wp:heading -->
 
 <!-- wp:heading -->
 <h2>What is Gutenberg?</h2>
 <!-- /wp:heading -->
 
 <!-- wp:paragraph -->
-<p><strong>Bold</strong> <em>Italic</em> <s>Striked</s> Superscript<sup>(1)</sup> Subscript<sub>(2)</sub></p>
+<p><strong>Bold</strong> <em>Italic</em> <s>Striked</s> Superscript<sup>(1)</sup> Subscript<sub>(2)</sub> <a href="http://www.wordpress.com" target="_blank" rel="noreferrer noopener">Link</a></p>
```
etoledom, line 16: Not sure if it can be a problem, but maybe we could change the `wordpress.com` references to `wordpress.org`? To be in the safe side.

Same in the contact info block.

Issue #2292, opened 2020-05-21
[GlobalStep] [Android] Unable to operate or handle all tags in HTML mode.


gglobalstep:
**Environment:**
Android: Samsung Galaxy s9 (v10), Samsung Galaxy S7 (v8), Samsung Galaxy Tab A (v9).
Build WPAndroid [version]: v14.9

**Steps To Reproduce:**
 1. Install and launch the application.
 2. Tap on "Log In".
 3. Log in to the app using valid credentials.
 4. Tap on "Create a new Post" symbol.
 5. Enter any text at title.
 6. Tap on (More)three dots symbol at top right side.
 7. Now tap on Switch to HTML mode.
 8. Enter any tag except paragraph.
 9. Observe that user is not able to add all other tags apart from Paragraph.

Note: Paragraph tag <p> gets automatically added when we enter text in HTML view.

**Actual Result:**
Unable to operate or handle all tags in HTML mode.

**Expected Result:**
User should be able to operate smoothly in HTML mode.

**Attachment:**
[#2292.zip](https://github.com/wordpress-mobile/gutenberg-mobile/files/4663095/2292.zip)

designsimply:
Screenshots from the video attachment:

<img width="200" alt="Screen Shot 2020-05-22 at May 22 1 23 55 PM" src="https://user-images.githubusercontent.com/1119271/82702621-8fe88e00-9c2f-11ea-9e81-f7e3cf3703d6.png"> <img width="200" alt="Screen Shot 2020-05-22 at May 22 1 24 03 PM" src="https://user-images.githubusercontent.com/1119271/82702616-8e1eca80-9c2f-11ea-8337-796783c9ecc6.png">

PR #2312, opened 2020-05-25
Allow writing custom HTML in HTML mode

Tug:
Fixes https://github.com/wordpress-mobile/gutenberg-mobile/issues/2292

This PR registers the classic block for freeform support when parsing a post. It does not provide an Edit feature for the classic block though

Gutenberg PR: https://github.com/WordPress/gutenberg/pull/22609

### Testing Instructions
- Boot WPAndroid
- Switch to HTML mode
- Type text
- Wait for a second
- Notice text is not replaced by `<p>text</p>`
- Switch back to visual
- Notice unsupported block is displayed

### Warning
- If a post with only HTML (no blocks) is saved it will load with aztec whenever it's opened again
- Switching back and forth between visual and code mode might transform the HTML into paragraphs (or other blocks) as the web does

PR submission checklist:


SergioEstevao:
I gave this a spin and in general, it's working ok, but there are still some edge cases:

- If you try to write something like `I like to be <strong>bold</strong>` and the auto-save occurs in the middle of writing the `strong` element you will get some strange HTML generation.

- When you switch back to the visual mode you will see the `unsupported block` but when you tap to remove it you see:

Screenshot 2020-05-25 at 21 38 06

It says: `Remove Classic` is this expected/wanted?

Tug:
@SergioEstevao Thanks for giving it a go. I do think Remove Classic is expected as it corresponds to the classic block. I have updated the PR to reflect that on the unsupported block ui.



> If you try to write something like I like to be <strong>bold</strong> and the auto-save occurs in the middle of writing the strong element you will get some strange HTML generation.

Yeah that's a problem, maybe we could update the autosave in WP apps to be triggered less often? We can't do much on the RN side as we just react to the `native-editor.persist-html` action.

Tug:
> We can't do much on the RN side as we just react to the native-editor.persist-html action.

One thing we could do though is send the HTML as it is whenever the parent apps asks for it, no need to parse and serialize back while we're in the HTML mode. If the post is reloaded and some HTML is invalid the block parser will try to fix that (which may fail but there are always the possibility of falling back to aztec in that case)

Tug:
@SergioEstevao I did just that. Could you give it another go please?

hypest:
> One thing we could do though is send the HTML as it is whenever the parent apps asks for it.

I feel that that'd be the expectation. *When in html mode*, the user has direct access to the raw post and I think it has to be the source of truth and stay unaltered. Switching back and forth between visual and html mode might be a different deal, but while editing in html mode I'd expect the edits to be kept as is.

PR #2323, opened 2020-05-26
Fix link insertion at start of lists.

### LINKED ISSUE #2319: Linked text disappears after pressing enter
**Describe the bug**
Under certain circumstances, adding a link to a selection of rich text and tapping Enter afterwards results in the linked text disappearing.

**To Reproduce**
1. Open the editor and add a block that uses rich text (paragraph, list, quote, etc)
2. Add text 
3. Select the text
4. Tap the link button in the toolbar and add any link
5. Close the link bottom sheet
6. Tap Enter on the keyboard and observe the link disappear

**Expected behavior**
Tapping enter after a linked portion of text should not result in the linked portion of text disappearing.

**Screenshots**
<kbd></kbd>

- WordPress for iOS 14.8.2
 - Device: iPhone 7 physical device, iPhone 11 simulator
 - OS: iOS 
 - Version 13.4.1, iOS 13.5

**Additional context**
- Not present on Android 
- If there's an item in the list _after_ the linked item, the bug is **not** present
- If there's an item in the list _before_ the linked item, the bug is **still** present

SergioEstevao:
Fixes #2319 

To test:
 - Follow the steps in the ticket above
 - Beside fixing the issue the caret selection should be positioned at the end of the links after insertion.

PR submission checklist:


Tug:
Just tested the PR. It seems that the cursor does move to the end when adding a new link but it doesn't work when converting existing text. Thus the bug described in https://github.com/wordpress-mobile/gutenberg-mobile/issues/2319 is still reproducible.

PR #2329, opened 2020-05-27
Keep no jitpack builds from dirtying submodule

mchowning:
WIP 

This is just a quick implementation of an idea I had about a direction we could go to make the no-jitpack builds a bit "cleaner". My intention here is to insure that the submodule directory isn't dirty when the build starts and also clean up the submodule directory after the build is done. I have only minimally tested this code, I'm more just looking for thoughts about the general direction at this point.

Aborting if the submodule is dirty could avoid us generating an invalid bundle file because of unexpected changes in the `libs/gutenberg-mobile` folder. Cleaning the submodule directory after the build avoids distracting users with the artifacts of the build process showing up in git as changes.

My thinking is that if we are able to feel confident that the submodule is clean for the build (with checks like those I've added in this PR), then we could go further and add local caching of the JS bundles, linking them to the relevant gutenberg-mobile submodule hash (this might even be able to replace the majority of the current logic in the `bundleUpToDateCheck` task since we could rely on a submodule hash check to know if we could re-use a particular bundle). That way if you built branch A, then build branch B (which has a different gutenberg-mobile submodule hash), then went back tried to build to branch A, gradle would just check our local cache of js bundles, see that we already had one for this git hash, and use that instead of making a new one. This caching has **not** been implemented yet, it's just an idea at this point.

#### Aborts build if submodule dirty

1. Make a change to a file in `libs/gutenberg-mobile` that is under source control
2. Attempt to build WPAndroid with `BUILD_GUTENBERG_FROM_SOURCE` set to false
3. Observe that the build fails with an error message about the gutenberg-mobile submodule not being clean.

#### Cleans submodule directory
1. Build WPAndroid with `BUILD_GUTENBERG_FROM_SOURCE` set to false
2. Verify that the resulting apk runs
3. Verify that the `libs/gutenberg-mobile` submodule is clean

PR submission checklist:

### DIFF
--- a/react-native-gutenberg-bridge/android/build.gradle
+++ b/react-native-gutenberg-bridge/android/build.gradle
@@ -181,6 +181,38 @@ if (buildGutenbergMobileJSBundle) {
     def bundleName = 'index.android.bundle'
     def jsRootDir = '../..'
 
+    // FIXME also check that submodule is properly up to date???
+    task abortIfGutenbergMobileSubmoduleNotClean(type: Exec) {
+        workingDir "../../"
+        commandLine "git", "status", "--porcelain"
+        standardOutput = new ByteArrayOutputStream()
+
+        // Storing this because we cannot access `standardOutput` in the doLast block
+        def output = { standardOutput.toString() }
+        doLast {
+            if (!output().isEmpty()) {
+                def message = "Attempting to build Gutenberg JS Bundle without a clean submodule. \
+Either clean the `libs/gutenberg-mobile` submodule (you probably want to do this) or \
+build Gutenberg from source by setting wp.BUILD_GUTENBERG_FROM_SOURCE to true."
+                throw new GradleException(message)
+            }
+        }
+    }
+
+    task removeGutenbergMobileSubmoduleChanges {
+        doLast {
+            println "Removing any changes in gutenberg-mobile project"
+            exec {
+                workingDir "../../"
+                commandLine "git", "checkout", "."
+            }
+            exec {
+                workingDir "../../"
+                commandLine "git", "clean", "-f"
+            }
+        }
+    }
+
     task bundleUpToDateCheck {
         description("Checks if the inputs to the javascript bundle and the bundle itself are unchanged. \
 If they are changed, the isBundleUpToDate flag is switched to false. That flag is used by other tasks.")
@@ -298,11 +330,13 @@ If they are changed, the isBundleUpToDate flag is switched to false. That flag i
         delete nodeFolder, npmFolder, yarnFolder
     }
 
-    preBuild.dependsOn(cleanupNodeModulesFolder)
+    preBuild.dependsOn(removeGutenbergMobileSubmoduleChanges)
+    removeGutenbergMobileSubmoduleChanges.dependsOn(cleanupNodeModulesFolder)
     cleanupNodeModulesFolder.dependsOn(backupHermesDebugAAR)
     backupHermesDebugAAR.dependsOn(backupHermesReleaseAAR)
     backupHermesReleaseAAR.dependsOn(copyJSBundle)
     copyJSBundle.dependsOn(buildJSBundle)
     buildJSBundle.dependsOn(yarn_install)
     nodeSetup.dependsOn(resetExtractedRNTools)
+    resetExtractedRNTools.dependsOn(abortIfGutenbergMobileSubmoduleNotClean)
 }


react-native-gutenberg-bridge/android/build.gradle
```diff
@@ -181,6 +181,38 @@ if (buildGutenbergMobileJSBundle) {
     def bundleName = 'index.android.bundle'
     def jsRootDir = '../..'
 
+    // FIXME also check that submodule is properly up to date???
```
hypest, line 184: What do you have in mind here in terms of "up to date"?

mchowning: I was just thinking to confirm that the proper submodule hash was checked out (not behind or ahead). Doing that removes some flexibility since it forces you to always use the submodule ref that is committed, but it also might avoid people seeing weird issues because they forgot to update the submodule after switching branches. If we're going to enforce that the submodule is clean it kind of feels like it makes sense to also verify that the right submodule is checked out, but I didn't implement it yet because I was a bit torn on it.

react-native-gutenberg-bridge/android/build.gradle
```diff
@@ -181,6 +181,38 @@ if (buildGutenbergMobileJSBundle) {
...
+        commandLine "git", "status", "--porcelain"
+        standardOutput = new ByteArrayOutputStream()
+
+        // Storing this because we cannot access `standardOutput` in the doLast block
+        def output = { standardOutput.toString() }
+        doLast {
+            if (!output().isEmpty()) {
+                def message = "Attempting to build Gutenberg JS Bundle without a clean submodule. \
+Either clean the `libs/gutenberg-mobile` submodule (you probably want to do this) or \
+build Gutenberg from source by setting wp.BUILD_GUTENBERG_FROM_SOURCE to true."
```
hypest, line 196: WDYT about adding where that flag resides? It might not be easy to know for someone not experienced with developing in gutenberg-mobile.

react-native-gutenberg-bridge/android/build.gradle
```diff
@@ -181,6 +181,38 @@ if (buildGutenbergMobileJSBundle) {
...
+                throw new GradleException(message)
+            }
+        }
+    }
+
+    task removeGutenbergMobileSubmoduleChanges {
+        doLast {
+            println "Removing any changes in gutenberg-mobile project"
+            exec {
+                workingDir "../../"
```
hypest, line 206: Got to admit, I'm feeling uncomfortable having a build script undo local changes to source folder. I understand though that our goal here is to clean up the `bundle` subfolder so, what do you think about specifying directly that folder to clean up? (Same for the `git clean -f` command.

react-native-gutenberg-bridge/android/build.gradle
hypest: Ah, I see.

Hmm, failing the build if the submodule hash is beind or ahead feels a bit restrictive. I mean, when building code, we understand that if git reports something is dirty (and it will do that if the sunmodule hash is different than the committed one) then that something is what will end up in the build. I trust that the developer will have the same insight for this case too. Especially devs not working on Gutenberg, they will easily notice that the dirty folder is not something they intended to be dirty.

Besides, there's already a hint printed out when there is a mismatch (https://gist.github.com/hypest/e06f6097065728b6db7b7c462f8fef1a/#file-submodulegithash-gradle-L12). We could try to make that look like a warning than just a hint if we're concerned it will not be noticed.

hypest:
Thanks for opening this PR @mchowning !

I have this sense that having a build script modify the state of files and folders under source control is like looking for trouble :). For instance, cleaning code the developer was meaning to use would be a bad surprise. Also, cleaning might make the build irreproducible.

I'm thinking, since the issue here is with the `bundle` subfolder, how about discussing a different approach: To avoid dirtying the checkout in the first place.

The only reason we still have the JS bundle committed to git is to "match" the same file we have for WPiOS. WPAndroid doesn't really need it commited. Then again, the `block.json` files are not needed for the build either so, we could remove them from git, or git-ignore them.

Finally, we don't need to update any translation files during the build. `yarn bundle` would update them but that's the design of the command we had for gutenberg-mobile alone where `yarn bundle` is supposed to be a command for during the release process. A typical WPAndroid build shouldn't modify any gutenberg-mobile code.

We could still go ahead with this PR if we limit the git-clean to the bundle folder, but this might actually be a good opportunity to rethink what the JS bundle command should do in the case of WPAndroid building.

mchowning:
Thanks for the review @hypest! I think you're right that this PR was a bit heavy-handed. I'm going to close this for the time being to see how the new build-system pans out, but I may reopen this later (with some changes).

PR #2333, opened 2020-05-28
Use available sim on local device tests

guarani:
I'm propose making running local Appium device tests easier by using the iOS runtime installed on a developer's machine instead of looking for the hard-coded one found in `caps.js` which might not be present:
https://github.com/wordpress-mobile/gutenberg-mobile/blob/06c5f4e16d2d4ea9f50784015fc90054aa29a433/__device-tests__/helpers/caps.js#L20

#### Upside
We now don't have to manually update this value every so often. 

#### Downside
With this change developers will run tests on whichever runtime they have installed which is not necessarily the same as other developers (this isn't always a bad thing). I've also tried to make the runtime used to run the tests more explicit by logging it the build output.

#### To test
1. Checkout the branch locally
2. Ensure a local test passes, e.g.:`yarn test:e2e:ios:local __device-tests__/gutenberg-editor-latest-posts.test.js`

PR submission checklist:

### DIFF
--- a/__device-tests__/helpers/utils.js
+++ b/__device-tests__/helpers/utils.js
@@ -102,17 +102,17 @@ const setupDriver = async () => {
 
 			// Set the iOS runtime to one of the available simulator runtimes
 			try {
-			  const allRuntimes = childProcess
-			    .execSync( 'xcrun simctl list runtimes --json' )
-			    .toString()
-			  const iOSRuntimes = JSON.parse(allRuntimes)
-			    .runtimes
-			    .filter( runtime => runtime.name.startsWith('iOS') )
-			  const chosenRuntimeVersion = iOSRuntimes[0].version // assume first in list should be used
-			  desiredCaps.platformVersion = chosenRuntimeVersion
-			  console.log( 'Using iOS simulator runtime version:', chosenRuntimeVersion )
+				const allRuntimes = childProcess.execSync( 'xcrun simctl list runtimes --json' ).toString();
+				const iOSRuntimes = JSON.parse( allRuntimes ).runtimes.filter( ( runtime ) =>
+					runtime.name.startsWith( 'iOS' )
+				);
+				const chosenRuntimeVersion = iOSRuntimes[ 0 ].version; // assume first in list should be used
+				desiredCaps.platformVersion = chosenRuntimeVersion;
+				// eslint-disable-next-line no-console
+				console.log( 'Using iOS simulator runtime version:', chosenRuntimeVersion );
 			} catch ( error ) {
-			  console.error( 'No compatible iOS simulator runtime found' )
+				// eslint-disable-next-line no-console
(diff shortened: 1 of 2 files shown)


__device-tests__/helpers/utils.js
```diff
@@ -99,6 +99,21 @@ const setupDriver = async () => {
...
 			desiredCaps = _.clone( iosLocal );
 			desiredCaps.app = path.resolve( localIOSAppPath );
+
+			// Set the iOS runtime to one of the available simulator runtimes
+			try {
+				const allRuntimes = childProcess.execSync( 'xcrun simctl list runtimes --json' ).toString();
+				const iOSRuntimes = JSON.parse( allRuntimes ).runtimes.filter( ( runtime ) =>
+					runtime.name.startsWith( 'iOS' )
+				);
+				const chosenRuntimeVersion = iOSRuntimes[ 0 ].version; // assume first in list should be used
```
guarani, line 109: Comparing version strings is harder than just an alphabetic sort – and I'm not sure if there will be point (or patch versions) released (e.g. 14.5.1) – so I'm going with the first option.

I have just one runtime version installed at the moment but I'm installing another to help verify if the `xcrun simctl list runtimes --json` command used here sorts it for us.

guarani: @SergioEstevao pinging you here in case you want to add something.

guarani: I found out we needed to sort the `simctl` runtime list ourselves so I've updated the PR. 
This ended needing a complex solution to a small problem, so it's up to you whether we discard this – I'm not too keen on adding lots of extra complexity for what seems like little gain. Let me know what you think.

mchowning:
We're preparing to cut the 1.30 release @guarani , so I'm going to bump the milestone for this to 1.31. If this needs to make the 1.30 release though, just let me know. 🙂

mchowning:
👋  @guarani ! We're close to merging the monorepo changes 🎉 , and we will be cutting the release branches as soon as the monorepo changes are merged to gutenberg-mobile. As a result, we may be cutting the 1.31 release branches earlier than normal.

It looks like these changes can wait until 1.32, so I'm going to bump this to the 1.32 milestone. Please don't merge this until after the 1.31 release is cut because we don't add more changes to the 1.31 release than are necessary—this release is already going to have a lot more changes than a normal release because of the monorepo changes. 

Of course, if it is important that this PR make the 1.31 release, just let me know and let's try to get this merged ASAP so it is in before we cut the release. 🙇

mchowning:
:wave: @guarani ! We're getting ready to cut the 1.32 branch, so I'm going to bump the milestone to 1.33. If this needs to make 1.32 though, go ahead and undo that change and please let us know in Slack. :bow:

guarani:
Closing this because it doesn't look like it brings enough benefit for the added complexity.

Issue #1651, opened 2019-12-04
Enter key does not work in captions


mchowning:
## Describe the bug
Due to a [recent change](https://github.com/WordPress/gutenberg/pull/18818#issuecomment-559818548), the enter key stopped working in captions on Android. It used to add new lines within the caption, but now the newline is immediately removed.

## To Reproduce
1. Create a caption
2. Press "Enter"
3. Observe that the cursor has not moved to a new line or a new block

![](https://user-images.githubusercontent.com/4656348/69877451-3e6ad200-1290-11ea-852e-ea5c88d1705e.gif)

## Expected behavior
It is unclear what the expected behavior should be. Previously, we were allowing the addition of new lines in captions. @hypest has noted, however, that this is inconsistent with the way that the "Enter" key is handled elsewhere (i.e., paragraph blocks) on mobile. For that reason, he has suggested that we may also want to have the enter key move to a new paragraph block from within captions. It should be noted, however, that this would be a change from web, which has the Enter key create new lines within captions but has the enter key create a new paragraph block elsewhere. What do you think is the best way to proceed here @iamthomasbishop ?

## Smartphone
Pixel, Android 10, gutenberg-mobile 1.18.0

iamthomasbishop:
> Expected behavior

@mchowning I remember having a discussion about this a while back, and IIRC we stuck close to the web's behavior. I must admit, it's not clear to me which behavior is the expected one for our users but personally, I agree that I usually expect `Enter` to close the caption and create a new block because more often than not I want to add a quick caption, hit enter, and move along to a new block.

Is there a way to have a single press of `enter` to create a new line, then a second (from a blank line) break out? Similar to a list? This would be best of both worlds, imo.

At the end of the day, if we could allow the ability to hit `Shift + Enter` for new lines we'd be golden, although I think we decided to punt on that at a few points in the project. 😞

hypest:
Past ticket related to the topic: https://github.com/wordpress-mobile/gutenberg-mobile/issues/426

mkevins:
I noticed that if I type rapidly enough (typing a bunch of letters and pressing enter lots of times), occasionally I am able to get a newline to "survive". Just noting here in case it helps.

mchowning:
I'm noticing a very similar can't-add-a-newline-with-enter issue with the button block as well. Again, it only happens on Android. I wouldn't be surprised if it had a similar underlying cause. cc: @lukewalczak

[...]

PR #2352, opened 2020-06-05
Issue/426 fix enter on captions

### LINKED ISSUE #426: Proposal: add a new default block (para) when tapping Enter on Image caption
I labelled this issue `Writing Flow`, because of the following:

Imagine I'm writing, hit enter, get a new block, keep writing, all looks good. Decide to insert an image and add a caption. I expect that adding a caption and hitting ENTER gets me a new Paragraph block to keep writing. Instead, it dismisses the keyboard and no new block is added, so I need to figure out what happened for a second and then if I want to keep writing I need to add a block myself.

I think we could try making it **add a new default block**, to _keep the writing flow going_.

On the Web, ENTER does not have a special treatment but only adds a new line to the caption:
gutenbergwebcaption

wdyt?

SergioEstevao:
Fixes #426 

To test:
 - Start the demo app
 - Add an image block
 - Set an image to it
 - Add a caption
 - When writing on the caption tap Enter on the end of the caption text
 - Check if new paragraph block is inserted
 - Do the same test as above using a video block.

PR submission checklist:


mchowning:
I'm seeing an issue with this on Android where pressing Enter both adds a `<br>` to the caption and creates a new paragraph block. I've looked into it a bit, but I haven't figured out yet what is causing this.

hypest:
Is this also fixing https://github.com/wordpress-mobile/gutenberg-mobile/issues/1651?

mchowning:
> Is this also fixing #1651

👍 Closed the issue

PR #2359, opened 2020-06-08
Add Release Checklist

mchowning:
I would like to try preserving our release checklist here in the github repo. Until now, we've just been copying the release checklist from the previous release. There are three main problems with doing this, and I think creating a release checklist template on Github solves them:

1. **It takes manual editing to "clean" the previous release's checklist.** Mostly unchecking checkboxes and removing the usernames that get attached to any "completed" items. This only takes maybe 5 minutes, but it's a tedious 5 minutes that we shouldn't have to lose everytime we do a release or hotfix.
2. **Checklist updates are made by a single person editing the checklist item without any review** (unless they specifically reach out and solicit feedback). This is troubling considering how important it is to not make mistakes when creating a release. Requiring a PR with review for any changes will help insure that our release checklist is correct. 
3. **There is no additional context for any of the release checklist items.** On more than one occasion, I have been a bit confused by a particular checklist item, and it was very hard to find any additional information the checklist item. With a template committed to Github, the PR for any changes serves as a helpful record of the reason behind the change that can be referred back to later. 

In this PR, I am basically just copying our current release checklist with only a few very minimal changes. There are additional improvements I think make sense, but I believe those make more sense as separate PRs.

Note: I couldn't find a way to make a markdown template with checklist items that could be pasted into a P2 post. If there is a way to do that, I do think we would be better off making this template in markdown because that would be _much_ easier to read and maintain imo.

### To Test

1. Copy and paste the release checklist into the block editor for an app release post.
2. Preview the resulting post and verify that it is correct.

PR submission checklist:

### DIFF
--- a/docs/Releasing.md
+++ b/docs/Releasing.md
@@ -29,7 +29,7 @@ After checking that everything looks good, you just need to push the changes:
 git push --tags
 ```
 
-# Release Checklist
+# Release Checklist Template
 
 Just copy this checklist and replace all occurrences of `X.XX.X` with the applicable release number, when we are ready to
 cut a new release.
@@ -38,6 +38,10 @@ cut a new release.
 <!-- wp:heading {"level":1} -->
 <h1>Gutenberg Mobile X.XX.X – Release Scenario</h1>
 <!-- /wp:heading -->
+
+<!-- wp:paragraph -->
+<p>This checklist is based on the <a href="https://github.com/wordpress-mobile/gutenberg-mobile/blob/develop/docs/Releasing.md#release-checklist">Release Checklist Template</a>. If you need a checklist for a new gutenberg-mobile release, please copy from that template.</p>
+<!-- /wp:paragraph -->
 
 <!-- wp:paragraph -->
 <p>+mobilegutenberg</p>
--- a/docs/Releasing.md
+++ b/docs/Releasing.md
@@ -116,7 +116,7 @@ cut a new release.
 <!-- /wp:paragraph -->
 
 <!-- wp:paragraph -->
-<p>o Create a new release and name it with the tag name from step 1. For Aztec-iOS, follow <a href="https://wp.me/p77Llu-buw">this post</a>. For Aztec-Android, releases are created via the <a href="https://github.com/wordpress-mobile/AztecEditor-Android/releases">GitHub releases page</a> by hitting the “Draft new release” button, put the tag name to be created in the tag version field and release title field, and also add the changelog to the release description. The binary assets (.zip, tar.gz files) are attached automatically after hitting “Publish release”.</p>
+<p>o Create a new release and name it with the tag name from step 1. For Aztec-iOS, follow <a href="https://github.com/wordpress-mobile/AztecEditor-iOS/blob/develop/Documentation/ReleaseProcess.md">this process</a>. For Aztec-Android, releases are created via the <a href="https://github.com/wordpress-mobile/AztecEditor-Android/releases">GitHub releases page</a> by hitting the “Draft new release” button, put the tag name to be created in the tag version field and release title field, and also add the changelog to the release description. The binary assets (.zip, tar.gz files) are attached automatically after hitting “Publish release”.</p>
 <!-- /wp:paragraph -->
 
 <!-- wp:heading {"level":3} -->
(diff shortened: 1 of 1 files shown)


docs/Releasing.md
```diff
@@ -27,4 +27,167 @@ After checking that everything looks good, you just need to push the changes:
...
+<!-- wp:paragraph -->
+<p>o Make sure there is no pending Aztec PR required for this Gutenberg release. Check the commit hash referred in the gutenberg repo is in the Aztec <code>develop</code> branch. If it's not, make sure pending PRs are merged before next steps.</p>
+<!-- /wp:paragraph -->
+
+<!-- wp:paragraph -->
+<p>o Open a PR on Aztec repo to update the <code>CHANGELOG.md</code> and <code>README.md</code> files with the new version name.</p>
+<!-- /wp:paragraph -->
+
+<!-- wp:paragraph -->
+<p>o Create a new release and name it with the tag name from step 1. For Aztec-iOS, follow <a href="https://wp.me/p77Llu-buw">this post</a>. For Aztec-Android, releases are created via the <a href="https://github.com/wordpress-mobile/AztecEditor-Android/releases">GitHub releases page</a> by hitting the “Draft new release” button, put the tag name to be created in the tag version field and release title field, and also add the changelog to the release description. The binary assets (.zip, tar.gz files) are attached automatically after hitting “Publish release”.</p>
```
mchowning, line 119: Do you think we could move the information about creating an Aztec-iOS release to somewhere in the Aztec-iOS repo @SergioEstevao ? That would allow us to update this to be a public link.

SergioEstevao: Something like this: https://github.com/wordpress-mobile/AztecEditor-iOS/pull/1294

docs/Releasing.md
```diff
@@ -27,4 +27,167 @@ After checking that everything looks good, you just need to push the changes:
...
\ No newline at end of file
+```
+
+# Release Checklist
+
+Just copy this checklist and replace all occurrences of `X.XX.X` with the applicable release number, when we are ready to
+cut a new release.
+
+```
+<!-- wp:heading {"level":1} -->
```
maxme, line 38: Can we add a message here and a link to this doc, so no one is tempted to copy/paste the p2 post directly. Something like: "If you need to release gutenberg-mobile, please use [the checklist from our GitHub repo](https://github.com/wordpress-mobile/gutenberg-mobile/blob/master/docs/Releasing.md)"

@SergioEstevao, can we also add "deprecated document, please visit https://github.com/wordpress-mobile/gutenberg-mobile/blob/master/docs/Releasing.md" in the release gdoc?

SergioEstevao: > @SergioEstevao, can we also add "deprecated document, please visit https://github.com/wordpress-mobile/gutenberg-mobile/blob/master/docs/Releasing.md" in the release gdoc?

@maxme My understanding is that this document will replace the one that you link, so were do we want to mark deprecate?

maxme: We can delete the gdoc, I want to make sure no one update it or use it.

docs/Releasing.md
mchowning: Thanks @SergioEstevao ! I approved that PR with a couple of minor comments. I also went ahead and updated this to refer to that doc in the Aztec-iOS repo in 73796b4.

docs/Releasing.md
mchowning: > Can we add a message here and a link to this doc, so no one is tempted to copy/paste the p2 post directly.

Good idea @maxme. Added in ee027c4.

docs/Releasing.md
SergioEstevao: Here is the final link: https://github.com/wordpress-mobile/AztecEditor-iOS/blob/develop/Documentation/ReleaseProcess.md

docs/Releasing.md
```diff
@@ -27,4 +27,171 @@ After checking that everything looks good, you just need to push the changes:
 
 ```
 git push --tags
```
SergioEstevao, line 29: Should we keep this information above? I don't think the version command even works now...

mchowning: `yarn version` still works. It's one of the steps in the checklist even. With that said, I don't think anyone is bumping the version outside of the release process, so I removed the portion that references that in 9e5ecf1. I left the portion that just discusses creating a bundle though. I don't feel strongly about keeping it, but it seemed like it was correct, so I left it in. But let me know what you think.

PR #2362, opened 2020-06-09
Update the VS code search settings to that seaching is easier

enejb:
Currently in VS code when you search for a specific string you get back results that include build files, and symlinked files so the string in question can appear in multiple places. 

This PR tries to fix that in the VS code editor since by removing specific folders from search.

To test:
- Pull in this PR, perform a search in VS code do you get back files as expected? 

PR submission checklist:

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -7,8 +7,7 @@
     // Use 'prettier-eslint' instead of 'prettier'. Other settings will only be fallbacks in case they could not be inferred from eslint rules.
     "prettier.eslintIntegration": true,
     "search.exclude": {
-        "**/bundle/android": true,
-        "**/bundle/ios": true,
+        "**/bundle/": true,
         "**/gutenber/packages/**/build": true,
         "**/gutenber/packages/**/build-modules": true,
         "**/gutenber/packages/**/build-types": true,
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -7,7 +7,10 @@
     // Use 'prettier-eslint' instead of 'prettier'. Other settings will only be fallbacks in case they could not be inferred from eslint rules.
     "prettier.eslintIntegration": true,
     "search.exclude": {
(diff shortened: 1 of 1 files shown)


.vscode/settings.json
```diff
@@ -6,4 +6,13 @@
 
     // Use 'prettier-eslint' instead of 'prettier'. Other settings will only be fallbacks in case they could not be inferred from eslint rules.
     "prettier.eslintIntegration": true,
+    "search.exclude": {
+        "**/bundle/android": true,
```
hypest, line 10: 👋 Enej! I usually exclude the App.js and App.js.map files only, not the whole folder, since those files are big/noisy and usually just are duplicate results found elsewhere. What do you think about going that route? Any particular reason to exclude the whole folder?

enejb: Nope, let me update it to that we exclude the whole folder.

enejb: Sorry I miss understood, I updated the code so that we only exclude the App.js and the App.js.map files.

.vscode/settings.json
```diff
@@ -6,4 +6,15 @@
...
     // Use 'prettier-eslint' instead of 'prettier'. Other settings will only be fallbacks in case they could not be inferred from eslint rules.
     "prettier.eslintIntegration": true,
+    "search.exclude": {
+        "**/bundle/android/App.js": true,
+        "**/bundle/android/App.js.map": true,
+        "**/bundle/ios/App.js": true,
+        "**/bundle/ios/App.js.map": true,
+        "**/gutenber/packages/**/build": true,
+        "**/gutenber/packages/**/build-modules": true,
+        "**/gutenber/packages/**/build-types": true,
```
hypest, line 16: Is `gutenber` a typo perhaps, missing the final `g`?

enejb: Yes that was a typo. Thanks for catching it.

Issue #2325, opened 2020-05-27
A more reliable communication mechanism between the main apps and the RN app


pinarol:
Let's investigate if we can make a more reliable communication mechanism between the main apps and the RN editor.

Currently we know that some events can fall through cracks if they are fired during the time editor is bootstrapping. This is causing such bugs:

https://github.com/wordpress-mobile/gutenberg-mobile/issues/2041

https://github.com/wordpress-mobile/gutenberg-mobile/issues/1928

chipsnyder:
Just as an FYI here is an interesting find from React-Core 

https://github.com/facebook/react-native/blob/1fa3a8e4752fb7b34aaa8b464d7a678427f65ed4/React/Modules/RCTEventEmitter.m#L67

There is then this logic at the bridge that attempts to send the events after the bridge has loaded https://github.com/facebook/react-native/blob/master/React/CxxBridge/RCTCxxBridge.mm#L1236-L1273

This is just the iOS side so far but I'm exploring working around these areas to see if there is something we can tweak here.

mkevins:
Hi @chipsnyder :wave: :smile: 

> There is then this logic at the bridge that attempts to send the events after the bridge has loaded wich seems like if allowed to run would solve some of our issues (probably not all though)

Is this `_loading` state referring to the JS runtime initialization state? I guess we will need to wait until after all the components with listeners have mounted, so even with eventEmitter calls queued / batched over the bridge during editor startup, we'll need to make sure all the blocks have mounted first (so sometime after the initial content is parsed and blocks various components are instantiated).

pinarol:
I see that all the PRs are merged, I think we can close this issue. Let me know if something is pending.

Issue #2041, opened 2020-03-20
Invalid image URL after Close/Re-opening Gallery multiple times during upload of multiple images


marecar3:
**Describe the bug**
Close/Re-opening a gallery block multiple times while images are uploading will result in an invalid URL.

**To Reproduce**

Follow instructions from this test case: - Gallery [TC008](https://github.com/wordpress-mobile/test-cases/blob/master/test-cases/gutenberg/gallery.md#tc008)
but keep in mind that you need to close/reopen gallery block multiple times during upload.

More details can be found here: https://github.com/wordpress-mobile/gutenberg-mobile/pull/2025#issuecomment-601241846

**Expected behavior**
The image URL should be valid when the upload is finished.

pinarol:
Resolved with associated PRs

Issue #1928, opened 2020-02-18
Media Upload fails to upload image when leaving and coming back in


chipsnyder:
**Describe the bug**
Close/Re-opening a page with ongoing image uploads in a Gallery Block causes the image to be marked for retry.

**To Reproduce**
1. Create a new Draft Page or select an existing page. 
2. Add a Gallery or Image Block
3. Select Add Media
4. Select "Choose From Device"
5. Select an Image
6. While the image is uploading; select back
7. Go right back into the Page

**Error:** Image gets locked into a mode where it's tagged to be uploaded
**Expected:** Image should upload successfully and reflect the proper state. 

https://github.com/wordpress-mobile/test-cases/blob/master/test-cases/gutenberg/gallery.md#tc001

**Expected behavior**

**Screenshots**
![FailedUpload_GalleryTC001 2020-02-18 09_38_50](https://user-images.githubusercontent.com/3384451/74745915-96cbc300-5232-11ea-9248-adcba01cb261.gif)

**Smartphone (please complete the following information):**
 - Device: Pixel 3a
 - OS: Android 10
 - Version API 29

[APK used to test](https://37929-9306568-gh.circle-artifacts.com/0/Artifacts/WordPress-pr-11331-build-37929.apk)

**Additional context**
Discovered during regression testing for [Release 1.23.0](https://github.com/wordpress-mobile/WordPress-Android/pull/11331)

pinarol:
What happens when we dismiss the retry dialog? Does it complete successfully?

chipsnyder:
For me when the app was uploading a new page and a new image it was just hanging on the upload for a new page and never completing (waited for a few minutes). However, if I can fix this issue by:
1) Tap the image
2) Select Retry
3) Tap the image
4) Select Cancel 

Then the page uploads successfully. Just revalidated a few flows and this looks like it's an issue on other media blocks as well not just gallery.

pinarol:
I can repro this with the same apk on Android:

- Add an image block
- Start an upload
- Close/Re-open the post while the upload is still going on

Expected: 

- The post loads immediately showing an upload progress on the Image block
- Image upload continues and end up with success

Reality: 
 
- The post displays white screen until the upload is complete(the upload is observable on device's status bar only)
- There's no progress indication on Image block
- Image block still shows the local image even after waiting some time

chipsnyder:
[...]

PR #2366, opened 2020-06-09
Queue Events for better media uploads

chipsnyder:
Fixes (the iOS part of): https://github.com/wordpress-mobile/gutenberg-mobile/issues/2325
`WordPress-iOS` https://github.com/wordpress-mobile/WordPress-iOS/pull/14311

### To test:
Follow the testing steps in https://github.com/wordpress-mobile/gutenberg-mobile/issues/2041 and https://github.com/wordpress-mobile/gutenberg-mobile/issues/1928

### PR submission checklist:


react-native-gutenberg-bridge/ios/RNReactNativeGutenbergBridge.swift
```diff
@@ -188,18 +204,28 @@ public class RNReactNativeGutenbergBridge: RCTEventEmitter {
         }
     }
 
-    private func shouldLog(with level: Int) -> Bool {
-        return level >= RCTGetLogThreshold().rawValue
+    public func connectionEstablished() {
+        guard !hasObservers else { return } // We have an established connection no need to continue.
+        hasObservers = true
```
etoledom, line 209: `hasObservers` and `isJSLoading` seems to be doing pretty much the same. Not sure if it's good idea to refactor everything on this same PR though. 

I'm happy to have a follow up PR with these changes.

react-native-gutenberg-bridge/ios/RNReactNativeGutenbergBridge.swift
```diff
@@ -1,10 +1,24 @@
...
 public class RNReactNativeGutenbergBridge: RCTEventEmitter {
     weak var delegate: GutenbergBridgeDelegate?
     weak var dataSource: GutenbergBridgeDataSource?
     private var isJSLoading = true
     private var hasObservers = false
+    private var queuedEvents = [GutenbergEvent]()
 
+    public override init() {
+        super.init()
+        NotificationCenter.default.addObserver(forName: Notification.Name("RCTContentDidAppearNotification"), object: nil, queue: nil) { (_) in
```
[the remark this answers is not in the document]
chipsnyder: Thanks @etoledom, I was struggling to find this!

mkevins:
Also tested, and really tried to force a race condition (tried multiple times via image and gallery block, images with various filesizes, and reopening the editor with and without metro). Everything is working as expected :tada: . Nice work Chip!

react-native-gutenberg-bridge/ios/RNReactNativeGutenbergBridge.swift
chipsnyder: I agree this would be a good idea. I started going down this path in this PR but as it was really focused on making sure this pattern worked I was hesitant to make all of those changes here. 

I just opened up a new issue to track this debit https://github.com/wordpress-mobile/gutenberg-mobile/issues/2390

react-native-gutenberg-bridge/ios/RNReactNativeGutenbergBridge.swift
```diff
@@ -1,10 +1,24 @@
+struct GutenbergEvent {
+    let name: String
+    let body: Any
+}
+
```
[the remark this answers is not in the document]
chipsnyder: That's a good idea, I think we should be able to combine them in a way to maybe add some associated values to the enum. I think that would be worth exploring as part of https://github.com/wordpress-mobile/gutenberg-mobile/issues/2390

chipsnyder:
Thanks for the review @mkevins!

@etoledom Will you take another pass and see if I hit all of your concerns? I think by "It'd be great to fix the warning before merging." you were referring to https://github.com/wordpress-mobile/gutenberg-mobile/pull/2366#discussion_r440234758 but let me know if I missed a different warning :)

react-native-gutenberg-bridge/ios/RNReactNativeGutenbergBridge.swift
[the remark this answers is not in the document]
etoledom: Great!

react-native-gutenberg-bridge/ios/RNReactNativeGutenbergBridge.swift
```diff
@@ -188,18 +204,28 @@ public class RNReactNativeGutenbergBridge: RCTEventEmitter {
...
 
-    override public func startObserving() {
-        super.startObserving()
-        hasObservers = true
+    public override func sendEvent(withName name: String!, body: Any!) {
+        DispatchQueue.main.async {
+            if self.hasObservers && self.queuedEvents.count == 0 {
+                super.sendEvent(withName: name, body: body)
+            } else {
+                let event = GutenbergEvent(name: name, body: body)
```
etoledom, line 221: Sorry! I meant the warning on this line. I wrote the comment but it disappeared (second time that happens to me this week ¯\\\_(ツ)_/¯ )

The warning says `Coercion of implicitly unwrappable value of type 'Any?' to 'Any' does not unwrap optional`

Would the override still work if we change the optionality of the parameters?
Actually on a second look, that `Any!` looks quite dangerous. This might crash if any queued event had no body.

It would be great if `name` is not optional and `body` is an actual non-implicitly-unwrapped optional, but open to possibilities too.

react-native-gutenberg-bridge/ios/RNReactNativeGutenbergBridge.swift
etoledom: Amazing! Thank you 🙏

react-native-gutenberg-bridge/ios/RNReactNativeGutenbergBridge.swift
etoledom: I just did a test and it doesn't crash. `Any` seems to accept a nil value even if it's not an optional itself (?)

But still... I wouldn't trust the runtime completely with this 😬

mkevins: > I just did a test and it doesn't crash. `Any` seems to accept a nil value even if it's not an optional itself (?)

Interesting! I wouldn't have expected that. Does that mean the struct's default initializer is not actually force-unwrapping the value? When you tested with a `nil`, do you know if the `GutenbergEvent` was initialized, or did the code branch directly to the `super.sendEvent`?

etoledom: > Perhaps there is some subtlety with permissiveness afforded for potential value or reference types? 🤔

This kinds of make sense in a way that `Optional` is basically a `Value type`, and we are passing `Optional<>.none` as `Any`. 

And `AnyObject` wouldn't accept it since it expects `Reference` types only, so it will try to unwrap the optional. It would probably not even compile.

mkevins: Thanks for the explanation Eduardo! I also found [this explanation](https://stackoverflow.com/a/34644253/882146), which is just as you mentioned. The discussion is interesting because it explains why you cannot directly assign nil to a var with type Any, since there is no inferred type for the generic in the Optional. That there is no error in the above case was surprising to me, but this explains it (kind of a language design edge case, I suppose).

etoledom: So @mkevins , what do you think: Should we avoid this edge case?

I think it would be good to try avoid it, but since it's clear now that it won't crash, maybe it's not a blocker for this PR. Still it would be good to clear the warning 😬

chipsnyder: > comment but it disappeared somehow

Oh, that is weird! Thanks for confirming though. 

>The warning says Coercion of implicitly unwrappable value of type 'Any?' to 'Any' does not unwrap optional

I totally missed this error thank you for catching it. Also thank you both for the investigation and knowledge share. That's really interesting on how `Any` and `Optionals` interact. 

> (kind of a language design edge case, I suppose).

Yeah this is neat, sometimes it's easy to forget that language design has its own edge cases to deal with.


> what do you think: Should we avoid this edge case?

To me, it seems like an easy fix to adjust. So I'll make that tweak today and push up a commit to this PR here shortly.

chipsnyder: This is actually an interesting situation. @etoledom I didn't run into the warning that you were seeing so I was wondering if it was a runtime flag, but I wasn't able to see it at runtime either. 

I do think it would be a good idea to be explicit about the optionality of the event parameters though so I did an update to address that. If you don't mind will you let me know if it fixes the warning?

XCode 11.5 
Screen Shot 2020-06-17 at 12 38 24 PM

mkevins: Fwiw, I didn't see the warning either (but I was exploring the code via the WordPress-iOS project, so I wonder if that's a factor?)

etoledom: Oh right, you will see it if you run the gutenberg example project on Xcode.

PR #2368, opened 2020-06-10
[Gutenberg] Add AccessibilityLabel and hint to AztecViewManager

### LINKED ISSUE #2364: When the page/post title is selected tell the user about that they are edditing the post title
**Describe the bug**
When a user is in VoiceOver mode and they create a new page to edit. They land on the Page Title field but it is not clear that they are editing the page title. 

**To Reproduce**
1. In the WP add enable VoiceOver, (using siri)
2. Create a new page. By selecting the + and double clicking. 
3. Notice that VoiceOver never metions that you are editing the Page title.
 
**Expected behavior**
When you land on the page 

**Screenshots**
Here is a quick video. 
https://cloudup.com/cNlMs6Fbu_F

 - Device: iPhone Pro 11
 - OS: iOS 13.5
 - Version: 15

**Additional context**

enejb:
Fixes #2364

Related GB PR: https://github.com/WordPress/gutenberg/pull/23057

To test:
1. Download WordPress-iOS. 
2. Make sure that he WordPress-iOS repo is on the same level as this directory. (gutenberg-mobile) 
> - gutenberg-mobile (this repo)
> - WordPress-iOS (https://github.com/wordpress-mobile/WordPress-iOS)

3. Run the following command on the WordPress-iOS repo
```
$ export LOCAL_GUTENBERG=1
bundle exec pod update
``` 
4. Load this PR. 
5. Build the iOS app using xcode and send it open it on your device. 
6. Navigate to the Pages screen. 
7. Enable VoiceOver using Siri. 
8. Navigate to a new post. 

Before:
See https://cloudup.com/cNlMs6Fbu_F

After: You should get this experience 
https://cloudup.com/cjd-RHZpyQM

PR submission checklist:

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -10,6 +10,7 @@
 * [***] Add support for mentions.
 * [***] Users can now individually edit unsupported blocks found in posts or pages.
 * [*] [iOS] Improved editor loading experience with Ghost Effect.
+* [*] Accessibility, update the title VoiceOver label to be more clear when you are editing the post or page title.
 
 1.31.1
 ------
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -3,6 +3,7 @@
 * [***] Media editing support in Media & Text block.
 * [***] New block: Social Icons
 * [*] Cover block placeholder is updated to allow users start the block with a background color
+* [*] Accessibility, update the title VoiceOver label to be more clear when you are editing the post or page title.
 
 1.32.0
 ------
@@ -10,7 +11,6 @@
 * [***] Add support for mentions.
 * [***] Users can now individually edit unsupported blocks found in posts or pages.
 * [*] [iOS] Improved editor loading experience with Ghost Effect.
-* [*] Accessibility, update the title VoiceOver label to be more clear when you are editing the post or page title.
 
 1.31.1
 ------


react-native-aztec/ios/RNTAztecView/RCTAztecViewManager.m
```diff
@@ -16,6 +16,9 @@ @interface RCT_EXTERN_MODULE(RCTAztecViewManager, NSObject)
 RCT_EXPORT_VIEW_PROPERTY(onActiveFormatsChange, RCTBubblingEventBlock)
 RCT_EXPORT_VIEW_PROPERTY(onActiveFormatAttributesChange, RCTBubblingEventBlock)
 
+RCT_EXPORT_VIEW_PROPERTY(accessibilityLabel, NSString)
+RCT_EXPORT_VIEW_PROPERTY(accessibilityHint, NSString)
+
```
etoledom, line 21: For this to work we should use `RCT_REMAP_VIEW_PROPERTY` instead.
But my question is, do we want to add extra labels and hints to the Aztec TextView?

Personally, I think it's better for the system to handle accessibility on UITextViews since it's quite complex.

As an example, adding the current label and hint, the VoiceOver will read on Aztec's empty post title:

Label: "Title, Empty. TextField, is editing. 0 width space, character mode, insertion point at end. 
Hint: "Update the title. Use the rotor to access misspelled words."

Maybe we are complicating the already complex TextView labels and hint, it might get in the way of the users instead of helping.

Having this label and hint in the wrapper view should be enough to give the user the info needed, then they can just start editing in the TextView after performing the action.

What do you think? @enejb

etoledom: Ok not it makes sense to me on a new post where the title is automatically focused, not having this is confusing. 🤔

enejb: Yeah I am not sure what the best approach is here. To me it also sounds like lot of text. 

I just wanted to make sure that the user know that they are editing a the post title when they get dropped in creating a new post or page.

etoledom: > For this to work we should use RCT_REMAP_VIEW_PROPERTY instead.

So, this was a mistake on my part:

`RCT_EXPORT_VIEW_PROPERTY` will work just fine, and the property will be set to the view's `accessibilityHint` property as expected.

`RCT_REMAP_VIEW_PROPERTY` is used when the JS name doesn't match the View's property name. This gives an opportunity to re-name the property.

enejb: I remove the `RCT_EXPORT_VIEW_PROPERTY` since it wasn't doing anything.

PR #2369, opened 2020-06-10
Set base typographic rhythm

### LINKED ISSUE #550: Improve base typographic rhythm
When doing some investigation during some recent Headings font work, I noticed that our base typographic rhythm is a bit cramped. I compared against the following:

* **Core GB**: 16 / 1.8
* **Reader (web/mobile)**: 17 / 1.6
* **Mobile GB (current)**: 16 / 1.375
* **Mobile GB (proposed)**: 16 / 1.5

For a visual reference, set on a canvas width of 375:

image

I'm tempted to defer to what Reader web/mobile has in terms of line-height, but considering the base font-size is 16 in our case, the 1.5 line-height ends up being a nice round 24pt, which aligns really well with our 16pt grid and our Block top/bottom margins (12pt).

screen shot 2019-02-06 at 4 16 25 pm

### Punch List

- [ ] Change base line-height of paragraph text to 1.5. If we have to define a static px value, let's go with 24.

Bonus points if we can set up a nice type system that scales well with Lists, Quotes, etc. when we get to those blocks.

hypest:
Fixes #550 

Adds support in ReactAztecText (Android only for now) for the `lineHeight` property. At the same time, setting the default fontSize to 16(px) and default lineHeight to 26(px).

Note: Since this PR changes the defaults for the RichText component, blocks that depend on it might change as well. For example the List block, Heading block and others.

Gutenberg PR: https://github.com/WordPress/gutenberg/pull/23079

WPAndroid PR for testing this with a prebuilt APK: https://github.com/wordpress-mobile/WordPress-Android/pull/12156

Keeping this PR as a draft until we decide whether to wait for the iOS side implementation to happen on this PR, or not.

To test:
1. Run the demo app and have a paragraph block with a handful of lines of text
2. Check that the line spacing is around 1.6x

### Screenshots

Before | After
-|-
Screenshot_1591827487|Screenshot_1591827786
Screenshot_1591828274|Screenshot_1591827994

PR submission checklist:

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -1,3 +1,7 @@
+1.31.0
+------
+* [**] More spacious basic typographic rhythm
+
 1.30.0
 ------
 * [**] Adds editor support for theme defined colors and theme defined gradients on cover and button blocks.
--- a/react-native-aztec/android/src/main/java/org/wordpress/mobile/ReactNativeAztec/ReactAztecManager.java
+++ b/react-native-aztec/android/src/main/java/org/wordpress/mobile/ReactNativeAztec/ReactAztecManager.java
@@ -338,6 +338,14 @@ public class ReactAztecManager extends BaseViewManager<ReactAztecText, LayoutSha
         }
     }
 
+    @ReactProp(name = ViewProps.LINE_HEIGHT, defaultInt = ViewDefaults.LINE_HEIGHT)
+    public void setLineHeight(ReactAztecText view, int lineHeight) {
+        // setLineHeight is not available on API 21 (the current app minimum) so,
+        // go with setLineSpacing instead, by using this formula:
+        // (font-size) * (line-height multiplier) = (line-height value)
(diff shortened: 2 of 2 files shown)


etoledom:
I'm implementing setting line height on iOS, it works but I haven't pushed since I'm seeing a few problems with it.

In general terms, it's good to let the system take care of line spacing, and set it manually just as exceptions when it's required.

In this case, having a font size that is smaller than the manually set line height is giving us some troubles on List Block and Preformated block:

| Before | After |
| --- | --- |
|  IMG_0104 | IMG_0103 |

The text inside a bigger line will fall to the bottom, having extra space at the top that is visible at the top of the preformatted block, and makes the bullets unaligned on list blocks.

cc @iamthomasbishop @SergioEstevao

hypest:
Thank you @etoledom !

Closing this PR until we go back working on https://github.com/wordpress-mobile/gutenberg-mobile/issues/550 as it has gone stale.

Issue #2326, opened 2020-05-27
Proposal: Copy/Cut/Paste blocks


pinarol:
1. Currently there’s no way to move a block inside/outside of nested blocks which is a big missing piece. Cut/Paste can be a quick way to support this.

2. When I style a button block I feel the need to just copy the same button near it. Currently you can only add a new Button block from scratch. Copy paste can be a quick way to cover these kinds of needs. Button block is just an example but we'll keep on adding more style options to the blocks.

Copy/Cut/Paste is a popular paradigm so I feel like we can have a quick win feature here without needing to complicate the editor.

What I have in mind is just putting the copy/cut options to the new block menu we have. (The one that also has the Trash icon.) 

Again if sth has been copied or cut, we could show options there like `Paste after this`, `Paste before this`.

cc @iamthomasbishop

iamthomasbishop:
> Currently there’s no way to move a block inside/outside of nested blocks which is a big missing piece. Cut/Paste can be a quick way to support this.

I agree w/ your general assessment, this would be a really natural thing to have. I think there are 2 main considerations in my mind — using the native Edit/Selection Menus (mostly tied to richText blocks) and copy/paste via the `•••` menu on the inline toolbar.

The obvious path on the RN side is the `•••` menu options. When the user taps `•••` we should show options to `Copy block` and `Paste block` (if there is one in the clipboard). If the user has copied a block to their clipboard, tapping on `Paste block` would show a sheet that is similar to the one we show when users long-press the inserter button. This would show options for `Paste block before`, `Replace current block` (if this is an empty block), and `Paste block after`. That flow would look like this:

<a href="https://user-images.githubusercontent.com/1200432/83049612-137bf380-a011-11ea-9cda-0b9fab64505f.png"><img src="https://user-images.githubusercontent.com/1200432/83049612-137bf380-a011-11ea-9cda-0b9fab64505f.png" width="726"></a>

Alternatively, using the native Edit/Selection Menus would work something like this: When the user taps a richText input, we inherently show the Edit/Selection Menu, which shows options for copy/paste/etc. If the block is empty, we skip straight to replacing the empty block with the one from the clipboard. If the block has contents, we show the position sheet (paste above, below, replace current block). That would look like this (example uses an empty paragraph block):

[...]

PR #2378, opened 2020-06-12
Copy, Cut, Paste, Duplicate blocks

geriux:
Fixes https://github.com/wordpress-mobile/gutenberg-mobile/issues/2326

`Gutenberg PR` -> https://github.com/WordPress/gutenberg/pull/23110
`WordPress iOS PR`-> https://github.com/wordpress-mobile/WordPress-iOS/pull/14392
`WordPress Android PR`-> https://github.com/wordpress-mobile/WordPress-Android/pull/12187

This PR adds the functionality to copy, cut, paste, and duplicate blocks. This would be very useful to be able to move blocks from inner blocks or to group blocks like Group or Columns.

To test check `Gutenberg` PR description.

PR submission checklist:


pinarol:
Let's not forget to add release notes, though.

geriux:
Hey there 👋 @iamthomasbishop !

Whenever you have a chance, could you please test this feature and give us your feedback?

Here you'll find the builds:

`WordPress iOS PR`-> https://github.com/wordpress-mobile/WordPress-iOS/pull/14309
`WordPress Android PR`-> https://github.com/wordpress-mobile/WordPress-Android/pull/12187

More info and screenshots [here](https://github.com/WordPress/gutenberg/pull/23110).

Thank you so much!

iamthomasbishop:
@geriux This is looking pretty solid already — nice work! After testing out pretty extensively, I've got a bunch of feedback which I'll try to keep succinct 🙂 I'm sure we'll need another round of review, but this should cover the big picture things.

## Overfow menu

- Can we hide the paste action when there isn't anything in the clipboard to paste? (rather than just disabling it)
- On iOS, do we have access to the native [Context Menu](https://developer.apple.com/design/human-interface-guidelines/ios/controls/context-menus/) component? That might help to condense the menu a bit and also if we can get the menu to appear from near the `•••` menu, we can have some proximity benefits. If we used that on iOS, perhaps it would be trying the [Menu](https://material.io/components/menus) component on Android for consistency.  ([Examples](https://cloudup.com/cAFUmlFv6_1) of how this might work)
- I'm not sold on the "Cut" option, and because web also doesn't have it, I vote we remove it to conserve space as we add more items.
- I think we might want the labels on the sheet slightly more descriptive — "Copy block" and "Duplicate block", "Paste block", so it doesn't conflict/confuse with the native side (eg. while copying one part of a broader text string, "copy" means copy only that text, whereas in this context "copy" means the whole block).

## Native select/copy/paste

- Are we able to connect with the native paste function so users can use the [native edit/selection menu components](https://cloudup.com/cLoKaPzDljt) to paste? For example, if I’ve copied a block from the overflow menu, tap on a rich text input to see the edit menu, then tap paste, the block I’ve copied is pasted. ([example on iOS](https://cloudup.com/cUEjrtFx7oP))

## Notice component

- Can we also show notices when blocks are *duplicated*, *pasted*, or *removed*?
- Instead of stacking multiple notices when there are multiple at a given moment, can we substitute them? For example, if I perform an action that triggers a notice and then another action that triggers a second, they currently stack on top of each other. I think it might feel smoother if the second one simply replaces the first one (in the same Y-position, via a cross-fade or instantly)
- I've dropped updated styling for the component in the [design docs here](https://zeroheight.com/226850145/p/955bb6-compact-notice). The key thing you'll see is that I'm attempting to replicate the iOS-system materials on iOS, which produces a background blur and translucency. I talked to @wppinar today about potentially using a third party library called [Blur](https://www.npmjs.com/package/@react-native-community/blur) if we're unable to access the system materials directly. Android uses solid colors, which we can fine-tune to use available colors if necessary.

## Unrelated

- For the overflow menu in general, I wonder if long-pressing on an un-selected block (or the surrounding UI on a selected block aka inline toolbar) should show the overflow/menu sheet with all of the options? [Here is an example flow](https://cloudup.com/cfMPe2_RN57)

pinarol:
> I'm not sold on the "Cut" option, and because web also doesn't have it, I vote we remove it to conserve space as we add more items.

Actually "Cut block" is implemented behind the scenes on web, but it is connected with Cmd+X. It is just not in the menu but the functionality is there. You can see this small notice on web when you do cmd+x:

Screen Shot 2020-06-18 at 10 52 51


And as you know you can drag&drop blocks inside/outside a nested block by pressing down the up/down buttons on web.

But we don't have a way to move a block inside/outside a nested block. In this sense Cut functionality is quite useful imo, and it is a quick win for us since we are already implementing Copy/Paste/Duplicate, wdyt?

mchowning:
👋  @geriux ! We're close to merging the monorepo changes 🎉 , and we will be cutting the release branches as soon as the monorepo changes are merged to gutenberg-mobile. As a result, we may be cutting the 1.31 release branches earlier than normal.

If these changes can wait until 1.32, please bump this to the 1.32 milestone and wait to merge this until after the release branches are cut. That way we don't add more changes to the 1.31 release than are necessary—this release is already going to have a lot more changes than a normal release because of the monorepo changes. If it is important that this PR make the 1.31 release, however, please try to merge this ASAP so it is merged before we cut the release. 🙇

geriux:
> 👋 @geriux ! We're close to merging the monorepo changes 🎉 , and we will be cutting the release branches as soon as the monorepo changes are merged to gutenberg-mobile. As a result, we may be cutting the 1.31 release branches earlier than normal.
> 
> Because every merge further delays the monorepo work (ends up requiring conflicts to be fixed and other work due to the monorepo feature branch and the squash policy in gutenberg), we want to cut the release branch with no further changes. For that reason, I'll bump this milestone to 1.32. Of course, if this has to make the 1.31 release, go ahead and move it back to the 1.31 milestone, and let me know. 🙇

Sounds good! Thanks for letting me know Matt 🙌

geriux:
> * I've dropped updated styling for the component in the [design docs here](https://zeroheight.com/226850145/p/955bb6-compact-notice). The key thing you'll see is that I'm attempting to replicate the iOS-system materials on iOS, which produces a background blur and translucency. I talked to @wppinar today about potentially using a third party library called [Blur](https://www.npmjs.com/package/@react-native-community/blur) if we're unable to access the system materials directly. Android uses solid colors, which we can fine-tune to use available colors if necessary.

Hey, @iamthomasbishop 👋 I've added the library to test how it'd work, here's a [light mode example](https://drive.google.com/file/d/1VQdMcrFK2qfu_Xt4HsATeQ4IhABoImCj/view?usp=sharing) and a [dark mode one](https://drive.google.com/file/d/13tmtABa7QazyTKF1b7lTkEFSNB55yEdb/view?usp=sharing).

Is that what you had in mind? If it is I can prepare a build for you to test. Also, to double check, this is only for iOS right?

Thanks!

iamthomasbishop:
@geriux That looks _great_, glad to see you were easily able to implement it! I'm not sure what values you're using, but it feels a little thin esp on light mode. I'm looking over the [library readme](https://github.com/react-native-community/react-native-blur) and it looks like we have access to various `blurType` properties — can you tell me which one you're using in the examples above?

iamthomasbishop:
> In this sense Cut functionality is quite useful imo, and it is a quick win for us since we are already implementing Copy/Paste/Duplicate, wdyt?

@pinarol Fair points you made above, let's keep `cut` for now.

geriux:
> @geriux That looks _great_, glad to see you were easily able to implement it! I'm not sure what values you're using, but it feels a little thin esp on light mode. I'm looking over the [library readme](https://github.com/react-native-community/react-native-blur) and it looks like we have access to various `blurType` properties — can you tell me which one you're using in the examples above?

I was using `light` and `dark`, here are other examples:

light (before) | regular (after)
-|-
 | 

dark (before) | prominent (after)
-|-
 | 

What do you think of those? If you want other examples let me know =)

Thanks!

iamthomasbishop:
@geriux I pulled these screenshots into Figma to compare against the system material equivalents I have.

* Dark mode: "Prominent" (RN) looks roughly equivalent to "Regular" (System), which is what I was going for 👍 
* Light mode: "Regular" (RN) looks roughly equivalent to "Ultralight" (System), which is a bit thin 👎 
Can we try "Prominent" in light mode as well and see if that aligns w/ the "Regular" system material in the same way?

<a href="https://user-images.githubusercontent.com/1200432/85459823-3cd46480-b568-11ea-9ea2-5faf9f73626e.png"></a>

Another thing I think I'd like to do is to remove the border-bottom from both iOS and Android — I thought it might be necessary for separation, but it doesn't quite look right esp on iOS. I'll update the component on the Figma + design docs side.

geriux:
>   Can we try "Prominent" in light mode as well and see if that aligns w/ the "Regular" system material in the same way?

Prominent (light mode) | Prominent (dark mode)
-|-
 | 

I think this one does look like your example, what do you think? I also removed the border bottom.

Let me know, thanks!

iamthomasbishop:
That looks great, let's roll with it.  Nice work @geriux !

Side note: I think this will only look _better_ when the Navigation Bars on the apps are all (eventually) changed to use the translucent materials. 😄

PR #2382, opened 2020-06-12
[iOS] Adding support for custom loading views

### LINKED ISSUE #1179: [Try] Improve editor loading experience
## Problem

Sometimes the editor can take some time to load, and when the content is loaded, it enters pretty jarringly. This is especially strong on iOS when posts have a lot of content.

## Context

At the moment, on Android we're showing a spinner, like so:

image

...whereas on iOS, we're showing nothing.

IMG_4018

This is especially jarring on iOS with long posts.

## Proposal

Let's improve the experience of a post as it's loading by either:

* Showing a progress spinner or similar on iOS – like we do on Android. This at least tells me something is in progress.

* Even better, we could show skeleton elements in place of the title and body. That would look something like this:

image

If we were to take a skeleton-loader approach, we should reference other places in the apps that follow this pattern, and ensure we are able to show this elegantly, to limit flashes of content.

etoledom:
This PR adds support for the client to set a custom loading view while the editor is starting up.

Fixes #1179

WPiOS PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/14310

To test:
- Check out WPiOS PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/14310
- Follow the steps described in that PR.

skeleton

PR submission checklist:

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -2,6 +2,7 @@
 ------
 * [***] Adds Copy, Cut, Paste, and Duplicate functionality to blocks
 * [***] Add support for mentions.
+* [*] [iOS] Improved editor loading experience with Ghost Effect.
 
 1.31.0
 ------


mchowning:
👋  @etoledom ! We're close to merging the monorepo changes 🎉 , and we will be cutting the release branches as soon as the monorepo changes are merged to gutenberg-mobile. As a result, we may be cutting the 1.31 release branches earlier than normal.

If these changes can wait until 1.32, please bump this to the 1.32 milestone and wait to merge this until after the release branches are cut. That way we don't add more changes to the 1.31 release than are necessary—this release is already going to have a lot more changes than a normal release because of the monorepo changes. If it is important that this PR make the 1.31 release, however, please try to merge this ASAP so it is merged before we cut the release. 🙇

PR #2383, opened 2020-06-13
Glue code to get content info metrics

hypest:
Expands #2380 to include the Android side of the call to get content information.

To test:
Use the WPAndroid PR to test: https://github.com/wordpress-mobile/WordPress-Android/pull/12191
 
PR submission checklist:

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -1,6 +1,6 @@
 1.31.0
 ------
-* [*] [iOS] Show content structure (word, paragraph, block, characters counts).
+* [*] Show content information (block, word and characters counts).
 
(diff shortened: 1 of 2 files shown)


SergioEstevao:
@hypest any possibility to add a way to check the information directly in the demo app?

hypest:
> any possibility to add a way to check the information directly in the demo app?

We could add a menu item on the RN Debug menu, similar to the "Show html" we have added in the past, but that's not a scalable solution. We needed the "Show html" one for the UI tests so we added it, but this one is probably not as important to have.

Unlike the iOS demo app, the Android demo app doesn't have chrome/toolbar around the editor.

PR #2385, opened 2020-06-15
[FEAT] react navigation

dratwas:
Fixes #
Gutenberg PR: https://github.com/WordPress/gutenberg/pull/23156
WP-ios builds: https://github.com/wordpress-mobile/WordPress-iOS/pull/14347
To test:

- [x] Any sheet that at its natural height is less than 50% ("half" snap point) should display at its natural height, not be forced to 50%.
- [x] Any sheet that has a natural height of more than 50% should by default snap to "half". The user is then able to swipe upward until the sheet reaches its natural height. (example)
- [x] If the sheet's natural height is larger than 100%, the sheet's header becomes fixed and the contents scroll beneath it
- [x] The ability to override the initial starting height for any given sheet. The usual default is 50%, but this would allow setting either a custom starting height — such as a specific snap point or allow the sheet to display at its natural height (for example, an actionSheet equivalent might want to display at natural height)
- [x] The ability to force a sheet to have a static max-height of 50%, which would allow scrolling within the container (like most of our current sheets)
- [x] The ability to include/exclude a scrim (semi-transparent backdrop)

Fixes: 
- [x] On wider viewports, the sheet isn't centered
- [x] Something's going on w/ sheets that have text input controls as they get pushed up so that their contents are sometimes pushed off-screen when in full-height
- [x] When in landscape (phone), tapping on the inserter forces the device into portrait mode
- [x] I think the full-height snap point can be up closer to the bottom of the status bar. For example, on iOS our native preview modal goes up to ~10pt below the status bar
- [x] I'm not sure if it's intentional, but some sheets look like their initial heights are ~25%. Can we follow the behavior I mentioned above

TODO:
- [ ] Update to the monorepo
- [ ] handle correctly the orientation changing
- [ ] The sheet height isn't adjusting when navigating to/from sub-sheets. I imagine this is something we'll want to fine-tune with this new sheet structure in mind. // cc @lukewalczak this is mostly relevant to the work we've done on the color picking UX
- [ ] Any time the user is scrolled to the top of the sheet, swiping down from anywhere on the sheet should either move to a snap point or dismiss the sheet, based on the momentum the gesture creates (a fast swipe downward should dismiss the sheet, whereas a slow and subtle swipe might snap to a point).

PR submission checklist:


iamthomasbishop:
@dratwas This is great to see, and I'm excited to work on it 😃 I've only tested iOS so far (via [this build](https://github.com/wordpress-mobile/WordPress-iOS/pull/14347)) and I have many thoughts, but for the sake of documentation and getting on the same page, I'm going to map out what the behavioral logic should be. Note: I understand that some of this is already in place — for example, I think the snap points need to be fine-tuned, but the capability is there, etc.

## Snap Points

We can think of snap points as magnets — points that the sheet UI will "snap" to based on its position, size, and movement. We have 3 snap points — *full*, *half*, and *closed*. 

- Full: The sheet's maximum size. If the contents scroll beyond this point, the sheet header becomes fixed and content scrolls beneath i
- Half: 50% of the screen height
- Closed

## Threshold

The imaginary dividing line exactly halfway between two snap points. When the user drags the sheet surface and releases, the Y-position in relation to the threshold determines which point to snap to upon release.

## Natural Height

The height of the whole sheet as a single unit, determined by the size of its contents and the surrounding UI.

## Standard Sheet Behavior

- Any sheet that at its natural height is less than 50% ("half" snap point) should display at its natural height, *not be forced to 50%*. ([example](https://cloudup.com/cj-iLrnjNAn))
- Any sheet that has a natural height of more than 50% should by default snap to "half". The user is then able to swipe upward until the sheet reaches its natural height. ([example](https://cloudup.com/cSaDZsg9QXn))
- If the sheet's natural height is larger than 100%, the sheet's header becomes fixed and the contents scroll beneath it ([example](https://cloudup.com/c_DeZUK99aT))
- Any time the user is scrolled to the top of the sheet, swiping down from *anywhere on the sheet* should either move to a snap point or dismiss the sheet, based on the momentum the gesture creates (a fast swipe downward should dismiss the sheet, whereas a slow and subtle swipe might snap to a point).

## Other Capabilities

Most sheets should follow the behavior outlined above, but there are situations where it would be nice to have the following capabilities:

- The ability to override the initial starting height for any given sheet. The usual default is 50%, but this would allow setting either a custom starting height — such as a specific snap point or allow the sheet to display at its natural height (for example, an `actionSheet` equivalent might want to display at natural height)
- The ability to force a sheet to have a static max-height of 50%, which would allow scrolling within the container (like most of our current sheets)
- The ability to include/exclude a scrim (semi-transparent backdrop)

## Feedback on Latest Build

- I'd like to match the behavior mentioned above WRT swiping-to-dismiss from anywhere when I'm scrolled to the top of a sheet (essentially replicate the same behavior as dragging down on the drag handle).
- On wider viewports, the sheet isn't centered ([screenshot](https://cloudup.com/cWD-_ieuguw))
- Something's going on w/ sheets that have text input controls as they get pushed up so that their contents are sometimes pushed off-screen when in full-height ([example](https://cloudup.com/cvTZVRCkz8z))
- When in landscape (phone), tapping on the inserter forces the device into portrait mode
- I think the full-height snap point can be up closer to the bottom of the status bar. For example, on iOS our native preview modal goes up to ~10pt below the status bar ([example](https://cloudup.com/cvcQxxchz2o))
- I'm not sure if it's intentional, but some sheets look like their initial heights are ~25%. Can we follow the behavior I mentioned above
- The sheet height isn't adjusting when navigating to/from sub-sheets. I imagine this is something we'll want to fine-tune with this new sheet structure in mind. // cc @lukewalczak this is mostly relevant to the work we've done on the color picking UX

pinarol:
Hey @dratwas, not sure if this is a known issue but I realized a weird behavior:

- Rotate device to landscape
- Tap (+) add block button in the toolbar

Result:

The screen goes Portrait again with the appearance of BottomSheet

Expected:

The screen should stay in Landscape mode

PR #2394, opened 2020-06-17
Split the bundle creation steps for release and code update.

SergioEstevao:
Change the bundle step to only create updated bundles and not get
updates strings from the web.
This way for code updates only we just need to run `yarn bundle`.
For the releases, we run `yarn release:bundle`.
This will accelerate the builds for WP-Android and for WP-iOS it will
allow deploy and test of PR that need bundle updates quicker.

To test:
 - Run `yarn bundle` and check that only the bundles are updated with local string
 - Run `yarn release:bundle` check that bundles are updated and we also get new strings that come from the web repo.

PR submission checklist:


mchowning:
This generally looks good to me. Nice idea! 👍 

My only concern is that I think it could be a bit misleading now that `yarn bundle` no longer does everything it used to. What would you think about keeping `bundle`'s behavior unchanged (or giving it a new name) and instead adding a new script (i.e., `jsBundle`) for just generating the JS bundles? Basically, I'm concerned that it would be easy to miss that releases need `release:bundle` instead of `bundle`.

SergioEstevao:
> My only concern is that I think it could be a bit misleading now that yarn bundle no longer does everything it used to. What would you think about keeping bundle's behavior unchanged (or giving it a new name) and instead adding a new script (i.e., jsBundle) for just generating the JS bundles? Basically, I'm concerned that it would be easy to miss that releases need release:bundle instead of bundle.

That is a good call, do you think we should update the regular WP-Android build to use `jsBuild`? or that can be problematic when doing production builds for Android?

mchowning:
> That is a good call, do you think we should update the regular WP-Android build to use jsBuild? or that can be problematic when doing production builds for Android?

If we stay with the current no-jitpack build I think it would be a _really_ nice speedup to development to make that change on WPAndroid. It's a slightly tricky change though (since we need to make sure that both CI builds and local release builds generated the full bundle), and it looks like there is a good chance we'll replace our current no-jitpack build with a bintray-based solution, so for the time being I would prefer to just leave things as they are on WPAndroid (assuming we end up not changing the behavior of `bundle`).

mchowning:
👋  @SergioEstevao  ! We're close to merging the monorepo changes 🎉 , and we will be cutting the release branches as soon as the monorepo changes are merged to gutenberg-mobile. As a result, we may be cutting the 1.31 release branches earlier than normal.

It looks like these changes can wait until 1.32, so I'm going to bump this to the 1.32 milestone. Please don't merge this until after the 1.31 release is cut because we don't add more changes to the 1.31 release than are necessary—this release is already going to have a lot more changes than a normal release because of the monorepo changes. 

Of course, if it is important that this PR make the 1.31 release, go ahead and move it back to the 1.31 milestone and let me know so I can try to help get this merged before we cut the release. 🙇

mchowning:
Bumping this to the 1.33 milestone since we're getting ready to cut the 1.32 branches.

With that said, this PR looks good, so it's got the 👍 to merge.

PR #2398, opened 2020-06-17
Create script to automate release process steps.

SergioEstevao:
This PR adds a script to the repo to automate all the steps in the creation of the step process.

The idea is that on the cut release day you go to the `develop` branch and run this command:

`./bin/release_automation.sh`

It assumes the following:
 
 - npm package `json` is installed. This is done by running `yarn install` before
 - the [GitHub CLI is pre-installed](https://cli.github.com/). This needs to be done manually before using the script.

To test:
 - Check the content of the script
 - If you want to give it a go run the following: `./bin/release_automation.sh`
 - Remember to delete the PR just created and the corresponding branch.

PR submission checklist:


mchowning:
Nice job @SergioEstevao ! I like this a lot. I just have a few suggestions:

1. I think that adding some "pre-checks" at the beginning of the script to make sure we're in the right state for running the script and creating a release would be useful:
    1. Check that the required tools (`json` and the github cli) are installed. Failing right away seems a lot better than failing only when we get to the step that needs the tool.
    2. Perhaps we should only allow the script to be run on certain branches (i.e., `develop`, `master`, or `release/*`), and to fail otherwise.
    3. Seems like we should check to make sure the branch is clean to guard against things like:
        1. Having local changes (which would get committed by the script)
        2. Having the wrong gutenberg submodule checked out and or having local changes in the gutenberg submodule (which could affect the bundles)
2. Currently the script continues if any step fails, what would you think about either using `set -e` or adding `|| exit 1` to each command in the script?

I'd be happy to make those changes myself if you agree they're a good idea.

bin/release_automation.sh
```diff
@@ -0,0 +1,24 @@
...
+# Generate updated podspecs
+./bin/generate-podspecs.sh
+# Commit changes
+git commit -a -m "Prepare Release $VERSION_NUMBER"
+# Read PR template
+PR_TEMPLATE=$(cat .github/PULL_REQUEST_TEMPLATE/release_pull_request.md)
+# Replace version number
+PR_BODY=${PR_TEMPLATE//v1.XX.Y/$VERSION_NUMBER}
+# Create PR in GitHub
+gh pr create -t "Release $VERSION_NUMBER" -b "$PR_BODY" -B master -l "release-process"
```
mchowning, line 24: What do you think about adding the `-d` flag so that this PR is a draft? That way we can mark it as ready once we've filled out the description.

SergioEstevao: Good suggestion!

SergioEstevao:
> I'd be happy to add those changes to this PR myself if you agree with them.

Feel free to do them directly on this PR :)

mchowning:
Pushed some updates @SergioEstevao ! Mostly the changes are just around trying to make the script more bullet-proof and giving obvious error messages. I tried to separate out my changes into distinct commits so that it is easy to revert any individual changes if we decide to do anything differently.

SergioEstevao:
@mchowning your changes are great! Thanks for that! 

We probably should wait for monorepo to be merged and them update this to use the new `npm` commands instead of `yarn`.

bin/release_automation.sh
```diff
@@ -0,0 +1,62 @@
...
+# Confirm branch is clean
+[ -z "$(git status --porcelain)" ] || { git status; printf "\nUncommitted changes found. Aborting release script...\n"; exit 1; }
+
+# Ask for new version number
+CURRENT_VERSION_NUMBER=$(./node_modules/.bin/json -f package.json version)
+echo "Current Version Number:$CURRENT_VERSION_NUMBER"
+read -p "Enter the new version number: " VERSION_NUMBER
+
+# Insure javascript dependencies are up-to-date
+yarn install || { echo "Error: 'yarn install' failed"; echo 1; }
```
mchowning, line 32: We should move this `install` after we update the version in the package.json file so that the package-lock.json file also gets the updated version as part of the `install`.

mchowning:
Let's update the Release Checklist template to reflect using this script too.

bin/release_automation.sh
```diff
@@ -0,0 +1,62 @@
...
+yarn bundle || { echo "Error: 'yarn bundle' failed"; exit 1; }
+
+# Make sure podfile is updated
+pod install --project-directory=./ios/ || { echo "Error: pod install failed"; exit 1; }
+
+# Generate updated podspecs
+./bin/generate-podspecs.sh || { echo "Error: generate-podspecs script failed"; exit 1; }
+
+# Commit changes
+git commit -a -m "Prepare Release $VERSION_NUMBER" || { echo "Error: failed to commit changes"; exit 1; }
```
mchowning, line 51: I feel like we may be better off doing separate commits after each step. That way it'll be easier to reverse engineer what happened if there are any problems. wdyt?

bin/release_automation.sh
```diff
@@ -0,0 +1,62 @@
...
+git switch -c "$RELEASE_BRANCH" || { echo "Error: could not create '$RELEASE_BRANCH' branch."; exit 1; }
+
+# Set version number in package.json
+yarn json -I -f package.json -e "this.version='$VERSION_NUMBER'" || { echo "Error: could not update version in package.json"; exit 1; }
+
+# Update the bundles
+yarn bundle || { echo "Error: 'yarn bundle' failed"; exit 1; }
+
+# Make sure podfile is updated
+pod install --project-directory=./ios/ || { echo "Error: pod install failed"; exit 1; }
```
mchowning, line 45: Had some issues with this command even after updating the path: p1592938673087400-slack-mobile-gutenberg-monorepo

Also, this now makes the changes in the `gutenberg/`repo, so we need to update the script to insure that gets committed too. This will essentially force us to move forward with having this script also handling the gutenberg side of the release like you suggested.

bin/release_automation.sh
```diff
@@ -0,0 +1,76 @@
...
+# Confirm branch is clean
+[ -z "$(git status --porcelain)" ] || { git status; printf "\nUncommitted changes found. Aborting release script...\n"; exit 1; }
+
+# Ask for new version number
+CURRENT_VERSION_NUMBER=$(./node_modules/.bin/json -f package.json version)
+echo "Current Version Number:$CURRENT_VERSION_NUMBER"
+read -p "Enter the new version number: " VERSION_NUMBER
+
+# Insure javascript dependencies are up-to-date
+npm install || { echo "Error: 'yarn install' failed"; echo 1; }
```
mchowning, line 32: What do you think we should do about the `package-lock.json` change that results from running `npm install`? In theory it shouldn't happen, but it feels inevitable that it will happen again at some point. I just think we don't want to have the script blindly commit that change.

We could abort the script, but that seems like too much since we wouldn't want this to stop us from cutting a release branch. Feels like it might be better to do something like discard the `package-lock.json` change and, if there were changes, save a message to print at the end of the script.

bin/release_automation.sh
```diff
@@ -0,0 +1,76 @@
...
+# Read PR template
+PR_TEMPLATE_PATH='.github/PULL_REQUEST_TEMPLATE/release_pull_request.md'
+test -f "$PR_TEMPLATE_PATH" || { echo "Error: Could not find PR template at $PR_TEMPLATE_PATH"; exit 1; }
+PR_TEMPLATE=$(cat "$PR_TEMPLATE_PATH")
+
+# Replace version number in PR template
+PR_BODY=${PR_TEMPLATE//v1.XX.Y/$VERSION_NUMBER}
+
+# Create PR in GitHub
+gh pr create -t "Release $VERSION_NUMBER" -b "$PR_BODY" -B master -l "release-process" -d || { echo "Error: Failed to create PR"; exit 1; }
```
mchowning, line 76: Think we should create a gutenberg PR now as well?

mchowning:
This is looking nice @SergioEstevao ! I left a couple of comments and pushed two changes:

1. 4c30d62 adds to the failure message if bundling fails. I had the bundling fail with a very unhelpful error message, and I needed to wipe out my node modules folders to fix it. I thought it would be nice to provide some guidance there.
2. c8c8969 adds a check for the '--testing' flag, which would skip the branch checks (is it a proper branch and is the branch clean) and also avoid creating a PR. I then changed my mind because the --testing flag's removal of the check for changes could result in changes unintentionally getting committed during testing (tweak the script, run a test run, and the tweaks to the script would be committed during the test run). Therefore, in 818a71d I removed the testing flag and instead prompted the user to confirm they wanted to create PRs before we run that part of the script.

Let me know what you think.

bin/release_automation.sh
mchowning: I'm not seeing this issue anymore. 👍

bin/release_automation.sh
SergioEstevao: Are the changes happing inside the GB-Mobile repo or the GB repo?

SergioEstevao:
@mchowning thanks for the changes, really like the prompt checking if we want to proceed when the branch is not the expected.

bin/release_automation.sh
SergioEstevao: Maybe not because of the issues you raised about changes that can show up in the gutenberg folder after the `npm install` ?

bin/release_automation.sh
mchowning: Anytime I run `npm install` I get this change to the `gutenberg-mobile/package-lock.json` file:

image

My concern is not with this particular change, but with the possibility of the script committing other not-so-innocent changes to that file without us realizing it.

bin/release_automation.sh
mchowning: The changes I saw were actually in the `gutenberg-mobile/` folder, which we are already making a PR for. I don't feel strongly about automating the gutenberg PR too actually. It was more just a thought.

bin/release_automation.sh
SergioEstevao: Humm never saw that happening, could it be something specific in your machine setup?
We shouldn't see any changes after 'npm install' but if we noticed them  we probably should be able to revert them with a `git revert` ?

mchowning: > Humm never saw that happening, could it be something specific in your machine setup?

Quite possibly, but a number of people have seen it, so it's definitely not just me: p1592831812099400-slack-mobile-gutenberg-monorepo

> We shouldn't see any changes after 'npm install'...

`npm install` will update the [version line](https://github.com/wordpress-mobile/gutenberg-mobile/blob/develop/package-lock.json#L3) of the `package-lock.json` file if it sees the version has changed, so we want to commit that change at some point in the script. Otherwise, every dev will get a diff on the `package-lock.json` file anytime they run `npm install`.

> ...but if we noticed them we probably should be able to revert them with a git revert ?

Do you mean after the script has finished, or as part of the script itself. My concern with doing it after the script has finished is that we wouldn't actually notice the change because I doubt anyone really does a thorough code review on the release PRs since they're huge even ignoring all the bundle changes. Doing it as part of the script itself seems like the better idea (i.e., only commit the update to the version line in the `package-lock.json`).

Actually, maybe the better approach is to do the `npm install` before we set the version, like we do now (and **not** move the `npm install` to after the version is set [like I suggested earlier](https://github.com/wordpress-mobile/gutenberg-mobile/pull/2398#discussion_r444437234)). After `npm install` we could just reset the `package-lock.json` file. Then when we set the version, we could manually set it for both the `package.json` and `package-lock.json` files.

mchowning:
Noticed while running this that in the gb repo the `packages/react-native-editor/ios/Podfile.lock` file had an updated hash for `react-native-video` committed as part of the "Update react-native-editor version to..." commit. Just noting that here because it surprised me a bit.

bin/release_automation.sh
SergioEstevao: We solved this by using `npm ci`

Issue #2399, opened 2020-06-17
Link Colors in DarkMode appear to be dark than intended


chipsnyder:
**Describe the bug**
It appears that the link colors in Dark Mode are darker than the expected value of `blue-30` (`#5198d9`)

**To Reproduce**
1. Set your device to use Dark Mode
2. Create a post with some links

**Expected behavior**
Link color in dark mode should be `blue-30` (`#5198d9`)

**Screenshots**
![image](https://user-images.githubusercontent.com/3384451/84932878-f59b2f00-b0a2-11ea-843a-4e0560804d00.png)

**Smartphone (please complete the following information):**
 - Device: Android
 - OS: [e.g. iOS 8.1,  Android X]

**Additional context**
This was discovered as part of the testing for WPAndroid 15.1

chipsnyder:
This as merged to 1.33

PR #2404, opened 2020-06-18
Correct Color for RichText links in Dark Mode

chipsnyder:
**Fixes** https://github.com/wordpress-mobile/gutenberg-mobile/issues/2399
**Related PRs:**
`gutenberg` https://github.com/WordPress/gutenberg/pull/23289
`AztecEditor-iOS` https://github.com/wordpress-mobile/AztecEditor-iOS/pull/1295
~`WordPress-Android` https://github.com/wordpress-mobile/WordPress-Android/pull/12226~
~`WordPress-iOS` https://github.com/wordpress-mobile/WordPress-iOS/pull/14351~

## Solution Description
As it turns out this error seems to be one that will randomly happen. Android uses the following process to set the text and style on RichText components:
- The [text is set](https://github.com/wordpress-mobile/gutenberg-mobile/blob/develop/react-native-aztec/android/src/main/java/org/wordpress/mobile/ReactNativeAztec/ReactAztecManager.java#L206) 
- This then updates the text [on AztexText](https://github.com/wordpress-mobile/gutenberg-mobile/blob/74f2bb2c58b1fcd9a10125b78673318d5ef8d0a4/react-native-aztec/android/src/main/java/org/wordpress/mobile/ReactNativeAztec/ReactAztecManager.java#L221) 
- The links are [then styled](https://github.com/wordpress-mobile/AztecEditor-Android/blob/develop/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt#L1191) as part of that call
- Which is then dependent on `linkFormatter` to [determine the span and set the style](https://github.com/wordpress-mobile/AztecEditor-Android/blob/a1386beed8d8215fe8ab79c501c1ec237ae8a785/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt#L1374) This needs both the text and for `linkFormatter` to be set correctly.

**For the link color**
- When the `linkTextColor` [is set](https://github.com/wordpress-mobile/gutenberg-mobile/blob/74f2bb2c58b1fcd9a10125b78673318d5ef8d0a4/react-native-aztec/android/src/main/java/org/wordpress/mobile/ReactNativeAztec/ReactAztecManager.java#L391) this updates `linkFormatter` with the proper attributes. 

**The Issue**
It appears that the `@ReactProp`'s don't enforce an order of operations[[1]](https://stackoverflow.com/questions/50135103/is-there-a-way-to-enforce-order-of-execution-of-props-on-a-native-component#). So if `linkTextColor` is processed before `text` everything will work fine. If not then the app will fall back to the default styles.

**The Fix**
To fix this I simply moved the `linkTextColor` to be part of the `text` attribute that is set so that we can enforce the order of operations. 

This is actually only needed for Android however to prevent diverging implementations I made the same changes to iOS. 

## To Test
- Set your device too dark mode
- Create a post/page or open one that has links. 

Color should be using `blue-30` (`#5198d9`)

| Before | After |
| --- | --- |
|<kbr></kbr> | <kbr></kbr> |

## PR submission checklist:


geriux:
Hey @chipsnyder 👋 thanks for the very well detailed PR! 

> To fix this I simply moved the linkTextColor to be part of the text attribute that is set so that we can enforce the order of operations.

I know this fixes the issue but I'm not 100% sure about passing it along with the [text prop](https://github.com/WordPress/gutenberg/pull/23289/files#diff-dda065334d6da1eb4149443ac3439f7eR787), but of course, we have the issue with React Native [not allowing](https://stackoverflow.com/questions/50135103/is-there-a-way-to-enforce-order-of-execution-of-props-on-a-native-component#) a way to enforce the order. My knowledge of Aztec is very limited so I was wondering what @mchowning thinks about it. Matt, whenever you have time would you mind checking this approach, please? I just don't wanna overlook something by taking this path 😅

react-native-aztec/android/src/main/java/org/wordpress/mobile/ReactNativeAztec/ReactAztecManager.java
```diff
@@ -202,6 +203,11 @@ public Map getExportedCustomDirectEventTypeConstants() {
 
     @ReactProp(name = "text")
     public void setText(ReactAztecText view, ReadableMap inputMap) {
+        if (inputMap.hasKey(LINK_TEXT_COLOR_KEY)) {
+            int color = Color.parseColor(inputMap.getString(LINK_TEXT_COLOR_KEY));
+            setLinkTextColor(view, color);
```
mchowning, line 208: This means that we will be constructing new `LinkFormatter` and `LinkStyle` objects on text change, right? Seems like that is not ideal from a performance perspective, although any impact would probably be negligible. Still, maybe it's worth only updating the link formatter when the link color has changed. I do not feel strongly about this.

react-native-aztec/android/src/main/java/org/wordpress/mobile/ReactNativeAztec/ReactAztecManager.java
```diff
@@ -386,7 +392,6 @@ public void setTextAlign(ReactAztecText view, @Nullable String textAlign) {
         }
     }
 
-    @ReactProp(name = "linkTextColor", customType = "Color")
     public void setLinkTextColor(ReactAztecText view, @Nullable Integer color) {
```
mchowning, line 395: Could we make this private now?

mchowning:
> Matt, whenever you have time would you mind checking this approach, please?

I think this approach makes sense. Seems like a good way to insure a proper order. I noted one small [performance-related concern](https://github.com/wordpress-mobile/gutenberg-mobile/pull/2404#discussion_r443758157), but I don't think it's a big deal.

An alternative approach that might work would be to reset the text any time the `linkTextColor` prop comes down, but I wouldn't be surprised if that had some unwanted side effects.

I haven't built this and tested it out, but looking at the code it looks good to me. 👍 

> Hey @chipsnyder 👋 thanks for the very well detailed PR!

+1000 to that! 🙂

react-native-aztec/android/src/main/java/org/wordpress/mobile/ReactNativeAztec/ReactAztecManager.java
chipsnyder: We could make this private now. 👍

react-native-aztec/android/src/main/java/org/wordpress/mobile/ReactNativeAztec/ReactAztecManager.java
chipsnyder: > Still, maybe it's worth only updating the link formatter when the link color has changed. 

So are you thinking something along the lines of this? 
```suggestion
        int color = Color.parseColor(inputMap.getString(LINK_TEXT_COLOR_KEY));
        if (view.linkFormatter.color != color) {
            setLinkTextColor(view, color);
```

*Note* I didn't test this suggestion out yet.

mchowning: Yes, that's pretty much what I was imagining.

chipsnyder:
> An alternative approach that might work would be to reset the text any time the linkTextColor prop comes down, but I wouldn't be surprised if that had some unwanted side effects.

I considered this a bit; however, other items (`selection` and `eventCount`) are passed along with the text so I was afraid of the possible side effects.

react-native-aztec/android/src/main/java/org/wordpress/mobile/ReactNativeAztec/ReactAztecManager.java
chipsnyder: I'm good with that 👍 I'll work on making that change.

chipsnyder:
@mchowning @geriux  This is good to go again. I made the requested changes and migrated the code to Gutenberg as part of updating it with the monorepo changes.

chipsnyder:
Also, I closed the main App PRs b/c of the monorepo work. So if it's not too much of a hassle can we test with the demo apps?

geriux:
> Also, I closed the main App PRs b/c of the monorepo work. So if it's not too much of a hassle can we test with the demo apps?

Hey Chip 👋  The code looks good! I wanted to test on the demo apps but I don't know how to reproduce the bug in develop for example, so then I can switch to this branch and make sure it's fixed. Are there any steps to test a before and after in the demo app? I remember I was able to do it back then but with the main WordPress Android App. Or do you want me to just test that the changes didn't break anything?

Let me know, thanks!

chipsnyder:
Hey @geriux,
Thanks for coming back around to retest this. This might be a bit difficult to reliably recreate due to the nature of the issue being a random order (I'm assuming based on iterating over the keys of a dictionary). I was able to reproduce it in the demo app but I think I just got lucky.

I think for retesting making sure nothing breaks and trying multiple times to refresh the app should help. 
Since the Monorepo stuff has merged to the Apps now though I'll go ahead and enable those builds and make new testing builds that way you can at least retest where you know you saw the issue.

WPAndroid: 
WPiOS:

geriux:
Hey @chipsnyder 👋 thanks for bringing back the testing PRs =)

On Android is working great 🎉 

On iOS I noticed the colors are different after these changes:

Before | After
-|-
 | 

This is not expected, right? Let me know, thanks!

chipsnyder:
> On iOS I noticed the colors are different after these changes:

Hey @geriux, good catch. Were you running the metro server when you saw this? I totally missed this difference when I checked the builds but it looks like I just failed to rerun the bundle after merging monorepo

geriux:
> Were you running the metro server when you saw this? 

I used the IPA from the WordPress iOS PR without metro. 

 > (I'll update the bundle and regenerate the builds)

Thank you!

chipsnyder:
@geriux those updated builds are available now ([Android](https://59473-9306568-gh.circle-artifacts.com/0/Artifacts/WordPress-pr-12226-build-59473.apk), [iOS](https://appcenter-filemanagement-distrib2ede6f06e.azureedge.net/dbccb9d8-66e2-465c-8845-e363620387fa/WordPress%20Alpha.ipa?sv=2019-02-02&sr=c&sig=T4YnFP%2BYnee%2BEDS1olUP8hFl2%2B4mOkviPjgcMEmcrDk%3D&se=2020-07-01T18%3A42%3A21Z&sp=r)) Sorry for that mistake 😅

chipsnyder:
Reopening to keep GB mobile and iOS in sync on the Pod Changes

chipsnyder:
Closing this again. With https://github.com/wordpress-mobile/WordPress-iOS/pull/14444 in place we can just follow the normal release process

PR #2422, opened 2020-06-24
Add Release notes doc

enejb:
This PR add a release note doc so that we can explain why and how to add a release note to the `RELEASE-NOTE.txt` file. 

Does the new doc make sense what else should we add to it? 
Anything needs to be removed? 

PR submission checklist:

### DIFF
--- a/docs/Release-notes.md
+++ b/docs/Release-notes.md
@@ -33,4 +33,4 @@ For example:
 Release notes also go through editorial before they are shared with the users. So, while adding release notes, we should 
 try to err on the side of more details so that they changes can be better summarized. 
 
-Release notes are currently translated into different languages before the are seeing by users. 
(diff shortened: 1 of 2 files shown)


docs/Release-notes.md
```diff
@@ -0,0 +1,36 @@
...
+
+For example: 
+
+    * [*] [Android] Disable ripple effect for Slider control
+    
+## Additional info
+Release notes also go through editorial before they are shared with the users. So, while adding release notes, we should 
+try to err on the side of more details so that they changes can be better summarized. 
+
+Release notes are currently translated into different languages before the are seeing by users.
```
mchowning, line 36: Looks like there's a small typo here: "before the are seeing by users."

enejb:
> WDYT?
Yes I think that is a great idea. Just pushed a change to update that.

PR #2424, opened 2020-06-24
Issue 331/mention support on @ keypress

SergioEstevao:
Fixes #331

Related PRs:

 - Gutenberg PR: https://github.com/WordPress/gutenberg/pull/22119
 - WP-iOS: https://github.com/wordpress-mobile/WordPress-iOS/pull/14088
 - WP-Android: https://github.com/wordpress-mobile/WordPress-Android/pull/12115

This PR changes the AztecView component to be able to intercept any keycode that is defined. It then configures uses this functionality to intercept the `@` keypress and start the mention UI.

To test:
 - Start the demo app
 - Add a Paragraph block
 - Tap the @ on the keyboard
 - Check if a mention to `matt` is added
 - Check if the mention is only inserted if the @ is inserted with a space before or at the start of the paragraph.

PR submission checklist:

- [x] I have considered adding unit tests where possible.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #2467, opened 2020-07-08
Release 1.32.0

SergioEstevao:
Release for Gutenberg Mobile 1.32.0

## Related PRs

- Gutenberg: https://github.com/WordPress/gutenberg/pull/23786
- WPAndroid: https://github.com/wordpress-mobile/WordPress-Android/pull/12384
- WPiOS: https://github.com/wordpress-mobile/WordPress-iOS/pull/14437

- Aztec-iOS: https://github.com/wordpress-mobile/AztecEditor-iOS/pull/
- Aztec-Android: https://github.com/wordpress-mobile/AztecEditor-Android/pull

## Extra PRs that Landed After the Release Was Cut

NA

## Changes

* [***] Adds Copy, Cut, Paste, and Duplicate functionality to blocks: https://github.com/wordpress-mobile/gutenberg-mobile/pull/2378
* [***] Add support for mentions. (Under a feature flag on the main apps): https://github.com/wordpress-mobile/gutenberg-mobile/pull/2424
* [***] Users can now individually edit unsupported blocks found in posts or pages.
* [*] [iOS] Improved editor loading experience with Ghost Effect: https://github.com/wordpress-mobile/gutenberg-mobile/pull/2382

## Test plan

- Use the main WP apps to test the changes above. 
- Check WPAndroid and WPiOS PRs if there are specific tests to run.
- Smoke test the main WP apps for [general writing flow](https://github.com/wordpress-mobile/test-cases/tree/master/test-cases/gutenberg/writing-flow).

## Release Submission Checklist

- [x] Release number was bumped
- [x] Aztec dependencies are pointing to a stable release
  - iOS: 'grep WordPressAztec-iOS RNTAztecView.podspec'
  - Android: 'grep aztecVersion react-native-aztec/android/build.gradle'
- [x] Gutenberg 'Podfile' and 'Podfile.lock' inside './ios/' are updated to the release number
- [x] Bundle package of the release is updated 
- [ ] Check if `RELEASE-NOTES.txt` is updated with all the changes that made it to the release


RELEASE-NOTES.txt
```diff
@@ -1,7 +1,14 @@
+1.32.0
+------
+* [***] Adds Copy, Cut, Paste, and Duplicate functionality to blocks
+* [***] Add support for mentions.
```
pinarol, line 4: > Add support for mentions

Should we put this in the release notes if this will only be available internally? I think we wouldn't want this to occur in App Store, right?

mchowning: Good point, we definitely don't want this in the release notes for the 15.3 release notes. Although, at the gutenberg-mobile level, the mentions capability is added in this release--we're just turning that capability off in release builds for the main apps. So I'm not really sure what the right approach is here. If the gutenberg-mobile release-notes were really like a library's release notes, I would say it should be included, but we've always kind of used these release notes to just note things that will be user-facing in the main apps, and mentions will not be user-facing in the main apps release. Basically, I don't feel strongly either way. Any thoughts @SergioEstevao ?

pinarol: Wait, I thought for a minute I am commenting under a WPiOS PR, my apologies for the confusion. I think it is fine to keep it in gb-mobile level.

hypest:
👋 looks like we didn't update the `package-lock.json` and [the version mentioned there](https://github.com/wordpress-mobile/gutenberg-mobile/blob/d6c6172219230043d3b2abea1eea946c4496bb57/package-lock.json#L3) is still the previous (1.31.1) one. It's not easy to manually merge package-lock.json conflicts so, may I suggest adding a step to the release process to ensure that the lock file has the proper version?

Without updating the lock file, `npm install` ends up introducing changes to it.

mchowning:
> looks like we didn't update the package-lock.json and the version mentioned there is still the previous (1.31.1) one. It's not easy to manually merge package-lock.json conflicts so, may I suggest adding a step to the release process to ensure that the lock file has the proper version?

This is addressed in https://github.com/wordpress-mobile/gutenberg-mobile/pull/2490.

PR #2469, opened 2020-07-08
Improve release script

SergioEstevao:
Some improvements to the release automation script and pull request templates.

Check the commit messages for the changes for more detail.

To test:
  - Run the script for a simulated release to see if all is working correctly.

PR submission checklist:

- [x] I have considered if this change warrants user-facing release notes [more info](docs/Release-notes.md) and have added them to `[RELEASE-NOTES.txt](RELEASE-NOTES.txt)` if necessary.

### DIFF
--- a/bin/release_automation.sh
+++ b/bin/release_automation.sh
@@ -48,6 +48,9 @@ npx json -I -f package.json -e "this.version='$VERSION_NUMBER'" || { echo "Error
 # Set version number in react-native-editor package.json
 npx json -I -f gutenberg/packages/react-native-editor/package.json -e "this.version='$VERSION_NUMBER'" || { echo "Error: could not update version in react-native-editor package.json"; exit 1; }
 
+# Make sure podfile is updated
+npm run core preios
+
 # Commit react-native-editor package changes
 cd gutenberg
 git commit -a -m "Update react-native-editor version to: $VERSION_NUMBER" || { echo "Error: failed to commit changes"; exit 1; }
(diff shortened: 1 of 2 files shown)


bin/release_automation.sh
```diff
@@ -48,6 +48,9 @@ npx json -I -f package.json -e "this.version='$VERSION_NUMBER'" || { echo "Error
 # Set version number in react-native-editor package.json
 npx json -I -f gutenberg/packages/react-native-editor/package.json -e "this.version='$VERSION_NUMBER'" || { echo "Error: could not update version in react-native-editor package.json"; exit 1; }
 
+# Make sure podfile is updated
+npm run core preios
```
mchowning, line 52: It doesn't feel right to me to run this after we bump the version numbers, but before we commit those bumped to the version numbers. Perhaps we should move this either (a) before the version numbers are bumped or (b) after the version numbers are committed. Then we could also add a step that checks if there were any podfile changes and, if there were, commits them separately?

SergioEstevao: @mchowning Do you feel we should always commit separately the changes independently of doing a) or b)?

mchowning: I think it would be nice to have a separate commit for each distinct stage in the release process in order to make it easier to look back and see exactly what the release script did at each step (i.e., when something goes wrong). I don't think this is a critical change that needs to block this PR though, so I'll go ahead and approve. There are a couple of release script updates I'm planning to work on today, so I'll maybe see about separating out the commits in that PR and then you can give it a look and see what you think.

PR #2479, opened 2020-07-10
Remove duplicate RNTAztecView podspec

ceyhun:
We currently have two `RNTAztecView.podspec` files. One in this repo and another in `gutenberg` repo under `packages/react-native-aztec`. They may get out of sync easily so will try to remove one of them.

Fixes #

To test:

PR submission checklist:

- [ ] I have considered if this change warrants user-facing release notes [more info](docs/Release-notes.md) and have added them to `[RELEASE-NOTES.txt](RELEASE-NOTES.txt)` if necessary.


SergioEstevao:
@ceyhun I tried to use this branch hash on WP-iOS main app, and the cocoapods was unable to fetch the dependency. 
Were you able to use it?

ceyhun:
@SergioEstevao In WPiOS's Podfile `pod 'RNTAztecView', options` is not needed after this, could it be that? https://github.com/wordpress-mobile/WordPress-iOS/blob/0f2e0706e5326deb46b1f3ec982da2d55c361da8/Podfile#L86

I'm also not sure if this is the right approach as it could easily be broken in the future.

cc @chipsnyder

SergioEstevao:
After removing the entry you suggested it the 'rake dependencies' step worked but then if you try to compile the project RNTAztec is showing has not available.

Any other ideas?

ceyhun:
I think in this case RNTAztecView is not a pod anymore but becomes a part of Gutenberg (as a subspec). I will try another idea that still keeps `RNTAztecView.podspec` in gutenberg-mobile bu it just loads the `RNTAztecView.podspec` from gutenberg. This way we shouldn't need any changes to WPiOS.

mchowning:
Would it be worthwhile to make a similar change for [Gutenberg.podspec](https://github.com/wordpress-mobile/gutenberg-mobile/blob/develop/Gutenberg.podspec)?

mchowning:
:wave: @ceyhun ! Just wanted to check in on the status of this PR. Can we move this forward?

ceyhun:
👋 @mchowning I'm not really happy with how this ended up. Right now, if a new attribute is added to the `gutenberg/packages/react-native-aztec/RNTAztecView.podspec` it will be missed in `gutenberg-mobile/RNTAztecView.podspec`. 

Maybe there's a better way (using a loop) to assign **all** the attributes of one to another instead of the very repetitive way I used here. However I'm not really that familiar with Ruby to make it work this way.

hypest:
Is this PR still relevant? If yes, perhaps reaching out for help from Platform9 folks could be useful. If not relevant anymore, can we close it?

ceyhun:
This would still be nice to have, I will reach out for help 👍

ceyhun:
Closing this, will try maybe another time after collecting some more thoughts.

PR #2484, opened 2020-07-14
Link picker via fetch

mkevins:
### Related PRs

`gutenberg`: https://github.com/WordPress/gutenberg/pull/23922
`WordPress-Android`: https://github.com/wordpress-mobile/WordPress-Android/pull/12438
`WordPress-iOS`: https://github.com/wordpress-mobile/WordPress-iOS/pull/14537

### Details

This is a PR to produce a test apk for design review of the link picker feature.

To test:

#### Pre-testing steps:

1. Install <s>[the apk](https://61223-9306568-gh.circle-artifacts.com/0/Artifacts/WordPress-pr-12438-build-61223.apk) [the apk](https://63686-9306568-gh.circle-artifacts.com/0/Artifacts/WordPress-pr-12438-build-63686.apk)</s> [the apk](https://64073-9306568-gh.circle-artifacts.com/0/Artifacts/WordPress-pr-12438-build-64073.apk) from the WordPress-Android PR.
2. Switch to a site which has some published posts and pages (site can be private)

#### Adding a link
1. Create a new post
2. Tap the first paragraph block
3. Tap the link button to add a link (expect the fields to be blank)
4. Tap the URL field (expect the modal to switch to full screen - picker / omni search mode)
5. Type some letters and wait for results (expect results to be displayed below)
6. Pick one of the results by tapping (expect previous screen to come back, populated with the item you picked)
7. Dismiss the modal (via back, or swipe down) (expect new link to appear)

#### Editing a link
1. Place the caret in an existing link (expect the link button to become highlighted)
2. Tap the link button to edit the link (expect the fields to be populated)
3. Tap the URL field (expect the modal to switch to full screen - picker / omni search mode)
4. Delete the existing text, type some letters and wait for results (expect results to be displayed below)
5. Pick one of the results by tapping (expect previous screen to come back, populated with the URL of the item you picked, but the link text should not be overwritten unless it was empty before)

#### Note some known issues in the current implementation
* Horizontal padding of the results header cell (the omni input) is more narrow than the results
* The bottom-most results are partially blocked by the keyboard
* On web, we wait until at least two characters are in the query before fetching results. Currently, I've left this requirement out (we fetch on every keystroke except an empty input), to make it easier to test (more results on smaller sites). This can be changed later, before merging.

#### URL "Input" (button)

On the main screen, I've made the button for editing the URL appear as a text input. When tapped, it opens the fullscreen modal for editing / searching the URL. When populated, that field appears with an ellipsizemode = "middle". Please let me know if this is acceptable:

Main screen (empty) | Picker screen (empty) | Main screen (populated)
-|-|-
main-screen-empty|picker-screen-empty|main-screen-filled

#### Direct entry suggestion

On web, there is a suggestion in the results for entering the URL "as is", however this appears at the bottom of the list, so it is hidden from view until you scroll all the way to the bottom. This is not ideal, IMO, and in our implementation, this would be worse, since every time we reach the bottom, we'd have new results fetched, pushing it further down. I've instead added this entry at the top. Let me know if this behavior is ok:

Mobile results screen | Web results screen
-|-
picker-screen-t| web-picker Direct entry not visible :point_up_2: web-direct-input-bottom Direct entry at the bottom of the results :point_up_2: 

#### Icons vs. pills

The web uses pills (on the right) to distinguish post "subtypes" such as "post" and "page", and also to distinguish other kinds of links (such as mailto: and tel: links). In earlier designs for mobile, we have icons (on the left), so I've added icons, but I'm not sure these are ideal for distinguishing between the types. I think we need to decide either on a set of icons that are easily distinguishable (especially between page / post) or maybe we can also go with "pills" like web. Here's how it looks now on web vs mobile:

Web "pills" | Mobile icons
-|-
web-subtype-pills|

Here's how it looks with neither icons, nor pills (mailto: link direct entry suggestion):

If we decide to use icons, it might be good to specify icons for the following types of links (which are supported by the current implementation):

* `post`: A published post
* `page`: A published page
* `mailto`: A directly entered "mailto:" link
* `tel`: A directly entered "tel:" link
* `internal`: A directly entered anchor link ("#")

PR submission checklist:

- [ ] I have considered if this change warrants user-facing release notes [more info](docs/Release-notes.md) and have added them to `[RELEASE-NOTES.txt](RELEASE-NOTES.txt)` if necessary.


iamthomasbishop:
@mkevins @sylvesterwillmot I had a chance to give the latest build a proper spin, and I noted a few additional things that I think could be ironed out. I also made some sketches (attached below) in an effort to visualize my suggestions.

## Labeling

- Can we match the placeholder labels on the table row placeholder and the placeholder on the expanded sheet? (`Search or type URL`)
- I believe the label on the expanded sheet should match the table row title (`Link to`)

## Initial Sheet

- Considering the transition from the initial sheet to the expanded one is utilizing the sheet-to-subsheet transition when you tap the `Link to` row (which might soon become a more prominent page-push transition), I think we should display a chevron/disclosure indicator on the right side of the row (see [Buttons block color settings example](https://cloudup.com/cHwwLS6UOxK))

## Full Sheet

- It looks like a few things are out of alignment, but I think this would likely be fixed by simply applying 16px of margin on the left and right of the row (eg. the icon and "URL" label, and placeholder label are all ~16 off). ([Example](https://cloudup.com/c9QHG94nnUY))
- Rather than display the `URL` title, would it make more sense to left-align the input placeholder/label in its place (at the 72 keyline, with the icon left of it)?  This would give the input more breathing room and also feel a little more balanced, I think. We could even remove the icon so that the text input would span full-width (minus 16px margins on each side, of course)
- Would it make sense to show a checkmark/"done" icon on the top right of the sheet header?
- The spinner that shows while searching is okay, but I think it'd be preferable to use the native Android spinner if possible, and perhaps bump it down slightly (or centered vertically in the space between the last result and the keyboard)

Here is what all of the above would look like (yellow lines == matching labels, pink == tap gesture):

<a href="https://user-images.githubusercontent.com/1200432/88728379-38671280-d0f7-11ea-8c2b-e8721a283ad0.jpeg"></a>

## Other Questions

- Will this also be applied to other instances of links, such as the `Button link URL` on Button, `Link to` on Image, etc?

mkevins:
Thanks for the awesome design feedback Thomas!

> ## Labeling
> * Can we match the placeholder labels on the table row placeholder and the placeholder on the expanded sheet? (`Search or type URL`)
> * I believe the label on the expanded sheet should match the table row title (`Link to`)

:heavy_check_mark: Done

> ## Initial Sheet
> * Considering the transition from the initial sheet to the expanded one is utilizing the sheet-to-subsheet transition when you tap the `Link to` row (which might soon become a more prominent page-push transition), I think we should display a chevron/disclosure indicator on the right side of the row (see [Buttons block color settings example](https://cloudup.com/cHwwLS6UOxK))

:heavy_check_mark: Done
 
> ## Full Sheet
> * It looks like a few things are out of alignment, but I think this would likely be fixed by simply applying 16px of margin on the left and right of the row (eg. the icon and "URL" label, and placeholder label are all ~16 off). ([Example](https://cloudup.com/c9QHG94nnUY))

:heavy_check_mark: Done

> * Rather than display the `URL` title, would it make more sense to left-align the input placeholder/label in its place (at the 72 keyline, with the icon left of it)?  This would give the input more breathing room and also feel a little more balanced, I think. We could even remove the icon so that the text input would span full-width (minus 16px margins on each side, of course)

I agree this makes more sense. I have updated without the icon. Let me know what you think, or if I should add it back. Thanks for the great suggestions!

> * Would it make sense to show a checkmark/"done" icon on the top right of the sheet header?

I think it does make sense. Following the example from the custom color picker, I added the checkmark for Android, and the word "Apply" for iOS. Do you think that makes sense?

> * The spinner that shows while searching is okay, but I think it'd be preferable to use the native Android spinner if possible, and perhaps bump it down slightly (or centered vertically in the space between the last result and the keyboard)

:heavy_check_mark: Done - I added a 150 px container for the spinner (centered vertically within), and it stops displaying the extra space when we know we've reached the end of results. Let me know what you think of it, or if it needs adjusting the height.
 

> ## Other Questions
> * Will this also be applied to other instances of links, such as the `Button link URL` on Button, `Link to` on Image, etc?

I believe this is the plan for further iterations. At the moment, there is still some ongoing work with the bottomsheet navigation, so I think it'd make sense for that to land before implementing in other bottomsheets, but I've had the other use-cases in mind while forming the interfaces, so it should be fairly straightforward to reuse the underlying components.

Thanks again for all your help @iamthomasbishop and @SylvesterWilmott !

mkevins:
Updated builds:
Android: https://63686-9306568-gh.circle-artifacts.com/0/Artifacts/WordPress-pr-12438-build-63686.apk
iOS: tbd

iamthomasbishop:
@mkevins thanks for the update and for taking care of all of those changes so quickly! Testing out the latest Android and now iOS builds, and it's getting really close! I have some updated suggestions, and again I'd be curious to hear what @sylvesterwilmott thinks — and as always, feel free to point out if you disagree with anything!

> I believe this is the plan for further iterations

Sounds good!

## Expanded sheet

- **Input Alignment:** I'm still not 100% sure about the alignment on the search input. I'd suggest we remove the margins on the left/right of the input so the input itself (and as a result, its border) is flush against the left/right edges of the screen. I would also keep the icon (either the previous "link" icon or a "search" icon) and align it to the 16px keyline — then align the label to the 72px keyline — 56px on iOS. *Note: It might make sense to apply a slightly different input style for iOS, but I think this should work for both as a start.*
- **Height on iOS:** is it possible to also use a not-quite-full screen height on the sheet on iOS like we're doing on Android? What I mean is that the top of the sheet would be flush agains the bottom of the Status Bar. Having the UI go full-screen feels odd because you can't swipe from the left to go back as you'd typically do in iOS when you see a `< Back` button. If we *have to* keep it full screen, maybe we can do one of the following to remove that friction:
    - Find a way to allow the natural swipe gesture to close the sheet, or
    - Change the `< Back` button to `Close`

## Search results

- Very much a nitpick and not a blocker, but Is it possible to apply a unique subtitle (rather than "add this link") on the subtitles for `mailto`, `tel`? Eg. "add this link" makes sense for a standard URL, but for `mailto` it would make more sense to be "add email" or "email address", and for `tel` something like "add telephone number" or "telephone number" might make more sense, imo
- Similar to the last bullet point, is it possible to show unique icons for `tel` and `mailto` (whereas right now, we're showing the "globe" icon — this isn't a blocker, but would be a nice refinement)
- Should we change the color of the subtitle on each search result row to a lighter/secondary color and potentially make the label slightly smaller?

## After selection

- After an item is chosen from the expanded sheet, can we possibly display a cleaner label in the small sheet? Specifically for a `post` or `page` result, I think I'd rather see the post/page title. I think showing the raw URL for a `url`, `tel` , and `mailto` is fine.

That's all I've got for now! It's getting super close, and very usable already, so well done @mkevins !

mkevins:
Hi @iamthomasbishop , thanks for taking another look :smile: 

> ## Expanded sheet
> * **Input Alignment:** I'm still not 100% sure about the alignment on the search input. I'd suggest we remove the margins on the left/right of the input so the input itself (and as a result, its border) is flush against the left/right edges of the screen. I would also keep the icon (either the previous "link" icon or a "search" icon) and align it to the 16px keyline — then align the label to the 72px keyline — 56px on iOS. _Note: It might make sense to apply a slightly different input style for iOS, but I think this should work for both as a start._

:heavy_check_mark: Done

> * **Height on iOS:** is it possible to also use a not-quite-full screen height on the sheet on iOS like we're doing on Android? What I mean is that the top of the sheet would be flush agains the bottom of the Status Bar. Having the UI go full-screen feels odd because you can't swipe from the left to go back as you'd typically do in iOS when you see a `< Back` button. If we _have to_ keep it full screen, maybe we can do one of the following to remove that friction:
>   
>   * Find a way to allow the natural swipe gesture to close the sheet, or

This one might be a little tricky with the current bottomsheet, but might be possible with the navigation in the bottomsheet that is currently underway. cc: @dratwas :wave: :smiley: any thoughts on this :point_up_2: ?

>   * Change the `< Back` button to `Close`

This is achievable, but to clarify, do you mean for it to close the bottomsheet altogether, or just modify the button text (and still return to the smaller link editing sheet)?

> ## Search results
> * Very much a nitpick and not a blocker, but Is it possible to apply a unique subtitle (rather than "add this link") on the subtitles for `mailto`, `tel`? Eg. "add this link" makes sense for a standard URL, but for `mailto` it would make more sense to be "add email" or "email address", and for `tel` something like "add telephone number" or "telephone number" might make more sense, imo

Thanks for the suggestion. I've updated the summaries as follows: for email: "Add this email link", for telephone: "Add this telephone link". I can also remove the word "link" from those, but it felt weird without it, imo.

> * Similar to the last bullet point, is it possible to show unique icons for `tel` and `mailto` (whereas right now, we're showing the "globe" icon — this isn't a blocker, but would be a nice refinement)

Certainly! I looked around a bit in the Icons package, and also Gridicons, but I didn't find anything obvious for these. Once we have some Icons to use, it'll be easy to add. Btw, currently, it should not show the globe icon for those kinds of links.. did you encounter that? If so, perhaps the input value was missing the `:`? i.e. "tel" will not be detected as a telephone link, but "tel:" will.


> ## After selection
> * After an item is chosen from the expanded sheet, can we possibly display a cleaner label in the small sheet? Specifically for a `post` or `page` result, I think I'd rather see the post/page title. I think showing the raw URL for a `url`, `tel` , and `mailto` is fine.

This one is a little tricky. The current behavior is that the post or page title will populate the "Link text" field upon returning, as long as it is empty (i.e. it will not overwrite that value if it is already populated). But, the values there directly correspond with the underlying rich text. We could hypothetically hide the full url, displaying only the the title instead, but I think we'll still need to display the link text. Some problems I can foresee are:

* Link text and "URL" or "link" (or whatever we call that hybrid field) might be "duplicate" in that case
* The post / page title doesn't uniquely identify the content (i.e. two can have the same title) - I'm not sure if this would be a major concern in practice though
* After dismissing the modal, the post / page title is not persisted with the link, so tapping the link button again would show different information in the hybrid field

Are any of the remaining items considered "blockers" for this iteration? Thanks again for all your helpful guidance! :smile: :+1:

mkevins:
Updated builds:
`WordPress-Android`: https://64073-9306568-gh.circle-artifacts.com/0/Artifacts/WordPress-pr-12438-build-64073.apk
`WordPress-iOS`: tbd

SylvesterWilmott:
Thanks for reviewing this @iamthomasbishop, all the changes look great, thanks @mkevins 

>>Change the < Back button to Close

>This is achievable, but to clarify, do you mean for it to close the bottomsheet altogether, or just modify the button text (and still return to the smaller link editing sheet)?

Agree with @iamthomasbishop here. The expanded sheet is not too dissimilar to a full screen dialog and as such should use a "Close" or "X". This would not close the entire sheet, the behaviour would remain the same as it currently is.

A couple of small things I noticed:

When tapping a search item, the touch feedback doesn't extend to the edge of the sheet as we would expect. See below for illustration

Screenshot 2020-07-31 at 17 04 27

Is there a subtitle/body text style we can use for the URL when searching? Currently, the title and the URL are both the same text size, by using a smaller size for the URL we would be able to fit more characters on the line.

mkevins:
Thanks for reviewing @SylvesterWilmott 

> Agree with @iamthomasbishop here. The expanded sheet is not too dissimilar to a full screen dialog and as such should use a "Close" or "X". This would not close the entire sheet, the behaviour would remain the same as it currently is.

:heavy_check_mark: Done.

> A couple of small things I noticed:
> 
> When tapping a search item, the touch feedback doesn't extend to the edge of the sheet as we would expect. See below for illustration
> 
> Screenshot 2020-07-31 at 17 04 27

Good observation. Since this is consistent with the current behavior for all bottomsheet cells, perhaps this can be addressed as a standalone issue? For example, here is the image size setting for image block:



> Is there a subtitle/body text style we can use for the URL in the search results rows? Currently, the title and the URL are both the same text size, by using a smaller size for the URL we would be able to fit more characters on the line.

:heavy_check_mark: Done.

Thanks again for taking the time to review and provide guidance. I've updated the test builds:

`WordPress-Android`: https://64269-9306568-gh.circle-artifacts.com/0/Artifacts/WordPress-pr-12438-build-64269.apk
`WordPress-iOS`: tbd

Would you mind making another pass, and letting me know if there are any blockers?

iamthomasbishop:
Agreed w/ all of @SylvesterWilmott's feedback.

> Good observation. Since this is consistent with the current behavior for all bottomsheet cells, perhaps this can be addressed as a standalone issue? For example, here is the image size setting for image block:

This is a known issue, discovered during some recent bottom sheets cell work -- I thought there was a ticket for it, but can't find one. Would you mind creating one @mkevins ?

I'm going to test the new builds and will drop subsequent feedback if I see anything else of note. Thank you!

iamthomasbishop:
@mkevins After poking around the test builds (android + iOS), I think we're in pretty good shape. @SylvesterWilmott's recommendations look great as implemented.

* I would still like to figure out the issue I mentioned regarding the height of the expanded sheet (showing it as "almost full-screen") before shipping, although it is a small thing. I believe we should be able to do that w/ the updates to bottom sheets w/ navigation mechanism, but I'll let @dratwas confirm.
* On Android, I can only drag the expanded sheet downward if I drag _directly on the drag handle_. This might be resolved with the arrival of the new sheets work, but I'm not positive. If it remains an issue, we might want to either hide the drag handle or make this sheet into a full-screen modal. Thoughts, @SylvesterWilmott ?
* I should have mentioned this in my last review, but would it be possible to add a subtle "clear"/"cancel" icon button on the right side of the text input (on the expanded sheet)? Something like [this](https://cloudup.com/c1eGWB8f8eu), perhaps? I think this icon is from the Material icon library, but we might also have it in the core icon library. I can provide an icon if needed.

dratwas:
Hey, since @mkevins is AFK i migrated this PR to the new navigation.

@iamthomasbishop 
> I would still like to figure out the issue I mentioned regarding the height of the expanded sheet (showing it as "almost full-screen") before shipping, although it is a small thing. I believe we should be able to do that w/ the updates to bottom sheets w/ navigation mechanism, but I'll let @dratwas confirm.

Do you mean something like this? If yes then it is easy to implement (already did it locally :) )
fullheightiosmargin

>On Android, I can only drag the expanded sheet downward if I drag directly on the drag handle. This might be resolved with the arrival of the new sheets work, but I'm not positive. If it remains an issue, we might want to either hide the drag handle or make this sheet into a full-screen modal. 

Unfortunately, this issue is not solved after adding navigation. We still can not drag inside the list to hide the bottom sheet on Android.

> I should have mentioned this in my last review, but would it be possible to add a subtle "clear"/"cancel" icon button on the right side of the text input (on the expanded sheet)? Something like this, perhaps? I think this icon is from the Material icon library, but we might also have it in the core icon library. I can provide an icon if needed.

I'm on it :) I just need to know what should happen after a click on that button. I bet that the value of text input should be set to empty string and the `clear` button should disappear. Am I right?
## iOS
Light Mode | Dark Mode
--- | ---
Simulator Screen Shot - iPhone 11 - 2020-08-27 at 17 52 17 | Simulator Screen Shot - iPhone 11 - 2020-08-27 at 17 50 36

## Android
Light Mode | Dark Mode
--- | ---
Screenshot_1598546038 | Screenshot_1598545919

iamthomasbishop:
@dratwas Thanks for the update!

> Do you mean something like this? If yes then it is easy to implement (already did it locally :) )

Precisely! Thank you!

> Unfortunately, this issue is not solved after adding navigation. We still can not drag inside the list to hide the bottom sheet on Android.

Bummer 😞 We should probably figure out a work-around on Android in the meantime (until we can fix this limitation). A few thoughts:

* Apply a max-height to the "expanded" link sheet, and make it scrollable
  * Pro: the sheet header would be more reachable
  * Con: the scroll container would be shorter
* (Android-only) Remove the drag handle (in this case, we might want to make the sheet 100% full-screen in this case, so it doesn't appear as if you can dismiss the sheet)

Just to confirm, can we drag-to-dismiss on iOS?

> I bet that the value of text input should be set to empty string and the clear button should disappear. Am I right?

Exactly. The text input would clear, the placeholder text would re-appear ("Search or type URL") with the cursor at the front, and the `clear` button would hide (until the user adds a character to the input, of course).

Just saw the screenshots you attached and the positioning looks good, but I'm not a huge fan of that specific icon -- do we have another option in our icons library, or do you want me to provide a custom one?

dratwas:
@iamthomasbishop 
>Just to confirm, can we drag-to-dismiss on iOS?

Yes :)

> Exactly. The text input would clear, the placeholder text would re-appear ("Search or type URL") with the cursor at the front, and the clear button would hide (until the user adds a character to the input, of course).

 👍 

> (Android-only) Remove the drag handle (in this case, we might want to make the sheet 100% full-screen in this case, so it doesn't appear as if you can dismiss the sheet)

I did a quick test. TBH I like this option more than setting max-height for both platforms :)



> Just saw the screenshots you attached and the positioning looks good, but I'm not a huge fan of that specific icon -- do we have another option in our icons library, or do you want me to provide a custom one?

This is the one I found in our icons - https://github.com/WordPress/gutenberg/blob/master/packages/icons/src/library/close-circle-filled.js

I couldn't find another so please provide the custom one :)

iamthomasbishop:
@dratwas I will circle back later this afternoon or first thing Monday for a more thorough review, but I made an icon that should work for both iOS and Android, because the one in our library feels awkward, imo.

You can grab the .svg from the [design docs](https://zeroheight.com/226850145/p/345d49-icons/b/78279c/i/18387247) -- the one in question is named `icon-cancel-circle-filled`.

dratwas:
Hey, @iamthomasbishop thanks for the icon!

>You can grab the .svg from the design docs -- the one in question is named icon-cancel-circle-filled.

I tried but I can download only the .`png`. It seems like there is no .`svg` asset.

There is Android test build https://69784-9306568-gh.circle-artifacts.com/0/Artifacts/WordPress-pr-12841-build-69784.apk

iamthomasbishop:
> I tried but I can download only the .png. It seems like there is no .svg asset.

@dratwas Ahh, my bad, I accidentally chose the default (.png) when adding the slice. An .svg should be available now!

dratwas:
Hey @iamthomasbishop i created new builds with the icon and fixes for the pagination.


[iOS](https://appcenter-filemanagement-distrib1ede6f06e.azureedge.net/6e40b764-5add-43b7-b3ce-124ebd8f3227/WordPress%20Alpha.ipa?sv=2019-02-02&sr=c&sig=e%2FC7Hj8Dp8S0pPGgVLUM9IfSz8WtMa1HxXUplGJEYt8%3D&se=2020-09-02T12%3A47%3A13Z&sp=r), [Android](https://70153-9306568-gh.circle-artifacts.com/0/Artifacts/WordPress-pr-12841-build-70153.apk)

Could you please make a review? :)

iamthomasbishop:
@dratwas Looking really good overall! A couple of tiny things:

## Android-only

* I'm seeing a weird issue when trying to submit the search query, where the keyboard hides but the sheet stays in place. ([video](https://cloudup.com/cRTNSKgZdfs)). 
* Also in this video ⬆️ , you'll see when the link is applied on the block, the string is doubled somehow ([screenshot](https://cloudup.com/cov2cck-6Kt)). This happens every time you apply a link from the sheet.
* We can probably use a slightly dimmer gray for the `cancel` icon (meaning lighter in light mode, darker in dark mode). ([example](https://cloudup.com/cCP501PkGVU) using current `$gray` (left) and one shade lighter, `$gray-lighten-10` (right))
* Now that we are allowing the sheet to go up to the status bar and also not allowing the user to swipe down to dismiss, is it possible to remove the border-radii from the top left/right corners of the sheet so that it's more obviously "docked" to the top?

## iOS-only

* Would it make sense to try a more "standard" input style on iOS? For example, either the standard white background w/ a gray fill around it, or a gray-filled-background input style ([examples](https://cloudup.com/csBqzxbFxh8)). Do you have any thoughts or opinions on this @SylvesterWilmott ? It would def match our search input in the apps a bit more closely. If this is something we'd like to do, I can provide the exact colors.

dratwas:
Thanks for your feedback @iamthomasbishop! I made a list below to track all of your suggestions. Could you please take a look and answer to the questions i asked in section `Need details`? :)

## Need details
- **Colors of the icon**
>We can probably use a slightly dimmer gray for the cancel icon (meaning lighter in light mode, darker in dark mode). (example using current $gray (left) and one shade lighter, $gray-lighten-10 (right))

1. Is it `Android-only` thing? 
2. Could you please provide the color for the dark mode as well? :) 
 
- **Issue with doubled link string**
>Also in this video ⬆️ , you'll see when the link is applied on the block, the string is doubled somehow (screenshot). This happens every time you apply a link from the sheet.

I can not see it on the video because the modal is open from the very beginning but - how you open the link settings?
Do you select the `Link` word and the open link settings or you open link settings, type title, select link and close the bottom-sheet? Asking because i can not reproduce it as well.

I tried something like this:
linkpickerandroid

## In progress

- **Issue with empty bottom space**
>I'm seeing a weird issue when trying to submit the search query, where the keyboard hides but the sheet stays in place. (video).

Hmmm i haven't seen this before but finally, i reproduced that after a lot of switches between screens.

## Done

- Now that we are allowing the sheet to go up to the status bar and also not allowing the user to swipe down to dismiss, is it possible to remove the border-radii from the top left/right corners of the sheet so that it's more obviously "docked" to the top?

## Need Decision
- Would it make sense to try a more "standard" input style on iOS? For example, either the standard white background w/ a gray fill around it, or a gray-filled-background input style (examples). Do you have any thoughts or opinions on this @SylvesterWilmott ? It would def match our search input in the apps a bit more closely. If this is something we'd like to do, I can provide the exact colors.

iamthomasbishop:
Thank you @dratwas!

> Is it Android-only thing?

The reason I put it under Android-only is that I think we're going to for-sure keep the current (in this build) input style, so I provided the color values w/ that in mind.

> Could you please provide the color for the dark mode as well?

Sure thing! I took a look at our current color variables, and I think we can use the `$dark-quaternary` fill for the "clear" icon. Also, I noticed that on iOS dark mode, we aren't using the updated neutral gray on the icons (we're using one of the "old" blue-ish grays -- might currently be using either the standard light mode `$toolbar-button` or `$gray` variables, from the looks of it). Can we change that to use the standard gray, the same one we're using on the block toolbar icons (not exactly sure which color variable we are currently using)?

(Note: there are other places that we will also need to update, but while we're here, let's fix these instances on the sheet 😄 )

> I can not see it on the video because the modal is open from the very beginning but - how you open the link settings?

Sorry the video might not have been super clear. The issue is happening at the same time as the keyboard-position issue -- when I have a link already. So to clarify, it's happening when I have an existing link, tap the "link to" cell to open the full sheet (to edit), then hit `✔️` on the keyboard ("done"). You'll see in the video that when I do this, the keyboard hides, the sheet transitions back to the top-level sheet, but its position isn't adjusted as the keyboard is dismissed and the string is doubled.

dratwas:
>Sure thing! I took a look at our current color variables, and I think we can use the $dark-quaternary fill for the "clear" icon. Also, I noticed that on iOS dark mode, we aren't using the updated neutral gray on the icons (we're using one of the "old" blue-ish grays -- might currently be using either the standard light mode $toolbar-button or $gray variables, from the looks of it). Can we change that to use the standard gray, the same one we're using on the block toolbar icons (not exactly sure which color variable we are currently using)?

Yeah, i was surprised that the color is passed by the `color` prop but should be passed by the `fill` prop. I thought that it is intentional and didn't fix that :) Basically, it affects all Cells

dratwas:
Hey @iamthomasbishop :)
I created new builds with fixes except one with "double link string" since i couldn't reproduce that one.

[Android](https://71186-9306568-gh.circle-artifacts.com/0/Artifacts/WordPress-pr-12841-build-71186.apk)

[iOS](https://appcenter-filemanagement-distrib2ede6f06e.azureedge.net/46205b11-a7e2-42bb-940f-d3bf763a7482/WordPress%20Alpha.ipa?sv=2019-02-02&sr=c&sig=%2BpvbkrJJZjvAEPxZmxFxEkgIcSvD7VOtQAA95eazoKI%3D&se=2020-09-09T19%3A32%3A47Z&sp=r)

Could you please review it one more time? :)

iamthomasbishop:
@dratwas Got a chance to review again, thanks for the updated builds! Here are my notes:

* I played around w/ the `clear` button a bit more in the flow and realized we might want to change the `close`/`⬅️` action (on the sheet header) to `cancel`/`X`. For example, if I open an existing link and clear the URL, I expect that "close" button to basically discard the changes (and "Apply" to, well, apply). If there are any objections or complications that might come with this change, I'm all ears. // cc @SylvesterWilmott what do you think?

* Only colors thing that I'm seeing is on Android: it looks like the `clear` icon is using a "stronger" gray than desired (iOS looks "right" in dark/light modes)

* On Android, when the sheet is retracted (from full height to short), the transition/cross-fade looks a tiny bit choppy. I'm not sure if it's a performance issue or something else, bc I'm not seeing it on iOS. It's pretty subtle. ([video](https://cloudup.com/cH1SKxPKfue) of choppiness -- apologies in advance, it's a bit hard to see bc the video is low-resolution 😬 )

dratwas:
Hey, @iamthomasbishop thanks for this review :)

> played around w/ the clear button a bit more in the flow and realized we might want to change the close/⬅️ action (on the sheet header) to cancel/X. For example, if I open an existing link and clear the URL, I expect that "close" button to basically discard the changes (and "Apply" to, well, apply). If there are any objections or complications that might come with this change, I'm all ears. // cc @SylvesterWilmott what do you think?

Done

> Only colors thing that I'm seeing is on Android: it looks like the clear icon is using a "stronger" gray than desired (iOS looks "right" in dark/light modes)

Hmmm, I use the same color for both platforms (these ones you provided in previous comments )
Light mode : $gray-lighten-10
Dark mode: $dark-quaternary

Should I change them for Android? If yes, please provide correct values :)

>On Android, when the sheet is retracted (from full height to short), the transition/cross-fade looks a tiny bit choppy. I'm not sure if it's a performance issue or something else, bc I'm not seeing it on iOS. It's pretty subtle. (video of choppiness -- apologies in advance, it's a bit hard to see bc the video is low-resolution 😬 )

On it, however, it doesn't happen every time, just sometimes, and is quite hard to find the root cause.

iamthomasbishop:
> Hmmm, I use the same color for both platforms (these ones you provided in previous comments )

Interesting, here's what I'm seeing in the latest test builds:

<a href="https://user-images.githubusercontent.com/1200432/92771267-58c5f600-f360-11ea-93c2-183bdd633e70.png"></a>

> On it, however, it doesn't happen every time, just sometimes, and is quite hard to find the root cause.

Yea, it's hard to pin down. It's a very subtle thing, if we need to create a separate ticket for it that's fine.

dratwas:
@iamthomasbishop
>Interesting, here's what I'm seeing in the latest test builds:

Hmmm, weird... I'm pretty sure that the corners are not rounded when the bottom-sheet is in full height mode on Android latest build. 

Latest build should looks like this:



>Yea, it's hard to pin down. It's a very subtle thing, if we need to create a separate ticket for it that's fine.

Will spend some time on it tomorrow and if I fail I will create an issue

iamthomasbishop:
@dratwas ahh, I wonder if the build I tested on Android wasn't the latest. I'll try to update again and double check.

iamthomasbishop:
@dratwas Sure enough, I wasn't on the latest build. Colors look good on Android as well 👍 

I'm still seeing the same actions in the sheet header -- to clarify, here's what I was trying to explain in my previous comment:

<a href="https://user-images.githubusercontent.com/1200432/92795178-cb8d9c00-f375-11ea-8ff6-8701a67c879e.png"></a>

iamthomasbishop:
@dratwas Are you still unable to reproduce the "double-string" issue? I'm able to reproduce it with no issue. Steps:

* Add text to a paragraph block
* Select text, tap `link` button in block toolbar
* Add URL (wordpress.org, for example), tap enter
* See link text is doubled (with only the second string as a link)

[Video](https://cloudup.com/c4Azpq8q61E)

dratwas:
Hey @iamthomasbishop thanks for the exact steps to reproduce. I fixed it and generated a new build with the latest changes.

[Android](https://72102-9306568-gh.circle-artifacts.com/0/Artifacts/WordPress-pr-12841-build-72102.apk)
[iOS - PR](https://github.com/wordpress-mobile/WordPress-iOS/pull/14809)

iamthomasbishop:
@dratwas @lukewalczak Just took the latest builds of iOS and Android for a spin, and I think it's looking good-to-go. I will spend some time next week documenting (in the form of a new GH ticket) the things we would like to attack in the next iteration — most importantly extending this UI/flow to other places that have link settings (image block, buttons, etc). Thanks y'all for all of the great work on this, stoked we're going to ship it!

geriux (approved):
LGTM! Approved via https://github.com/WordPress/gutenberg/pull/23922#pullrequestreview-489578268

PR #2517, opened 2020-07-28
Update release script to push gutenberg release branch

mchowning:
Updating the release script to push the gutenberg release branch. I think we should do this because we update the gutenberg branch with the version number and the updated submodule hash is committed to gb-mobile release branch, but the referenced commit won't be pushed to the gb repository unless/until it is manually pushed. Until that is done no one else would be able to build the release branch.

To test:
Verify that the release script will push the gb branch.

PR submission checklist:


mchowning:
:wave: @SergioEstevao ! I'm inclined to update this PR to go further and just make a gutenberg PR as well. WDYT?

PR #2551, opened 2020-08-14
Fix for the Classic Block "null" inside the Unsupported Block editor (from v1.32.0)

etoledom:
This PR includes a fix for the Unsupported Block editor to better support the Classic block.

This PR is based on v1.32.0 so to land it as a hotfix onto WPiOS v15.3. There's an [accompanying PR](https://github.com/wordpress-mobile/gutenberg-mobile/pull/2550) based on v1.33.1 to land the same fix on WPiAndroid v15.4 (notice the difference in the WP apps and versions).

Gutenberg PR: https://github.com/WordPress/gutenberg/pull/24571

To test:

Follow the steps outlined in the GB PR above

PR submission checklist:


hypest:
✅ Verified that we need to base off 1.32.0 to land a hotfix to WPiOS v15.3
✅ Verified using GB hash b9ce09
⏳ Verify fix is working on the iOS demo app

package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "gutenberg-mobile",
-  "version": "1.31.1",
+  "version": "1.32.0",
```
hypest, line 3: We'll need to make sure this change propagates through these hotfix PRs.

ceyhun:
Will target WPiOS 15.4 with 1.33.2 and 15.5 with 1.34.1 instead.

PR #2554, opened 2020-08-17
Update release automation script to create WPiOS and WPAndroid PRs

ceyhun:
Here are the changes made to the script

- Replaced "npx json" usage with "jq" for faster execution
- Added string formatting, execute and abort utils
- The script can now create WPiOS and WPAndroid PRs

To test:
1. Run the script and use an unusually high version number 
1. Check if Gutenberg, GB-Mobile, WPiOS and WPAndroid PRs are created as drafts
1. Check if PRs have correct descriptions and changes
1. Close the draft PRs

PR submission checklist:


bin/release_automation.sh
```diff
@@ -58,73 +58,76 @@ fi
...
 GB_RELEASE_BRANCH="rnmobile/release_$VERSION_NUMBER"
+ohai "Create Git branch '$GB_RELEASE_BRANCH' in gutenberg."
 cd gutenberg
-git switch -c "$GB_RELEASE_BRANCH" || { echo "Error: could not create '$GB_RELEASE_BRANCH' branch."; exit 1; }
+execute "git" "switch" "-c" "$GB_RELEASE_BRANCH" 
 cd ..
 
 # Set version numbers
-for file in 'package.json' 'package-lock.json' 'gutenberg/packages/react-native-editor/package.json'; do
-    npx json -I -f "$file" -e "this.version='$VERSION_NUMBER'" || { echo "Error: could not update version in ${file}"; exit 1; }
```
mchowning, line 71: We may be able to remove [the json dependency from the package.json file](https://github.com/wordpress-mobile/gutenberg-mobile/blob/develop/package.json#L42) now.

bin/release_automation.sh
```diff
@@ -34,19 +34,19 @@ if [[ ! "$CURRENT_BRANCH" =~ "^develop$|^main$|^release/.*" ]]; then
 fi
 
 # Confirm branch is clean
-[[ -z "$(git status --porcelain)" ]] || { git status; printf "\nUncommitted changes found. Aborting release script...\n"; exit 1; }
+[[ -z "$(git status --porcelain)" ]] || { git status; abort "Uncommitted changes found. Aborting release script..."; }
 
 # Ask for new version number
-CURRENT_VERSION_NUMBER=$(./node_modules/.bin/json -f package.json version)
+CURRENT_VERSION_NUMBER=$(jq '.version' package.json --raw-output)
```
mchowning, line 40: Might be worth adding an early check that jq is available so that if it's not available the script can abort with a clear error message.

mchowning:
Really like these improvements! 😄 

The only thought I really have is that it might be nice to add some documentation to the new utility functions that are being added. At least for me, it was far from obvious what, for example, "execute" was used for. I'm far from a bash expert, so some of that is on me, but I feel like it might be nice to just add a bit of context for those methods.

bin/release_automation.sh
ceyhun: That's great!

bin/release_automation.sh
ceyhun: Makes sense 👍

ceyhun:
> The only thought I really have is that it might be nice to add some documentation to the new utility functions that are being added. At least for me, it was far from obvious what, for example, "execute" was used for. I'm far from a bash expert, so some of that is on me, but I feel like it might be nice to just add a bit of context for those methods.

Sure! I'll add some "docs" as comments to top of the functions 👍

PR #2578, opened 2020-08-27
Rename main to trunk in the release script

maxme:
We're renaming the `main` branch to `trunk`. I updated the release script and replaced `main` with`trunk`.


cameronvoell (approved):
Looks good, thanks for updating here!

PR #2582, opened 2020-08-28
Add/experimental block grid block

enejb:
Block Experiments PR: https://github.com/Automattic/block-experiments/pull/136 (Updates to the layout grid blocks edit) 

Here we are adding the `block-experiments` [repo](https://github.com/Automattic/block-experiments/) that includes the layout-grid block. In order to add support for that block in the gutenberg mobile editor.

This PR just add the basic read only block and doesn't include support for all the different block attributes that the layout block current supports. 

This PR add the following:
1. The block-experiments repo as a submodule. 
2. The layout-grid block as a block so that it can be added to the block editor.

**How this PR has been tested.** 
It has been tested on the demo app only.

To test:
Add the new repo:
```
git submodule init
git submodule update
```

Start the demo app
```
npm run start:reset
npm run core ios
```

Now you should be able to add the layout grid block to the demo app. 
See

After it is inserted it looks like this. 

PR submission checklist:


src/index.js
```diff
@@ -23,6 +23,11 @@ addAction( 'native.render', 'gutenberg-mobile', ( props ) => {
 	setupJetpackEditor(
 		props.jetpackState || { blogId: 1, isJetpackActive: true }
 	);
+	if ( __DEV__ ) {
+		// I am not sure this is the right way to do this. 
+		// But if I do an import instead we end up with an error.
+		require( './block-experiments-setup' );
```
dcalhoun: I believe relying upon `require` may be necessary as [Metro does not support dynamic `import`s](https://github.com/facebook/metro/issues/52) currently.

AmandaRiu (approved):
Tested on iOS and Android and all works as expected! 👍 :shipit:

dcalhoun (approved):
LGTM. Tested on iOS and Android. Worked as expected.

Issue #2444, opened 2020-06-29
[Modal Layout Picker] Filter layouts based on supported blocks. 


chipsnyder:
### Is your feature request related to a problem? Please describe.
This task is part of the effort to rework the Starter Page Template Picker for new pages. This part of the initiative will focus on modifying the API call [/wpcom/v2/sites/{site}/block-layouts](https://github.com/wordpress-mobile/gutenberg-mobile/issues/2413) to include the [supported_blocks](https://github.com/wordpress-mobile/gutenberg-mobile/issues/2415)
 query parameter which will be used to only include layouts that have met a server defined threshold of supported blocks.

### Goals

+ Create a mechanism that takes exposes the same array of [coreBlocks](https://github.com/WordPress/gutenberg/blob/9b6fee52af1066294e4212c0affd87dfb197bc9b/packages/block-library/src/index.native.js#L64-L114) defined in Gutenberg and exposes those to the App.
    + For now, this can be as simple as duplicating the list in [coreBlocks](https://github.com/WordPress/gutenberg/blob/9b6fee52af1066294e4212c0affd87dfb197bc9b/packages/block-library/src/index.native.js#L64-L114) to the apps.
    + Ideally this would be a mechanism that would be shared by the block registration as well as the list passed to the apps but that can be separated out to a future enhancement.
+ Use this array to modify the API call [/wpcom/v2/sites/{site}/block-layouts](https://github.com/wordpress-mobile/gutenberg-mobile/issues/2413) with the query paramert `supported_blocks` that was developed as part of https://github.com/wordpress-mobile/gutenberg-mobile/issues/2415

### Additional context

<details>
<summary>Blueprint Flow - iOS</summary>
<img src="https://user-images.githubusercontent.com/3384451/85329020-bce2c780-b49f-11ea-88f3-5db30c13af03.jpg" />
</details>

<details>
<summary>Blueprint Flow - Android</summary>
<img src="https://user-images.githubusercontent.com/3384451/85889853-40f7c080-b7ba-11ea-819b-21f075afc772.jpg" />
</details>

PR #2589, opened 2020-09-01
Expose list of supported blocks to be consumed by the mobile apps. 

chipsnyder:
**Fixes** https://github.com/wordpress-mobile/gutenberg-mobile/issues/2444
**Related PRs:**
`gutenberg:` https://github.com/WordPress/gutenberg/pull/24976
`WordPress-iOS` https://github.com/wordpress-mobile/WordPress-iOS/pull/14778

## To test:
- This isn't called by the demo apps at this time but can be tested as part of the testing steps in: https://github.com/wordpress-mobile/WordPress-iOS/pull/14778

## PR submission checklist:


chipsnyder:
cc @antonis

src/block-support/supported-blocks.json
```diff
@@ -0,0 +1,36 @@
...
+		"core/cover",
+		"core/social-link",
+		"core/social-links",
+		"jetpack/contact-info",
+		"jetpack/email",
+		"jetpack/phone",
+		"jetpack/address"
+	],
+	"devOnly": [ "core/code" ],
+	"iOSOnly": [ "core/pullquote" ],
```
mchowning: I think `core/pullquote` should be in "common" now: https://github.com/WordPress/gutenberg/pull/24921

src/block-support/supported-blocks.json
```diff
@@ -0,0 +1,36 @@
...
+		"core/freeform",
+		"core/button",
+		"core/spacer",
+		"core/shortcode",
+		"core/buttons",
+		"core/latest-posts",
+		"core/verse",
+		"core/cover",
+		"core/social-link",
+		"core/social-links",
```
mchowning, line 27: Instead of having to duplicate all the "core" blocks both here and in gutenberg's `packages/block-library/src/index.native.js` file, could we just have the gutenberg JS expose the list of "core" blocks to the native code via the JS bridge? Then we would only need to worry about non-core blocks in this file and avoid having to keep the list of supported blocks in sync. 

Having said that, I realize that this is a temporary solution that we intend to refactor later, so I'm open to leaving this as-is.

mchowning (commented):
This is looking pretty good, just left a couple of comments.

src/block-support/supported-blocks.json
chipsnyder: I agree I would very much like to achieve that goal however I'm a bit stuck on how we might be able to do it. The list here contains the block name but the [registration](https://github.com/WordPress/gutenberg/blob/master/packages/block-library/src/index.native.js#L194) ends up taking the block its self and [using](https://github.com/WordPress/gutenberg/blob/5dc434cd951c51d7965179fb84719ebb0ee219b8/packages/block-library/src/index.native.js#L129) the `metadata`, `settings`, and `name`. This is probably just a result of me still catching up with JS but I'm currently not sure what the best way would be to convert the block names to what we need. 

We would need to do something like what's below. I'm just not entirely sure how to handle the pseudo-code part

```
const registerBlock = ( blockName ) => {
	if ( ! blockName ) {
		return;
	}

        //  pseudo-code 
        const block =  import * from `./${blockName.split("/").slice(-1)}`
        // end pseudo-code 

	const { metadata, settings, name } = block;
	registerBlockType( name, {
		...metadata,
		...settings,
	} );
};
```

mchowning: > I'm currently not sure what the best way would be to convert the block names to what we need.

Instead of converting the names into blocks, could we have gutenberg convert the blocks into names for us [along these lines](https://github.com/WordPress/gutenberg/commit/9da350162f8c9f6fea602cee10f34dd36c8ea1fc) and give the native side access to those names?

chipsnyder: That would be a good way to go as well. The only hardship here is we need this before the editor starts (or maybe the editor never even startes). So ideally we would have a script that could start up a headless editor export the results of that `console.log` to a file then we'd be able to consolidate in some way that way. I hit a bit of a dead end on that though 😞

mchowning: > The only hardship here is we need this before the editor starts (or maybe the editor never even starts).

Ah yes, I didn't think about that. Good point!

mchowning: I'm just brainstorming at this point, but would it work to just hardcode the block names to a variable in the same file in gutenberg? That way we would just be keeping two variables in the same file in sync instead of keeping two files across two repos in sync?

To be clear, I do **_not_** consider this a blocking issue for this PR. Just talking through different ideas.

chipsnyder: >To be clear, I do not consider this a blocking issue for this PR. Just talking through different ideas.

No problem at all I appreciate the extra thoughts!

>just hardcode the block names to a variable in the same file in gutenberg?

This could be viable. We would want Gutenberg and Gutenberg-mobile to have a list of supported blocks. `Gutenberg` would serve up the list that handles all of the `core` blocks and `Gutenberg-mobile` would handle the subset of `jetpack` blocks that it supports. With that pattern we might want to say something along the lines of "plugins should provide a list of blocks supported on mobile" and we consolodate those in `Gutenberg-mobile`

chipsnyder: > So ideally we would have a script that could start up a headless editor export the results of that console.log to a file then we'd be able to consolidate in some way that way. I hit a bit of a dead end on that though 😞

I think there is a way to do this with Webpack but I'm not 100% sure yet

mchowning (approved):
Tested and this works well! 👍 

Obviously we want to improve the duplication, but this works perfectly well and I think it's a great step in the right direction. Nice job @chipsnyder !

PR #2610, opened 2020-09-11
Overwrite Missing block UI Text

etoledom:
WPiOS PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/14880
WPAndroid PR: https://github.com/wordpress-mobile/WordPress-Android/pull/13011
gutenberg PR: https://github.com/WordPress/gutenberg/pull/25238

This PR implements the changes from https://github.com/WordPress/gutenberg/pull/25238 and overwrites the default Missing block alert text with text that references WordPress.com and Jetpack.

This also allows us to change these texts dynamically depending on the UBE-related capabilities sent from the native apps, after calling `updateCapabilities()`.

To test:
Please check https://github.com/wordpress-mobile/WordPress-iOS/pull/14880

PR submission checklist:


marecar3:
Hey @etoledom 👋 
I have marked this PR as ready for review as it seems that we are ready? Thanks.

cc @Tug @guarani @mchowning

src/ui-strings.js
```diff
@@ -0,0 +1,52 @@
...
+	} else {
+		unsupportedBlocksExplanation = undefined; // Show default text.
+	}
+
+	const shouldOverwriteButtonTitle =
+		unsupportedBlockEditor.supported === false &&
+		unsupportedBlockEditor.switchable;
+	const unsupportedBlocksButton = shouldOverwriteButtonTitle
+		? __( 'Open Jetpack Security settings' )
+		: undefined;
```
guarani, line 45: Nitpick: It seems like instead of using the extra `shouldOverwriteButtonTitle` variable, we could use the if-else statement above to set `unsupportedBlocksButton`.

src/ui-strings.js
```diff
@@ -0,0 +1,52 @@
...
+import { __ } from '@wordpress/i18n';
+
+export const uiStrings = () => {
+	const { getSettings } = select( 'core/block-editor' );
+	const unsupportedBlockEditor = {
+		supported:
+			getSettings( 'capabilities' ).unsupportedBlockEditor === true,
+		switchable:
+			getSettings( 'capabilities' ).unsupportedBlockEditorSwitch === true,
+	};
```
guarani: Nitpick: An alternative here could be to use:
```
	const { 
		unsupportedBlockEditor = false,
		unsupportedBlockEditorSwitch = false
	} = getSettings( 'capabilities' ).unsupportedBlockEditor;
```

guarani (approved):
The changes here look good. I added some comments on things that I had suggestions about.

src/ui-strings.js
etoledom: This would make us declare `unsupportedBlocksButton` as `let` right?
I thought `shouldOverwriteButtonTitle` name would also give some context of the reasoning for the `undefined`.

Would you say something like this?
```javascript
let unsupportedBlocksButtonTitle = undefined;
if ( unsupportedBlockEditor.supported === false && unsupportedBlockEditor.switchable ) {
    unsupportedBlocksButtonTitle = __( 'Open Jetpack Security settings' );
}
```

src/ui-strings.js
etoledom: `unsupportedBlockEditor.supported` and `unsupportedBlockEditor.switchable` I think gives a nice context on their usage. Like:

```javascript
if (
 	unsupportedBlockEditor.supported === false &&
 	unsupportedBlockEditor.switchable
 )
```
vs
```javascript
if (
 	unsupportedBlockEditor === false &&
 	unsupportedBlockEditorSwitch
)
```

What do you think?

etoledom: (By the way, I think this whole file is a mess anyway, and not much extendible if we add more strings in the future. We will probably split it in the future)

src/ui-strings.js
```diff
@@ -0,0 +1,52 @@
...
+ */
+import { addStrings } from '@wordpress/components';
+import { select } from '@wordpress/data';
+import { __ } from '@wordpress/i18n';
+
+export const uiStrings = () => {
+	const { getSettings } = select( 'core/block-editor' );
+	const unsupportedBlockEditor = {
+		supported:
+			getSettings( 'capabilities' ).unsupportedBlockEditor === true,
```
mchowning, line 12: Just asking out of curiosity. We have a lot of explicit bool checks in this file (i.e., `someValue === true` and `someValue === false`). Are we doing that to guard against a cases where `someValue` could be some other truth-y or false-y values, or is it a style preference to do the explicit bool check instead of doing something like `someValue` (or even `!!someValue` if we needed a bool) and `!someValue`, or maybe there's another reason?

Just asking because I tend to use the `!`s when I can (I find them more readable), and I'm curious if there are considerations I may not be considering when I do that. I'm not asking because I think we need to change it here. 🙂

mchowning (approved):
LGTM!

src/ui-strings.js
guarani: Ideally, those attributes (`unsupportedBlockEditor` and `unsupportedBlockEditorSwitch`) would be descriptive enough on their own so that they can be deconstructed and used as-is without needing to be renamed. But I know it's hard here to find a good name, so I'm happy to leave it as-is if you want to.

src/ui-strings.js
guarani: Yes I think I had something like that in mind.

src/ui-strings.js
```diff
@@ -0,0 +1,52 @@
+/**
+ * WordPress dependencies
+ */
+import { addStrings } from '@wordpress/components';
+import { select } from '@wordpress/data';
+import { __ } from '@wordpress/i18n';
+
+export const uiStrings = () => {
+	const { getSettings } = select( 'core/block-editor' );
+	const unsupportedBlockEditor = {
```
Tug, line 10: Nitpick: I think just using 2 variables instead of grouping them in an object is just simpler and easier to understand, especially since you're assigning an `unsupportedBlockEditor` to a `supported` property

src/ui-strings.js
```diff
@@ -0,0 +1,52 @@
...
+		? __( 'Open Jetpack Security settings' )
+		: undefined;
+
+	return {
+		'missing-block-detail': unsupportedBlocksExplanation,
+		'missing-block-action-button': unsupportedBlocksButton,
+	};
+};
+
+addStrings( uiStrings );
```
Tug, line 53: Nitpick: I would expect that, given its name, addStrings would accept a list of strings as parameter. The fact that it accepts a function returning a mapping of (string id, string value) makes it a bit hard to grasp imo.

Tug (commented):
Couldn't we have achieved that feature by just adding a hook in `missing/edit.native.js`:
```js
const missingBlockDetail = applyFilters( 'native.missing_block_detail', __( 'We are working hard to add more blocks with each release.' ) );

...

<Text style={ [ infoTextStyle, infoDescriptionStyle ] }>
	{ missingBlockDetail }
</Text>
```
Then in gb-mobile:
```
addFilter( 'native.missing_block_detail', ( defaultValue ) => {
	// uiStrings() logic goes here
} );
```
?

Same for the action button.

etoledom:
> Couldn't we have achieved that feature by just adding a hook

@Tug - Nice tip! Indeed looks a lot simpler.

The only thing we need from it is that it has to refresh and return the correct String after updating the capabilities.
I will test this out and probably make the chance on an upcoming PR if that's OK with you 👍

PR #2617, opened 2020-09-14
Include strings from Jetpack in bundle

mchowning:
Fixes #2615 

Updating our bundle scripts to ensure that the strings from the jetpack/ repo are included. We are first retrieving the strings from the jetpack repo in a `jetpack-$PLATFORM.pot` file. I am then merging that into the `gutenberg-$PLATFORM.pot` file along with the strings from gutenberg. That `gutenberg-$PLATFORM.pot` file is then passed to the relevant `./bin/po2XXXX.js` script.

Note that I created this branch off of `trunk` in order to avoid having additional string changes on account of the updates to the submodules on `develop`. I anticipate that before merging this PR I'll want to rebase this PR to move the package.json changes on top of the appropriate branch (probably `develop`) and drop the strings changes I have committed in 68a00fd.

PR submission checklist:

- [X] I have considered adding unit tests where possible.
- [X] I have considered if this change warrants user-facing release notes [more info](https://github.com/wordpress-mobile/gutenberg-mobile/blob/develop/docs/Release-notes.md) and have added them to [RELEASE-NOTES.txt](https://github.com/wordpress-mobile/gutenberg-mobile/blob/develop/RELEASE-NOTES.txt) if necessary.

PR #2622, opened 2020-09-16
Generate translation files from bundles instead of native sources

### LINKED ISSUE #2615: Strings from Jetpack are not included in the bundle
**Describe the bug**
When we run our bundle script we are not including strings from jetpack in the resulting bundle.

**To Reproduce**
1. `npm run genstrings`
2. Observe that the strings files (i.e., `bundle/android/strings.xml`) do not contain relevant strings from the jetpack repo: for example, ["Link address to Google Maps"](https://github.com/Automattic/jetpack/blob/f6f313d268c336491c8bca1afcf3e5d3663597ce/extensions/blocks/contact-info/address/edit.native.js#L42)). 

**Expected behavior**
Relevant strings from the jetpack repos should be included in our bundles for translation.

Tug:
Fixes #2615
This is an alternative to #2617

This PR contains a change in wp-cli.phar, that is being reviewed here https://github.com/wp-cli/i18n-command/pull/228
It allows makepot to correctly parse our bundled JS and extract translation strings.

Using bundle as the source of truth for strings will make our i18n process less fragile, as any code that is included in it will be used to determine the strings to be translated. In return, we lose the position of those strings in our code as well as translators comments, something that has already been done on gutenberg to automate translations.

TODO:
- [ ] Investigate wrong encoding for emoji https://github.com/wordpress-mobile/gutenberg-mobile/pull/2622/files#diff-fadcda321726c553aff8098335f23ab6R101
- [ ] Investigate fixing translators comments by creating a dev build that preserve comments

PR submission checklist:


mchowning:
I like this approach better than what I have suggested in #2617. Assuming we get the translations and encoding issues you mention resolved, this feels like a much more robust approach to take going forward. Nice work @Tug ! 👍

mchowning:
Our translations pipeline has undergone a lot of changes (some inspired by this PR), so I'm going to close this.

PR #2626, opened 2020-09-16
Update README to correct initial setup instructions

ceyhun:
This repo doesn't have an `.nvmrc` file on it's own. This means the instruction `nvm install --latest-npm` would install the latest `node` version, which is not what we want. The submodule `gutenberg` repo has an `.nvmrc` which is kept up-to-date and we are making `nvm` use that one by adding these new steps that `cd` into `gutenberg` before installing `node`.


hypest (approved):
LGTM!

PR #2629, opened 2020-09-16
Update content hash item limit to prevent Watchman crawl failed errors

chipsnyder:
`Related-PR` https://github.com/WordPress/gutenberg/pull/25385

## Fixes:
While running the metro server locally we'll see the error:

```
Loading dependency graph...jest-haste-map: Watchman crawl failed. Retrying once with node crawler.
  Usually this happens when watchman isn't running. Create an empty `.watchmanconfig` file in your project's root folder or initialize a git or hg repository in your project.
  Error: Watchman error: too many pending cache jobs. Make sure watchman is running for this project. See https://facebook.github.io/watchman/docs/troubleshooting.html.
```

This error will occur when our file size has outgrown the default cache size of metro (more info in this [SO answer](https://stackoverflow.com/a/57448116/1350218)) This increases the content hash size limit to prevent the error.

## To test:
1. Run `npm run clean`
1. Run `npm start --reset-cache`

**Expect** The Metro server to start without displaying the cache failure error

## PR submission checklist:

### DIFF
--- a/.watchmanconfig
+++ b/.watchmanconfig
@@ -1,1 +1,3 @@
-{}
+{
+	"content_hash_max_items": 400000
+}


.watchmanconfig
```diff
@@ -1 +1,3 @@
-{}
+{
+	"content_hash_max_items": 400000
```
chipsnyder, line 2: This number was obtained by running `find . | wc -l` from the gutenberg-mobile root

chipsnyder:
Hey @hypest Would I be able to get you to take a look at this PR as well? It's the same change as https://github.com/WordPress/gutenberg/pull/25385 but handles it for the GB-mobile side.

hypest:
> Hey @hypest Would I be able to get you to take a look at this PR as well?

👋 Chip, sure!

I still am not able to reproduce the issue, and this time I tried lowering the `content_hash_max_items` all the way down to just `40` and still can't reproduce. The default for that setting [seems to be 131072](https://github.com/facebook/watchman/issues/737#issuecomment-518717721) and I'd expect to see the issue if I have it too low. What do you think?

chipsnyder:
> What do you think?

That seems odd to me that a clean and a reset isn't able to fix the issue for you. Do you happen to have a file in `/etc/watchman.json`? (https://facebook.github.io/watchman/docs/config.html)

I would think lowering the number to `40` would cause the error but perhaps there is a lower bound as well. 🤷‍♂️

hypest:
> isn't able to fix the issue for you.

To be clear, it's not that the fix doesn't work for me but rather, I don't see the original issue so can't verify that the fix actually works.

The PR doesn't seem to break the build on _my side_ anyway though so, I'm OK with merging since it fixes it for you. If the issue ends up not being fixed anyway we can revisit. How's that sound?

chipsnyder:
>If the issue ends up not being fixed anyway we can revisit. How's that sound?

I think that's a good plan :) It's an easy enough line to remove if it does cause trouble.

hypest (approved):
LGTM!

Also, reminder to update to resolve the merge conflict.

PR #2645, opened 2020-09-18
Fix image block render in Xcode 12

guarani:
This PR fixes #25431 which is a bug that prevents Image blocks from rendering their images. This only affects builds made using Xcode 12.

To test: See https://github.com/WordPress/gutenberg/pull/25470

⚠️ I opted not to include release notes because while this is a user facing bug fix, it only affects builds made using Xcode 12 and the App Store versions of WPiOS are still built using Xcode 11.

PR submission checklist:


ceyhun (approved):
Approved via https://github.com/WordPress/gutenberg/pull/25470

PR #2652, opened 2020-09-22
Image block - Update image url 

### LINKED ISSUE #25418: Duplicating an Image block not working consistently
**Describe the bug**
Sometimes when creating an Image block (without setting the size to full-screen) depending on the image size it would fill the entire space when it shouldn't since its smaller than the container.

If that block gets duplicated the new one will show correctly but it looks weird that the same block has two different sizes after duplicating.

<kbd></kbd>

**To reproduce**

Using this [image](https://images.pexels.com/photos/1851164/pexels-photo-1851164.jpeg?auto=compress&cs=tinysrgb&dpr=2&h=750&w=1260)

- Open the app with metro running
- Add an image block
- Upload the image from above
- **Expect** to see the image filling the entire space (since it doesn't happen always a few tries might be needed)
- Once you have the image block with the image filling the block space, duplicate it
- **Expect** to see the duplicated block with a smaller size

**Expected behavior**
The image block should show the correct size and duplicated ones should match

**Editor version (please complete the following information):**
Current version

 - Device: iPhone 11 Pro simulator
 - OS: 13
- Demo App Version (develop)

geriux:
Fixes https://github.com/WordPress/gutenberg/issues/25418

`Gutenberg PR` -> https://github.com/WordPress/gutenberg/pull/25529

An issue [was reported](https://github.com/WordPress/gutenberg/issues/25418
) of the Image block not keeping its image size after duplicating, after some investigation it turns out that if the block is not selected while an upload is in process it won't request the image data, sizes, and resized URL when it finishes.

On the web editor, it works differently because when the post request is made to upload the image it returns the resized URL and data, where in mobile it just returns the full size URL of the image.

This PR handles the case of updating the final resized URL of the image when the block is not selected and there's an ongoing upload.

PR submission checklist:


chipsnyder (approved):
:shipit: https://github.com/WordPress/gutenberg/pull/25529#pullrequestreview-495706534

PR #2654, opened 2020-09-22
Try running scheduled device tests targeting gutenberg master

Tug:
WIP

This PR adds new scheduled device tests to CircleCI. The same exact set of tests is running but it targets gutenberg master instead of the default ref so we can spot regressions as they appear in gutenberg.

PR submission checklist:


jhnstn:
Just noticed this branch is still running daily on Circle CI and failing. Closing since it's stale.

PR #2684, opened 2020-10-01
Fix "ENOMEM: not enough memory, write" error on CI with Android E2E tests

ceyhun:
Fixes `ENOMEM: not enough memory, write` error on CI with Android E2E tests

Example failed run: https://app.circleci.com/pipelines/github/wordpress-mobile/gutenberg-mobile/8518/workflows/2254d8bc-a0d5-45f7-8a9c-fd2e14c8874e/jobs/45314

Used `machine` instead of `docker` as executer which has `7.5GB` instead of `4GB` of RAM. The error during bundling step of JS code was fixed by this, but the ubuntu image used for the machine doesn't come with android tooling pre-installed. So to build the apk used the previous docker image but ran it inside the ubuntu machine.
Tried another approach using [persist_to_workspace](https://circleci.com/docs/2.0/configuration-reference/#persist_to_workspace) to save the JS bundle created in the `machine` executor and use it in a `docker` executer later with [attach_workspace](https://circleci.com/docs/2.0/configuration-reference/#attach_workspace), but found out that this needs re-installing the npm packages inside the `docker` executor container again which may not be the most efficient way, so went with the current solution instead.

To test:
- CI should be green.

PR submission checklist:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -233,37 +233,37 @@ jobs:
 workflows:
   gutenberg-mobile:
     jobs:
-      - checks:
-          name: Check Correctness
-          check-correctness: true
-      - checks:
-          name: Test iOS
-          platform: ios
-          check-tests: true
-      - checks:
-          name: Test Android
-          platform: android
-          check-tests: true
-      - ios-device-checks:
-          name: Test iOS on Device - Canaries
-          is-canary: "-canary"
+      # - checks:
+      #     name: Check Correctness
+      #     check-correctness: true
+      # - checks:
+      #     name: Test iOS
+      #     platform: ios
+      #     check-tests: true
+      # - checks:
+      #     name: Test Android
+      #     platform: android
+      #     check-tests: true
+      # - ios-device-checks:
+      #     name: Test iOS on Device - Canaries
+      #     is-canary: "-canary"
       - android-device-checks:
           name: Test Android on Device - Canaries
           is-canary: "-canary"
-      - Optional UI Tests:
-          type: approval
-          filters:
-            branches:
-              ignore:
-                - develop
-      - ios-device-checks:
-          name: Test iOS on Device - Full
-          requires: [ "Optional UI Tests" ]
-      - android-device-checks:
-          name: Test Android on Device - Full
-          requires: [ "Optional UI Tests" ]
-      - android-native-unit-tests:
-          name: Android Native Unit Tests
+      # - Optional UI Tests:
+      #     type: approval
+      #     filters:
+      #       branches:
+      #         ignore:
+      #           - develop
+      # - ios-device-checks:
+      #     name: Test iOS on Device - Full
+      #     requires: [ "Optional UI Tests" ]
+      # - android-device-checks:
+      #     name: Test Android on Device - Full
+      #     requires: [ "Optional UI Tests" ]
+      # - android-native-unit-tests:
+      #     name: Android Native Unit Tests
 
   ui-tests-full-scheduled:
     jobs:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -93,37 +93,41 @@ jobs:
       is-canary:
         type: string
         default: ""
-    docker:
-    - image: circleci/android:api-29-node@sha256:71d61d6c21b29948d57120f476a83cc322a280979bce355c5a0ad771293ca380
-    environment:
-      JAVA_OPTS: '-Xms512m -Xmx2g'
-      GRADLE_OPTS: '-Xmx3g -Dorg.gradle.daemon=false -Dorg.gradle.jvmargs="-Xmx2g -XX:+HeapDumpOnOutOfMemoryError"'
+    machine:
+      image: ubuntu-1604:202007-01
```
hypest: 👋 Ceyhun, for documentation purposes, can you elaborate on the particular ubuntu version used here? For example, why not a more recent LTS perhaps? Thanks!

.circleci/config.yml
```diff
@@ -93,37 +93,41 @@ jobs:
...
-            echo 'export TEST_ENV=sauce' >> $BASH_ENV
-      - run:
-          name: Bundle Android and Generate debug .apk file for testing
-          command: npm run test:e2e:bundle:android && npm run core test:e2e:build-app:android
+            docker run --rm -it \
+            --volume $(pwd):/home/circleci/project \
+            --workdir /home/circleci/project \
+            --env JAVA_OPTS='-Xms512m -Xmx2g' \
+            --env GRADLE_OPTS='-Xmx3g -Dorg.gradle.daemon=false -Dorg.gradle.jvmargs="-Xmx2g -XX:+HeapDumpOnOutOfMemoryError"' \
+            circleci/android@sha256:71d61d6c21b29948d57120f476a83cc322a280979bce355c5a0ad771293ca380 \
```
hypest: Not a blocker but, looks like we are referencing this docker image in a couple of places (here and in line 146) and I'm thinking if it makes sense to have a constant defined somewhere with this name, to avoid potentially changing the one but not the other?

hypest (commented):
Awesome work here @ceyhun and a cool solution!

I only left a couple of minor comments, have a look.

Also, I re-triggered the CI even though it's green to verify that it's consistently green.

.circleci/config.yml
ceyhun: No real reason actually :) other than it being used in the docs here: https://circleci.com/docs/2.0/executor-types/#using-machine 
Let me try the latest version there: `ubuntu-2004:202010-01`

.circleci/config.yml
ceyhun: Good idea! But on second thought maybe it's not necessary to pin this image (which was done to pin node version) as we might not need node when building the apk. Let me try to remove the hash instead.

.circleci/config.yml
```diff
@@ -110,7 +110,7 @@ jobs:
             --workdir /home/circleci/project \
             --env JAVA_OPTS='-Xms512m -Xmx2g' \
             --env GRADLE_OPTS='-Xmx3g -Dorg.gradle.daemon=false -Dorg.gradle.jvmargs="-Xmx2g -XX:+HeapDumpOnOutOfMemoryError"' \
-            circleci/android@sha256:71d61d6c21b29948d57120f476a83cc322a280979bce355c5a0ad771293ca380 \
+            circleci/android:api-29-node \
```
hypest: Ah, much more readable than the hash 😃 

Wonder though, shouldn't we do the same for https://github.com/wordpress-mobile/gutenberg-mobile/blob/a4cdb95b0e4f1553420b36b6826eb3b2a46bc4e8/.circleci/config.yml#L146?

And while at it, can we put that name as a string [pipeline parameter](https://circleci.com/docs/2.0/pipeline-variables/#pipeline-parameters-in-configuration) so it's defined in a single place but used in many?

ceyhun: `api-29-node` takes the latest version of the image with that tag, but in this case we need a previous version because the node version is updated in the latest. We don't have to use the same docker image for both because for the one in this PR the node version is not important. But for consistency I will roll back the latest change and update them both to use the same image from a parameter 👍

hypest: Ah, I see. Note that no strong feelings about having the 2 images be exactly the same. If we end up with two different than the only thing I'd prefer would be to have a pipeline parameter for each to make it clearer which is which and why.

hypest (approved):
LGTM!

I'll try to re-trigger CI to make sure the green there is consistent and we can merge!

PR #2690, opened 2020-10-05
Add gb mobile specific initial html

guarani:
This replaces the default HTML demo content with just non-core blocks (at the moment that's just the `jetpack/contact-info` block).
The related PRs are:
- https://github.com/WordPress/gutenberg/pull/25844 removes these non-core blocks from Gutenberg
- https://github.com/WordPress/gutenberg/pull/25865 allows us to add core blocks back here

To test:
1. Run `npm run start:reset` from the `gutenberg-mobile` folder
2. Run `npm run core ios` or `npm run core android` to open the demo app
3. Expect to see ~~just non-core blocks in the editor (no core blocks)~~ Jetpack blocks followed by core blocks

PR submission checklist:


src/index.js
```diff
@@ -28,6 +29,7 @@ addFilter( 'native.block_editor_props', 'gutenberg-mobile', ( editorProps ) => {
 		return {
 			...editorProps,
 			initialTitle,
+			initialData: initialHtml,
```
ceyhun: Maybe it's better to append `initialHtml` from gutenberg to this like: `initialHtmlGutenberg + initialHtml` so we have both when running from gb-mobile.

mzorz: sounds like a good idea! would like to be able to see all blocks together

guarani: I was thinking of how we could achieve this:
1. Add another filter (`addFilter`) that executes after [gutenberg's filter](https://github.com/WordPress/gutenberg/blob/51233ab706a9630e997cc9a508f72a9fe99a27ca/packages/react-native-editor/src/index.js#L63) — override the title and append the Jetpack blocks to the content
2. Export Gutenberg's `initial-html` and import it here into gutenberg-mobile

For (1), the challenge is how to allow the main apps to set the title and content without them being overriden. For (2), in my tests merely importing the `react-native-editor` instantiates it, so I couldn't simply export `initial-html` from inside gutenberg and import it in gutenberg-mobile

Do y'all have any thoughts on how this can be done? Maybe I'm overlooking something.

mzorz: > I couldn't simply export initial-html from inside gutenberg and import it in gutenberg-mobile

This seems the most natural thing to do to me, but I have trouble understanding why (2) is not possible, I think I don't fully understand how exporting that in GB is not import-able in mobile-gb?

guarani: Looks like (2) works, I just needed to make a few tweaks. PR incoming.

mchowning:
Code looks good. Nice work @guarani ! 👍

Tested the gutenberg-mobile demo app (running `npm run core android` from `gutenberg-mobile/`) and everything is working well. I wanted to also check that these changes didn’t break anything when building the gutenberg demo app (running `npm run native android` from `gutenberg-mobile/gutenberg`), but there the metro bundler crashes when the app starts. This doesn't appear to be a regression with your branch though because it's also happening on `master` for me.

I'm thinking that if you or someone else can confirm that the gutenberg demo app is crashing for you too on both your branch and `master` (want to make sure I don't just have some weird issue locally), I would be fine to approve and merge your changes and either (a) fix the gutenberg demo app crash in a separate PR or (b) open an issue to track that issue.

guarani:
Thanks for taking a look, @mchowning!

> Tested the gutenberg-mobile demo app (running npm run core android from gutenberg-mobile/) and everything is working well. 

I actually run the packager manually (although the way you mentioned is usually quicker). When I run it your way though, I get an error, so I created an issue now: https://github.com/WordPress/gutenberg/issues/25883.

> I wanted to also check that these changes didn’t break anything when building the gutenberg demo app (running `npm run native android` from `gutenberg-mobile/gutenberg`), but there the metro bundler crashes when the app starts.

I wonder if this is the same error.

On the iOS side of things, there's actually another error I came across today (but hadn't reported and there is a workaround), here's the issue I just created for it: https://github.com/WordPress/gutenberg/issues/25882

mzorz:
> I wanted to also check that these changes didn’t break anything when building the gutenberg demo app (running npm run native android from gutenberg-mobile/gutenberg), but there the metro bundler crashes when the app starts.

I ran things locally here after first running `npm run core android` on `gutenberg-mobile` successfully, and it just worked well here (that is, I couldn't observe the mentioned crash @mchowning ).

Also, code LGTM @guarani 👍  - FWIW feel free to merge if Matt's problem gets sorted out somehow

mchowning:
Thanks for creating the issues @guarani . 🙇 

> I actually run the packager manually (although the way you mentioned is usually quicker). When I run it your way though, I get an error.

I forgot about `npm run native start:reset` and was either letting `npm run native android` start the packager or I was manually starting it with `react-native start -- --reset-cache`, neither of which was working. Thanks for the reminder about `npm run native start:reset` in the issue. With that, I am able to confirm that your changes also work with the gutenberg demo app. 😄

mchowning (approved):

guarani:
> I ran things locally here after first running `npm run core android` on `gutenberg-mobile` successfully, and it just worked well here (that is, I couldn't observe the mentioned crash @mchowning ).

Thanks for checking, @mzorz! Looks like the error some of us get affects both gutenberg and gutenberg-mobile, although I only reported it in gutenberg: https://github.com/WordPress/gutenberg/issues/25883. Perhaps it's related to caching of something similar and that's why only some of us are seeing it.

> Also, code LGTM @guarani 👍 - FWIW feel free to merge if Matt's problem gets sorted out somehow

I'm just waiting for CircleCI and then I will merge (I've also made the updates you suggested in https://github.com/WordPress/gutenberg/pull/25865 👍)

PR #2766, opened 2020-10-30
Move the setupJetpackEditor() call order

mzorz:
This PR moves the order in which the jetpack blocks get registered from `'native.pre-render'` action to  `'native.render'` action, with the aim of having the jetpack blocks listed last in the block picker as per default (and let core blocks be first with the default order at first). Note the order in which they appear may be subject to change later on a per user basis, depending on the [frecency](https://github.com/WordPress/gutenberg/blob/master/packages/block-editor/src/store/selectors.js#L1396-L1566) calculations when getting items from the store to the inserter. 

**Before**
See the jetpack blocks (Contact info, Story) appear first
blockorderBefore

**After**
See the jetpack blocks (Contact info, Story) appear last
blockorderNow


etoledom (approved):
Working great as expected! 🎉

PR #2768, opened 2020-11-02
Added story block in supported block list

mzorz:
Adds the story block to the supported list.

To test:
N/A.


mzorz:
Decided to keep this unmerged until there is a need to include the Story block in templates cc @aforcier

PR #2770, opened 2020-11-02
Stories block: hide/show based on Editor capability settings being set, only after it's been loaded

mzorz:
We observed the settings may not be ready exactly when the `native.render` hooks get run but rather right after that execution cycle. Hence, we're only checking for the actual settings to be loaded by using `setTimeout` without delay

To test:
- verify the tests are now passing with `npm run test src/test/index.js`

Note: I've considered and tried adding unit tests for the specific functionality added but, since we haven't found a way to mock select/dispatch easily yet - will get back at this at a later point.

PR submission checklist:


src/jetpack-editor-setup.js
```diff
@@ -44,16 +44,18 @@ export default ( jetpackState ) => {
 
 	const jetpackData = setJetpackData( jetpackState );
 
-	const mediaFilesCollectionBlock = useSelect( ( select ) => {
-		return select( 'core/block-editor' ).getSettings( 'capabilities' )
-			.mediaFilesCollectionBlock;
-	}, [] );
+	// eslint-disable-next-line @wordpress/react-no-unsafe-timeout
```
etoledom, line 54: Maybe we could write down the reason why are we using a timeout here too for future references

etoledom (approved):
Working great as expected 🎉 

Tested on Android, iOS and unit tests passing locally too
Great job @mzorz ! 🙏

src/jetpack-editor-setup.js
mzorz: Good idea!  added here 6eb17d5dc

PR #2771, opened 2020-11-03
Scroll to selected block that goes off the screen 

### LINKED ISSUE #735: Editor doesn’t scroll with the block/cursor when I move a block
**Expectation**: When I move a block up and down in a post, I expect the editor to scroll so that the selected block and the cursor are always in view.

**Actual**: If I move the block up or down so that the block goes out of the visible area, the editor does not scroll automatically. I have to swipe to scroll the block back into view. (This is especially annoying if I’m trying to move a block a ways up or down in the post, as I have to alternate between moving the block and swiping to see it again.)

**To repro**:
1. Create or open a post with a lot of content.
2. Select a block.
3. Tap the up/down arrows to move the block until it goes off screen. Result: The editor doesn’t scroll with the block.

jd-alexander:
Fixes #735

To test:
Visit this [PR](https://github.com/WordPress/gutenberg/pull/26596)
PR submission checklist:


jd-alexander:
I am going to look into why these tests are failing.

ceyhun:
Bumping the milestone to 1.42, feel free to ping me if it needs to make it to 1.41 this week.

guarani (commented):
Just noting here that this isn't pending a review from my side as noted here: https://github.com/WordPress/gutenberg/pull/26596#issuecomment-728342331

PR #2960, opened 2021-01-05
Bump GB hash and update package-lock.json, plus fix Memory issue on CI

hypest:
PR to bump the Gutenberg submodule hash and update the `package-lock.json` file, mitigating the lock file issue spotted by [the Dependabot PR](https://app.circleci.com/pipelines/github/wordpress-mobile/gutenberg-mobile/9759/workflows/b2ab46b0-6c5a-4b06-9942-f0c66f1f6b2f/jobs/51677/parallel-runs/0/steps/0-108).

To test:
* CI should be green across the board, including the extended e2e test job

PR submission checklist:


hypest:
All tests green except the Correctness check, which failed with more changes detected in `package-lock.json`. Will run another `npm install` and commit.

ceyhun:
After removing and re-adding `@wordpress/eslint-plugin` to fix missing plugin errors there were still `import/no-unresolved` and `import/no-extraneous-dependencies` errors [on CI](https://circleci.com/gh/wordpress-mobile/gutenberg-mobile/51717). This is because `metro` doesn't resolve packages the same way `eslint-plugin-import` does.

Tried using `metro-resolver` as a custom `eslint-plugin-import` resolver, but it didn't work, as the `eslint-plugin-import` resolvers use different function parameters than `metro-resolver`. 

Finally I managed to resolve all the packages with [eslint-import-resolver-node](https://github.com/benmosher/eslint-plugin-import/tree/master/resolvers/node), by adding `gutenberg`, `gutenberg/node_modules` to the folders to be looked in for packages.

But all the errors were now `import/no-extraneous-dependencies`. This requires adding various dependencies to gutenberg-mobile's `package.json`. But this doesn't make sense as we have already the whole `gutenberg` repo as a dependency. So I added `gutenberg/package.json` as a path to also get checked for imported packages. Now everything should be green.

_Side note:_ Seems like both the rules are [disabled for React Native code](https://github.com/WordPress/gutenberg/pull/27965#issuecomment-754674096) in Gutenberg.

package.json
```diff
@@ -15,7 +15,7 @@
     "@babel/runtime": "^7.7.7",
     "@react-native-community/cli": "^3.0.4",
     "@wordpress/babel-preset-default": "file:./gutenberg/packages/babel-preset-default",
-    "@wordpress/eslint-plugin": "file:./gutenberg/packages/eslint-plugin",
+    "@wordpress/eslint-plugin": "file:gutenberg/packages/eslint-plugin",
```
fluiddot, line 18: I was wondering if we could apply the same path format in the rest of packages that use `file:`.

.eslintrc.js
```diff
@@ -19,5 +19,17 @@ module.exports = {
 	],
 	extends: [
 		"plugin:@wordpress/eslint-plugin/recommended",
-	]
+	],
+	settings: {
+		'import/resolver': {
+			'node': {
+				'moduleDirectory': ['node_modules', 'gutenberg', 'gutenberg/node_modules']
```
fluiddot: Would it make sense to add `jetpack` here too?

fluiddot:
Not sure why but when I run `npm install` the `package-lock.json` changes, did it happen to you @ceyhun?

Screenshot 2021-01-05 at 18 17 05

package.json
ceyhun: This was done by `npm` after I ran: `npm uninstall @wordpress/eslint-plugin && npm install ./gutenberg/packages/eslint-plugin --save-dev` to update the sub-dependencies. I did the same with other `@wordpress/` packages now and I realised maybe some of them are not necessary and uninstalled them instead 🙂

.eslintrc.js
ceyhun: Currently metro [doesn't check](https://github.com/wordpress-mobile/gutenberg-mobile/blob/a53c06da987bda6de57b53aede55065c989c8413/metro.config.js) the `node_modules` folder in there, I don't think it would make sense to add it.

Also I removed `gutenberg` from there as it was not working as I expected. Building `gutenberg` packages before linting instead did the trick for some reason ~~because it symlinked the `@wordpress/` packages inside `gutenberg/node_modules`~~.

ceyhun:
>Not sure why but when I run npm install the package-lock.json changes, did it happen to you @ceyhun?

This happened some time ago, but doesn't happen anymore. I think it could be related to node/npm version mismatch. Can you please try again after running `nvm install --latest-npm` inside `gutenberg/`.

fluiddot:
> > Not sure why but when I run npm install the package-lock.json changes, did it happen to you @ceyhun?
> 
> This happened some time ago, but doesn't happen anymore. I think it could be related to node/npm version mismatch. Can you please try again after running `nvm install --latest-npm` inside `gutenberg/`.

I've cleared the `node_modules` folders and now when running `npm install` the file didn't change, thanks!

fluiddot (commented):
I'm afraid I'm getting the `import/no-unresolved` error again, however before the latest changes it was working, @ceyhun are you getting it too?

By the way, I saw that one of the PR checks `Check Correctness` is failing, looks like the machine that is running the process runs out of memory 🤔 :
`Error: ENOMEM: not enough memory, read`

fluiddot:
> I'm afraid I'm getting the import/no-unresolved error again, however before the latest changes it was working, @ceyhun are you getting it too?

I've run `npm run lint` and it's working, I think the error I saw was caused by the ESLint plugin in VSCode 🙇 .

fluiddot (approved):

.eslintrc.js
ceyhun: Found the reason: https://github.com/WordPress/gutenberg/pull/22088

ceyhun:
> I'm afraid I'm getting the import/no-unresolved error again, however before the latest changes it was working, @ceyhun are you getting it too?

Did you try running `npm run build:packages` inside `gutenberg/` before running `npm run lint`? 

> By the way, I saw that one of the PR checks Check Correctness is failing, looks like the machine that is running the process runs out of memory 🤔 :
Error: ENOMEM: not enough memory, read

I saw it too, this could be a problem :( To debug it I rerun the job with SSH, but it passed this time. It means that this job could be flaky. I would like to reproduce this again and then solve it for good.

fluiddot:
>Did you try running npm run build:packages inside gutenberg/ before running npm run lint?

Good point!. Yeah I think that was the cause of the issue indeed because if I run `npm run clear:packages` and `npm run lint` I get the original errors. Running `npm run build:packages` before `npm run lint` works like charm, thanks!

ceyhun:
This might need another look now @fluiddot. Managed to fix the memory error by using `machine` instead of `docker` as executor, which has 7.5GB instead of 4GB of RAM. 
The last 3 builds have passing `Check Correctness` jobs. Feel free to re-run all the jobs to test another time.

fluiddot:
> This might need another look now @fluiddot. Managed to fix the memory error by using `machine` instead of `docker` as executor, which has 7.5GB instead of 4GB of RAM.
> The last 3 builds have passing `Check Correctness` jobs. Feel free to re-run all the jobs to test another time.

Good job! I'll re-run them one more time to double check.

I was checking the environment used in Gutenberg PR checks for building the packages and I saw that there we're using [GitHub hosted runners](https://docs.github.com/en/free-pro-team@latest/actions/reference/specifications-for-github-hosted-runners#supported-runners-and-hardware-resources) which have 7 GB of RAM so probably we need that amount of RAM for this process.

fluiddot (approved):
`Check Correctness` check passed 🎊 !

hypest:
Thanks for the added work on this @ceyhun and for the review @fluiddot ! ❤️

PR #3063, opened 2021-01-27
Documentation: Update README.md with Note about Android Studio/Xcode

### LINKED ISSUE #3062: Documentation: Note for Android Studio/Xcode Configurations
Android Studio is listed as a prerequisite in the README.md but there isn't any note about the fact some changes are needed to the default set up in order to get the demo app working correctly with it. I ended up falling down a bit of a rabbit hole with this as I had Android Studio already and had been successfully using it for app development, but couldn't figure out why I couldn't get the demo app working.

The specific configurations I needed to make are listed in React Native's documentation here: https://reactnative.dev/docs/environment-setup

I think it'd be useful to add a quick note about this in our README.md, it'd have saved me a bit of time.

SiobhyB:
This PR updates the README.md with a reference to React Native's documentation, which details configurations that are needed in Android Studio and Xcode in order for the demo apps to run.

Fixes #3062

PR submission checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,7 +87,7 @@ To see a list of all of your available iOS devices, use `xcrun simctl list devic
 
 ### Troubleshooting
 
-If the app doesn't run on your first attempt, it may help to double-check the set up of your development environment against the latest requirements in [React Native's documentation](https://reactnative.dev/docs/environment-setup). With Android Studio, for example, you will need to configure the ANDROID_HOME environment variable and ensure that your version of JDK matches the latest requirements.
+If the Android emulator doesn't start correctly, or compiling fails with `Could not initialize class org.codehaus.groovy.runtime.InvokerHelper` or similar, it may help to double check the set up of your development environment against the latest requirements in [React Native's documentation](https://reactnative.dev/docs/environment-setup). With Android Studio, for example, you will need to configure the `ANDROID_HOME` environment variable and ensure that your version of JDK matches the latest requirements.
 
 Some times, and especially when tweaking anything in the `package.json`, Babel configuration (`.babelrc`) or the Jest configuration (`jest.config.js`), your changes might seem to not take effect as expected. On those times, you might need to clean various caches before starting the packager. To do that, run the script: `npm run start:reset`. Other times, you might want to reinstall the NPM packages from scratch and the `npm run clean:install` script can be handy.
 
--- a/README.md
+++ b/README.md
@@ -87,6 +87,8 @@ To see a list of all of your available iOS devices, use `xcrun simctl list devic
 
 ### Troubleshooting
 
+If the app doesn't run on your first attempt, it may help to double-check the set up of your development environment against the latest requirements in [React Native's documentation](https://reactnative.dev/docs/environment-setup). With Android Studio, for example, you will need to configure the ANDROID_HOME environment variable and ensure that your version of JDK matches the latest requirements.
+
 Some times, and especially when tweaking anything in the `package.json`, Babel configuration (`.babelrc`) or the Jest configuration (`jest.config.js`), your changes might seem to not take effect as expected. On those times, you might need to clean various caches before starting the packager. To do that, run the script: `npm run start:reset`. Other times, you might want to reinstall the NPM packages from scratch and the `npm run clean:install` script can be handy.
 
 ## Developing with Visual Studio Code
--- a/README.md
+++ b/README.md
@@ -12,10 +12,12 @@ For a developer experience closer to the one the project maintainers current hav
 * [nvm](https://github.com/creationix/nvm)
 * Node.js and npm (use nvm to install them)
 * [Yarn](https://yarnpkg.com/)
-* [AndroidStudio](https://developer.android.com/studio/) to be able to compile the Android version of the app
+* [Android Studio](https://developer.android.com/studio/) to be able to compile the Android version of the app
 * [Xcode](https://developer.apple.com/xcode/) to be able to compile the iOS app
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -12,11 +12,13 @@ For a developer experience closer to the one the project maintainers current hav
 * [nvm](https://github.com/creationix/nvm)
 * Node.js and npm (use nvm to install them)
 * [Yarn](https://yarnpkg.com/)
-* [AndroidStudio](https://developer.android.com/studio/) to be able to compile the Android version of the app
+* [Android Studio](https://developer.android.com/studio/) to be able to compile the Android version of the app
 * [Xcode](https://developer.apple.com/xcode/) to be able to compile the iOS app
 * CocoaPods(`sudo gem install cocoapods`) needed to fetch React and third-party dependencies.
 * [Carthage](https://github.com/Carthage/Carthage#installing-carthage) for appium to be able run iOS UI tests
 
+There may be some configurations needed in Android Studio and Xcode, depending on your setup. Please refer to [React Native's documentation](https://reactnative.dev/docs/environment-setup)
```
dcalhoun:
```suggestion
There may be some configurations needed in Android Studio and Xcode, depending on your setup. Please refer to [React Native's documentation](https://reactnative.dev/docs/environment-setup).
```

dcalhoun (approved):
This looks good! 💯 Thanks for adding clarity to the setup documentation. I offered one small suggestion related to punctuation.

README.md
SiobhyB: Thank you! Your review made me realise that the last part of the sentence, `for the latest requirements for each development environment.`,  was somehow missing. I've added that now!

hypest:
Thanks for the PR Siobhan!

I think it will be useful for others (as well as our future selves) to know what were the issues you did encounter the the now added documentation solves. Do you think we could add some brief descriptions of the issues and what helped fixing them, either in the PR description here or in the original GitHub ticket?

We could also assess how "popular" those issues can be and actually go ahead and mention them in the README directly, in the [Troubleshooting section](https://github.com/wordpress-mobile/gutenberg-mobile#troubleshooting) for example.

SiobhyB:
@hypest, sure! I don't recall all of the errors but, with Android Studio, the emulator didn't originally work because I hadn't added the needed lines to my `$HOME/.bash_profile` and then the app wouldn't load due to the version of JDK I was running. This was the error that helped me figure out there was a conflict with my JDK version:

```
Could not initialize class org.codehaus.groovy.runtime.InvokerHelper
```

 With Xcode, there was some kind of conflict with the way Ruby's installed on my machine. I kept getting the following error and [this guide](https://www.moncefbelyamani.com/the-definitive-guide-to-installing-ruby-gems-on-a-mac/) helped me:

```
You don't have write permissions for the /Library/Ruby/Gems/2.6.0 directory
```

React Native's documentation is pretty comprehensive when it comes to how both Android Studio and Xcode should be set up, so I suspect being pointed towards it would have been enough for me to avoid the above errors. I ended up wading through a sea of StackOverflow posts when, in fact, the documentation had everything I needed, I just wasn't aware of it.

I've added some extra details to the troubleshooting section, let me know what you think!

hypest:
> ... but, with Android Studio, the emulator didn't originally work because I hadn't added the needed lines to my $HOME/.bash_profile and then the app wouldn't load due to the version of JDK I was running.

and

> I ended up wading through a sea of StackOverflow posts when, in fact, the documentation had everything I needed, I just wasn't aware of it.

Ah, I see. Yeah, unfortunately, there's no "single" way to setup things (one could use the Oracle provided JDK for example, not the OpenJDK from homebrew on Macos), and it one way or the other might conflict with WPAndroid's setup. I feel like being too prescriptive might do harm so, these kind of errors are probably best be left out of the "requirements" for gutenberg-mobile. That said, hinting about things in the Troubleshooting section is fair and helpful.

README.md
```diff
@@ -85,6 +87,8 @@ To see a list of all of your available iOS devices, use `xcrun simctl list devic
 
 ### Troubleshooting
 
+If the app doesn't run on your first attempt, it may help to double-check the set up of your development environment against the latest requirements in [React Native's documentation](https://reactnative.dev/docs/environment-setup). With Android Studio, for example, you will need to configure the ANDROID_HOME environment variable and ensure that your version of JDK matches the latest requirements.
```
hypest: > If the app doesn't run on your first attempt

👋 @SiobhyB ! This feels too "open" to me and judging from your [comment here](https://github.com/wordpress-mobile/gutenberg-mobile/pull/3063#issuecomment-769698986), it might not even be an app issue but a development environment issue (say, if the Android emulator itself doesn't work).

What do you think about adding specific examples so the context is narrower? Something like this for example:
`If the Android emulator doesn't start correctly, or compiling fails with 'Could not initialize class org.codehaus.groovy.runtime.InvokerHelper' or similar, it may help to double-check...`

SiobhyB: Thank you for your feedback @hypest! I did lean towards keeping it a bit more open for the reasons you touched on in your last comment, I get the sense that there are many, many possible setup errors that can't all be covered in detail in the docs. I really like your suggestion, though, as it's more specific but still broad enough. I've updated. :)

hypest:
Looks good to me at this stage! I took the liberty to re-trigger the failed CI jobs as those seemed like flakiness fails. I'll LGTM when they get green!

hypest (approved):
LGTM!

hypest:
All green, reviews are 👍, and since there's no assignee set to govern who merges, I'll go ahead and merge at this point (internal ref p9ugOq-Kc-p2).

SiobhyB:
Excellent, thank you! :)

Issue #3061, opened 2021-01-27
Reusable blocks can't be edited through UBE


fluiddot:
**Describe the bug**
Due to [recent changes in the editing flow on the web version](https://github.com/WordPress/gutenberg/pull/27887), editing Reusable blocks through UBE is not working.

The main change is that saving the content is now done in the publish process so there's no longer the `Edit/Save` buttons in the block UI to apply the changes as we had previously.

<details><summary><strong>New editing flow - Web version</strong></summary>

https://user-images.githubusercontent.com/14905380/105991755-7f34c580-60a4-11eb-897f-ef6a74505bc3.mp4

</details>

**To Reproduce**
Since in development mode Reusable blocks are rendered, there're two ways to reproduce it.

<details><summary><strong>Development mode (block content rendered)</strong></summary>

## Steps
1. Go to a post that has some Reusable blocks included the content.
2. Tap twice on a Reusable block.
3. Tap on `Edit using web editor`.
4. Modify the content of the block.
5. Tap on `Continue` or ✔️ button to apply the changes.
6. Check that the content is not being modified.

</details>

<details><summary><strong>Production mode (block content rendered as a Missing block)</strong></summary>

## Steps
1. Go to a post that has some Reusable blocks included the content.
2. Tap twice on a Reusable block.
3. Tap on `Edit using web editor`.
4. Modify the content of the block.
5. Tap on `Continue` or ✔️ button to apply the changes.
6. Since the content is not rendered in production try to edit the block again.
7. Check that the content is not being modified.

</details>

**Expected behavior**
Reusable block content should be editable through UBE.

**Screenshots**

https://user-images.githubusercontent.com/14905380/105991439-0d5c7c00-60a4-11eb-8b55-2a829f8533a3.mp4

**Smartphone (please complete the following information):**
 - Device: Pixel 4 (emulator)
 - OS: Android 11
 - Version alpha-269

**Additional context**
No.

fluiddot:
Addressed via https://github.com/wordpress-mobile/gutenberg-mobile/pull/3067.

PR #3067, opened 2021-01-28
Disable Unsupported block editor for Reusable blocks

fluiddot:
Addresses https://github.com/wordpress-mobile/gutenberg-mobile/issues/3061

`gutenberg` PR: https://github.com/WordPress/gutenberg/pull/28552

## To test:

Since these changes are related to unsupported blocks, it requires a quick setup to be done in the **web version editor**:
1. Create a post/page.
2. Add a block that is not supported in the native editor like GIF (`jetpack/gif`).
3. Create a Reusable block and add it.
4. Save the post/page.

The following flows have been used for testing these changes:

<details><summary>Reusable block (dev mode only)</summary>

**This has to be tested in development mode.**

### Steps
1. Go to a post/page that has a Reusable block.
2. Tap on it two times.
3. Check that there's no option to edit it through the web editor.

</details>

<details><summary>Missing block - UBE supported block</summary>

### Steps
1. Go to a post/page that has a block that is not supported in the native editor.
2. Tap on it two times.
3. Check that the option of editing it through the web editor is displayed.

</details>

<details><summary>Missing block - UBE unsupported block</summary>

**This has to be tested in release mode or by disabling the Reusable block from the block list in [`block-library` package](https://github.com/WordPress/gutenberg/blob/master/packages/block-library/src/index.native.js#L228).** 

### Steps
1. Go to a post/page that has a Reusable block.
2. Tap on it two times.
3. Check that there's no option to edit it through the web editor.

</details>

PR submission checklist:

### DIFF
--- a/src/strings-overrides.js
+++ b/src/strings-overrides.js
@@ -56,24 +56,3 @@ addFilter(
 			: defaultValue;
 	}
 );
-
-addFilter(
-	'native.reusable_block_action_button',
-	'native/reusable_block',
-	( defaultValue ) => {
-		const { getSettings } = select( 'core/block-editor' );
-		const isUnsupportedBlockEditorSupported =
-			getSettings( 'capabilities' ).unsupportedBlockEditor === true;
-		const canEnableUnsupportedBlockEditor =
-			getSettings( 'capabilities' ).canEnableUnsupportedBlockEditor ===
-			true;
-
-		const shouldOverwriteButtonTitle =
-			isUnsupportedBlockEditorSupported === false &&
-			canEnableUnsupportedBlockEditor;
-
-		return shouldOverwriteButtonTitle
-			? __( 'Open Jetpack Security settings' )
-			: defaultValue;
-	}
-);


guarani:
@fluiddot should this "address" instead of "fix" https://github.com/wordpress-mobile/gutenberg-mobile/issues/3061? The issue's expected behavior is:
> Reusable block content should be editable through UBE.

And from my understanding here, this PR disables the UBE in this scenario to avoid a broken experience but doesn't address fixing the UBE for reusable blocks.

fluiddot:
> @fluiddot should this "address" instead of "fix" #3061? The issue's expected behavior is:
> 
> > Reusable block content should be editable through UBE.
> 
> And from my understanding here, this PR disables the UBE in this scenario to avoid a broken experience but doesn't address fixing the UBE for reusable blocks.

Yeah, makes more sense to use `address` than `fix` since it's not really going to fix it, thanks for pointing that up!

guarani (commented):
### Reusable block (dev mode only)

✅ This worked great, I saw the "Reusable block aren't supported message as expected. 

### Missing block - UBE supported block

⚠️ I tried this a couple of time because the first time I opened the UBE on a GIF block I got a login screen. The site in question is a private Simple WP.com site. After switching the site to public then back to private, I couldn't reproduce it again 🤷. I recorded a video of it when it failed, could you double check to make sure this works for you?

| |
| - |
|  |

### Missing block - UBE unsupported block

❌ This scenario is to ensure that the UBE is disabled on release builds. In Xcode, I switched the current "WordPress" scheme (Edit Scheme > Run > Info tab > Build Configuration) to Release and re-ran the app. As expected, the editor rendered my reusable block as an unsupported block, but I was able to open the UBE on it:

| |
| - |
|  |

fluiddot:
> ### Missing block - UBE supported block
> ⚠️ I tried this a couple of time because the first time I opened the UBE on a GIF block I got a login screen. The site in question is a private Simple WP.com site. After switching the site to public then back to private, I couldn't reproduce it again 🤷. I recorded a video of it when it failed, could you double check to make sure this works for you?

I bumped into this multiple times, actually it happens in every webview that connects to WP.com, like for example when viewing a post. I checked with other devs and looks like it's known issue and only happens in development.

This reminded me that I should have opened an issue regarding this (here is the [issue](https://github.com/wordpress-mobile/WordPress-iOS/issues/15739)).

> ### Missing block - UBE unsupported block
> ❌ This scenario is to ensure that the UBE is disabled on release builds. In Xcode, I switched the current "WordPress" scheme (Edit Scheme > Run > Info tab > Build Configuration) to Release and re-ran the app. As expected, the editor rendered my reusable block as an unsupported block, but I was able to open the UBE on it:

I think for testing the bundle in release mode it requires more steps sorry. Actually I think it's not really needed to change the scheme of the Xcode project since the only thing we need is to have is the RN bundle in production mode.

The extra steps are:
1. Run `npm run bundle:ios` in `gutenberg-mobile` project.
2. Copy `bundle/ios` folder from `gutenberg-mobile` project to `Pods/Gutenberg/bundle/ios` in `WordPress-iOS` project.
3. Build and run Xcode.

Another option would be to create a couple of test branches in both `gutenberg-mobile` and `WordPress-iOS` repositories and push the JS bundle for iOS, this way we would have a build to test with the bundle in production mode.

guarani (approved):
Thanks for the response @fluiddot! I'd missed how Release mode in Xcode wouldn't mean release mode for the running packager 😅. Tested again using the latest commit 64ebb69f1e6f544486acffb115863d273a551a06 using both platforms and all good!

### Reusable block (dev mode only)

✅ Works great.

### Missing block - UBE supported block

✅ Works great (fwiw I couldn't reproduce the issue I saw previously where I got the log in screen).

> I bumped into this multiple times, actually it happens in every webview that connects to WP.com, like for example when viewing a post. I checked with other devs and looks like it's known issue and only happens in development.

I've seen this problem on WPiOS with View Post and View Site web views (https://github.com/wordpress-mobile/WordPress-iOS/issues/15389 and https://github.com/wordpress-mobile/WordPress-iOS/issues/15342 come to mind), but I'm not sure yet if this case here with the UBE is the same issue or different. I'll check out your new issue after this and see if there's anything I can add there.

I forgot to mention that when I ran into the login screen, I downloaded the WPiOS App Store version 16.5 and couldn't reproduce the issue there, which made me think it was an issue related to this PR or something that had recently landed on the `develop` branch.

### Missing block - UBE unsupported block

> I think for testing the bundle in release mode it requires more steps sorry. Actually I think it's not really needed to change the scheme of the Xcode project since the only thing we need is to have is the RN bundle in production mode.

Thanks! Makes sense, I'd forgotten about that.

✅ Works great, I got the `'Reusable Block' is not fully-supported` message.
Just a thought here (definitely not a blocker). For unsupported blocks that the UBE **can** handle, we also use the same "not fully-supported" messaging. It might make it clearer to users if we use different wording here such as `'Reusable Block' is not yet supported`. Otherwise it looks very similar to the other bottom sheet, just minus the option to edit with the UBE.

fluiddot:
>I've seen this problem on WPiOS with View Post and View Site web views (wordpress-mobile/WordPress-iOS#15389 and wordpress-mobile/WordPress-iOS#15342 come to mind), but I'm not sure yet if this case here with the UBE is the same issue or different. I'll check out your new issue after this and see if there's anything I can add there.

As far as I investigated I think it's quite related because the cause is not being authenticated in WP.com. When I investigated further this issue some weeks ago, I spotted that when the webviews try to authenticate in WP.com, it's responding with empty cookies and the error message: `Please enter the verification code generated by your authenticator mobile application.`.

> I forgot to mention that when I ran into the login screen, I downloaded the WPiOS App Store version 16.5 and couldn't reproduce the issue there, which made me think it was an issue related to this PR or something that had recently landed on the develop branch.

Yeah I'm not sure when this started but looks like it's only happening in development, I couldn't manage to reproduce it in release versions.

> Just a thought here (definitely not a blocker). For unsupported blocks that the UBE can handle, we also use the same "not fully-supported" messaging. It might make it clearer to users if we use different wording here such as 'Reusable Block' is not yet supported. Otherwise it looks very similar to the other bottom sheet, just minus the option to edit with the UBE.

Good point! I doubted about changing the message to be honest because as you comment, it might look not clearer enough compared to the message we present for the other blocks. I'll review it and create an issue for changing the text in next versions, thanks!

fluiddot:
Reference updated to `gutenberg` `master` branch.

PR #3113, opened 2021-02-05
Temporary gutenberg update to test media files type

bjtitus:
Fixes an error when the media files delegate methods are called on iOS.

See https://github.com/wordpress-mobile/WordPress-iOS/pull/15811
See https://github.com/WordPress/gutenberg/pull/28766

<a href="https://user-images.githubusercontent.com/3250/107052017-66a26a80-678a-11eb-9a90-80b6468f8d14.png"></a>

To test:
* Open a post containing a stories block
* Attempt to edit the block
* With this change, no error should be shown
* Without this change, an error screen is shown (see above)

PR submission checklist:

- [x] ~I have considered adding unit tests where possible.~ Didn't see a place to test this
- [x] !I have considered if this change warrants user-facing release notes.~ Unreleased on iOS.


bjtitus:
@wendyjchen @mzorz I think the two of you were taking a look at this issue cropping up in iOS where an error screen is shown when editing Stories. This is the necessary fix to the Swift type used in the Gutenberg delegate methods. The actual editing implementation isn't in WPiOS yet, so nothing will happen when hitting the editing button after merging this (as opposed to the error showing).

bjtitus:
@guarani After some struggling with npm issue, I _think_ the bundle finally worked correctly. Could you check and see that this looks right?

guarani (approved):
Tested via the WPiOS PR and looks good!

fluiddot (commented):
👋 I realised that this PR has introduced the [`yarn.lock` file](https://github.com/wordpress-mobile/gutenberg-mobile/pull/3113/files?file-filters%5B%5D=.lock#diff-51e4f558fae534656963876761c95b83b6ef5da5103c4adef6768219ed76c2de), does this mean that we can start using it as an alternative to NPM?

PR #3116, opened 2021-02-08
Add a "Set as Featured Image" Button to Image Block (Android Only)

### LINKED ISSUE #1011: [Proposal] Allow easier setting of Featured Image
## Problem

IMG_0450

The current Featured Image flow has a few issues:

* Featured Image is only available in Post Settings screens, which means it can get buried and not easy to locate

* If you've set an image from the post body as a Featured Image, it's not obvious which image is selected because there is no visual cue

## Solution

To solve the above issues, we should we should supply an intuitive "shortcut" of sorts to achieve this – in context. This could be on the first image of the post, which is typically the image users expect to use.

* Add a quick action button to the Image block settings sheet

* Add an indicator over the image that is being used as the Featured Image – similar indication that we have on GIFs in Aztec

* Optionally, if the Featured Image is already set, we could display a dialog that confirms whether the user wants to replace the current image.

It might look something like this:

IMG_0451

## Aztec

If we do any work on the Aztec end, I'd recommend we simply add the `Featured` badge to the image that's being used as the Featured Image.

SiobhyB:
Partial fix for: #1011

`gutenberg`: https://github.com/WordPress/gutenberg/pull/31705
`WordPress-Android`: https://github.com/wordpress-mobile/WordPress-Android/pull/14503

Internal P2 detailing progress: p9ugOq-1D7-p2

This PR adds a **Set as Featured** button to the image block's settings, with the purpose being to make it simpler for users to set a featured image within the post's editor. Users will also be able to **Remove as Featured** directly from the block's settings. 

_Note: This PR introduces the **Set as Featured** button to Android devices only, support for iOS will follow in future iterations, too._

_**To test:** Please refer to the Gutenberg PR as the central place for steps to tests and screenshots._

PR submission checklist:


hypest (approved):
LGTM @SiobhyB ! Only thing needed is get the Gutenberg PR merged and this can then be updated and merged when CI green!

Issue #2971, opened 2021-01-07
Publish Android react-native-gutenberg-bridge to Bintray only when gutenberg updates


mokagio:
To get to an earliest testable version quickly, the first attempt of publishing react-native-gutenberg-bridge to Bintray runs on every commit on every PR.

That's obviously wasteful; we should publish a new AAR only when the `gutenberg` code changes.

I don't think CircleCI exposes that sort of information from the GitHub PR, but it should be relatively easy to do a `curl` of the GitHub APIs to find that out, something like in [this post](https://javorszky.co.uk/2019/03/11/set-up-circle-ci-to-run-phpcs-on-only-changed-files-in-a-pr/).

---

Clearly, this is not a blocker to get to something testable. Still, we should address is sooner rather than later.

mokagio:
cc @oguzkocer

PR #3230, opened 2021-03-08
Publish React Native Bridge to Bintray from CI

mokagio:
- [ ] Before merging this PR, we need to merge https://github.com/WordPress/gutenberg/pull/29488 into Gutenberg's `trunk` and update the submodule has here.
- [x] Before merging this PR, it needs to implement the versioning strategy specified [here](https://github.com/wordpress-mobile/gutenberg-mobile/pull/3262/files#diff-11096f9d861fef21abf4673f7b86333f1d5dfb14018b31593c81d88f6f00f446R37-R43)

---

This PR adds CI jobs to build and publish the AAR for the `react-native-bridge` package (which lives in Gutenberg).

I originally planned to open this against the `gutenberg-composite-build` project WIP branch, but given this doesn't change anything in the codebase itself other than the Gutenberg submodule hash, I think `develop` is appropriate.

To test: 

1. Verify that CI builds
2. Head over to https://github.com/wordpress-mobile/WordPress-Android/pull/14215 which uses the binary published from here and:
  - Verify it builds in CI
  - Verify you can run the app and use the editor locally, making sure the `localGutenbergMobile` path in your `local-builds.gradle` is commented out, otherwise it'll kick off a composite build, which won't use the binary.

---

PR submission checklist:

- [x] I have considered adding unit tests where possible. (N.A.)
- [x] I have considered if this change warrants user-facing release notes [more info](https://github.com/wordpress-mobile/gutenberg-mobile/blob/develop/Release-notes.md) and have added them to [RELEASE-NOTES.txt](https://github.com/wordpress-mobile/gutenberg-mobile/blob/develop/RELEASE-NOTES.txt) if necessary. (N.A.)


.circleci/config.yml
```diff
@@ -253,6 +253,89 @@ jobs:
...
+          name: Get cache id
+          command: |
+            echo $CIRCLE_SHA1 > cache_id
+            cat cache_id
+      - save_cache:
+          name: Cache React Native Android Bridge JS Bundle
+          key: android-js-bundle-{{ checksum "cache_id" }}
+          paths:
+            # The build process for the react-navite-bridge requires the App.js
+            # bundle and also access to the node_modules folder.
```
mokagio: Should I a link to the line of code that requires `node_modules` diff on GitHub once the PR for that is merged?

.circleci/config.yml
```diff
@@ -253,6 +253,89 @@ jobs:
...
+    machine:
+      image: << pipeline.parameters.linux-machine-image >>
+    steps:
+      - checkout
+      - checkout-submodules
+      - npm-install
+      - run:
+          name: Build JS Bundle
+          command: |
+            npm run bundle:android
```
mokagio:
```suggestion
          command: npm run bundle:android
```

.circleci/config.yml
```diff
@@ -253,6 +253,89 @@ jobs:
...
+              PR_NUMBER=$(basename $CIRCLE_PULL_REQUEST)
+              # Making the assumption that $CIRCLE_SHA1 will always be
+              # available.
+              VERSION="$PR_NUMBER-$CIRCLE_SHA1"
+            elif [[ "$branch" != "trunk" && "$branch" != "develop" ]]; then
+              # This happens on the first push of a new branch, when there
+              # isn't a PR open for it yet.
+              echo "IGNORED WHILE DEVELOPING THE FEATURE - Running on a feature branch with no open PR: skipping Bintray upload"
+              # exit 0
+              VERSION="mokagio-ci-test-$CIRCLE_SHA1"
```
mokagio: Need to fix this before merging turning to "Ready for review"

mokagio:
I'm not sure where all those `package-lock.json` changes came from... I checked out a new branch from `develop`, pointed Gutenberg to the same hash as here, deleted the `node_modules` folder, and run `npm install` and nothing changed in the file 🤔

mokagio:
@oguzkocer @mchowning @cameronvoell I think this is ready to go folks.

I'm confused by the errors I'm getting on the "Check Correctness" is like no matter how I spin the `package-lock.json` there's always something that doesn't work. 😭 

I _think_ this might have to do with the different Gutenberg version compared to `develop`? I'm not sure how to triage this, can you point me in the right direction? Thanks!

.circleci/config.yml
```diff
@@ -253,6 +253,87 @@ jobs:
                 include_job_number_field: false
                 include_project_field: false
                 failure_message: ':red_circle: Scheduled tests failed on iOS device!'
+  build-android-js-bundle:
+    # This job builds the JS bundle that the Android react-navite-bridge
+    # bridges and saves it in the CI cache to make it available to other jobs,
+    # in particular the one building the AAR for binary distribution.
```
oguzkocer: @mokagio I am not sure if I understand what you mean here, could you take a look at it? There are also a couple spelling mistakes.

.circleci/config.yml
oguzkocer:
```suggestion
            # The build process for the react-native-bridge requires the App.js and the node_modules folder.
```

We technically don't need the `node_modules` once the bundle is created. However, without it [configuration gets a bit messy](https://github.com/WordPress/gutenberg/pull/29488#discussion_r590635389). I don't think it's worth adding a link here and I am hoping we can work towards not needing the `node_modules` folder.

.circleci/config.yml
```diff
@@ -253,6 +253,87 @@ jobs:
...
+              PR_NUMBER=$(basename $CIRCLE_PULL_REQUEST)
+              # Making the assumption that $CIRCLE_SHA1 will always be
+              # available.
+              VERSION="$PR_NUMBER-$CIRCLE_SHA1"
+            elif [[ "$branch" != "trunk" && "$branch" != "develop" ]]; then
+              # This happens on the first push of a new branch, when there
+              # isn't a PR open for it yet.
+              echo "Running on a feature branch with no open PR: skipping Bintray upload"
+              exit 0
+            fi
```
oguzkocer: @loremattei Could you review this part?

@mokagio @loremattei Excuse my lack of knowledge of CircleCI configuration, but is there a way to not run the tasks at all if we are not going to publish the binary? I don't know how often that'll happen, but this task takes a long time (mostly due to `npm install`), so if we can stop the build as early as possible, we can potentially save some decent time. At the very least, I think we can move some of these checks to an earlier step, right?

Please feel free to ignore that ^ question, if the exit path is more of a sanity check.

oguzkocer (commented):
@mokagio This looks mostly good to me. I have left a few minor comments, let me know what you think!

.circleci/config.yml
mokagio: @oguzkocer, sure!

This is the only spelling mistake I see.

```suggestion
    # This job builds the JS bundle that the Android react-native-bridge
    # bridges and saves it in the CI cache to make it available to other jobs,
    # in particular the one building the AAR for binary distribution.
```

What are the other(s)?

Aside from that, does this wording sound clearer?

```suggestion
    # This job builds the Android version of the JS bundle and saves it
    # in the CI cache to make it available to other jobs that might need
    # it, namely the job building the AAR for the Android version of the
    # react-native-bridge package.
```

.circleci/config.yml
mokagio: @oguzkocer wow, I totally missed this! Thank you!

This is a port of the code from the `WordPress-Utils-Android`, where [we do everything in a single job](https://github.com/wordpress-mobile/WordPress-Utils-Android/blob/develop/.circleci/config.yml#L39-L101). I missed the possible wastefulness of running the previous step if we're not on the right branch.

> is there a way to not run the tasks at all if we are not going to publish the binary? I don't know how often that'll happen

There definitely is a way to do this. I don't know how often that'll happen either, mainly because I'm not sure whether we should build on every PR and merge to the main branch, or whether we want a manual step to trigger the publishing. @mchowning @cameronvoell can you chime in on what would be the best workflow?

I'm leaning towards always building but not making those jobs a [required status check](https://docs.github.com/en/github/administering-a-repository/managing-a-branch-protection-rule) for the PR merge. The reason I prefer this over a manual step is that it's seamless. I don't want people to have to remember to do things (I'm the first to forget tasks all the time unless I write them down) or to have to wait for a binary to be ready before being able to use it if it can be avoided.

As I discuss in https://github.com/wordpress-mobile/gutenberg-mobile/issues/2971, eventually it would be good to have some way to know whether the AAR should be built, but I don't know where to start for there and I think it fits into a dedicated PR.

mokagio: I have a working version of a setup where we only do all the steps from `npm install` if necessary.

Unfortunately, I had to use the Ubuntu machine for that, so there's a bit more overhead. I was hoping to be able to skip the checkout step as well, but I haven't figured out how. Finally, it's quite slow because it doesn't access the npm cache. I plan to work on that later on.

That's all to say, it works, but it can still be improved and made faster. [As a first step, it might be enough to get us over the line](https://wiki.c2.com/?MakeItWorkMakeItRightMakeItFast).

Keen to hear what you think of this approach. cc @ceyhun.

https://github.com/wordpress-mobile/gutenberg-mobile/compare/mokagio/publish-to-bintray-with-script...mokagio/publish-to-bintray-with-script-demo-2

One last thing to add is that if we're concerned about CI time, we could use the manual step (like we do for the Installable Builds for the mobile apps) while we iron out the performance.

.circleci/config.yml
loremattei: Oh yes, this new version looks clearer to me.

.circleci/config.yml
```diff
@@ -253,6 +253,87 @@ jobs:
...
+    # Running on a Ubuntu machine to have more memory compared to the Docker +
+    # Node image that CircleCI offers (circleci/node)
+    #
+    # See: https://github.com/wordpress-mobile/gutenberg-mobile/pull/3038#issuecomment-765334675
+    machine:
+      image: << pipeline.parameters.linux-machine-image >>
+    steps:
+      - checkout
+      - checkout-submodules
+      - npm-install
```
loremattei: The name is misleading :-) This actually runs `npm ci` as expected 👍

.circleci/config.yml
loremattei: I probably miss some context about the workflow we want to follow 😅, but I wonder if simply adding a filter to the workflow to make it run only on `develop` and `trunk` branches would work here. 
IIRC PRs _should_ be built regardless of any filters, so you shouldn't need this check at all. 

To answer your question: you can exit any job gracefully using `circleci-agent step halt`.

.gitmodules
```diff
@@ -1,6 +1,6 @@
 [submodule "gutenberg"]
 	path = gutenberg
-	url = ../../WordPress/gutenberg.git
+	url = git@github.com:oguzkocer/gutenberg.git
```
loremattei: I'm pretty sure this is intentional to test this PR, but I'm leaving this comment just to make sure this is not something which fell between the cracks :-)

.circleci/config.yml
ceyhun: > Keen to hear what you think of this approach. cc @ceyhun.

@mokagio That looks great and seems to also have very few commands! 
I wanted to also mention that have the [android-device-checks](https://github.com/wordpress-mobile/gutenberg-mobile/blob/89a07397df85a6a4f116f81865f4cb8cfc4a446b/.circleci/config.yml#L114-L137) job where we, with a counterintuitive approach, ran a docker container inside the Ubuntu machine just to build the APK and carried out other tasks back in Ubuntu. It allowed us to overcome the memory issues and also to skip installing Android tooling. Not sure if that makes sense here, or even if it can speed things up 🙂 One piece to improve in that job could be `sudo chown -R circleci:circleci .` part to only include the folders that are really needed.

.gitmodules
mokagio: Correct & great idea!

.circleci/config.yml
mokagio: > I probably miss some context about the workflow we want to follow 😅, but I wonder if simply adding a filter to the workflow to make it run only on develop and trunk branches would work here.

To be honest, I'm not sure what workflow we want to follow exactly either. @mchowning @cameronvoell do you mind chiming in on this? What makes more sense for the way Gutenberg mobile operate? cc @oguzkocer 

I'm also curious of how we should version the binaries built from `develop` or `trunk`. In WordPress-Utils-Android, we have a convention to always bump the version before merging into `develop` or `trunk`. Is there something similar here? Otherwise, we might want to start the AAR publishing from a tag.

I think versioning is an important but orthogonal problem. We can merge this and start using it in the real world even if the versioning workflow is not ironed out.

> IIRC PRs should be built regardless of any filters, so you shouldn't need this check at all.

Yeah, that's correct. That checks is to skip publishing packages for branches that are pushed without a PR open, for which CI starts a build anyway. The problem with publishing a package for these branches, which I estimate are either test branches or branches in the early WIP state, is that if we don't prepend the PR number to the version, then we don't have a way to automatically delete them with [the garbage collector](https://github.com/Automattic/bintray-garbage-collector/).

mokagio: @loremattei, also, thanks for this

> To answer your question: you can exit any job gracefully using `circleci-agent step halt`.

I knew there was something like that but couldn't find where we discussed about it.

.circleci/config.yml
```diff
@@ -253,6 +253,104 @@ jobs:
...
+          name: Evaluate whether to run
+          command: |
+            # This is a partial duplication of the logic to compute the version
+            # to use for the AAR from below. It's useful to have it here
+            # because there are conditions under which we shouldn't run the job
+            # that can't be captured in a CircleCI filter.
+            #
+            # Exiting here saves a non-trivial amount of CI time because of how
+            # heavy the repo is to checkout (it takes at least 1 minute) and
+            # how time consuming running `npm ci` is.
```
mokagio, line 281: As per comment, I tried having this in a dedicated script, but the repo checkout takes at least 1 minute.

On the other hand, the builds that would actually exit from here are a small % of the total builds running on this repo 🤔 Maybe by taking that into consideration the cost in CI has less weight than the duplication cost here?

Keen to hear your thoughts. It's a straightforward change to make, but it might be best to make it in a followup PR, just so we can get this moving along.

.circleci/config.yml
```diff
@@ -253,6 +253,104 @@ jobs:
...
+              echo "Build initiated from a tag: proceeding..."
+            fi
+      - checkout
+      - checkout-submodules
+      - run:
+          # Setting up Android before fetching the Node dependencies because
+          # this step is faster, so if it fails we'll learn about it sooner and
+          # avoid wasting cycles.
+          name: Setup Android Tooling
+          command: .circleci/setup-android-on-ubuntu.sh
```
mokagio, line 303: @ceyhun suggested looking into spinning up a Docker container to avoid this manual setup.

I definitely want to look into it but opted for this approach which I already know works for this first version (which has already been dragging along for a while 😞 )

.circleci/config.yml
```diff
@@ -253,6 +253,104 @@ jobs:
...
+                # CIRCLE_SHA1 should always be available.
+                VERSION="$PR_NUMBER-$CIRCLE_SHA1"
+              elif [[ $CIRCLE_BRANCH == "develop" ]]; then
+                VERSION="$CIRCLE_BRANCH-$CIRCLE_SHA1"
+              else
+                # This happens on the first push of a new branch, when there
+                # isn't a PR open for it yet.
+                echo "Running on a feature branch with no open PR."
+                echo "The checks previous to this step should not have let this happen."
+                exit 1
```
mokagio, line 345: I decided to fail the build here because if this happen there's something wrong in the setup and we should know about it sooner rather than later.

.circleci/config.yml
```diff
@@ -301,6 +399,15 @@ workflows:
           filters:
             branches:
               only: /^dependabot/submodules/.*/
+      - android-build-and-publish-react-native-bridge:
+          name: Build Android RN Bridge & Publish to Bintray
+          filters:
+            branches:
+              # Don't run on trunk, otherwise, run on every other branch
+              ignore: trunk
```
mokagio, line 407: The reason I set this to not run on `trunk` is that from my understanding of our workflows discussions, we never need to access a build from that branch.

.circleci/config.yml
```diff
@@ -253,6 +253,104 @@ jobs:
...
+            if [[ -z ${CIRCLE_TAG+x} ]]; then
+              if [[ -n "$CIRCLE_PULL_REQUEST" ]]; then
+                echo "Build initiated from a branch with a pull request: proceeding..."
+              elif [[ $CIRCLE_BRANCH == "develop" ]]; then
+                echo "Build initiated on $CIRCLE_BRANCH: proceeding..."
+              else
+                # We end up here on the first push of a new branch, when there
+                # isn't a PR open for it yet.
+                echo "Build initiated from a feature branch with no pull request: aborting."
+                circleci-agent step halt
```
mokagio, line 291: [This build](https://app.circleci.com/pipelines/github/wordpress-mobile/gutenberg-mobile/10976/workflows/b36695d1-4e5b-4fe1-9333-da8769a99a7d/jobs/57726) is an example of this scenario

image

Unfortunately, CircleCI doesn't treat halts differently and the UI might be misleading as it looks like the build passed.

.circleci/config.yml
```diff
@@ -301,6 +399,15 @@ workflows:
...
               only: /^dependabot/submodules/.*/
+      - android-build-and-publish-react-native-bridge:
+          name: Build Android RN Bridge & Publish to Bintray
+          filters:
+            branches:
+              # Don't run on trunk, otherwise, run on every other branch
+              ignore: trunk
+            tags:
+              # Run for every tag
+              only: /.*/
```
mokagio, line 410: Here's an example of build from a tag: [`mokagio-test-20210317-173215`](https://app.circleci.com/pipelines/github/wordpress-mobile/gutenberg-mobile/10980/workflows/2b2eaf57-d638-42ce-80bf-def2720ba820/jobs/57750).

mokagio (commented):
@mchowning @cameronvoell @ceyhun @loremattei here's an updated version of the workflow that incorporates the versioning workflow discussed internally and most of your suggestions. 

Can I ask you to have another look?

I updated [the test PR on WordPress Android](https://github.com/wordpress-mobile/WordPress-Android/pull/14215) to use the binary from [the latest CI build](https://app.circleci.com/pipelines/github/wordpress-mobile/gutenberg-mobile/10979/workflows/d358107d-abde-482e-b5be-7bef619b418c/jobs/57748), if you want to verify it builds and works.

.circleci/config.yml
mchowning: I believe that all of our release tags are made on `trunk`. I may just not be understanding how this CI job is used.

image

mokagio: Thanks @mchowning!

Every time a tag is pushed, this workflow will run and upload an AAR with the version being the tag name. See [this build ](https://app.circleci.com/pipelines/github/wordpress-mobile/gutenberg-mobile/10980/workflows/2b2eaf57-d638-42ce-80bf-def2720ba820/jobs/57750) for an example: it started from GitHub receiving the tag `mokagio-test-20210317-173215` and uploaded [a binary to Bintray with version `mokagio-test-20210317-173215`](https://bintray.com/wordpress-mobile/maven/react-native-gutenberg-bridge/mokagio-test-20210317-173215).

Do you ever need to reference in WordPress Android a commit from gutenbreg-mobile's `trunk` that has not a matching tag? If that's the case, I can easily make builds on `trunk` behave like builds on `develop`: publish a `trunk-< commit SHA1 >` version to Bintray.

mchowning: > Do you ever need to reference in WordPress Android a commit from gutenbreg-mobile's trunk that has not a matching tag? 

I don't think so. As long as all the tags get pushed we should be fine. 👍

.circleci/config.yml
ceyhun: Thanks @mokagio 🙇  This already looks awesome and we can surely improve it later once we have it in place and working first 👍

oguzkocer (approved):

mokagio:
@cameronvoell @mchowning @oguzkocer the tests still fail, even after syncing with `develop` and the latest https://github.com/WordPress/gutenberg/pull/29488, which is synced with Gutenberg's `trunk`.

This is the failure on both steps:

```
FAILURE: Build failed with an exception.

* What went wrong:
Could not determine the dependencies of task ':app:testDebugUnitTest'.
> Could not resolve all task dependencies for configuration ':app:debugUnitTestRuntimeClasspath'.
   > Could not find org.wordpress-mobile:hermes-debug-mirror:0.61.5.
     Searched in the following locations:
       - file:/home/circleci/.m2/repository/org/wordpress-mobile/hermes-debug-mirror/0.61.5/hermes-debug-mirror-0.61.5.pom
       - file:/home/circleci/project/gutenberg/node_modules/react-native/android/org/wordpress-mobile/hermes-debug-mirror/0.61.5/hermes-debug-mirror-0.61.5.pom
       - https://dl.google.com/dl/android/maven2/org/wordpress-mobile/hermes-debug-mirror/0.61.5/hermes-debug-mirror-0.61.5.pom
       - https://jcenter.bintray.com/org/wordpress-mobile/hermes-debug-mirror/0.61.5/hermes-debug-mirror-0.61.5.pom
       - https://jitpack.io/org/wordpress-mobile/hermes-debug-mirror/0.61.5/hermes-debug-mirror-0.61.5.pom
     Required by:
         project :app > project :android:react-native-bridge
```

<details>
<summary>diff of the error output between the two steps to show it's the same</summary>

```diff
4,5c4,5
< Could not determine the dependencies of task ':app:testDebugUnitTest'.
< > Could not resolve all task dependencies for configuration ':app:debugUnitTestRuntimeClasspath'.
---
> Could not determine the dependencies of task ':app:processDebugResources'.
> > Could not resolve all task dependencies for configuration ':app:debugRuntimeClasspath'.
```

</details>

🤔 

I bet the issue has to do with this project not having the S3 repository where our mirror lives. Let's see if I can do something about it...

mokagio:
All green in CI 🎉 Thanks @oguzkocer for the help.

I'm going to turn this "Ready for review" even though the "_Before merging this PR, we need to merge WordPress/gutenberg#29488 into Gutenberg's `trunk` and update the submodule has here_" hasn't been checked yet.

@mchowning @cameronvoell let me know if you think we should merge before or after the Gutenberg counterpart lands into Gutenberg's `trunk`. I'm writing this in my Friday EOD, so feel free to take the PR over and merge it if you want, without waiting for me.

mchowning (approved):

PR #3253, opened 2021-03-12
Merge release 1.48.1 to develop

fluiddot:
The purpose of this PR is to merge into `develop` all the changes of `1.48.1` release.

PR submission checklist:


fluiddot:
It's weird that after generating the bundle it added so many new localized strings 🤔 . I tried to generate the bundle in `develop` branch and generated the same output so I think it's safe to keep them.

jd-alexander:
@fluiddot just verifying here, that there does not need to be any Gutenberg ref bumps in the diff.

fluiddot:
> @fluiddot just verifying here, that there does not need to be any Gutenberg ref bumps in the diff.

That's a good question, I decided to leave the same Gutengerg reference we currently have in `develop`, but maybe it makes sense now that we have the Gutenberg PR merged to point to that commit, wdyt?

jd-alexander:
> That's a good question, I decided to leave the same Gutengerg reference we currently have in develop, but maybe it makes sense now that we have the Gutenberg PR merged to point to that commit, wdyt?

Yeah, it wouldn't hurt. But also when I think about it, the next gb-mobile PR that's created will be based on `trunk` in `gutenberg` so that is going to bring in the changes due to it being a commit hash that after the `1.48.1` merge to `trunk` so it should be fine either way.

jd-alexander (approved):
Looking good. LGTM 🚢

PR #3301, opened 2021-03-23
Fix broken documentation links

dcalhoun:
The stale links likely referenced files relocated during the monorepo merge. The example test file used is copied from the documentation found in WordPress/gutenberg.

To test:
1. Open relevant documentation files on GitHub.
1. Click changed links to verify destination is correct.

PR submission checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,7 +121,7 @@ Then, open `chrome://inspect` in Chrome to attach the debugger (look into the "R
 
 ## Writing and Running Unit Tests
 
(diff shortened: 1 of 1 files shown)


guarani (approved):
Thanks for updating, I see just one other link that's not working, which is the link for the "the one in the Gutenberg project" text in the last section. I'm guessing the correct link should be:
https://github.com/WordPress/gutenberg/blob/trunk/packages/eslint-plugin/README.md

While we're at it, we could change the license link, which is at the bottom of this page, from a specific commit to a branch, wdyt?

PR #3326, opened 2021-03-31
Update houndci to support es6

enejb:
Updates hound ci to support es6. 

Currently, hound ci doesn't seem to support es6. See https://github.com/wordpress-mobile/gutenberg-mobile/pull/2582

I am hoping this PR properly configures hound ci so that when it runs it validates es6 syntext as expected. 

To test:
Does hound CI work as expected? 

PR submission checklist:


.hound.yml
```diff
@@ -0,0 +1,4 @@
+javascript:
+  enabled: false
+eslint:
+  enabled: true
```
dcalhoun, line 4: [This Hound documentation](https://intercom.help/hound/en/articles/2138530-eslint) would imply we need to add a pointer to our own eslint configuration file. Did you find Hound worked without explicitly adding the path?
```suggestion
eslint:
  enabled: true
  config_file: .eslintrc.json
```

dcalhoun (commented):
I'm not highly familiar with HoundCI configuration, but overall this approach seems sound to me.

>We currently only support JSON or YAML formats for .eslintrc since we parse and merge configs.

I was puzzled by the move from `.eslintrc.js` to `.eslintrc.json`, but after finding the related [documentation](https://intercom.help/hound/en/articles/2138530-eslint) and issues (houndci/hound#1630, houndci/hound#1793) it makes sense now.

>Currently, hound ci doesn't seem to support es6. See WordPress/gutenberg#29670

Was the above from your PR description intended to link to https://github.com/wordpress-mobile/gutenberg-mobile/pull/2582#pullrequestreview-620289101? The current link didn't appear to include any references to Hound.

.hound.yml
enejb: I can't say I am able to confirm either way.

enejb:
It looks like hound currently doesn't support plugins. So we can't have the same eslint config as before. 
See https://github.com/houndci/hound/issues/1838

dcalhoun:
> It looks like hound currently doesn't support plugins. So we can't have the same eslint config as before.
> See [houndci/hound#1838](https://github.com/houndci/hound/issues/1838)

@enejb bummer. So, what do you think next steps of exploration are? I am a bit confused as to the state of things. 

I suppose prior to https://github.com/wordpress-mobile/gutenberg-mobile/pull/3326/commits/39c1c072d2fd3ecad46ebbadabaf959c4a8462d7 Hound was relying upon [the default](https://intercom.help/hound/en/articles/2138530-eslint), built-in [ESLint config](https://gist.githubusercontent.com/gylaz/fd520c53a49b3b0f736b324309d50862/raw/eed4c46d7cb61db910992986ad951d5a49b2c664/eslintrc). I suppose the alternative path forward could be to revert https://github.com/wordpress-mobile/gutenberg-mobile/pull/3326/commits/39c1c072d2fd3ecad46ebbadabaf959c4a8462d7 and rely upon that? The downside would be that Hound wouldn't really be checking our code against the ESLint rules we use. It would, presumably, unblock your issue though, correct?

enejb:
I think we should be more explicit about the rules that we want to check. It is a bummer that we can't have the plugins that we want but I think having a eslint config that we use specifically for Hound would help us control that better vs using the default. 

I still wasn't able to replace the error that I am seeing in https://github.com/wordpress-mobile/gutenberg-mobile/pull/2582. 
I was hoping that https://github.com/wordpress-mobile/gutenberg-mobile/pull/3326/commits/ab05abd13c631f4e60dc0d2d321bfdeaad555a39 would trigger it.

Do you understand what the difference between hound and the "Check Correctness" tests that seems to run `eslint . --ext .js` command are? 

How are they different? If they serve the same purpose maybe we can disable hound for js files? 
Since to me they seem to be taken care of by `Check Correctness` already?

dcalhoun:
> I think we should be more explicit about the rules that we want to check. It is a bummer that we can't have the plugins that we want but I think having a eslint config that we use specifically for Hound would help us control that better vs using the default.

@enejb agreed. I'm just hesitant to maintain a duplicate ESLint config to mimic `@wordpress/eslint-plugin`. I fear that it would become outdated quickly. 

> I still wasn't able to replace the error that I am seeing in #2582.
> I was hoping that [ab05abd](https://github.com/wordpress-mobile/gutenberg-mobile/commit/ab05abd13c631f4e60dc0d2d321bfdeaad555a39) would trigger it.

My thought is that the original import error doesn't show up currently because that after https://github.com/wordpress-mobile/gutenberg-mobile/pull/3326/commits/39c1c072d2fd3ecad46ebbadabaf959c4a8462d7 Hound is now throwing [the plugin error](https://github.com/wordpress-mobile/gutenberg-mobile/pull/3326#pullrequestreview-625710654) prior to hitting the ES6 import error. 

Prior to https://github.com/wordpress-mobile/gutenberg-mobile/pull/3326/commits/39c1c072d2fd3ecad46ebbadabaf959c4a8462d7, my thought is that the error no longer showed up because your PR included [Hound configuration to enable ESLint](https://github.com/wordpress-mobile/gutenberg-mobile/blob/a0d2c4bab4186f0f64f8c31a07b220104da9a644/.hound.yml#L4), which included [the default ESLint config](https://gist.githubusercontent.com/gylaz/fd520c53a49b3b0f736b324309d50862/raw/eed4c46d7cb61db910992986ad951d5a49b2c664/eslintrc) that enabled ES6.  
 
> Do you understand what the difference between hound and the "Check Correctness" tests that seems to run `eslint . --ext .js` command are?
> 
> How are they different? If they serve the same purpose maybe we can disable hound for js files?
> Since to me they seem to be taken care of by `Check Correctness` already?

No, I don't fully understand why we would have both Hound and "Check Correctness" for JavaScript. Given Hound's lack of support for ESLint plugins, I wonder if the shortest and best path forward is to disable checking JavaScript with Hound. That would leave "Check Correctness" in place that leverages `@wordpress/eslint-plugin`.

Thoughts?

enejb:
My vote would be to disable it for now. 
But I don't really have a good understanding of the differences. 
maybe @ceyhun, @cameronvoell, and @hypest might have a better idea on the next steps here.

enejb:
Since we disabled Hound I am closing this PR.

dannyrb:
FWIW, the hound maintainers have been fairly responsive when contacted. Its mostly a matter of them "allow listing" plugins

PR #3372, opened 2021-04-13
Add/npm experimental block grid block

enejb:
This PR is an experiment and an alternative to the https://github.com/wordpress-mobile/gutenberg-mobile/pull/2582. 

It does the following. 
1. It adds the https://github.com/Automattic/block-experiments/pull/136 PR repo into the npm_modules folder. So that it can be accessed by the mobile editor. 
2. It adds a new `npm run start:block-experiments` command that adds a parallel folder called `block-experiments` to metro watch folder list. So that we develop in the `block-experiment` repository that lived parallel to the `mobile-gutenberg` repo. 

To test:
1. Load this PR. Run `npm install`
2. Run `npm run start:reset` notice that you now see the `layout grid` block in the block inserter. You are able to insert the block into the page.
3. Clone the Block Experiment repo in a parallel folder to gutenberg-mobile. 
```
-- gutenberg-mobile
-- block-expriements
```
with  `git clone git@github.com:Automattic/block-experiments.git` 
4. Switch to the `add-mobile-app-support` branch in the `block-experiments` repo. Make a change to the code base so that you will notice that this is the repo being used. 
Such as updating the name of the block on https://github.com/Automattic/block-experiments/blob/master/blocks/layout-grid/src/index.js#L45 
5. Run `npm run start:block-experiments` from the gutenberg-mobile 
6. Notice that the layout grid block now has a new different name.

PR submission checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,7 +50,8 @@
 		"sprintf-js": "^1.1.1",
 		"typescript": "4.1.3",
 		"wd": "^1.11.1",
-		"yarn": "^1.22.10"
+		"yarn": "^1.22.10",
+		"block-experiments": "Automattic/block-experiments#add-mobile-app-support"
 	},
 	"scripts": {
 		"postinstall": "npm ci --prefix gutenberg && cd jetpack/projects/plugins/jetpack && yarn install --frozen-lockfile --ignore-engines",
--- a/package.json
+++ b/package.json
@@ -58,7 +58,7 @@
 		"start": "echo \"\\x1b[33mThe start command is not available in this project. It is strongly recommended to use \\x1b[1:33mstart:reset\\x1b[0m\\x1b[33m to perform some cleanup when starting the metro bundler.\nOr you may use \\x1b[1:33mstart:quick\\x1b[0m\\x1b[33m for a quicker startup, but this may lead to unexpected javascript errors when running the app.\\x1b[0m\"",
 		"start:reset": "npm run core clean:runtime && npm run start:quick -- --reset-cache",
 		"start:quick": "react-native start --config ./metro.config.js",
-		"start:block-experiments": "cross-env LOCAL_REPO=../block-experiments-local react-native start --config ./metro.config.js",
+		"start:block-experiments": "cross-env LOCAL_REPO=../block-experiments react-native start --config ./metro.config.js",
 		"core": "cd gutenberg && npm run native",
 		"prern-bundle": "patch-package --patch-dir gutenberg/packages/react-native-editor/metro-patch",
 		"rn-bundle": "react-native bundle",
--- a/package.json
+++ b/package.json
@@ -58,6 +58,7 @@
 		"start": "echo \"\\x1b[33mThe start command is not available in this project. It is strongly recommended to use \\x1b[1:33mstart:reset\\x1b[0m\\x1b[33m to perform some cleanup when starting the metro bundler.\nOr you may use \\x1b[1:33mstart:quick\\x1b[0m\\x1b[33m for a quicker startup, but this may lead to unexpected javascript errors when running the app.\\x1b[0m\"",
 		"start:reset": "npm run core clean:runtime && npm run start:quick -- --reset-cache",
 		"start:quick": "react-native start --config ./metro.config.js",
+		"start:block-experiments": "cross-env LOCAL_REPO=../block-experiments-local react-native start --config ./metro.config.js",
 		"core": "cd gutenberg && npm run native",
 		"prern-bundle": "patch-package --patch-dir gutenberg/packages/react-native-editor/metro-patch",
 		"rn-bundle": "react-native bundle",
--- a/metro.config.js
+++ b/metro.config.js
@@ -4,6 +4,10 @@ const gutenbergMetroConfigCopy = {
 	...require( './gutenberg/packages/react-native-editor/metro.config.js' ),
 };
 
+if ( process.env.LOCAL_REPO ) {
+	gutenbergMetroConfigCopy.watchFolders.push( path.resolve( __dirname, process.env.LOCAL_REPO ) );
+}
+
 gutenbergMetroConfigCopy.resolver.extraNodeModules = new Proxy(
 	{},
 	{
--- a/package-lock.json
+++ b/package-lock.json
@@ -4803,6 +4803,30 @@
 						"regenerator-runtime": "^0.13.4"
 					}
 				},
+				"@wordpress/element": {
+					"version": "2.20.1",
+					"resolved": "https://registry.npmjs.org/@wordpress/element/-/element-2.20.1.tgz",
+					"integrity": "sha512-aHfv3rnY8yea3VvfabPpyQXuYTQLKMcOSRxG+ydxptQNlQjEBe+KtCXyqr5g7bpqEBT5waFSr3IbJKk+G+hZxA==",
+					"dev": true,
+					"requires": {
+						"@babel/runtime": "^7.13.10",
+						"@types/react": "^16.9.0",
+						"@types/react-dom": "^16.9.0",
+						"@wordpress/escape-html": "^1.12.1",
+						"lodash": "^4.17.19",
+						"react": "^16.13.1",
+						"react-dom": "^16.13.1"
+					}
+				},
+				"@wordpress/escape-html": {
+					"version": "1.12.1",
+					"resolved": "https://registry.npmjs.org/@wordpress/escape-html/-/escape-html-1.12.1.tgz",
+					"integrity": "sha512-moudz8yUXiI2dGi+OWcNxKffbpF7B+XCG0dt9W3HRzSwalWAeo4CNzA5wSve1BYk8wMX8BiSGO+8ew+2Yr+cug==",
+					"dev": true,
+					"requires": {
+						"@babel/runtime": "^7.13.10"
+					}
+				},
 				"diff": {
 					"version": "4.0.2",
 					"resolved": "https://registry.npmjs.org/diff/-/diff-4.0.2.tgz",
@@ -4888,6 +4912,30 @@
 						"regenerator-runtime": "^0.13.4"
 					}
 				},
+				"@wordpress/element": {
+					"version": "2.20.1",
+					"resolved": "https://registry.npmjs.org/@wordpress/element/-/element-2.20.1.tgz",
+					"integrity": "sha512-aHfv3rnY8yea3VvfabPpyQXuYTQLKMcOSRxG+ydxptQNlQjEBe+KtCXyqr5g7bpqEBT5waFSr3IbJKk+G+hZxA==",
+					"dev": true,
+					"requires": {
+						"@babel/runtime": "^7.13.10",
+						"@types/react": "^16.9.0",
+						"@types/react-dom": "^16.9.0",
+						"@wordpress/escape-html": "^1.12.1",
+						"lodash": "^4.17.19",
+						"react": "^16.13.1",
+						"react-dom": "^16.13.1"
+					}
+				},
+				"@wordpress/escape-html": {
+					"version": "1.12.1",
+					"resolved": "https://registry.npmjs.org/@wordpress/escape-html/-/escape-html-1.12.1.tgz",
+					"integrity": "sha512-moudz8yUXiI2dGi+OWcNxKffbpF7B+XCG0dt9W3HRzSwalWAeo4CNzA5wSve1BYk8wMX8BiSGO+8ew+2Yr+cug==",
+					"dev": true,
+					"requires": {
+						"@babel/runtime": "^7.13.10"
+					}
+				},
 				"lodash": {
 					"version": "4.17.21",
 					"resolved": "https://registry.npmjs.org/lodash/-/lodash-4.17.21.tgz",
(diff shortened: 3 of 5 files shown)


mkevins:
Hi Enej :wave: :smile: 

I tried `npm i`, then `npm run start:reset`, but I'm encountering a module resolution issue in bundling:

```
error: bundling failed: Error: Unable to resolve module `@wordpress/element` from `node_modules/block-experiments/blocks/layout-grid/src/grid/edit.native.js`: @wordpress/element could not be found within the project or in these directories
```

Have you encountered this before?

**Edit**

It seems that it's looking in `~/gutenberg-mobile/gutenberg/node_modules/@wordpress`, which does _indeed_ have an `element` directory, but maybe it's looking for `@wordpress/element` within that `@wordpress` directory for some reason :thinking: I'm curious what is different about our environments that it is working well for you. I did not try yet with `npm run start:block-experiments`, so perhaps that is a clue?

chipsnyder:
I'm running into similar errors as @snapfractalpop 

```
error: bundling failed: Error: Unable to resolve module `@wordpress/element` from `node_modules/block-experiments/blocks/layout-grid/src/grid/save.js`: @wordpress/element could not be found within the project or in these directories:
  {path to my directory}/gutenberg-mobile/gutenberg/node_modules/@wordpress
```
There weren't many options available to me to try and debug the issue we've had problems with `<div>` in the past which is [used by the referenced file  ](https://github.com/Automattic/block-experiments/blob/add-mobile-app-support/blocks/layout-grid/src/grid/save.js#L31) so I'm not sure if that's connected or not.

package.json
```diff
@@ -50,13 +50,15 @@
 		"sprintf-js": "^1.1.1",
 		"typescript": "4.1.3",
 		"wd": "^1.11.1",
-		"yarn": "^1.22.10"
+		"yarn": "^1.22.10",
+		"block-experiments": "Automattic/block-experiments#add-mobile-app-support"
```
hypest, line 54: 👋 Enej! I'm generally weary of using dependency refs that are "moving" and can lead to un-reproducible builds when new commits have landed in the referenced branch.

I have the feeling that maybe this is an attempt to bypass another problem though, the one about how to put the proper/desired commit hash in here anyway, in particular how to maintain it as one works on the branch. 🤔

mkevins: I can't speak for Enej's intentions here, but my undestanding of a workflow here is that a package reference could be "floating" on an in-progress PR, but that ultimately, hashes, or better yet, tags would be used for whatever is committed in the dependency, and updated w/ semver accordingly in the related PR. Also, assuming the configuration for npm dependencies can be made to work, I think if this "floating" approach would be accompanied by lock file versioning, which could serve as a signal too, for reviewers to run `npm i` (similar to how we currently run `git submodule update`).

mkevins:
I tinkered a bit more with this configuration, first by actually installing `@wordpress/element` in the experiments package (tried as both a dep and peer dep) and this had no effect. It also occurred to me that `element` might be treated differently, since it is included in a pragma plugin for "generating import statements" (since `@wordpress/block-editor` was imported first, but triggered no such error). On that hunch, I tried "manually" resolving the package via the `extraNodeModules` proxy getter, which then resulted in a new resolution error for `@wordpress/warning`. Some kind of progress :laughing: . The plot thickens: `@wordpress/warning` is _also_ included in the pragma plugin. So, manually resolving that get's us to a new error.

With these changes, it builds:

<details>
<summary>Patch</summary>

```diff
diff --git a/metro.config.js b/metro.config.js
index 3c082b7c..13772874 100644
--- a/metro.config.js
+++ b/metro.config.js
@@ -11,8 +11,20 @@ if ( process.env.LOCAL_REPO ) {
 gutenbergMetroConfigCopy.resolver.extraNodeModules = new Proxy(
 	{},
 	{
-		get: ( target, name ) =>
-			path.join( process.cwd(), `gutenberg/node_modules/${ name }` ),
+		get: ( target, name ) => {
+			switch ( name ) {
+				case `@wordpress/element`:
+					return path.join( process.cwd(), `gutenberg/packages/element` );
+				case `@wordpress/warning`:
+					return path.join( process.cwd(), `gutenberg/packages/warning` );
+				case `@wordpress/react-native-bridge`:
+					return path.join( process.cwd(), `gutenberg/packages/react-native-bridge` );
+				case `@wordpress/react-native-aztec`:
+					return path.join( process.cwd(), `gutenberg/packages/react-native-aztec` );
+				default:
+					return path.join( process.cwd(), `gutenberg/node_modules/${ name }` );
+			}
+		},
 	}
 );
 ```
</details>

But, I'm not convinced it's working correctly, since, for example, I did not see the layout grid block in the inserter. :thinking: Also, I noticed that the `gutenberg/node_modules/@wordpress` directory has lots of symlinks to `../../packages/moduleName`, so I'm sure I'm missing some context on how metro is resolving these currently. In any case, maybe this patch will provide additional clues, if nothing else.

**Comment by hypest** [comment]
> Also, I noticed that the gutenberg/node_modules/@wordpress directory has lots of symlinks to ../../packages/moduleName, so I'm sure I'm missing some context on how metro is resolving these currently

FWIW, from my [past experience with symlinks](https://github.com/wordpress-mobile/gutenberg-mobile/pull/154), Metro is usually very smart and finds packages in any nested folder even without directly pointing to those in a config file, but Jest isn't that smart. Not sure if in this case the symlinks are maintained for Jest?

File: package.json (hunk 1 above)
Comment (hypest) [reply package.json:54]: Indeed, the process moves to maintaining the lock file instead of the package.json file. I feel that that workflow is more fragile, especially since the lock file can actually get out of hand with all the changes that can happen in it. Having to rely on the lock file becoming essentially a source file is awkward, but I understand it might just come down to personal preference.

Comment (enejb) [reply package.json:54]: @hypest Thanks for bringing your concerns up. I also share them. :) 

This PR was definitely an experimental approach. Trying to see what it would take to make things work. So I thought it would be ok to use the branch name for now. 

Like @mkevins suggested I think we would want to manage the code with a tag or a commit hash.

An example of how this is currently done can be found in calypso. 
https://github.com/Automattic/wp-calypso/blob/trunk/apps/editing-toolkit/package.json#L143
In order to update the package. The process seems to be that a new tag gets created on the and then a new commit in the calypso repo. [For example](https://github.com/Automattic/wp-calypso/commit/b3d3062e61115fe35a55f577c8eaa9d86cfaa254#diff-6fe967ae8f9263d474fe549abca821d4e9835e8488601853df9a1a26ace570af)

The current release process of the block-experiments repo already does use tags and has a build process to create the different plugins that eventually end up on WordPress.org. I think it would be valuable to see if we can integrate into this build process to create new versions of the plugin that then gets included in our repo. 

We could even try to use the .org repo directly by using links such as 
https://downloads.wordpress.org/plugin/layout-grid.1.5.zip. 

The first step there would be to add a `package.json` in that zip. I just tried adding 
"layout-grid": "https://downloads.wordpress.org/plugin/layout-grid.1.5.zip" to our `packages.json`
and got the following error. 

```
npm ERR! code ENOPACKAGEJSON
npm ERR! package.json Non-registry package missing package.json: layout-grid@https://downloads.wordpress.org/plugin/layout-grid.1.5.zip.
npm ERR! package.json npm can't find a package.json file in your current directory.
```

However, some of the concerns about properly versioning things are definitely a concern. 
I am not sure how to best address them besides more developer education. Making sure that we are on the same page about how we release software. Such as creating releases and versioning things. 
For example, code published in NPM doesn't necessarily reflect a particular Github tag.

hypest: > accompanied by lock file versioning, which could serve as a signal too, for reviewers to run `npm i`

👋 @mkevins , just to be sure, do you happen to mean `npm ci` (instead of just `i`) there so to ensure the lock file is enforced?

> Making sure that we are on the same page about how we release software.

Good point. I'd still file my concerns under the "development flow" actually, not the "release flow" per se, though it's true that when dealing with libraries spread out in different repos/infrastructures, there's an element of "releasing" that happens anyway during developing. Generally, I'm leaning more towards having non-standard release flows if that serves the development flows better, since these days development still happens way more oftenly than releasing. When the project matures enough that development has "concluded" then we should lean to making releasing easier instead. Again, this is just my thought process, it doesn't mean we should abandon experimenting like here in this PR 👍.

mkevins: :wave: @hypest 

> just to be sure, do you happen to mean `npm ci` (instead of just `i`) there so to ensure the lock file is enforced

Well, it depends. When "mapping" git submodules and npm dependencies conceptually, I look at the lock file as similar to the git submodule hash, perhaps garnering a bit more semantic information from the dependency graph. For reproducibility (especially CI) `npm ci` is perfect, since it treats the lock file as the source of truth.

`npm i`, on the other hand, allows the versions to "float", which could be handy if used with that intention (or better yet, `npm up <some-package>`), but when used without deliberation can be a source of unwanted "drift". For example, different versions of `npm` may resolve the graph in slightly different ways from a clean state, and so things like `.nvmrc` and `nvm use` become helpful for consistency. In the case where any versions are updated, this would be signalled by diff in the lock file.

mkevins:
> FWIW, from my [past experience with symlinks](https://github.com/wordpress-mobile/gutenberg-mobile/pull/154), Metro is usually very smart and finds packages in any nested folder even without directly pointing to those in a config file, but Jest isn't that smart. Not sure if in this case the symlinks are maintained for Jest?

Thanks for linking to that for context! I see that even back then, it seems that `@wordpress/element` was treated differently. I suspect this could be to do with the order in which things are resolved, because of [its inclusion in the pragma plugin](https://github.com/WordPress/gutenberg/blob/1de0d7192509281eb050a76df7b52fb41e036724/packages/babel-preset-default/index.js#L68), but I haven't investigated that deeply enough to be sure. There is also [this logic](https://github.com/WordPress/gutenberg/blob/1de0d7192509281eb050a76df7b52fb41e036724/webpack.config.js#L43) which removes the `@wordpress` scope in generating entry points. Packages beginning with `react-native` are omitted (two lines up) - these same packages are in the patch above, so I think it is somehow related. :thinking:

package.json
hypest: Oh I see. Yeah, intentionality is key. I'd expect folks to invoke `npm i` exactly when the intention is to allow for upgrades.

I'm trying to think of the PR review flow and who and why would want to upgrade a dep while the PR is in progress. I understand that the author has of course perfect intention to upgrade when needed. The reviewer most probably not, unless they are helping with updating the PR, otherwise I'd assume the reviewer will always go for `npm ci` to pull what the author intents to merge and nothing more.

In that view, the lock file does indeed play the same role as the submodule hash; it will actually be a hash pointer itself too for most of the PR's lifecycle :).

I guess my concern with the lock file is that it tends to pick up changes in other (3rd party) deps, obscuring the intentional "dep bump" change. That is, it can make the reviewing harder. Contrast that to a submodule bump where the change is super isolated and easy to spot.

I wonder if we can work around that "busyness" of the lock file by using https://docs.npmjs.com/cli/v7/commands/npm-shrinkwrap to pinpoint exactly the dependency versions/bumps we want for the particular deps. Like, essentially we want some deps to be closer to the project than the average dep and subdep, and perhaps maintaining a list for that special set of "1st connection" deps in the shrinkwrap file is a better indication of intentionality?

I still kinda wonder what shortcoming of the git submodules we are trying to sidestep but, yeah, what do y'all think about that shrinkwrap idea above?

mkevins: > I'm trying to think of the PR review flow and who and why would want to upgrade a dep while the PR is in progress. I understand that the author has of course perfect intention to upgrade when needed. The reviewer most probably not, unless they are helping with updating the PR, otherwise I'd assume the reviewer will always go for `npm ci` to pull what the author intents to merge and nothing more.

There are a lot of options here. The one most similar to the current submodule hash update would be to use `npm ci`. In that flow, all dependencies are "locked" in place (transitive deps as well as the package under development). In that case, the PR author will need to always update the lock-file in the associated `gutenberg-mobile` PR, just as we do currently with the submodule hash.

Another idea is to allow the commitish to float, and the source of truth be "upstream". In that case, the PR author need not update the `gutenberg-mobile` PR (unless there are source changes there), and instead the experiments branch would be referenced. In that case, to pull the latest, the reviewer would run `npm up block-experiments`. This would also bring in updates to transitive dependencies for that package, if they are specified as floating there, and those changes (if they exist) will be visible in the lock-file diff. This could be desirable, for example, if the direct dependency (in this case `block-experiments`) specified only patch versions to float (i.e. `~` and not `^`), as it would bring bug fixes and sometimes important security updates.

On the other hand, exact versions could be specified in the direct dependency as well. Since this is under our control, this offers a different kind of flexibility, which may serve us well in scaling, since the choice to allow a version to float is very granular. I can imagine a scenario where the direct dependency has a related (transitive) dependency under development, and in that case, it could be helpful to have everything locked _except_ that dependency. The depth of this is entirely configurable on a PR by PR basis, which is an attractive quality.

Regarding `npm-shrinkwrap.json`, that is also a configuration to keep in mind. As I understand it, when we have specified an _exact_ version in our `package.json` file (i.e. a direct dependency), we are still not guaranteed a deterministic outcome via `npm i`, since the transitive dependencies that package brings (hoisted or not) can be specified as floating in the `package.json` of the direct dependency. While, in this case, this is entirely in our control, shrinkwrap offers locking of the transitive dependencies, even without locking them with exact versions in the `package.json` of the direct dependency. I see this as more of a feature for library development though, to ensure a library version is deterministic, which is crucial for release versions.

> I still kinda wonder what shortcoming of the git submodules we are trying to sidestep

I don't think we're trying to sidestep any shortcomings with git submodules; that approach has been working well. I think the idea here is to explore another avenue for dependency management, which offers some interesting options / flexibility which may also serve us well as we scale.

Issue #3104, opened 2021-02-04
Switching from Solid to Gradient doesn't scroll to end when having customised gradient


fluiddot:
**Describe the bug**
Related to test case: [Color Settings - Test Cases - TC003 - Scroll color palette to the end](https://github.com/wordpress-mobile/test-cases/blob/trunk/test-cases/gutenberg/color-settings.md#scroll-color-palette-to-the-end).

When having a customised gradient, switching from Solid to Gradient doesn't scroll the colors list to the end.

**To Reproduce**
1. Go to a post/page
2. Add `Buttons` block
3. Tap on gear button
4. Tap on `Background color` option
5. Tap on `Gradient` option
6. Tap one gradient color
7. Tap on `Customize Gradient`
8. Change `Gradient Type` to `Radial` (it also works changing other options)
9. Tap on back arrow button
10. Tap on `Solid` option
11. Tap on `Gradient` option again
12. Observe that the scroll doesn't go to the end where the custom gradient is shown

**Expected behavior**
The scroll should go to to the end showing the custom gradient color when switching between Solid and Gradient options.

**Screenshots**

https://user-images.githubusercontent.com/14905380/106886020-b417cc80-66e3-11eb-97ad-71cb90db8656.mp4

**Smartphone (please complete the following information):**
 - Device: Samsung Galaxy S20 FE 5G
 - OS: Android 10
 - Version 16.5

chipsnyder:
Related to https://github.com/wordpress-mobile/gutenberg-mobile/issues/2882

dcalhoun:
@chipsnyder @fluiddot I have found that **_the "Custom" button is now no longer displayed at all when customizing the gradient_**, in addition to not scrolling to the end. I was about to create a new issue, but then found this one. Happy to create a separate issue if that is more appropriate. 

<details>
<summary>Details of missing "Custom" button issue</summary>

## Description
When customizing the gradient of a block, the color list does not add the "Custom" button to the end of the color list. It also does not remove the "Custom" button when switching from a customized gradient to a predefined gradient. This was found in the steps within the [GB Mobile TC005 color settings test case](https://github.com/wordpress-mobile/test-cases/blob/trunk/test-cases/gutenberg/color-settings.md#tc005). 

## Step-by-step reproduction instructions
1. Launch block editor on WPiOS or WPAndroid. 
2. Add a Button block. 
3. Open Button block settings. 
4. Tap Background Color. 
[...]

PR #3394, opened 2021-04-20
Fixes color picker segmented control rendering and scrolling issues

antonis:
**Fixes**: https://github.com/WordPress/gutenberg/issues/30878 and https://github.com/wordpress-mobile/gutenberg-mobile/issues/3104

`gutenberg`: https://github.com/WordPress/gutenberg/pull/30994
`WordPress-Android`: https://github.com/wordpress-mobile/WordPress-Android/pull/14492
`WordPress-iOS`: https://github.com/wordpress-mobile/WordPress-iOS/pull/16366

To test:
Follow the steps on the [`gutenberg` pr](https://github.com/WordPress/gutenberg/pull/30994)

PR submission checklist:


cameronvoell:
Hey @antonis. 👋🏻 We will cut the 1.52.0 release Friday 30 April. Please let me know if you're aiming to include this PR in 1.52.0. Otherwise perhaps we can add the milestone label for 1.53.0 Thanks! 🙇

bundle/ios/GutenbergNativeTranslations.swift
```diff
@@ -27,39 +27,26 @@ private func dummy() {
     _ = NSLocalizedString("50 / 50", comment: "")
     _ = NSLocalizedString("70 / 30", comment: "")
     _ = NSLocalizedString("9:16", comment: "")
-    _ = NSLocalizedString("A calendar of your site’s posts.", comment: "")
```
dcalhoun: Do you know if these removals are of consequence? I wonder if they are accurate removals or potentially erroneous changes occurring due to outdated i18n files. Possibly related to https://github.com/wordpress-mobile/gutenberg-mobile/pull/3423? WDYT?

dcalhoun (approved):
LGTM. Approved via https://github.com/WordPress/gutenberg/pull/30994#pullrequestreview-653922751. I left one question regarding the bundle translation strings, but I think this is ready to merge.

bundle/ios/GutenbergNativeTranslations.swift
geriux: I think we should only merge the Gutenberg reference update and not the rest, I'm guessing they were only present to be able to test the builds right @antonis? If so we can just update this PR to only have the GB commit hash change before merging.

antonis: Thank you for reviewing this @dcalhoun! I really appreciate the time you put on this 🙇 

> Do you know if these removals are of consequence? I wonder if they are accurate removals or potentially

Nice catch. I'm not sure why the changes were introduced. It should be related with #3423. I've only run `bundle` and not `prebundle` btw 😕 

> I think we should only merge the Gutenberg reference update and not the rest

I would agree with @geriux on this. I would just update the reference. The rest was just needed to be able to produce the App builds for testing.

PR #3442, opened 2021-04-30
Fix Issue with Search block stealing focus after media updated for image block

### LINKED ISSUE #31351: Search block: The block gets selected after adding a media item on another block
## Description

After adding the search block into a post, it gets automatically selected after adding or replacing media items on other media blocks like the image block.

## Step-by-step reproduction instructions

1. Add a search block
2. Add an image block
3. Select an image to be inserted in the image block
4. Observe that after the image is added the search block gets selected

## Expected behaviour

The image block should be kept selected after adding an image.

## Actual behaviour

The search block gets selected instead of the image block.

## Screenshots or screen recording (optional)

https://user-images.githubusercontent.com/14905380/116595087-3c90bc80-a923-11eb-928b-a3bc3185a24f.mp4

## WordPress information
- WordPress version: N/A
- Gutenberg version: N/A
- Are all plugins except Gutenberg deactivated? N/A
- Are you using a default theme (e.g. Twenty Twenty-One)? N/A

## Device information
- Device: iPhone 8 (simulator)
- Operating system: iOS 14.2
- WordPress app version: `develop`

AmandaRiu:
Fixes https://github.com/WordPress/gutenberg/issues/31351. Search block was stealing the focus away from the Image block after the image was updated (iOS only). The fix itself was really straightforward - at some point I defaulted the isPlaceholderSelected state variable to true but this is not necessary and was forcing the search input element of the Search block to gain focus and take focus from the image block. Defaulting this to false fixed the issue. While this change was tiny, it required fixes to both the Unit Test snapshot and the e2e tests.

**Gutenberg PR**: https://github.com/WordPress/gutenberg/pull/31393

To test:
- See test instructions in the Gutenberg PR

PR submission checklist:


package-lock.json
```diff
@@ -4664,7 +4664,7 @@
 					}
 				},
 				"prettier": {
-					"version": "npm:prettier@2.2.1-beta-1",
+					"version": "npm:wp-prettier@2.2.1-beta-1",
```
dcalhoun: We likely need to revert this change. I believe it is erroneous. 

This particular change has been discussed a few times before. My current (untested) understanding is that `rm -rf node_modules && cd gutenberg && npm run distclean && cd ../ && npm install` may resolve the issue.

dcalhoun (changes requested):

package-lock.json
AmandaRiu: That worked beautifully! Thanks for the tip. Fixed in 013ec06950d09a6c0176268e48dfe2b94256d33b

AmandaRiu:
Hey @dcalhoun! Thanks for the review. I've fixed `package-lock.json` as well as the conflict. 👍

dcalhoun (approved):
Approved via https://github.com/WordPress/gutenberg/pull/31393#pullrequestreview-650497778.

RELEASE-NOTES.txt
```diff
@@ -6,6 +6,7 @@ Unreleased
 * [***] New Block: Search Block [#3210]
 * [**] The media upload options of the Image, Video and Gallery block automatically opens when the respective block is inserted. [https://github.com/wordpress-mobile/gutenberg-mobile/pull/2700]
 * Image block: Add a "featured" banner. (Android only) [https://github.com/wordpress-mobile/gutenberg-mobile/pull/3371]
+* [*] Fixed a bug where the Search block was stealing focus from the Image block upon updating image asset [https://github.com/wordpress-mobile/gutenberg-mobile/pull/3442]
```
fluiddot, line 9: I was checking the [release notes of `1.52.0`](https://github.com/wordpress-mobile/gutenberg-mobile/releases/tag/v1.52.0) and I didn't see this change, looks like this got merged after the `1.52.0` code freeze so maybe we should move it to the `Unreleased` section, what do you think @AmandaRiu ?

AmandaRiu: I think that makes sense. How would I go about doing this now? Should I open a PR?

fluiddot: Yep, you could open a PR to update it. Btw, I saw [this entry is also in `1.52.0` in Gutenberg](https://github.com/WordPress/gutenberg/blob/trunk/packages/react-native-editor/CHANGELOG.md#1520-1) so we could also make the update there.

PR #3444, opened 2021-05-01
Add/experimental grid block all controls

enejb:
Adds the submodule Block experiements. 

Related PR: https://github.com/Automattic/block-experiments/pull/187 

iOS: https://github.com/wordpress-mobile/WordPress-iOS/pull/16423

Android PR:
https://github.com/wordpress-mobile/WordPress-Android/pull/14578

To test:
run 
1. `git submodule update` 
2. Start metro. `npm run start`
3. Add the layout grid block.
4. Follow the rest of the instruction on https://github.com/Automattic/block-experiments/pull/187 

PR submission checklist:

- [x] I have considered if this change warrants user-facing release notes - No release notes just yet since we are not releasing the new block to the public)


src/index.js
```diff
@@ -23,6 +23,9 @@ addAction( 'native.render', 'gutenberg-mobile', ( props ) => {
 	setupJetpackEditor(
 		props.jetpackState || { blogId: 1, isJetpackActive: true }
 	);
+	if ( __DEV__ ) {
+		require( './block-experiments-setup' );
```
mkevins, line 27: I'm wondering why this needs to be required dynamically. Could we instead `import { registerBlock as registerLayoutGridBlock } from '../block-experiments/blocks/layout-grid/src';` here, and invoke it under `__DEV__`? If this is for the incidental idempotency given by `require.cache`, maybe we can explicitly make this idempotent by setting a flag in the file-level closure here?

src/block-experiments-setup.js
```diff
@@ -0,0 +1,5 @@
+// This file is to set up the jetpack/layout-grid block that currently lives in block-experiments/blocks/layout-grid
+
+import { registerBlock } from '../block-experiments/blocks/layout-grid/src';
```
mkevins, line 3: I just had a thought about the relative import path used here. Not suggesting a change, but wondering if there's a way to scope this (similar to how we resolve `@wordpress` dependencies) such that a future change in project structure wouldn't require any changes here. Something like `'@something/blocks'` :thinking: 

If there's too much "magic" in the module resolution configuration, perhaps we'd still achieve a similar effect by importing from an index at a higher level in `../block-experiments`? It feels like "reaching deep into" the path here will be brittle to structure changes in the dependency. (This comment might belong there as well as here :sweat_smile: ) Wdyt?

src/index.js
enejb: I believe that a `require` is necessary because metro doesn't support require imports just yet. See https://github.com/wordpress-mobile/gutenberg-mobile/pull/2582#discussion_r613574059

src/block-experiments-setup.js
enejb: I think that is a great idea but I am not sure how to go about doing that. Any suggestions?

src/index.js
mkevins: That comment refers to why we would use `require()` instead of `import()`, which are both dynamic. What I'm wondering is, why are we aiming for this to be dynamic? I.e. why not a static import at the top of the file, and dynamically register the block under `__DEV__`?

src/block-experiments-setup.js
mkevins: The second idea would just involve creating an `index.js` (or possibly `index.native.js`) file in `../block-experiments/blocks/` and re-exporting the block registration there. This would shift the responsibility of maintaining the deeper path traversals to the dependency, which probably makes more sense, since that's where the structure changes would occur.

Regarding the module resolution magic, there is [a plugin](https://github.com/tleunen/babel-plugin-module-resolver) that might be worth investigating for this purpose, but I haven't tried it out. It essentially allows you to create an alias for the imports.

geriux: > Regarding the module resolution magic, there is a plugin that might be worth investigating for this purpose, but I haven't tried it out. It essentially allows you to create an alias for the imports.

I've used that plugin in every React native project I've worked on, it's very useful. Also very [easy to set up](https://dev.to/karanpratapsingh/cleaning-up-imports-using-module-resolver-in-react-native-58g8) (usually).

src/index.js
enejb: Good question! I think part of the reason why I included it like this is so that we don't load the js script when the flag is not set. but since we are not downloading the js I guess this part doesn't really matter.

mkevins: It would be cool to explore something like this down the road for lazy-loading / code splitting.

enejb:
Closing this since it was already merged.

enejb:
I am closing this since it is not needed any more.

PR #3485, opened 2021-05-12
Updating release notes for transform block.

illusaen:
Adding release notes for transform block.

PR submission checklist:

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -3,6 +3,7 @@ Unreleased
 
 1.52.1
 ------
+* [**] Added transform block capability [#3321]
 * [*] Fixes for the generated localized strings files.
 
 1.52.0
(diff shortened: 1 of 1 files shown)


RELEASE-NOTES.txt
```diff
@@ -3,6 +3,7 @@ Unreleased
 
 1.52.1
 ------
+* [**] Added transform block capability [#3321]
```
mchowning: I thought this was in 1.52.0. Was this in the hotfix because of the bug where it was left disabled on debug builds? I thought that bug was on the native side and had nothing to do with gutenberg-mobile. I may just be misremembering though. 🤔

illusaen: Actually you're completely right, it was in 1.52.0. I just saw the 52 and somehow didn't notice the 1.

mchowning (approved):

Issue #3466, opened 2021-05-07
Localized strings files are missing strings and adding extra ones


fluiddot:
**Describe the bug**
For each platform we generate a specific file with the localized strings:
- [iOS: GutenbergNativeTranslations.swift](https://github.com/wordpress-mobile/gutenberg-mobile/blob/develop/bundle/ios/GutenbergNativeTranslations.swift)
- [Android: strings.xml](https://github.com/wordpress-mobile/gutenberg-mobile/blob/develop/bundle/android/strings.xml)

I investigated further the process of generating these files and I noticed the following issues:

## Missing strings
Comparing the files generated in the latest releases, I saw that some strings were missing, for example, the string `Block copied` is present in version [`1.51.1`](https://github.com/wordpress-mobile/gutenberg-mobile/blob/8611e4978ba6072cf6f77ccb6cb4b2adc1e7ba62/bundle/ios/GutenbergNativeTranslations.swift#L32) but not in [`1.52.0`](https://github.com/wordpress-mobile/gutenberg-mobile/blob/v1.52.0/bundle/ios/GutenbergNativeTranslations.swift), [here is the diff](https://github.com/wordpress-mobile/gutenberg-mobile/commit/8645885e0549eac2425502b66dd088c268c0bd97#diff-d5d2ed0bfbc4117c67ed9c3822eda75a06c4cd100b1d51961a0df2bbe9718db5L32) of the commit that introduced the change.

I debugged the scripts that we use for extracting the strings from the codebase and looks like that some files are not including their strings. Specifically, when we run the command with the debug option (`npm run makepot:ios -- --debug`), we can see that some files can't be parsed (see attached logs). This is caused due to the use of the optional chaining operator that is not supported by the script ([this is known bug](https://github.com/wp-cli/i18n-command/issues/218)) we use to extract the strings ([`wp i18n make-pot`](https://developer.wordpress.org/cli/commands/i18n/make-pot/)).

<details><summary><strong>Parse file errors generated by npm run makepot:ios</strong></summary>

```
Debug: Could not parse file block-editor/src/components/block-icon/index.native.js: Unexpected: . (line 23, column 11) (2.163s)
Debug: Could not parse file block-editor/src/components/block-list/block-list-item.native.js: Unexpected: . (line 211, column 16) (2.18s)
Debug: Could not parse file block-editor/src/components/block-list/block-selection-button.native.js: Unexpected: . (line 63, column 29) (2.184s)
[...]

PR #3493, opened 2021-05-12
Exclude non-native files when extracting localization strings

fluiddot:
**Fixes the case of extra strings of https://github.com/wordpress-mobile/gutenberg-mobile/issues/3466.**

## Description

I added the following paths to the excluded argument of the `make:pot` commands, this way we prevent adding extra strings from non-native files.

New excluded paths:
- `build/*`
- `build-module/*`
- `build-style/*`
- `*.js`
- `*.php`

In theory, this change shouldn't be required because the [`--include` argument should only include the specified paths](https://developer.wordpress.org/cli/commands/i18n/make-pot/#options), however, this is not working as expected.

Besides, just by excluding `*.js` should only extract strings from `*.native.js` files, unfortunately, it's not and it's including by default the `build*` folders 🤷‍♂️ .

### Extra strings are not added
1. Run command `npm run genstrings`
2. Observe that the localization strings files are not including a high number of strings

### Only extract strings from native files
1. Run command `npm run makepot:ios -- --debug`
2. Observe the output log and verify that only `.native.js` files are being parsed

**Note:** Some files are giving parse failures, this is caused by the use of the optional chaining operator that is not supported by the script that extracts the strings.

PR submission checklist:


dcalhoun (approved):
LGTM. I verified that in this branch `npm run genstrings` generated _significantly_ fewer translations string changes and `npm run makepot:ios -- --debug` _only_ included `*.native.js` files in comparison to the `develop` branch.

PR #3575, opened 2021-06-02
Switch to pnpm needed by the Jetpack repo

hypest:
Fixes the Jetpack build issues noticed in https://github.com/wordpress-mobile/gutenberg-mobile/pull/3572.

For context, the Jetpack build has switched to pnpm instead of yarn: https://github.com/Automattic/jetpack/blob/master/docs/yarn-upgrade.md

### Changes:

1. Stop using `yarn` for the Jetpack side build.
1. Running `pnpm` directly via `npx`. I tried a few different ways to install it on CI but failed. Maybe there's still a way but I mostly given up and went with npx. Happy to revise. Long story short, resolving the module from inside the .pnpm folder since the normally "exposed" paths are actually symlinks and Metro won't work OK with symlinks.
1. Had to [add more logic to the Metro package resolver](https://github.com/wordpress-mobile/gutenberg-mobile/pull/3575/files#diff-eb288b08cb881cff653283c14cb9630dc0924a1751950e1562a8ca2c0b39694bR20-R33) (and thank you @ceyhun for the guidance on how we use the Proxy there!) to help Metro find the Jetpack package dependencies as installed by pnpm. Happy to find a simpler way to point Metro to the right place if there is one!

### To test:
* CI jobs should be green on this PR
* Demo app should run normally

PR submission checklist:

### DIFF
--- a/metro.config.js
+++ b/metro.config.js
@@ -9,28 +9,25 @@ gutenbergMetroConfigCopy.resolver.extraNodeModules = new Proxy(
 	{},
 	{
 		get: ( target, name ) => {
-			const gutenberg_folder = path.join(
+			const gutenbergFolder = path.join(
 				process.cwd(),
 				`gutenberg/node_modules/${ name }`
 			);
-			if ( fs.existsSync( gutenberg_folder ) ) {
-				return gutenberg_folder;
+			if ( fs.existsSync( gutenbergFolder ) ) {
+				return gutenbergFolder;
 			}
 
 			// let's try find the mobile in the Jetpack submodule. We'll try the .pnpm folder.
-			const pnpm_module_dir = path.join(
+			const moduleFolderPnpm = path.join(
 				process.cwd(),
 				`./jetpack/node_modules/.pnpm/node_modules/${ name }`
 			);
 
 			// pnpm uses symlinks so, let's find the target
-			const symlink_target = fs.readlinkSync( pnpm_module_dir );
+			const symlinkTarget = fs.readlinkSync( moduleFolderPnpm );
 
 			// the target is still using paths relative to the parent folder of the module, let's find the real path.
-			const real_path = path.resolve(
-				pnpm_module_dir + '/../' + symlink_target
-			);
-			return real_path;
+			return path.resolve( moduleFolderPnpm + '/../' + symlinkTarget );
 		},
 	}
 );
--- a/metro.config.js
+++ b/metro.config.js
@@ -17,7 +17,7 @@ gutenbergMetroConfigCopy.resolver.extraNodeModules = new Proxy(
 				return gutenbergFolder;
 			}
 
-			// let's try find the mobile in the Jetpack submodule. We'll try the .pnpm folder.
+			// let's try find the module in the Jetpack submodule. We'll try the .pnpm folder.
 			const moduleFolderPnpm = path.join(
 				process.cwd(),
 				`./jetpack/node_modules/.pnpm/node_modules/${ name }`
--- a/metro.config.js
+++ b/metro.config.js
@@ -1,4 +1,5 @@
 const path = require( 'path' );
+const fs = require( 'fs' );
 
 const gutenbergMetroConfigCopy = {
 	...require( './gutenberg/packages/react-native-editor/metro.config.js' ),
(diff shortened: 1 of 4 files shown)


metro.config.js
```diff
@@ -7,8 +8,27 @@ const gutenbergMetroConfigCopy = {
...
+		get: ( target, name ) => {
+			const gutenbergFolder = path.join(
+				process.cwd(),
+				`gutenberg/node_modules/${ name }`
+			);
+			if ( fs.existsSync( gutenbergFolder ) ) {
+				return gutenbergFolder;
+			}
+
+			// let's try find the mobile in the Jetpack submodule. We'll try the .pnpm folder.
```
AmandaRiu: np: Should "mobile" be replaced with "module" in this comment?

hypest: Oops, my hands see "mobiles" everywhere 😆, thanks for spotting! Fixed with https://github.com/wordpress-mobile/gutenberg-mobile/pull/3575/commits/785bfe2aab960b3e39a085cec235d96c53d09c5d.

package.json
```diff
@@ -53,7 +53,7 @@
 		"yarn": "^1.22.10"
 	},
 	"scripts": {
-		"postinstall": "npm ci --prefix gutenberg && cd jetpack/projects/plugins/jetpack && yarn install --frozen-lockfile --ignore-engines",
+		"postinstall": "npm ci --prefix gutenberg && cd jetpack/projects/plugins/jetpack && npx pnpm install",
```
dcalhoun, line 56: >Running `pnpm` directly via `npx`. I tried a few different ways to install it on CI but failed. Maybe there's still a way but I mostly given up and went with npx. Happy to revise. Long story short, resolving the module from inside the .pnpm folder since the normally "exposed" paths are actually symlinks and Metro won't work OK with symlinks.

Is there a reason to avoid `npx` from your perspective? Or is it merely a desire to use the "default" flow of `npm install` package and `npm <binary>`?

My only alternative thought would be to explore [relying upon `npm bin`](https://www.rockyourcode.com/run-locally-installed-npm-packages-without-global-install/) to locate the `pnpm` executable (e.g. `$(npm bin)/pnpm install`). Although, I'm not sure how relevant that is to the issues you were seeing installing `pnpm` on CI.

metro.config.js
```diff
@@ -7,8 +8,27 @@ const gutenbergMetroConfigCopy = {
...
+			const moduleFolderPnpm = path.join(
+				process.cwd(),
+				`./jetpack/node_modules/.pnpm/node_modules/${ name }`
+			);
+
+			// pnpm uses symlinks so, let's find the target
+			const symlinkTarget = fs.readlinkSync( moduleFolderPnpm );
+
+			// the target is still using paths relative to the parent folder of the module, let's find the real path.
+			return path.resolve( moduleFolderPnpm + '/../' + symlinkTarget );
```
dcalhoun, line 30: >Had to [add more logic to the Metro package resolver](https://github.com/wordpress-mobile/gutenberg-mobile/pull/3575/files#diff-eb288b08cb881cff653283c14cb9630dc0924a1751950e1562a8ca2c0b39694bR20-R33) (and thank you @ceyhun for the guidance on how we use the Proxy there!) to help Metro find the Jetpack package dependencies as installed by pnpm. Happy to find a simpler way to point Metro to the right place if there is one!

For better or worse, I do believe this is likely as "simple" as it gets until [Metro supports symlinks](https://github.com/facebook/metro/issues/1).

metro.config.js
```diff
@@ -7,8 +8,27 @@ const gutenbergMetroConfigCopy = {
...
 	{},
 	{
-		get: ( target, name ) =>
-			path.join( process.cwd(), `gutenberg/node_modules/${ name }` ),
+		get: ( target, name ) => {
+			const gutenbergFolder = path.join(
+				process.cwd(),
+				`gutenberg/node_modules/${ name }`
+			);
+			if ( fs.existsSync( gutenbergFolder ) ) {
```
dcalhoun, line 16: 💯

dcalhoun (approved):
LGTM. Verified the CI checks pass and Demo app runs as expected. 🎉

package.json
hypest: > Is there a reason to avoid npx from your perspective? Or is it merely a desire to use the "default" flow of npm install package and npm <binary>?

The latter, I'd intuitively prefer to use something like `pnpm install` with `pnpm` already installed like we had with `yarn`. Also, since this is a "postinstall" phase, feels weird to have to _install_ something here. Not sure if that's a antipattern.

> My only alternative thought would be to explore relying upon npm bin to locate the pnpm executable (e.g. $(npm bin)/pnpm install). Although, I'm not sure how relevant that is to the issues you were seeing installing pnpm on CI.

Aha, that could be an option to try. At this stage, we can probably start with `npx ...` since we don't really need pnpm elsewhere, if you also are satisfied with the current approach too. We should revise if we found ourselves needing it normally installed. WDYT?

metro.config.js
hypest: Cool, thanks for adding your "stamp" on it David! 🙇✅

package.json
dcalhoun: > Aha, that could be an option to try. At this stage, we can probably start with `npx ...` since we don't really need pnpm elsewhere, if you also are satisfied with the current approach too. We should revise if we found ourselves needing it normally installed. WDYT?

Agreed. I think we should move forward with `npx`. 👍🏻

PR #3590, opened 2021-06-05
[Android] Fix Gboard enter detection in EnterPressedWatcher

### LINKED ISSUE #3316: Enter detection fails in EnterPressedWatcher for Gboard keyboard
## Description

The mechanism to detect whether the `[ENTER]` key was pressed is not reliable for users with the Gboard keyboard. The Gboard keyboard exhibits some unusual behavior regarding the `TextWatcher` interface methods, which are breaking some assumptions in the logic we are using for key detection. This manifests as various bugs, and is likely not limited to the `[ENTER]` key (e.g. I believe `[BACKSPACE]` detection may be affected in a similar manner).

### Example bug

On a Pixel 3a (physical device) with the Gboard keyboard:

1. Create a new post
2. Add a paragraph block
3. Type a word (e.g. "Hello") without a space
  * Observe that the keyboard suggestions
  * Observe that the word is underlined
4. Press `[ENTER]`
5. Observe that the block does not split - instead, a newline is added to the paragraph block

#### Expected

The block should split and form a new paragraph block below.

## Deeper details

The `EnterPressedWatcher` class works by implementing the `TextWatcher` interface, overriding the following methods:

* `beforeTextChanged`
* `onTextChanged`
* `afterTextChanged`

and tracking state during a user's manipulation of the `EditText` input. These methods allow us to observe the mutations to the `CharSequence`, and to provide a means to reconcile what has changed, and respond in some way after evaluating the changes at each stage of the update. Typically, these changes are fairly straightforward, and these methods are invoked with predictable index boundaries for a given text input operation, but here is where Gboard exhibits some unusual behavior.

Here is a table describing some observations that surprised me:

Consider a scenario where we have the input with text `Hello worl`, and we then press `d` on the keyboard:

Comparison |`TextWatcher` Interface Method | `text: CharSequence` | `start: Int` | `count: Int` | `after: Int` | `before: Int`
-|-|-|-|-|-|-
Expected | `beforeTextChanged` | "Hello worl" | 10 | 0 | 1 | N/A
Actual | `beforeTextChanged`| "Hello worl" | 6 | 4 | 5 | N/A
Expected | `onTextChanged` | "Hello world" | 10 | 1 | N/A | 0
Actual | `onTextChanged`| "Hello world" | 6 | 5 | N/A | 4

The surprising part is that some characters from the original text are being replaced when nothing was selected (just a normal caret at the end of the input). It seems instead that Gboard is actually replacing the last 4 characters with those same 4 characters _plus_ the typed text. It is even possible to notice a small underline style applied to the characters that will be included in the replacement. Before this, I would have only expected `beforeTextChanged` to be called with a `count` > 0 if there was a selection prior to the new characters being typed (or if characters were being deleted).

Note that the result is the same (and that the sum is still 11 for `start + after` and `start + count` for `beforeTextChanged` and `onTextChanged` respectively).

Interestingly, after selecting one of the suggestions offered by Gboard, the `onEnter` behavior was works normally (and the underlined text goes away).

AmandaRiu:
Fixes #3316 

Gutenberg PR: https://github.com/WordPress/gutenberg/pull/32471

To test:
See testing instructions in the gutenberg PR. 

PR submission checklist:


RELEASE-NOTES.txt
```diff
@@ -6,6 +6,7 @@ Unreleased
 ------
 * [*] Image block: "Set as featured" button within image block settings. (Android only). [https://github.com/wordpress-mobile/gutenberg-mobile/pull/3116]
 * [***] Audio block now available on WP.com sites on the free plan. [https://github.com/wordpress-mobile/gutenberg-mobile/pull/3523]
+* [**] Fixed an issue where pressing enter inside a text-based block was not creating a new block when using Gboard [https://github.com/wordpress-mobile/gutenberg-mobile/pull/3590]
```
dcalhoun: I believe this new note should be in the above "Unreleased" section, if I am not mistaken. It currently is listed under the `1.55.0` release, which I do not believe included this fix.

dcalhoun (dismissed):

RELEASE-NOTES.txt
AmandaRiu: Oh good catch! this PR had hung around for a while and I didn't think about it. Fixed in 06eee7b2aaa09100ac9c3afd035b7a37f750d581

mkevins (approved):
Tested through the gutenberg PR. LGTM!

RELEASE-NOTES.txt
mkevins: Thanks for wrangling this!

PR #3600, opened 2021-06-09
Cache node_modules for CI tasks

dcalhoun:
Attempt to improve task speed by [caching long-running dependency installations](https://circleci.com/docs/2.0/caching/#npm-node).

To test: Verify CI checks succeed. 

PR submission checklist:


dcalhoun:
After further research and consideration, caching `node_modules` does not seem like a worthwhile approach. 

* We use `npm ci` rather than `npm install`. The former removes `node_modules` before installing, negating any value in caching the directory. 
* Given the complexity of `node_module` directories, there may be unforeseen issues and cryptic errors attempting to cache them. (p9ugOq-20o-p2#comment-4218)

~It may be more appropriate to attempt to share dependency installations between tasks by ["selectively persisting workspaces"](https://circleci.com/blog/six-optimization-tips-for-your-config/) to attach later tasks to an earlier task workspace that contains the results of `npm ci`.~ Utilizing workspaces for an incalculable number of files (e.g. `node_modules`) is [not intended or recommended](https://discuss.circleci.com/t/2-0-persist-to-workspace-does-not-preserve-symlinks/14338/4). `save_cache` is the recommended approach, which is what this PR originally explored.

hypest:
> We use npm ci rather than npm install. The former removes node_modules before installing, negating any value in caching the directory.

Aha, I would assume that we won't be performing any `npm ci` or `npm install` at all, if we had the cache, so, not sure if that's a blocker? I mean, the node_modules from the cache would stay there, we wouldn't execute a `npm ci` at all. WDYT? 🤔

dcalhoun:
> Aha, I would assume that we won't be performing any `npm ci` or `npm install` at all, if we had the cache, so, not sure if that's a blocker? I mean, the node_modules from the cache would stay there, we wouldn't execute a `npm ci` at all. WDYT? 🤔

You are absolutely correct. My I original plan was to use `save_cache` to speed up `npm` installs in subsequent tasks (i.e. each task still ran `npm` install itself). However, abstracting `npm ci` into a single, shared task was the approach I explored with workspaces before abandoning it due to other reasons. I'm not sure why my brain did not combine the two approaches. 

I will explore combining `save_cache` with a single, shared install dependencies task. 👍🏻 Thanks!

dcalhoun:
I do not believe I will work on this in the near term and would likely leverage a different branch if I did. So, I will close this PR to clean up a bit.

PR #3606, opened 2021-06-10
Disable CI npm audit

dcalhoun:
Remove audit that is likely not monitored within CI task logs to reduce duration of CI task runs. Alternatively, these audits should be monitored locally or via depedabot tooling. (p9ugOq-20o-p2#comment-4209)

To test: Verify CI tasks succeed. 

PR submission checklist:


hypest (approved):
It's not immediately clear how much of a speed improvement this gets us (probably something in the range or 10-20secs?) but I'll take it! 🎉

dcalhoun:
> It's not immediately clear how much of a speed improvement this gets us (probably something in the range or 10-20secs?) but I'll take it! 🎉

Yes, it is likely negligible, but it is something, right? 😄

PR #3623, opened 2021-06-16
Increase flexibility of CI shallow git clone

dcalhoun:
Previously, the CI configuration cloned the repository with a depth of 1. In the context where the CI job was attempting to check out an older commit of a branch, this would cause an error. This might occur for a scheduled job on a shared branch when a merge occurs after the scheduled job begins. This commit increases the depth to reduce the likelihood of this circumstance. Relates to #3601.

To test: All CI checks on this PR succeed. After this merge, all scheduled CI jobs should succeed as well. 

PR submission checklist:

- [x] I have considered adding unit tests where possible.
- [x] I have considered if this change warrants user-facing release notes [more info](https://github.com/wordpress-mobile/gutenberg-mobile/blob/develop/Release-notes.md) and have added them to [RELEASE-NOTES.txt](https://github.com/wordpress-mobile/gutenberg-mobile/blob/develop/RELEASE-NOTES.txt) if necessary.

PR #3601, opened 2021-06-09
Shallow clone project for CI tasks

dcalhoun:
Decrease CI task duration by reducing depth of git operations. (p9ugOq-20o-p2#comment-4215)

- [CircleCI: Shallow Repository Cloning](https://support.circleci.com/hc/en-us/articles/360045096514-Shallow-Repository-Cloning)
- [CircleCI: How do I modify the checkout step?](https://support.circleci.com/hc/en-us/articles/115015659247-How-do-I-modify-the-checkout-step-)
- [Example: CircleCI Shallow Clone](https://discuss.circleci.com/t/shallow-clone-for-circleci-2-0-builds/20200)

| Before | After |
| --- | --- |
| checkout-before | checkout-after |

To test: Verify CI tasks succeed. 

PR submission checklist:

- [x] I have considered adding unit tests where possible.
- [x] I have considered if this change warrants user-facing release notes [more info](https://github.com/wordpress-mobile/gutenberg-mobile/blob/develop/Release-notes.md) and have added them to [RELEASE-NOTES.txt](https://github.com/wordpress-mobile/gutenberg-mobile/blob/develop/RELEASE-NOTES.txt) if necessary.

PR #3624, opened 2021-06-16
CI fetches commit reference directly

dcalhoun:
The CI configuration is set to clone with a depth of 1. If the CI tasks then attempts to reference an older commit in the relevant branch, an error occurs. This might occur when a branch merge occurs between a CI tasks beginning and the point it attempts to fetch the HEAD of the branch. This change configures the shallow clone to fetch the specific commit reference, rather than the HEAD of the relevant branch. Relates to #3601. Supersedes #3623. 

To test: All CI checks on this PR succeed. After merging this work, all scheduled CI checks should succeed. 

PR submission checklist:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -84,7 +84,7 @@ commands:
             if [ -n "$CIRCLE_TAG" ]; then
               git fetch --depth 1 --force --tags origin "refs/tags/$CIRCLE_TAG"
             else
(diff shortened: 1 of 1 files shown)


hypest (approved):
LGTM, very clean fix and still the speedy git checkout! Let's merge as soon as CI gets green!

PR #3640, opened 2021-06-18
Fix incorrect block insertion point after blurring post title

### LINKED ISSUE #3636: New-block indicator and newly added block is in the wrong position after selecting the Title
After selecting the Title and then some other block, and then try to add a new block from the block picker, the blue line/indicator is at the wrong place.

**To Reproduce**
1. In the demo app
2. Tap on the Title
3. Then tap on a block a couple of blocks away
4. Then tap on the "+" toolbar button to add a new block from the picker
5. 💥 Notice that the new-block indicator is right below the title instead of being below the currently selected block. See video below too.
6. Select a new block from the picker and notice that it's added at the wrong place

https://user-images.githubusercontent.com/1032913/122574498-611e3080-d058-11eb-8d59-cc94500cd88e.mp4

**Expected behavior**
The new-block indicator should be below the selected block, not the Title.

 - Device: Pixel 2XL
 - OS: Android 11
 - Version on `develop` hash 11c7f704
 - Verified happening on WPAndroid 17.5 (production at the time of writing)

dcalhoun:
* **Gutenberg:** https://github.com/WordPress/gutenberg/pull/32831

Fixes #3636. Selecting the title selection status within the same `useSelect` for reusable blocks caused the title selection status to become stale, due to the dependency array for the `useSelect` hook. The staleness caused the block insertion point to show up in the incorrect location after blurring the title text input.

To test: See https://github.com/WordPress/gutenberg/pull/32831. 

PR submission checklist:

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -1,5 +1,6 @@
 Unreleased
 ------
+* [**] Fix incorrect block insertion point after blurring the post title field. [https://github.com/WordPress/gutenberg/pull/32831]
 * [*] Tablet view fixes for inserter button. https://github.com/wordpress-mobile/gutenberg-mobile/pull/3602
 * [**] Fixed an issue where pressing enter inside a text-based block was not creating a new block when using Gboard [https://github.com/wordpress-mobile/gutenberg-mobile/pull/3590]
 


RELEASE-NOTES.txt
```diff
@@ -1,5 +1,6 @@
 Unreleased
 ------
+* [**] Fix incorrect block insertion point after blurring the post title field. [https://github.com/WordPress/gutenberg/pull/32831]
```
fluiddot, line 3: I see we usually reference the Gutenberg Mobile PR instead of the Gutenberg one so what do you think about updating this entry with the following suggestion?

```suggestion
* [**] Fix incorrect block insertion point after blurring the post title field. [https://github.com/wordpress-mobile/gutenberg-mobile/pull/3640]
```

fluiddot (approved):
Approved via https://github.com/WordPress/gutenberg/pull/32831#pullrequestreview-688585991 🎊 .

PR #3629, opened 2021-06-17
Add Sentry SDK

fluiddot:
`gutenberg` PR: https://github.com/WordPress/gutenberg/pull/32768
`WordPress-iOS`PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/16700
`Automattic-Tracks-iOS` PR: https://github.com/Automattic/Automattic-Tracks-iOS/pull/183

This PR adds the Sentry React native SDK and customize the initialization to accomplish the following items:
1. Use the same options used for initializing the SDK in the main apps.
2. Only enable the needed Sentry integrations.
3. Add `beforeSend` callback to attach the same scope defined in the main apps and prevent sending events as it's determined in the main apps ([code reference](https://github.com/Automattic/Automattic-Tracks-iOS/blob/36db74cceaf98e107e0c713eaae81eb4139c2059/Automattic-Tracks-iOS/Crash%20Logging/CrashLogging.swift#L53-L90)).

**NOTE: The Sentry SDK will be only enabled on iOS.**

To test:
Follow the testing instructions from `gutenberg` PR: https://github.com/WordPress/gutenberg/pull/32768.

PR submission checklist:

- [x] I have considered adding unit tests where possible.
- [ ] I have considered if this change warrants user-facing release notes [more info](https://github.com/wordpress-mobile/gutenberg-mobile/blob/develop/Release-notes.md) and have added them to [RELEASE-NOTES.txt](https://github.com/wordpress-mobile/gutenberg-mobile/blob/develop/RELEASE-NOTES.txt) if necessary.

PR #3659, opened 2021-06-24
Fix Gutenberg Mobile dependencies resolution

fluiddot:
I added a new dependency to the `package.json` file of Gutenberg Mobile in [this PR](https://github.com/wordpress-mobile/gutenberg-mobile/pull/3629) and I noticed that Metro was throwing an error due to not being able to resolve it.

After checking the extra node modules resolver, I realized that the dependencies that are not resolved via the Gutenberg submodule were considered to be located in the Jetpack submodule, including the ones defined in Gutenberg Mobile. To address this, I basically added a condition that checks if the path exists before considering it a dependency within Jetpack.

Additionally, I removed the dependency `email-validator` from Gutenberg Mobile because it's already in the Jetpack submodule, in fact, it's not even referenced in the Gutenberg Mobile code.

As a side note, these are the dependencies resolved from the Jetpack submodule:
- `@automattic/color-studio`
- `email-validator`

To test:
- CI jobs should be green on this PR
- Demo app should run normally
- As a double-check, verify that the Contact info block works as expected because it's the one that imports the `email-validator` package.

PR submission checklist:

### DIFF
--- a/metro.config.js
+++ b/metro.config.js
@@ -18,7 +18,7 @@ gutenbergMetroConfigCopy.resolver.extraNodeModules = new Proxy(
 				return gutenbergFolder;
 			}
 
-			// If not exists, let's try find the module in the Jetpack submodule. We'll try the .pnpm folder.
+			// Try to find the module in Jetpack's .pnpm folder.
 			const moduleFolderPnpm = path.join(
 				process.cwd(),
 				`./jetpack/node_modules/.pnpm/node_modules/${ name }`
--- a/metro.config.js
+++ b/metro.config.js
@@ -9,6 +9,7 @@ gutenbergMetroConfigCopy.resolver.extraNodeModules = new Proxy(
 	{},
 	{
 		get: ( target, name ) => {
+			// Try to find the module in the Gutenberg submodule.
 			const gutenbergFolder = path.join(
 				process.cwd(),
 				`gutenberg/node_modules/${ name }`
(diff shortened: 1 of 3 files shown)


metro.config.js
```diff
@@ -17,17 +18,21 @@ gutenbergMetroConfigCopy.resolver.extraNodeModules = new Proxy(
 				return gutenbergFolder;
 			}
 
-			// let's try find the module in the Jetpack submodule. We'll try the .pnpm folder.
+			// If not exists, let's try find the module in the Jetpack submodule. We'll try the .pnpm folder.
```
hypest: Nit: want to phrase this as "If not there," or "If it doesn't exist,"? Alternatively, how about replacing the whole sentence with just "Try to find the module in Jetpack's .pnpm folder instead."?

fluiddot: I finally replaced it with `Try to find the module in Jetpack's .pnpm folder.` in [this commit](https://github.com/wordpress-mobile/gutenberg-mobile/pull/3659/commits/2d317c0d62aa560aefb028cc79b12c965f6be909), wdyt?

hypest (approved):
LGTM! Let's auto-merge when CI green.

✅ CI jobs should be green on this PR. It was green on https://github.com/wordpress-mobile/gutenberg-mobile/pull/3659/commits/0837df742146a78fbcf62b1964e2c00a679cab5b already.
✅ Demo app should run normally
✅ As a double-check, verify that the Contact info block works as expected because it's the one that imports the email-validator package. ➡️ I verified that when the email is well-formed, the html output includes a `mailto:` link, while if not well-formed it's just plain text, just like [expected by the code](https://github.com/Automattic/jetpack/blob/ba2dd6c30be137baad96aafb2466d0aed7e182e5/projects/plugins/jetpack/extensions/blocks/contact-info/email/save.js#L11-L28).

metro.config.js
hypest: 👍 works for me!

PR #3716, opened 2021-07-09
Fix invisible Android activity indicator

### LINKED ISSUE #3708: [Android] RN ActivityIndicator component is blank and does not show up unless explicit color prop is provided
After RN 0.64 upgrade, on Android [ActivityIndicator](https://reactnative.dev/docs/activityindicator) is blank and does not show up unless explicit `color` prop is provided.

See https://github.com/facebook/react-native/pull/30057

A possible fix is already implemented in Embed block [here](https://github.com/WordPress/gutenberg/blob/rnmobile/embed-fetching-preview/packages/block-library/src/embed/embed-loading.native.js#L22-L27). We should extract that into it's own component and 
check the code to find where `ActivityIndicator` is being used and replace them with this new component.

dcalhoun:
* **Gutenberg:** https://github.com/WordPress/gutenberg/pull/33337

Fixes wordpress-mobile/gutenberg-mobile#3708. Apply a patch for the React Native `ActivityIndicator` rendering invisible on Android. This applies the same patch that was previously applied to the Embed block to to React Native itself.

The core issue in React Native appears to be [wider reaching](https://git.io/JcNlc), but I chose to [limit the scope](https://git.io/JcNla) of the patch to the one issue we are currently facing (i.e. invisible activity indicators), as I was concerned applying a deeper, untested change might introduce regressions elsewhere.

This patch can be removed once we upgrade to a React Native version including fixes for the related issues: 

- https://github.com/facebook/react-native/issues/30056
- https://github.com/facebook/react-native/issues/29495
- https://github.com/facebook/react-native/issues/29378.

To test: See https://github.com/WordPress/gutenberg/pull/33337.

PR submission checklist:


ceyhun (approved):
Approved via https://github.com/WordPress/gutenberg/pull/33337

Issue #3471, opened 2021-05-10
Introduce allow-list and disallow-list for blocks


hypest:
The block editor will be used in various contexts, some of which (example: in the comments box in the WordPress apps) will need to restrict the block types offered.

**Describe the solution you'd like**
Let's introduce some form of allow-list for the list of blocks the editor should offer, and a disallow-list for the blocks that need to be unavailable.

For the time being, I'm thinking that those list should relate to the block picker, while the editor could still support any block in terms of rendering if found in the markup.

aerych:
>  I'm thinking that those list should relate to the block picker, while the editor could still support any block in terms of rendering if found in the markup.

Love it

ttahmouch:
## (Stupid) Questions

+ Will offered block types need restriction based on account, site, page, post, story, etc.?
+ Presumably, this means that the block types available to select would be managed via an HTTP API; not just locally, right?
+ ![image](https://user-images.githubusercontent.com/3058263/127704102-49ce8c32-562e-4155-a81e-6d42e3a1a2bf.png)

fluiddot:
Thanks @ttahmouch for asking these questions, here are the answers:

> Will offered block types need restriction based on account, site, page, post, story, etc.?

That's a good question and to be honest, I don't have a clear answer, from my POV, the offered blocks could be restricted by multiple variables but for the purpose of this task, I think it's out of the scope since we'll only provide the mechanism to restrict them and not the logic to determine the offered blocks.

> Presumably, this means that the block types available to select would be managed via an HTTP API; not just locally, right?

It will depend on the case, but in any case, this will be handled by the client that embeds the Gutenberg editor, not the editor per se. Following with my previous answer, the logic for determining the available blocks is out of the scope.

Issue #3804, opened 2021-08-05
Supported-blocks contain blocks that are conditionally enabled


chipsnyder:
**Describe the bug**
Our [supported-blocks.json](https://github.com/wordpress-mobile/gutenberg-mobile/blob/d215136be2bec40b386dec39dca4f43a29e3b0e2/src/block-support/supported-blocks.json#L32-L35) file contains blocks that are conditionally supported based on the site settings and plugin availability.
 
This data is provided to the app to help filter out layouts that the current app version wouldn't be able to support. 
We should update this list to only include those particular blocks when they are turned on via the [capabilities](https://github.com/WordPress/gutenberg/blob/27942d70e1cfaba14bc4c8d28933ff3047d56ef0/packages/react-native-bridge/ios/GutenbergBridgeDelegate.swift#L18-L30) flag. 

This might be a good exercise to tackle alongside an effort to move jetpack references out of the capabilities list.

**To Reproduce**
1. Open a site without Jetpack connected
2. Add a new page in the WP Mobile Apps
3. Notice that Jetpack-blocks are included in the network request for supported blocks

**Expected behavior**
The network request shouldn't include blocks that aren't supported by the site. 

**Additional context**
I had started thinking about one approach here but I wasn't able to get to it yet: 

We could start by adding capability specific flags that look like:
```
"blockCapabilities": {
		"reusableBlock": [
			"core/block"
		],
		"contactInfoBlock": [
			"jetpack/contact-info",
			"jetpack/email",
			"jetpack/phone",
			"jetpack/address"
		],
		"isAudioBlockMediaUploadEnabled": [
			"core/audio"
		],
		"mediaFilesCollectionBlock": [
			"jetpack/story"
		]
}

```

chipsnyder:
We have a design `contact-with-info` that's currently marked beta only. This uses the `jetpack/contact-info` block. Once we solve this we can remove the beta flag from that design.

PR #3666, opened 2021-06-25
[WIP] Contact Info block integration test

fluiddot:
**Related to issue:** https://github.com/wordpress-mobile/gutenberg-mobile/issues/3673

This PR adds an integration test for the Contact Info block (Jetpack).

Additionally, it updates the Jest configuration to able the integration tests:
- Add test helper to the module name mapper.
- Fixes transpiled packages calculation.

PR submission checklist:

- [ ] I have considered adding unit tests where possible.
- [ ] I have considered if this change warrants user-facing release notes [more info](https://github.com/wordpress-mobile/gutenberg-mobile/blob/develop/Release-notes.md) and have added them to [RELEASE-NOTES.txt](https://github.com/wordpress-mobile/gutenberg-mobile/blob/develop/RELEASE-NOTES.txt) if necessary.

PR #3811, opened 2021-08-08
TT | 3471 | "Introduce allow-list and disallow-list for blocks."

ttahmouch:
## Description

This is meant to be treated as a `draft`.

### Idea

The general idea behind this work is that there is a desire to allow the Gutenberg Block Editor to restrict the block types it offers in various contexts it is presented in to the customer. It came from [this feature request](https://github.com/wordpress-mobile/gutenberg-mobile/issues/3471).

The changes found here should, in theory, only impact the blocks presented in the block picker, and not affect the blocks displayed within the editor if someone were to manually modify the markup.

The "requirements" that I thought seemed reasonable to me would be:

0. If you pass in neither an allow list nor a disallow list, then proceed by including all registered blocks, i.e., `core/`, `jetpack/`, or otherwise.
1. If you pass in an allow list, and not a disallow list, then proceed by only including what is in the allow list from the entire set of registered blocks, i.e., `core/`, `jetpack/`, or otherwise.
2. If you pass in a disallow list, and not an allow list, then proceed by including the entire set of registered blocks, i.e., `core/`, `jetpack/`, or otherwise, sans the blocks in the disallow list.
3. If you pass in both an allow and disallow list, then proceed by ignoring the disallow list and and only including the blocks that result from the allow list detailed in `1` (as it is the more restrictive case).
    + However, this use case isn't realistic in my opinion. The consumer should only provide an allow or disallow list whichever is more convenient for the context.
    + An allow list is preferred when the amount of blocks we'd like to present to the customer is small.
    + A disallow list is preferred when the amount of blocks we'd like to present is large.

#### Previous Attempt

I made a [previous attempt](https://github.com/WordPress/gutenberg/pull/33914) to implement the concept of allowing and disallowing block types. I tried ensuring that registration of the disallowed blocks never occured using `registerBlockType` if the block name *wasn't* included the provided allow list, or if the block name *was* provided in the disallow list.

This attempt had some unintended side-effects. These are detailed in [this P2 post](https://appsp2.wordpress.com/2021/08/08/3471-introduce-allow-list-and-disallow-list-for-blocks/#previousattempt) if you would like more context.

#### Remaining Work

- [ ] Document the interfaces with JSDoc and include useful contextual knowledge.
- [x] Automate tests to reflect the requirements defined here (since it seems that this approach is okay to move forward with. I may still need some help understanding the best possible testing strategy from a functional or integration test perspective as opposed to basic unit testing.)
- [x] Consider adding a comment in the modified test explaining the reason for using the inline require (as noted by @fluiddot).
- [ ] Refine the PR description for the current approach and update the testing instructions.
- [ ] Update PR to "Ready for Review" when the PR description and testing is "ready."
- [x] Change implementation to ensure `hideBlocks` gets ignored if `showBlocks` and `hideBlocks` are provided together (if that makes sense).

## Demo

| Providing empty sets of `showBlocks` and `hideBlocks`. | Providing only the `jetpack/contact-info` block type in the `showBlocks` set while `capable`. |
| --- | --- |
|  |  |

| Providing only the `jetpack/contact-info` block type in the `hideBlocks` set while `capable`. | Providing only the `jetpack/contact-info` block type in the `showBlocks` set while `uncapable`. |
| --- | --- |
|  |  |

To test: TBD

PR submission checklist: TBD


src/allowed-blocks-setup.js
```diff
@@ -0,0 +1,54 @@
...
+
+export default (
+	props = {},
+	dependencies = {
+		dispatch: _dispatch,
+		getBlockTypes: _getBlockTypes,
+		map: _map,
+		setTimeout,
+		without: _without,
+	}
```
fluiddot: I see that this function is only being used in [this line](https://github.com/wordpress-mobile/gutenberg-mobile/pull/3811/files#diff-bfe9874d239014961b1ae4e89875a6155667db834a410aaaa2ebe3cf89820556R28) which only passes the `props` argument, so I'm wondering in which case we would need the `dependencies` argument 🤔.

src/allowed-blocks-setup.js
```diff
@@ -0,0 +1,54 @@
...
+
+		if ( uniqueHideBlocks.length > 0 ) {
+			dispatch( 'core/edit-post' ).hideBlockTypes( uniqueHideBlocks );
+			dispatch( 'core/edit-post' ).showBlockTypes( differenceShowBlocks );
+		}
+
+		if ( uniqueShowBlocks.length > 0 ) {
+			dispatch( 'core/edit-post' ).showBlockTypes( uniqueShowBlocks );
+			dispatch( 'core/edit-post' ).hideBlockTypes( differenceHideBlocks );
+		}
```
fluiddot: I'm wondering if we would need to explicitly show blocks if by default they're already showed 🤔 . If it's not required, I think it would be great if we reduce the number of calls by only calculating the block list that we need to hide and then only dispatch one action with it, something like `dispatch( 'core/edit-post' ).hideBlockTypes( blocksToHide );`, what do you think?

fluiddot (commented):
> The only drawback that I wasn't able to figure out when spiking this concept was how to listen for a completion of all block type registrations. There doesn't appear to be a specific event defined for that. So I just ensured that I waited until after all core/* and jetpack/* block types were registered in order to determine if the previously registered block types should be hidden. This unfortunately meant that I couldn't dispatch SHOW or HIDE actions to the Redux Store from the main initialization flow, but rather I had to setTimeout to ensure it happened at a future tick of the event loop after all possible block type registrations occured. This appeared to be the same basic flow that was used to determine if the jetpack/story or jetpack/contact-info block types should be hidden. I'm not fond of this implementation, but I don't think there is a robust way to handle listening for all registrations being completed at the moment.

The Jetpack blocks registration is conditioned by the capabilities, as per [this comment](https://github.com/wordpress-mobile/gutenberg-mobile/blob/157892a769373ad09ba183aa2f1e1c651610c278/src/jetpack-editor-setup.js#L55-L60), looks like the `capabilities` object obtained from the `getSettings` selector might not return the capabilities at the time the Jetpack setup is done and therefore a `setTimeout` is used. To be honest, I think the wait is not necessary and most likely there's a way to address it because the capabilities are passed through the initial props so they should be available, being this said, I wouldn't consider this flow as the best one to be based on.

Additionally and also related to Jetpack blocks, I think we're not considering the case of having a Jetpack block that was hidden by the capabilities and included in the allow list, in this case, we would be displaying a block that should be disabled, right?

src/allowed-blocks-setup.js
ttahmouch: I added an optional second argument that defaults to the real implementation of all dependency functions to allow for dependency injection to supply mock functions should they be needed in Jest unit tests. This argument would never be used in the application implementation.

Does that make sense?

fluiddot: I don't feel particularly comfortable with adding extra code that is only intended to be used by unit tests. Besides, I think with Jest we would be able to mock them without adding new code, wdyt?

ttahmouch: As we discussed in a separate thread, I have removed the `dependencies` argument from the function signature, @fluiddot .

src/allowed-blocks-setup.js
ttahmouch: I think this is redundant, @fluiddot . I originally included it for comprehensiveness. I have since removed it to ensure we're not unnecessarily dispatching and causing rerenders of the block editor UI.

ttahmouch:
I made these changes as noted in my commit log, @fluiddot . I would appreciate your feedback. I tried to explain my rationale behind the changes in our other thread, but I can bring that to the PR if it's easier to track. I'm also not opposed to trying a different approach to the implementation. This approach just appears to work well.

+ `src/allowed-blocks-setup.js`
	+ Added `jetpack/layout-grid` as an "Available Jetpack Block." However, it's availability is irrelevant since it is not registered with `registerJetpackBlock` but rather `registerBlockType`.
	+ Added a centralized entry point for registration and hiding of block types called `setupBlocks` that should be executed in the `gutenberg-mobile` `native.render` action as early as possible in the editor initialization.
	+ Removed the redundant dispatching of the `SHOW_BLOCK_TYPES` action for the inverse of the block types that are hidden with the dispatching of the `HIDE_BLOCK_TYPES` action to @fluiddot 's point.
	+ Ensured that all `jetpack/` block types export their own `registerBlock` entry point function to make registration conditional for non-`core/` block types.
	+ Added `registerJetpackBlocksIfCapable` to ensure that `jetpack/contact-info`, `jetpack/layout-grid`, and `jetpack/story` are only registered if their respective `capabilities` passed through the editor intial `props` indicate they should be enabled for the customer.
	+ Simplified the registration of all `jetpack/` block types previously done in `setupJetpackEditor` and `setJetpackData` in the new `setupJetpackBlocks` function. The `setTimeout` to ensure `jetpack/` blocks are conditionally hidden on next tick is no longer needed if not attempting to select the `capabilities` from the `core/block-editor` store, but rather just leverage the `props` passed directly into the initialization of the editor. The inline `require` is no longer needed since it now exports functions used to register lazily as opposed to implied registration when the module is loaded. `toggleBlock` is also no longer needed as the hiding of all block types, `core/`, `jetpack/`, or future types has been unified.
	+ Removed the unnecessary `dependencies` argument from the `setupAllowedBlocks` function signature. We can just leverage Jest module mocking for dependency injection when necessary.
	+ Removed the unnecessary `setTimeout` from the `setupAllowedBlocks` function as it no longer has to wait for `jetpack/` block types to be hidden first since that flow can now be done on the current tick of the event loop instead of a future tick waiting for the redux store state to have changed.
+ `src/block-experiments-setup.js`
	+ Removed the module to consolidate block type registration and hiding into `src/allowed-blocks-setup.js`.
+ `src/block-support/supported-blocks.json`
	+ Added `jetpack/layout-grid` and `jetpack/story` to the list of supported blocks. This appears to only be used in the iOS Swift React Native Bridge.
+ `src/index.js`
	+ Consolidated the entry points used for registering and hiding `jetpack/contact-info`, `jetpack/layout-grid`, and `jetpack/story` to `setupBlocks` from `setupJetpackEditor` and `setupBlockExperiments`.
+ `src/jetpack-editor-setup.js`
	+ Removed the module to consolidate block type registration and hiding into `src/allowed-blocks-setup.js`.
+ `src/test/index.js`
	+ Updated the staging of the existing test that ensures `jetpack/` block types register successfully during the editor initialization.

src/allowed-blocks-setup.js
```diff
@@ -0,0 +1,94 @@
...
+/**
+ * WordPress dependencies
+ */
+import { dispatch } from '@wordpress/data';
+import { getBlockTypes } from '@wordpress/blocks';
+/**
+ * Internal dependencies
+ */
+import { JETPACK_DATA_PATH } from '../jetpack/projects/plugins/jetpack/extensions/shared/get-jetpack-data';
+import { registerBlock as registerJetpackLayoutGridBlock } from '../block-experiments/blocks/layout-grid/src';
```
fluiddot, line 10: ~~The [block experiments](https://github.com/Automattic/block-experiments) are not part of Jetpack so probably we shouldn't use the word `Jetpack` here.~~

**EDIT:** The layout grid block is part of Jetpack as its id references it: `jetpack/layout-grid`.

src/allowed-blocks-setup.js
```diff
@@ -0,0 +1,94 @@
...
+import {
+	registerContactInfoBlock as registerJetpackContactInfoBlock,
+	registerStoryBlock as registerJetpackStoryBlock,
```
fluiddot, line 13: I'm curious about why you renamed the register functions, what benefit do you think we could get by adding the word `Jetpack`?

src/allowed-blocks-setup.js
```diff
@@ -0,0 +1,94 @@
...
+		capabilities: {
+			layoutGridBlock = false,
+			mediaFilesCollectionBlock = false,
+			contactInfoBlock = false,
+		} = {},
+	} = props;
+
+	if ( layoutGridBlock ) {
+		registerJetpackLayoutGridBlock();
+	}
```
fluiddot, line 49: I think this block is not part of Jetpack so I'd consider renaming the function `registerJetpackBlocksIfCapable` to something probably more generic or split it into two functions, one for block experiments and another for Jetpack, wdyt?

src/allowed-blocks-setup.js
```diff
@@ -0,0 +1,94 @@
...
+	'contact-info': { available: true },
+	'layout-grid': { available: true },
+	story: { available: true },
+};
+
+const setJetpackData = ( {
+	isJetpackActive = false,
+	userData: tracksUserData = null,
+	siteFragment = null,
+	blogId: wpcomBlogId = 1,
```
fluiddot, line 27: I'm wondering if we need to include the default value `1` here, as far as I checked in the [original code](https://github.com/wordpress-mobile/gutenberg-mobile/blob/6fea169a3c04e6fa441c5369d60539df2a407ca9/src/jetpack-editor-setup.js#L24-L34), I see we didn't have it, should we keep it?

src/allowed-blocks-setup.js
```diff
@@ -0,0 +1,94 @@
...
+import { registerBlock as registerJetpackLayoutGridBlock } from '../block-experiments/blocks/layout-grid/src';
+import {
+	registerContactInfoBlock as registerJetpackContactInfoBlock,
+	registerStoryBlock as registerJetpackStoryBlock,
+} from '../jetpack/projects/plugins/jetpack/extensions/editor.native';
+
+// Please also consider updating ./block-support/supported-blocks.json
+const availableJetpackBlocks = {
+	'contact-info': { available: true },
+	'layout-grid': { available: true },
```
fluiddot, line 19: As in previous comments, I think that the `layout-grid` block is not part of Jetpack so I would either rename this constant or split it into two.

src/allowed-blocks-setup.js
```diff
@@ -0,0 +1,94 @@
...
+	story: { available: true },
+};
+
+const setJetpackData = ( {
```
fluiddot: This function is not really setting data but getting so probably we should consider renaming it, wdyt?

src/allowed-blocks-setup.js
```diff
@@ -0,0 +1,94 @@
...
+	if ( contactInfoBlock ) {
+		registerJetpackContactInfoBlock();
+	}
+};
+
+export const setupJetpackBlocks = ( props = {} ) => {
+	const { jetpackState = { blogId: 1, isJetpackActive: true } } = props;
+	const { isJetpackActive = false } = jetpackState;
+
+	global.window[ JETPACK_DATA_PATH ] = setJetpackData( jetpackState );
```
fluiddot: I see that in the original code this is only done if Jetpack is active, not sure about the potential side-effects of changing it but probably we should consider keeping the flow as the original:
https://github.com/wordpress-mobile/gutenberg-mobile/blob/6fea169a3c04e6fa441c5369d60539df2a407ca9/src/jetpack-editor-setup.js#L41-L45

src/allowed-blocks-setup.js
```diff
@@ -0,0 +1,94 @@
+/**
+ * External dependencies
+ */
+import { map, without } from 'lodash';
```
fluiddot: We don't have `lodash` as a dependency in Gutenberg Mobile ([reference](https://github.com/wordpress-mobile/gutenberg-mobile/blob/develop/package.json)), it's only used in the Gutenberg repository, so I'm wondering if we should use it here. If we decide to use it, we should include the dependency although probably we could replace easily the functionality of `map` and `without`, wdyt?

src/block-support/supported-blocks.json
```diff
@@ -29,10 +29,12 @@
 		"core/file",
 		"core/search",
 		"core/block",
+		"jetpack/address",
 		"jetpack/contact-info",
 		"jetpack/email",
+		"jetpack/layout-grid",
 		"jetpack/phone",
-		"jetpack/address"
+		"jetpack/story"
```
fluiddot: This list is used to filter for starter page templates that meet a certain threshold of supported blocks ([more info in this issue](https://github.com/wordpress-mobile/gutenberg-mobile/issues/2444)). I guess it's ok to include `jetpack/layout-grid` and `jetpack/story` but it would be nice if we confirm that this won't produce undesired side-effects.

cc @chipsnyder @mchowning in case you have better context for this file.

src/test/index.js
```diff
@@ -18,8 +18,7 @@ describe( 'Test Jetpack blocks', () => {
 			() => jest.fn()
 		);
 
-		const setupJetpackEditor = require( '../jetpack-editor-setup' ).default;
-		setupJetpackEditor( { blogId: 1, isJetpackActive: true } );
+		require( '../allowed-blocks-setup' );
```
fluiddot: Usually, we import the modules with `import` instead of `require` so if possible it would be great to change this part.

Besides, I'm wondering if we could add unit tests to cover the logic of the allow/disallow lists, wdyt?

src/allowed-blocks-setup.js
```diff
@@ -0,0 +1,94 @@
...
+		registerJetpackLayoutGridBlock();
+	}
+
+	if ( mediaFilesCollectionBlock ) {
+		registerJetpackStoryBlock();
+	}
+
+	if ( contactInfoBlock ) {
+		registerJetpackContactInfoBlock();
+	}
```
fluiddot, line 57: I noticed that originally we were hiding these blocks instead of preventing registering them:
https://github.com/wordpress-mobile/gutenberg-mobile/blob/6fea169a3c04e6fa441c5369d60539df2a407ca9/src/jetpack-editor-setup.js#L68-L69
https://github.com/wordpress-mobile/gutenberg-mobile/blob/6fea169a3c04e6fa441c5369d60539df2a407ca9/src/jetpack-editor-setup.js#L47-L53

From my POV, I think it makes sense to register them conditionally to the props but I'm wondering if there was a reason to do it that way.

@illusaen do you have any insights regarding this topic?

fluiddot (commented):
Thanks @ttahmouch for making the changes on the PR 🙇.

I've added some comments in the code but I think that the PR is ready to be reviewed so please when you're available, let me know if you could change it to "Ready for review". It would be also nice if you could also wrap up the PR by explaining the current approach and provide some testing instructions for reviewers to use as a guide. 

From my side, I went ahead and tested the PR and except for a couple of issues everything is working fine 🎊 , here are the insights:

- I tested disabling some of the capabilities used in the Jetpack blocks and including the disabled blocks in the allow-list and I confirmed that they weren't displayed ✅ .
- I tested the requirements described in the PR's description and they were fulfilled except one, here are the details:

> If you pass in neither an allow list nor a disallow list, then proceed by using all core blocks.

This works as expected ✅ .

> If you pass in an allow list, and not a disallow list, then proceed by only including what is in the allow list from the entire set of core blocks.

This works as expected ✅ .

> If you pass in a disallow list, and not an allow list, then proceed by including the entire set of core blocks sans the blocks in the disallow list.

This works as expected ✅ .

> If you pass in both an allow and disallow list, then proceed by removing blocks named in the disallow list first, then keeping the blocks named in the allow list second.

For this case, I included the same block in both allow/disallow lists and I noticed that is not displayed so looks like that this requirement is not being fulfilled ❌.

I used the following values:
```
showBlocks: [ 'core/paragraph' ],
hideBlocks: [ 'core/paragraph' ],
```

src/block-support/supported-blocks.json
mchowning: I _think_ this should be fine, but let's cc: @antonis too since Chip isn't on this project anymore.

chipsnyder: I'll still bounce in. All of the layouts are also tagged on the backend to be marked to have them included so if there is an issue we could remove that flag. However worth noting is this issue I raised. https://github.com/wordpress-mobile/gutenberg-mobile/issues/3804

The Jetpack blocks aren't be stripped out for sites that don't support JetPack. So it would actually be safer to not include them until that list matches the site's capabilities. But only safer in terms of if we want to expose those layouts to users.

ttahmouch:
> Thanks @ttahmouch for making the changes on the PR 🙇.

You're welcome, @fluiddot . Thank you for being so patient through this process. 🙏

> I've added some comments in the code but I think that the PR is ready to be reviewed so please when you're available, let me know if you could change it to "Ready for review". It would be also nice if you could also wrap up the PR by explaining the current approach and provide some testing instructions for reviewers to use as a guide.

Thanks for adding comments throughout the code. I appreciate it. I didn't feel comfortable changing its status from `draft` until I knew that this approach was going to be the one we intend to move forward with. I didn't want to add unnecessary automated tests for an approach that may not realistically be used.

I will try to make the PR description clearer with respect to the current approach and update the testing instructions. Apologies if it was confusing to anyone.

> From my side, I went ahead and tested the PR and except for a couple of issues everything is working fine 🎊 , here are the insights:
> 
> * I tested disabling some of the capabilities used in the Jetpack blocks and including the disabled blocks in the allow-list and I confirmed that they weren't displayed ✅ .
> * I tested the requirements described in the PR's description and they were fulfilled except one, here are the details:
> 
> > If you pass in neither an allow list nor a disallow list, then proceed by using all core blocks.
> 
> This works as expected ✅ .
> 
> > If you pass in an allow list, and not a disallow list, then proceed by only including what is in the allow list from the entire set of core blocks.
> 
> This works as expected ✅ .
> 
> > If you pass in a disallow list, and not an allow list, then proceed by including the entire set of core blocks sans the blocks in the disallow list.
> 
> This works as expected ✅ .

Thank you for testing and providing feedback regarding your insights. 🙏 🙇

I'm glad these cases worked as intended.

> > If you pass in both an allow and disallow list, then proceed by removing blocks named in the disallow list first, then keeping the blocks named in the allow list second.
> 
> For this case, I included the same block in both allow/disallow lists and I noticed that is not displayed so looks like that this requirement is not being fulfilled ❌.
> 
> I used the following values:
> 
> ```
> showBlocks: [ 'core/paragraph' ],
> hideBlocks: [ 'core/paragraph' ],
> ```

The reason this is no longer working as intended I believe is simply because I am no longer dispatching the `SHOW` action as the PR once was. So it never removes the `core/paragraph` block type from the hide list in the store state in your example.

That being said this has always been the use case that has never made particularly intuitive sense to me. If the consumer says "I want you to take the set of all registered block types, reduce it by hiding a few of them, and then reduce it further by only showing a few of them" it would likely make practical sense for either:

0. the `hideBlocks` to just go ignored altogether and only acknowledge the `showBlocks` since it is more restrictive.
1. taking a stricter stance by simply not allowing both `hideBlocks` and `showBlocks` to be provided in unison.

I'd lean in favor of the former over the latter. I do think that ignoring the `hideBlocks` and letting `showBlocks` supersede actually makes sense. WDYT?

It really only matters in your edge case that they're providing the same block type in both sets.

**Practical Example?**
```
// Registered Blocks [ 'core/audio', 'core/paragraph', 'core/video' ]
// First, reduce set to [ 'core/audio', 'core/video' ].
// Then, reduce remaining set to [ 'core/audio' ].
hideBlocks: [ 'core/paragraph' ],
showBlocks: [ 'core/audio' ],
```

**Impractical Example?**
```
// Registered Blocks [ 'core/audio', 'core/paragraph', 'core/video' ]
// First, reduce set to [ 'core/audio', 'core/video' ].
// Then, reduce remaining set to [ 'core/paragraph' ]?
hideBlocks: [ 'core/paragraph' ],
showBlocks: [ 'core/paragraph' ],
```

- [ ] I will try to make the PR description clearer with respect to the current approach and update the testing instructions.
- [ ] I will try to update the automated tests since it seems that this approach is okay to move forward with. (I may still need some help understanding the best possible testing strategy from a functional or integration test perspective as opposed to basic unit testing.)
- [ ] I will change it to "Ready for Review" when I add automated testing and update the description.
- [ ] I will make the change to ensure `hideBlocks` gets ignored if `showBlocks` and `hideBlocks` are provided together (if that makes sense).

src/allowed-blocks-setup.js
ttahmouch: I can probably easily replace `map` and `without` with basic array `map` and `filter` to eliminate the need for the dependency. Good call on it not being included in the `package.json`. I apologize for the oversight. It felt like we were using it everywhere so I assumed it was a dependency in `gutenberg-mobile` directly and continued to use it to match style. It isn't my goto approach for array operations.

- [x] Replace dependence on `lodash` with basic array operations.

src/allowed-blocks-setup.js
ttahmouch: Okay. So explain something to me then, @fluiddot . Why is the name of the block type `jetpack/layout-grid` and not something like `experiment/layout-grid` or similar? I knew it wasn't part of the actual `jetpack` package, but the namespace of the block type didn't make practical sense to me.

- [x] Disassociate `layout-grid` from `jetpack` here and in other places if it makes sense to do so?

src/allowed-blocks-setup.js
ttahmouch: Simply, that it makes it more expressive to the reader that this is not a `core/` block type, but rather a third-party `jetpack/` type. That may not be a compelling reason, but it made it easier for me to provide a visual cue when reading the code using the `register*Block` functions that `contact-info`, `layout-grid`, and `story` were not `core/` types. Does that make sense?

src/allowed-blocks-setup.js
ttahmouch: That crossed my mind at one point as well and I simply forgot. Thanks for reminding me. I couldn't think of a good name at the time though. Naming is hard.

`toJetpackData`, `mapToJetpackData`, or similar aren't really that much more expressive even if they don't read like an accessor method name. WDYT?

- [x] Rename the function to something more accurate.

src/allowed-blocks-setup.js
ttahmouch: We sort of did have it just not in this way where it would be used if the value provided was `undefined`. We had a default for the entire object, i.e., `{ blogId: 1, isJetpackActive: true }`. I agree that it doesn't make practical sense for it to be `1` as the default though. Would `0` make more sense as a default falsey value? Although there may legitimately be a blog with `id` of `0`. So maybe `NaN`? I'm not particularly fond of simply leaving it `undefined`, but that might actually make practical sense as well. WDYT?

- [x] Adjust the default value of `blogId` to not be `1`.

src/allowed-blocks-setup.js
ttahmouch: I couldn't see any adverse side-effects of changing it because all usages of `getJetpackData` where the `JETPACK_DATA_PATH` is not set on the `global` object have reasonable default values, and the usages of the fields on the Jetpack Data are guarded.

That being said I think I just answered my own question that `wpcomBlogId` should probably remain `undefined` as a default value, and I understand the reluctance with wanting to set the field on the `global` object even if `{jetpack: { is_active: false }}` so I'll revert to not setting it at all like it was previously doing.

```
if (
    typeof window === 'object' &&
    typeof window.Jetpack_Editor_Initial_State === 'object' &&
    typeof window.Jetpack_Editor_Initial_State.tracksUserData === 'object' &&
    typeof window.Jetpack_Editor_Initial_State.wpcomBlogId !== 'undefined'
) {
```
```
if (
    window &&
    window.Jetpack_Editor_Initial_State &&
    window.Jetpack_Editor_Initial_State.siteFragment
) {
```

- [x] Revert to not setting the `JETPACK_DATA_PATH` unless Jetpack is `active`.

src/block-support/supported-blocks.json
ttahmouch: Thanks for the feedback, everyone. 🙏

- [x] Revert this list back since it sounds like it may be safer. (I was unsure exactly how to check when adding them to the supported list originally.)

src/test/index.js
ttahmouch: I don't personally prefer to use inline `require`s. The test style was already written as such. I just knew that test was going to fail because of my deletion of that module. My "fixing" that test from failing was neither an indication that I considered this PR tested with respect to the dis/allow lists, nor an indication of my preference in importing modules.

That being said, the unfortunate part of that test is that there wouldn't be a simple way to move it from being an inline `require` to an `import` operator at the top of the test module because the module being required is the subject under test and the test is asserting that the `jetpack` block collection is implicitly being registered when `import './shared/block-category';` happens from `jetpack/projects/plugins/jetpack/extensions/editor.native.js`.

I can still adjust the implementation details of the test, but I'd have to give some thought to what that may even look like as I'm unsure of the practicality of the given test at the moment.

WDYT?

src/allowed-blocks-setup.js
fluiddot: Good point, I noticed that the block references `jetpack` in its id so indeed looks it's part of Jetpack although it's not exposed from the Jetpack repository 🤔 , sorry for the misleading. Being that said, I agree with using the Jetpack reference in this block so I'll update the rest of my comments on the PR.

src/allowed-blocks-setup.js
fluiddot: Thanks for the clarification, I think it makes sense to use this naming as it provides a visual cue 👍 .

src/allowed-blocks-setup.js
fluiddot: I'm solving this conversation as I noticed that the layout grid block is part of Jetpack as its id references it: `jetpack/layout-grid`.

src/allowed-blocks-setup.js
fluiddot: Right, I see that we were setting this default value here:
https://github.com/wordpress-mobile/gutenberg-mobile/blob/6fea169a3c04e6fa441c5369d60539df2a407ca9/src/index.js#L25

To be honest I don't have enough context on why we use `1` as the default blog id but probably it won't produce any undesired side-effect if we keep it. I found that this was introduced in [this PR](https://github.com/wordpress-mobile/gutenberg-mobile/pull/1934) long ago but it doesn't provide any clue about the value.

On the Gutenberg repository, when the editor is initialized we use also the value `1` as the default value for the post id so maybe this value is the standard default value for these type of ids:
https://github.com/WordPress/gutenberg/blob/0ebe9b3a9183a907d6eaf26c3f9f1860bd29dd2e/packages/react-native-editor/src/index.js#L48

fluiddot: Since we're already setting the default value to `1` [here](https://github.com/wordpress-mobile/gutenberg-mobile/blob/6fea169a3c04e6fa441c5369d60539df2a407ca9/src/index.js#L25), I think it's ok to keep it so I'm resolving this conversion 👍 .

src/allowed-blocks-setup.js
fluiddot: I'm solving this conversation as I noticed that the layout grid block is part of Jetpack as its id references it: `jetpack/layout-grid`.

src/allowed-blocks-setup.js
fluiddot: Yep, naming is hard 😄 , to be honest, I don't have a better name in mind but I'd probably avoid using `set` in the name as it could be a bit misleading. Being this said, `mapToJetpackData` looks slightly better so I think it's a good alternative.

src/allowed-blocks-setup.js
fluiddot: No worries, I also thought that it could be used in Gutenberg Mobile until I checked its dependencies 😄 .

src/test/index.js
fluiddot: Oh I see, ok, let's keep it this way for now. However, I'd consider adding a comment in the code explaining the reason for using the inline require, this way when we tackle this in the future we'll have a better context, wdyt?

fluiddot:
> Thanks for adding comments throughout the code. I appreciate it. I didn't feel comfortable changing its status from `draft` until I knew that this approach was going to be the one we intend to move forward with. I didn't want to add unnecessary automated tests for an approach that may not realistically be used.
> 
> I will try to make the PR description clearer with respect to the current approach and update the testing instructions. Apologies if it was confusing to anyone.

No worries, ok, it makes sense, let's change it to "ready for review" once the PR is updated 👍 .

> The reason this is no longer working as intended I believe is simply because I am no longer dispatching the `SHOW` action as the PR once was. So it never removes the `core/paragraph` block type from the hide list in the store state in your example.
> 
> That being said this has always been the use case that has never made particularly intuitive sense to me. If the consumer says "I want you to take the set of all registered block types, reduce it by hiding a few of them, and then reduce it further by only showing a few of them" it would likely make practical sense for either:
> 
> 1. the `hideBlocks` to just go ignored altogether and only acknowledge the `showBlocks` since it is more restrictive.
> 2. taking a stricter stance by simply not allowing both `hideBlocks` and `showBlocks` to be provided in unison.
> 
> I'd lean in favor of the former over the latter. I do think that ignoring the `hideBlocks` and letting `showBlocks` supersede actually makes sense. WDYT?

Yeah, since providing both lists is an edge case, I think ignoring `hideBlocks` in case `showBlocks` is defined makes sense. One thing it would be great is to add a comment in the code explaining this behavior to prevent potential misleading, wdyt?

> * [ ]  I will try to update the automated tests since it seems that this approach is okay to move forward with. (I may still need some help understanding the best possible testing strategy from a functional or integration test perspective as opposed to basic unit testing.)

If you're referring to UI tests with "automated tests", not sure if we'll be able to cover this feature with them because, as far as I know, the initial props can't be set on this type of test.

src/allowed-blocks-setup.js
```diff
@@ -0,0 +1,94 @@
...
+		registerJetpackBlocksIfCapable( props );
+	}
+};
+
+export const setupAllowedBlocks = ( props = {} ) => {
+	const { showBlocks = [], hideBlocks = [] } = props;
+	const blocks = map( getBlockTypes(), 'name' );
+	const uniqueHideBlocks = [ ...new Set( hideBlocks ) ];
+	const uniqueShowBlocks = [ ...new Set( showBlocks ) ];
+	const differenceHideBlocks = without( blocks, ...uniqueShowBlocks );
```
mchowning: Would it be worth moving lines 77 and 80 inside the `if ( uniqueShowBlocks.length > 0 )` block since it looks like that is the only place they are used?

mchowning:
> Yeah, since providing both lists is an edge case, I think ignoring hideBlocks in case showBlocks is defined makes sense. One thing it would be great is to add a comment in the code explaining this behavior to prevent potential misleading, wdyt?

Is this maybe highlighting that we could be modeling this a bit better? Did we consider not having two lists and instead just providing one with a show/hide toggle (this might have been what you meant @ttahmouch when you [suggested](https://github.com/wordpress-mobile/gutenberg-mobile/pull/3811#issuecomment-899980204) "taking a stricter stance by simply not allowing both hideBlocks and showBlocks to be provided in unison")? I'm thinking of something like:

```
{
  blocks: [string],
  blockAction: "disallow" | "onlyAllow"
}
```

Feels like that would both make the behavior more transparent and simplify the implementation. I apologize if you already considered this, and I just missed it. I do realize this PR is in a relatively late stage of revision, and I am fine with the ignore-one-of-the-lists-if-both-are-defined since it sounds like that is what you prefer.

src/allowed-blocks-setup.js
ttahmouch: No problem. It was just confusing why it would be related to `jetpack` but not be offered from the `jetpack` package. Is it something that should change ownership eventually from `block-experiments` to `jetpack`?

I'm unsure, but `block-experiments` feels like inherent technical debt if an "experiment" results in long-term production use instead of leveraging something like a feature toggle in the `jetpack` package for "beta" features.

src/allowed-blocks-setup.js
ttahmouch: Yeah, that's basically what I gathered as well when reading through the code. It honestly didn't seem intuitive to me that the default value be something like `1` as I would assume `1` could literally be the second post ever made. Something like `NaN` would probably provide a better cue, but we are defaulting to `1` if the `post` is not passed as a `prop` to the `Editor` component. 🤷 

Thanks for sending the links over.

src/allowed-blocks-setup.js
ttahmouch: Done.

src/allowed-blocks-setup.js
ttahmouch: Done.

src/allowed-blocks-setup.js
ttahmouch: Done.

src/block-support/supported-blocks.json
ttahmouch: Done.

Admittedly, this part wasn't completely clear to me.
> But only safer in terms of if we want to expose those layouts to users.

src/allowed-blocks-setup.js
ttahmouch: - [x] Would it be worth moving lines 77 and 80 inside the if ( uniqueShowBlocks.length > 0 ) block since it looks like that is the only place they are used?

I inadvertently did this when updating them to not rely on `lodash`.

src/test/index.js
ttahmouch: - [ ] Consider adding a comment in the code explaining the reason for using the inline require, this way when we tackle this in the future we'll have a better context.

Sounds good to me.

ttahmouch: I haven't started refining the tests yet. However, I'm consolidating remaining checklist items in the PR description. So I'm resolving this conversation in the meantime. I hope that is okay with you, @fluiddot . It's becoming difficult to manage a lot of threads.

ttahmouch:
> If you're referring to UI tests with "automated tests", not sure if we'll be able to cover this feature with them because, as far as I know, the initial props can't be set on this type of test.

I was referring to mostly if we could test the integration of the registration flow with the hiding flow instead of just unit testing the individual functions in isolation. I usually prefer to integration test, or higher-level functional test, over isolated unit tests with everything mocked since it more accurately reflects the tested use cases from the customers' user experience. It's just a personal preference.

To be pedantic, I don't even necessarily mean functional testing as "running the application and interacting with the visual elements on the screen from a 'black box'" with something like Selenium/Appium/Espresso/XCUITest. Even something that simulates testing like that from the component layer, e.g., `@testing-library/react-native`, would apply.

Does this make sense, @fluiddot ?

ttahmouch:
> > Yeah, since providing both lists is an edge case, I think ignoring hideBlocks in case showBlocks is defined makes sense. One thing it would be great is to add a comment in the code explaining this behavior to prevent potential misleading, wdyt?
> 
> Is this maybe highlighting that we could be modeling this a bit better? Did we consider not having two lists and instead just providing one with a show/hide toggle (this might have been what you meant @ttahmouch when you [suggested](https://github.com/wordpress-mobile/gutenberg-mobile/pull/3811#issuecomment-899980204) "taking a stricter stance by simply not allowing both hideBlocks and showBlocks to be provided in unison")? I'm thinking of something like:
> 
> ```
> {
>   blocks: [string],
>   blockAction: "disallow" | "onlyAllow"
> }
> ```
> 
> Feels like that would both make the behavior more transparent and simplify the implementation. I apologize if you already considered this, and I just missed it. I do realize this PR is in a relatively late stage of revision, and I am fine with the ignore-one-of-the-lists-if-both-are-defined since it sounds like that is what you prefer.

It's not my "preference." I did think that intuitively reading the interface as `show` or `hide` would be more descriptive as to the intent behind the property when initializing the editor. I do agree that having a single list would simplify the implementation details of the block hiding flow, but it also just felt like a micro-optimization for a use case that I didn't think would ever practically exist. It's just something that I thought would make a good case for explaining in a JSDoc.

I am not opposed to changing the interface for the benefit of the consumer of the block editor. I just don't feel strongly that the approach using a single list would necessarily make the interface more intuitive since practically speaking the use case is not realistic.

If the consumer only wanted to show a few blocks, then they should prefer the `showBlocks` interface. If the consumer only wanted hide a few blocks, then they should prefer the `hideBlocks` interface. Realistically, providing both sets is effectively the same as providing the more restrictive of the two.

Does that make sense, @mchowning ?

### Aside

If we did attempt it with an approach similar to `{blocks: [], blockAction: ""}`, then I think we'd have to also handle a different edge case or two. We'd realistically want to have the default `blockAction` be `undefined` or `"default"`, or however we'd name it, to indicate that we'd neither like to allow nor disallow a subset of blocks, but what happens if a `blocks` set is still provided? Do we ignore it (treat it as empty)? Or do we default the `blockAction` to an action other than `"default"`, e.g., "onlyAllow"?

It may just be shifting the concern? Does that make sense?

To be clear, I like the concept and appreciate your feedback. I'm happy to make changes if the way that I've been implementing it so far is not particularly intuitive. I just took a first pass at it as a draft. It wasn't necessarily something I was married to and intending to be my final pass.

mchowning:
> I am not opposed to changing the interface for the benefit of the consumer of the block editor. I just don't feel strongly that the approach using a single list would necessarily make the interface more intuitive since practically speaking the use case is not realistic.

What you say makes sense, and I'm totally fine sticking with the current object structure. I personally like a single list a bit better, but I think that this is something that is firmly in the realm of personal preference (and your approach is probably the more javascript-y™ one), so let's stick with it.

fluiddot:
> > If you're referring to UI tests with "automated tests", not sure if we'll be able to cover this feature with them because, as far as I know, the initial props can't be set on this type of test.
> 
> I was referring to mostly if we could test the integration of the registration flow with the hiding flow instead of just unit testing the individual functions in isolation. I usually prefer to integration test, or higher-level functional test, over isolated unit tests with everything mocked since it more accurately reflects the tested use cases from the customers' user experience. It's just a personal preference.
> 
> To be pedantic, I don't even necessarily mean functional testing as "running the application and interacting with the visual elements on the screen from a 'black box'" with something like Selenium/Appium/Espresso/XCUITest. Even something that simulates testing like that from the component layer, e.g., `@testing-library/react-native`, would apply.
> 
> Does this make sense, @fluiddot ?

It makes totally sense, I just wanted to make sure that we were talking about integration tests and not UI tests, thanks for the clarification 👍  .

I like the idea to use `@testing-library/react-native`, in fact, we recently added a guideline for integration tests using this library ([reference](https://github.com/WordPress/gutenberg/blob/trunk/docs/contributors/code/native-mobile-integration-test-guide.md)) but it's meant to be used in the tests defined in the Gutenberg repository only. However, if you would like to use it, some time ago I worked on [a PR](https://github.com/wordpress-mobile/gutenberg-mobile/pull/3666/files#diff-1e058ca1442e46581b13571fb8d261f9e1f5657e26c96634d4c1072f0f0347f1) that enable them in Gutenberg Mobile, so feel free to take those changes in case they help you in this purpose.

Being this said, I foresee that the integration tests might take extra time so maybe we could add them later on and in the meantime address the rest of the subtasks:

> - [ ] I will try to make the PR description clearer with respect to the current approach and update the testing instructions.
> - [ ] I will make the change to ensure `hideBlocks` gets ignored if `showBlocks` and `hideBlocks` are provided together (if that makes sense).

What do you think?

ttahmouch:
> Being this said, I foresee that the integration tests might take extra time so maybe we could add them later on and in the meantime address the rest of the subtasks:
> 
> > * [ ]  I will try to make the PR description clearer with respect to the current approach and update the testing instructions.
> > * [ ]  I will make the change to ensure `hideBlocks` gets ignored if `showBlocks` and `hideBlocks` are provided together (if that makes sense).
> 
> What do you think?

That's totally reasonable, @fluiddot .

I finally had a few minutes to think about what tests I intend to write, and this is a crude list of them just in case you'd like to chime in. I hope they make sense since I wrote them pretty quickly.

### Integration Tests

Given `native.render` hook is emitted,
When `setupBlocks` takes place,
Then ...

+ Ensure that the `jetpack/layout-grid` block is registered if capable.
+ Ensure that the `jetpack/layout-grid` block is not registered if incapable.
+ Ensure that the `jetpack/story` block is registered if capable.
+ Ensure that the `jetpack/story` block is not registered if incapable.
+ Ensure that the `jetpack/contact-info` block is registered if capable.
+ Ensure that the `jetpack/contact-info` block is not registered if incapable.
+ Ensure that Jetpack Data is set if Jetpack is Active.
+ Ensure that Jetpack Data is not set if Jetpack is Inactive.
+ Ensure that the `HIDE` action is not dispatched if neither `hideBlocks` nor `showBlocks` types are provided.
+ Ensure that the `HIDE` action is dispatched with hide types if `hideBlocks` types are provided and `showBlocks` types are not provided.
+ Ensure that the `HIDE` action is dispatched with inverse show types if `showBlocks` types are provided and `hideBlocks` types are not provided.
+ Ensure that the `HIDE` action is dispatched with inverse show types if `showBlocks` types are provided and `hideBlocks` types are provided.

src/test/index.js
```diff
@@ -1,27 +1,481 @@
...
+ * Mobile editor, and `defaultHooks.doAction( 'native.render', props )` is the `native.render` hook that would normally be
+ * listened for when the editor is initializing. So all the tests ensure that when certain `props` are passed in from the
+ * `native.render` hook that the underlying Redux Store state is reduced expectedly. This is deliberately done to keep
+ * the tests as black-box as possible such that the implementation details of the application code may be changed more
+ * freely without a tight coupling on the code that tests it.
+ */
+describe( 'Gutenberg Mobile Initialization', () => {
+	beforeAll( () => {
+		jest.unmock( '@wordpress/api-fetch' );
+		jest.unmock( '@wordpress/blocks' );
```
fluiddot, line 31: I'm wondering why we need to un-mock these packages, have you experienced any issues when running the tests?

src/test/index.js
```diff
@@ -1,27 +1,481 @@
...
+	} );
+
+	describe( 'Jetpack Blocks', () => {
+		it( `should register Jetpack block collection.`, () => {
+			// Arrange
+			const expectedCollectionTitle = 'Jetpack';
+			const props = { jetpackState: { isJetpackActive: false } };
+			require( '../index' );
```
fluiddot, line 39: I haven't tested it but would it be possible to import this only once as the rest of the import statements, located at the top of the file? I don't see anything on [that file](https://github.com/wordpress-mobile/gutenberg-mobile/blob/a6fe7103497041bd37902d4415786618d4678c63/src/index.js) that would require to import it on every test 🤔, am I right?

fluiddot (commented):
Thank you very much @ttahmouch for making these changes and adding such thorough integration tests for covering all the logic 🎊 .

I added a couple of comments but since this PR will probably remain paused for a while, there's no rush on reviewing them.

PR #3817, opened 2021-08-10
Color settings - Update label and show the selected color value

### LINKED ISSUE #3766: Improve footer copy on color picker BottomSheet
**Describe the bug**

We've heard that some users assume that the "Select a color" label is an action button. I believe we should phrase this slightly more clearly and potentially change what is displayed depending on whether or not the user has a color selected. 

**To Reproduce**

1. Add any block that supports colors (paragraph, button, group, etc)
2. Open the block's settings by tapping on the settings/cog icon near the block's content
3. Tap on one of the color setting cells (i.e. text color or background color)
4. Notice that the footer of the BottomSheet has a text label that says "Select a color" (this is where the confusion arises)

<a href="https://user-images.githubusercontent.com/1200432/127396928-3e42d9fa-5d81-4a56-a494-aff8b1405595.jpeg"></a>

**Proposed solution(s)**

I propose that we change the label to “Select a color above” (other options considered were "Select color above", "No color selected"). This should be a bit clearer, as it directs users to the color-selection control.

We're also not displaying the selected color's hex value (like we are on the "custom color" picker — in monospace font and the ability to long-press to copy the associated hex value) when the user has a color selected, but it would be good to do so; both for consistency and visibility.

Here's an overview of the interaction:

image

geriux:
Fixes https://github.com/wordpress-mobile/gutenberg-mobile/issues/3766

`Gutenberg PR` -> https://github.com/WordPress/gutenberg/pull/33973

PR submission checklist:


antonis (approved):
Approved via https://github.com/WordPress/gutenberg/pull/33973#pullrequestreview-726486893 🎉

PR #3858, opened 2021-08-20
Add fallback cache key to NPM install steps

fluiddot:
This PR addresses the issue that happens randomly on CI jobs leading to the error `cb() never called!` when installing the dependencies.

Using a fallback cache key will assure that the NPM cache is always present, which as far as we know, looks like this is the only way to prevent the issue. It's important to note that this practice is not recommended because will lead to an increase of the NPM cache size over time (more info in [this interesting article](https://glebbahmutov.com/blog/do-not-let-npm-cache-snowball/)). Nevertheless, since the frequency of this issue is increasing and is slowing down the development on most of PRs, I think it's worth the sacrifice.

**To test:**
- Run all CI jobs.
- Verify that all of them pass.

PR submission checklist:


antonis (approved):
Thank you for tackling this issue @fluiddot 🙇 
The configuration change LGTM and works as advertised 🎉

PR #3867, opened 2021-08-24
Enable Inserter Search

jhnstn:
Fixes #
Companion PR https://github.com/WordPress/gutenberg/pull/34129

To test:
See companion PR

PR submission checklist:


mchowning (approved):
Great work! Just need those release notes and we can :shipit: !

mchowning (approved):
Retested both demo app and WPAndroid release build and re-approved in: https://github.com/WordPress/gutenberg/pull/34129#pullrequestreview-738557525

geriux:
Hey 👋 FYI this PR deleted some localization strings that we should keep 😅

There are two issues when generating the strings that were already reported and we are waiting for an update of the library:

- https://github.com/wp-cli/i18n-command/issues/280
- https://github.com/wp-cli/i18n-command/issues/281

mchowning:
Thanks for catching this @geriux ! Do you think we should just revert all the string changes and let the strings get handled during the release process? I see there was one string addition, but I don't think it has anything to do with this feature.

geriux:
> Do you think we should just revert all the string changes and let the strings get handled during the release process? I see there was one string addition, but I don't think it has anything to do with this feature.

I think we could revert all of the changes yes, that addition is for a feature that is currently under development (DEV flag) so it's safe to do so. 😃

PR #3903, opened 2021-09-01
Register Jetpack Tiled Gallery block

guarani:
Fixes https://github.com/wordpress-mobile/gutenberg-mobile/issues/3904

Related:
- https://github.com/Automattic/jetpack/pull/20925
- https://github.com/WordPress/gutenberg/pull/34472

## To test

Run the demo app locally and verify that the Tiled Gallery block can be inserted into a post. Note that the block itself isn't meant to work just yet.

1. Run the packager by executing `npm run start:reset` in one terminal window
2. Run the demo app for iOS in another terminal window, `npm run core ios`
3. Verify the block can be inserted via the block inserter, the block inserter search, and the slash inserter
4. Run the demo app for Android, `npm run core android`
5. Repeat step 3
6. Run the integration tests and verify it works as expected, `npm run test src/jetpack/test/tiled-gallery.js`

PR submission checklist:

### DIFF
--- a/src/block-support/supported-blocks.json
+++ b/src/block-support/supported-blocks.json
@@ -32,7 +32,8 @@
 		"jetpack/contact-info",
 		"jetpack/email",
 		"jetpack/phone",
-		"jetpack/address"
+		"jetpack/address",
+		"jetpack/tiled-gallery"
 	],
 	"devOnly": [ "core/code" ],
 	"iOSOnly": [],
--- a/src/jetpack/test/tiled-gallery.js
+++ b/src/jetpack/test/tiled-gallery.js
@@ -1,0 +1,74 @@
+/**
+ * External dependencies
+ */
+import {
+	getEditorHtml,
+	initializeEditor,
+	fireEvent,
+	waitFor,
+	within,
+} from 'test/helpers';
+
+/**
+ * WordPress dependencies
+ */
+import { getBlockTypes, unregisterBlockType } from '@wordpress/blocks';
+import { registerCoreBlocks } from '@wordpress/block-library';
+
+/**
+ * Internal dependencies
+ */
+import setupJetpackEditor from '../../jetpack-editor-setup';
+
+beforeAll( () => {
+	// Register all core blocks
+	registerCoreBlocks();
+	// Register Jetpack blocks
+	setupJetpackEditor( { blogId: 1, isJetpackActive: true } );
+} );
+
+afterAll( () => {
+	// Clean up registered blocks
+	getBlockTypes().forEach( ( block ) => {
+		unregisterBlockType( block.name );
+	} );
+} );
+
+describe( 'Tiled Gallery block', () => {
+	it( 'inserts the block', async () => {
+		const {
+			getByA11yLabel,
+			getByTestId,
+			getByText,
+			debug,
+		} = await initializeEditor( {
+			initialHtml: '',
+			capabilities: { tiledGalleryBlock: true },
+		} );
+
+		// Open the inserter menu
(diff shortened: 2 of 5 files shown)


guarani:
👋 @fluiddot, on this new Jetpack Tiled Gallery block, I wanted to add integration tests, so I cherry-picked your commits from https://github.com/wordpress-mobile/gutenberg-mobile/pull/3666 here.
This PR targets a feature branch, not `develop`. My thinking is that when the time comes to merge the feature branch to `develop`, we can look at how to handle it best. In the meantime, I just wanted to give you a heads-up and say thanks for doing this work! I'm open to doing this differently of course, just let me know if you think there's a better way to do this.

src/block-support/supported-blocks.json
```diff
@@ -32,7 +32,8 @@
 		"jetpack/contact-info",
 		"jetpack/email",
 		"jetpack/phone",
-		"jetpack/address"
+		"jetpack/address",
+		"jetpack/tiled-gallery"
```
guarani, line 36: Context for this change https://github.com/wordpress-mobile/gutenberg-mobile/pull/2589

fluiddot:
> 👋 @fluiddot, on this new Jetpack Tiled Gallery block, I wanted to add integration tests, so I cherry-picked your commits from #3666 here.
> This PR targets a feature branch, not `develop`. My thinking is that when the time comes to merge the feature branch to `develop`, we can look at how to handle it best. In the meantime, I just wanted to give you a heads-up and say thanks for doing this work! I'm open to doing this differently of course, just let me know if you think there's a better way to do this.

Thanks for the heads up 🙇 !

I'd like to spare some time to do a final review on #3666 and merge it, it won't add too many tests for the Contact info block but I think it's worth it anyway. If this feature gets merged before, I think it would be ok as it will introduce the support for integration tests in Gutenberg Mobile.

src/jetpack/test/tiled-gallery.js
```diff
@@ -0,0 +1,74 @@
...
+			getByTestId,
+			getByText,
+			debug,
+		} = await initializeEditor( {
+			initialHtml: '',
+			capabilities: { tiledGalleryBlock: true },
+		} );
+
+		// Open the inserter menu
+		fireEvent.press( await waitFor( () => getByA11yLabel( 'Add block' ) ) );
```
SiobhyB, line 50: While putting research into another PR, I found that [@testing-library/user-event](https://github.com/testing-library/user-event) is a package that's built on top of `fireEvent` and has some additional methods. I added more details in my comment at https://github.com/WordPress/gutenberg/pull/34431#discussion_r701418921, including the fact `userEvent` is being used on the web side of Gutenberg. Are there any blockers or reasons for not using it over `fireEvent` for native tests, too? This isn't a blocker but more curiosity on my side!

SiobhyB (approved):
LGTM, the tests run smoothly. 🎉 I only added one comment with a question, but no blockers.

src/jetpack/test/tiled-gallery.js
SiobhyB: It was shared in the other thread (https://github.com/WordPress/gutenberg/pull/34431#discussion_r701925251) that `userEvent` isn't currently supported in React Native's testing library, which answers my question here. :) I'll go ahead to mark this as resolved.

guarani: Thanks for sharing the link to those comments on the other PR! I wasn't aware of this and it looks interesting. I'd just been using `fireEvent.press` because I'd seen it used elsewhere, but good to know it is getting improvements to make the event more like an actual user press.

PR #3761, opened 2021-07-27
Enforce to use the node version specified by Gutenberg on CircleCI workflows

fluiddot:
I noticed that currently, the node version doesn't match in most cases with the one we try to install in CircleCI workflows:
Screenshot 2021-07-27 at 13 30 34

This PR adds a new command (`install-node-version`) to CircleCI configuration to handle the installation of the node version specified by Gutenberg via NVM. This command will be used in all workflows to consolidate the use of a unique node version along with all the processes.

**NOTE: This issue was spotted on https://github.com/wordpress-mobile/gutenberg-mobile/pull/3760, I cherry-picked these changes in that branch to verify that fix the original problem.**

To test:
- Verify that all PR checks pass.
- Verify that all workflows use the same node version `14.17.3`.

PR submission checklist:

- [x] I have considered adding unit tests where possible.
- [x] I have considered if this change warrants user-facing release notes [more info](https://github.com/wordpress-mobile/gutenberg-mobile/blob/develop/Release-notes.md) and have added them to [RELEASE-NOTES.txt](https://github.com/wordpress-mobile/gutenberg-mobile/blob/develop/RELEASE-NOTES.txt) if necessary.

PR #3990, opened 2021-09-20
Fix error "nvm not found" in CircleCI workflows

fluiddot:
Fixes the error `nvm: command not found` that is shown in some steps of the CircleCI workflows like "Check correctness":
https://app.circleci.com/pipelines/github/wordpress-mobile/gutenberg-mobile/14186/workflows/9eb80d13-b000-4ad3-b494-b2999c204598/jobs/77249

The issue was caused when running bash scripts, as far as I checked, before executing the script, it was running the command `nvm use default` that is set in the Bash environment (`$BASH_ENV`) but failed due to not having the command `nvm` defined. This was fixed by adding the NVM loading command to the Bash environment that assures that NVM will be always available.

**To test:**
Verify that all CI checks pass and that the error is no longer shown in the logs.

**Last "Check correctness" run job:**
https://app.circleci.com/pipelines/github/wordpress-mobile/gutenberg-mobile/14161/workflows/a43133cd-e79f-43d4-9b27-97998319ab71/jobs/77133?invite=true#step-110-3

PR submission checklist:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -119,7 +119,7 @@ commands:
           command: |
             curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.37.2/install.sh | bash
             if [ -z "${NVM_DIR}" ] ; then
-              echo 'Appending nvm source string to BASH_ENV'
+              echo 'Default NVM_DIR to .nvm in home'
               echo 'export NVM_DIR="$HOME/.nvm"' >> $BASH_ENV
             fi
             echo '[ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"' >> $BASH_ENV
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -131,7 +131,7 @@ commands:
             # Set the installed version as the default one
             nvm alias default $(nvm current)
             # Enforce to use the default version in the rest of workflow
-            echo 'nvm use default' >> $BASH_ENV
+            echo 'nvm use default --silent' >> $BASH_ENV
 
 parameters:
   android-docker-image:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -131,7 +131,7 @@ commands:
             # Set the installed version as the default one
             nvm alias default $(nvm current)
             # Enforce to use the default version in the rest of workflow
-            echo 'nvm use default --silent' >> $BASH_ENV
+            echo 'nvm use default' >> $BASH_ENV
 
 parameters:
   android-docker-image:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -121,8 +121,8 @@ commands:
             if [ -z "${NVM_DIR}" ] ; then
               echo 'Appending nvm source string to BASH_ENV'
               echo 'export NVM_DIR="$HOME/.nvm"' >> $BASH_ENV
-              echo '[ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"' >> $BASH_ENV
             fi
+            echo '[ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"' >> $BASH_ENV
       - run:
           name: Install node version specified in Gutenberg using nvm
           command: |


.circleci/config.yml
```diff
@@ -131,7 +131,7 @@ commands:
             # Set the installed version as the default one
             nvm alias default $(nvm current)
             # Enforce to use the default version in the rest of workflow
-            echo 'nvm use default' >> $BASH_ENV
+            echo 'nvm use default --silent' >> $BASH_ENV
```
fluiddot: I decided to silent the output of this command because it was logging `Now using node v14.17.6 (npm v6.14.15)` on every `run` step of the workflows. Besides, for the "Check correctness" workflow, it was logging it two times ([reference](https://app.circleci.com/pipelines/github/wordpress-mobile/gutenberg-mobile/14158/workflows/309e8e7d-b45f-4a46-985b-ed5bd060f1e2/jobs/77114?invite=true#step-110-3)), one when running the "Run checks" job step and then when running the Bash script `bin/ci-checks-js.sh`.

.circleci/config.yml
```diff
@@ -121,8 +121,8 @@ commands:
             if [ -z "${NVM_DIR}" ] ; then
               echo 'Appending nvm source string to BASH_ENV'
               echo 'export NVM_DIR="$HOME/.nvm"' >> $BASH_ENV
-              echo '[ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"' >> $BASH_ENV
             fi
+            echo '[ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"' >> $BASH_ENV
```
hypest, line 125: From what I can tell, this diff is only moving the line to outside the conditional, right? Does that mean that currently the conditional actually fails? If that's the case, I wonder if we need to improve the conditional or revise it?

fluiddot: Yes, it's only moving the line outside, in theory, the `[ -z "${NVM_DIR}" ]` condition is not required, as it's described in [NVM readme](https://github.com/nvm-sh/nvm/tree/v0.37.0#install--update-script). However, when I introduced these lines in [this PR](https://github.com/wordpress-mobile/gutenberg-mobile/pull/3761/files), I added the condition to prevent path conflicts in case NVM was already installed.

For the line moved, I originally thought that if `$NVM_DIR` is defined, it implied that NVM would be accessible along the workflow, however, this issue uncovered that it's possible to have NVM installed but not loaded at the time when the commands of the Bash environment (`$BASH_ENV`) are executed. For this reason, I decided to move it outside.

hypest: Thanks for the context Carlos! I see, it makes sense to me.

Nit, now that the conditional only has the $NVM_DIR definition clause, the inline comment (`Appending nvm source string to BASH_ENV'`) doesn't read accurate anymore. Want to change it to something like "Default NVM_DIR to .nvm in home" or something?

.circleci/config.yml
hypest: No strong feelings but I wonder if the versions callout could be helpful during debugging of the jobs (when needed) and the output is not too verbose.

Yes, the double output in `bin/ci-checks-js.sh` is not great, but not too bad anyway. If anything, I'd try to figure out how to optimize to only have to call `nvm` once in that script, but feels like out of scope for this PR.

.circleci/config.yml
fluiddot: > Nit, now that the conditional only has the $NVM_DIR definition clause, the inline comment (Appending nvm source string to BASH_ENV') doesn't read accurate anymore. Want to change it to something like "Default NVM_DIR to .nvm in home" or something?

Good point, that message should now only reference the `NVM_DIR`, I'll change it to "Default NVM_DIR to .nvm in home".

.circleci/config.yml
fluiddot: > No strong feelings but I wonder if the versions callout could be helpful during debugging of the jobs (when needed) and the output is not too verbose.

Right, it could be useful for debugging, I'll revert the change.

fluiddot: > Yes, the double output in bin/ci-checks-js.sh is not great, but not too bad anyway. If anything, I'd try to figure out how to optimize to only have to call nvm once in that script, but feels like out of scope for this PR.

I tried to debug this by connecting via SSH to CircleCI but I couldn't figure out a way to prevent it, it would be great if we investigate it further in the future although I think it's a minor issue.

.circleci/config.yml
fluiddot: This suggestion was added in [this commit](https://github.com/wordpress-mobile/gutenberg-mobile/pull/3990/commits/24adc32c8e143f7715d0112fdcaa6abe1ab4b663).

.circleci/config.yml
fluiddot: This suggestion was added in [this commit](https://github.com/wordpress-mobile/gutenberg-mobile/pull/3990/commits/d0fdfa122f3c568241360dcb9c890a30d43fa18d).

hypest (approved):
LGTM!

mkevins:
Thanks for addressing this Carlos!

PR #4065, opened 2021-10-01
Fix missing translation strings

enejb:
Currently when you run the build script. We end up with deleted translations. 
This is because we accidentally exclude all the *.js files even though we also say that we should include them. This results in the translations that are only defined in the mobile files to not be removed. 

This Pr fixes this by making sure that *.js files are included as expected.

To test:
- Run `npm run genstrings:ios ` - notice that the translations files stay the same.
- RUn `npm run genstrings:android` - notice that the translations files stay the same.

PR submission checklist:


ttahmouch (approved):
I couldn't think of an edge case where it would make sense to exclude `web` `.js` modules like we discussed the other day when we discovered this. However, I also didn't understand intuitively the point of even having the `exclusion` list at all if the `inclusion` would suffice.

I think @fluiddot may know best since I think that's who modified the `exclusion` list if I remember correctly.

**Note:** I haven't actually attempted to play around with the `npm` scripts at all locally yet. I just stepped through troubleshooting with @enejb on his workstation when pairing the other day.

fluiddot:
I originally included the `*.js` pattern into the exclude list due to [this issue](https://github.com/wp-cli/i18n-command/pull/262), [in the `Extra strings` section of the PR that introduced the change you can find more info](https://github.com/wordpress-mobile/gutenberg-mobile/issues/3466). I've noticed that this issue is already fixed, so probably this change is safe to be merged 👍.

However, I'd like to note that the missing translations are not caused due to file exclusion but parsing issues, specifically the following issues:
- [Files with comments in JSX are not being parsed if placed before an inline expression](https://github.com/wp-cli/i18n-command/issues/280)
- [Files with React's fragment short syntax are not being parsed](https://github.com/wp-cli/i18n-command/issues/281)

As you can see, the issues are already addressed but not released in a new version of the WP-CLI command, so we're not using them yet.

enejb:
Thanks for clarifying this @fluiddot

After further testing, this doesn't do what we thought it would do. :( 
I am going to close this for now.

guarani (commented):
Thanks for working on the fix!

When I run `npm run genstrings:ios` I see strings being removed.

fluiddot:
> Thanks for clarifying this @fluiddot
> 
> After further testing, this doesn't do what we thought it would do. :( I am going to close this for now.

One option that we could try, in the meantime a new WP-CLI version is released, is to install a development version of the [`i18-command`](https://github.com/wp-cli/i18n-command) that includes the fixes I mentioned in [the previous comment](https://github.com/wordpress-mobile/gutenberg-mobile/pull/4065#issuecomment-932352175). I noticed that there's a way to [install plugins](https://developer.wordpress.org/cli/commands/plugin/install/), so maybe there's a way to install the command using a GitHub URL that points to a commit hash 🤔 .

enejb:
The idea that @ttahmouch  and I had was to try to build the translations from a special bundle file. 
This way all the translations would be included. (Jetpack and block-experiments as well) in order to get that to work, we would have to make sure that all the comments are preserved. We were making some progress with the creating a special bundle but the process right now to get the translations definitely seems a bit complex. 
> js => php => pot => (swift or xml) 

It might be worth looking into if we can just get the metro to generate the swift and xml files directly. 🤔

fluiddot:
> The idea that @ttahmouch and I had was to try to build the translations from a special bundle file. This way all the translations would be included. (Jetpack and block-experiments as well) in order to get that to work, we would have to make sure that all the comments are preserved. We were making some progress with the creating a special bundle but the process right now to get the translations definitely seems a bit complex.
> 
> > js => php => pot => (swift or xml)
> 
> It might be worth looking into if we can just get the metro to generate the swift and xml files directly. 🤔

That's interesting, it would be great if we could fix and improve the translation process. One thing that we need to take into account is that the localization strings files we generate (the swift and xml files) are partial, as they only contain the strings that are not shared with the web version, in fact, most of the translated strings come from the [Gutenberg's Glotpress project](https://translate.wordpress.org/projects/wp-plugins/gutenberg/).

Let me know if I could help you with that task as I've dealt (and suffered 😅) with the translations process some time ago.

PR #4066, opened 2021-10-01
Setup Redux middleware for analytics tracking

### LINKED ISSUE #4009: Set up Redux Middleware in Gutenberg Mobile
To start tracking user interactions in Gutenberg, we will use Redux middleware to tap into Redux events. 

The middleware will live in Gutenberg Mobile

jhnstn:
## Related PRs

* https://github.com/WordPress/gutenberg/pull/35272
* https://github.com/wordpress-mobile/WordPress-Android/pull/15406
* https://github.com/wordpress-mobile/WordPress-iOS/pull/17243

## Description

Fixes #4009. This adds the ability to send Redux actions over the RN bridge for processing on the host apps.

To test: See [WPiOS](https://github.com/wordpress-mobile/WordPress-iOS/pull/17243) and [WPAndroid](https://github.com/wordpress-mobile/WordPress-Android/pull/15406) PRs. 

PR submission checklist:


src/analytics/events.js
```diff
@@ -0,0 +1,5 @@
+export const INSERT_BLOCK = 'insertBlock';
```
jhnstn: Tracking additional Redux actions starts here. First export the action name as a const and add to the appropriate Redux namespace below.

src/analytics/trackers.js
```diff
@@ -0,0 +1,17 @@
...
+ * Internal dependencies
+ */
+import * as EVENTS from './events';
+
+export default {
+	// Track block insertion
+	[ EVENTS.INSERT_BLOCK ]( eventData ) {
+		const { innerBlocks: inner_blocks, name: block_name } = eventData;
+		sendEventToHost( EVENTS.INSERT_BLOCK, { inner_blocks, block_name } );
+	},
```
jhnstn: After the event is added to `events.js` add an event handler here to set up the event properties that get sent up to the host apps.

src/analytics/from-redux.js
```diff
@@ -0,0 +1,41 @@
...
+		dispatch: ( namespace ) => {
+			const namespaceName =
+				typeof namespace === 'object' ? namespace.name : namespace;
+			const actions = { ...registry.dispatch( namespaceName ) };
+			const trackedEvents = REDUX_TRACKING[ namespaceName ];
+
+			if ( trackedEvents ) {
+				trackedEvents.forEach( ( actionName ) => {
+					const originalAction = actions[ actionName ];
+					const tracker = eventTrackers[ actionName ] || ( () => {} );
```
jhnstn: I'm initializing the tracker to a noop function in case event is tracked but the tracker is not set up. The `try` should cover this case so we might want the error to show up so we know  the event is not fully tracked.

src/analytics/trackers.js
```diff
@@ -0,0 +1,17 @@
+/**
+ * WordPress dependencies
+ */
+import { sendEventToHost } from '@wordpress/react-native-bridge';
+
+/**
+ * Internal dependencies
+ */
+import * as EVENTS from './events';
```
jhnstn: I was thinking we could just define all the events here and import into the redux middleware module. The only tricky part is the redux action namespaces. It feels better to sort that out in the `events.js` file.

src/analytics/from-redux.js
dcalhoun: As you said, because of the `try`, I would think a no-op is unnecessary. I would vote we remove the unnecessary no-op. 
```suggestion
					const tracker = eventTrackers[ actionName ];
```

src/analytics/index.js
```diff
@@ -0,0 +1,4 @@
+/**
+ * Internal dependencies
+ */
+export default from './from-redux';
```
dcalhoun: This is likely personal preference, so _please_ feel free to disagree or disregard this. 🙇🏻 

Is there specific value to this file? From my perspective, it feels a bit like indirection for the sake of module "ceremony." I feel it might be a bit easier to follow the logic if more of it was colocated. WDYT?

src/analytics/trackers.js
dcalhoun: > I was thinking we could just define all the events here and import into the redux middleware module. The only tricky part is the redux action namespaces. It feels better to sort that out in the `events.js` file.

This is likely personal preference, so _please_ feel free to disagree or disregard this. 🙇🏻 

Colocating the events and their respective trackers in a single file would be my preference. I would argue keeping most of the logic in one place may make the overarching design easier to follow, particularly given how small/simple its usage is currently (e.g. a single event). 

Currently, I do not perceive the value or long-term vision of separate `events.js`, `trackers.js`, and `from-redux.js` files. Would you elaborate a bit on what value you picture the split bringing please?

src/analytics/trackers.js
dcalhoun: In regards to the work managed by the tracker and https://github.com/wordpress-mobile/WordPress-Android/pull/15406#discussion_r720445968, I propose we handle all of the sanitization of events and properties within the tracker. This might reduce the likelihood that analytic events are erroneously ignored due to their lack of correct formatting. I have seen analytic events [silently fail due to improper property casing](https://github.com/wordpress-mobile/WordPress-iOS/pull/17130). 

Additionally, I think we should send the entirety of the event name to the host, e.g. `editor_block_inserted` rather than `insertBlock` or `block_inserted`. I think this will set us up to one day leverage [associated values](https://docs.swift.org/swift-book/LanguageGuide/Enumerations.html#ID148) to "automatically" convert the strings to enum cases. 

For example, we might rename Redux actions and convert property casing to match the patterns expected by the host app. WDYT? 

```js
import { snakeCase } from "lodash";

const REDUX_TRACKING = {
  "core/block-editor": {
    insertBlock: ({ innerBlocks, name: blockName }) => {
      tracksRecordEvent("editor_block_inserted", { blockName, innerBlocks });
    },
  },
};

function tracksRecordEvent(eventName, properties = {}) {
  const santizedEventName = snakeCase(eventName);
  const sanitizedProperties = Object.keys(properties).reduce((acc, key) => {
    acc[snakeCase(key)] = properties[key];
    return acc;
  }, {});
  sendEventToHost(santizedEventName, sanitizedProperties);
}
```

src/analytics/trackers.js
```diff
@@ -0,0 +1,17 @@
...
+
+export default {
+	// Track block insertion
+	[ EVENTS.INSERT_BLOCK ]( eventData ) {
+		const { innerBlocks: inner_blocks, name: block_name } = eventData;
```
dcalhoun: From my testing, sending the entirety of `innerBlocks` will likely not work, as it is an object with nested properties. I believe Tracks only receives strings, integers, and booleans. 

If we choose to include this property, we likely need to map over `innerBlocks` and send a simplified value like `name`. E.g. `inner_blocks: 'core/paragraph, core/image'`. Similar to the current `unsupported_blocks` [property included](https://github.com/WordPress/gutenberg/blob/83bbd27d50a7c7bd22ad1f8aaf4b1768efab09dd/packages/editor/src/components/provider/index.native.js#L256-L258) in the editor session [start event](https://github.com/wordpress-mobile/WordPress-iOS/blob/fafd3cb4f0f4121b5e405f78f9bced4442c24384/WordPress/Classes/ViewRelated/Post/PostEditorAnalyticsSession.swift#L40-L44). 

However, blocks within `innerBlocks` can contain their own `innerBlocks`, so how deep do we care to track? What would the format of the property value be if we attempted to track multiple levels? 

I might propose we drop `innerBlocks` altogether for a first iteration. I'm not too sure how useful a string of inner block names will be for analysis. WDYT?

dcalhoun (commented):
Great work! Thank you for putting this together. 🙌🏻 

I left a few comments, but most of them are personal preference and not blockers. The only blockers would be `innerBlocks` and that we need to make sure all the sibling branches ultimately align on things like the pattern of the string values sent to the host app. I would imagine we will wait to merge all of the PRs together after they are tested together. 

I also tested this work (with a minor event name change from `insertBlock` to `editor_block_inserted`) against my current https://github.com/wordpress-mobile/WordPress-iOS/pull/17243 and https://github.com/WordPress/gutenberg/pull/35272 work — the event arrived in Tracks with the expected `block_name` property. 🎉

src/analytics/from-redux.js
mchowning: Feels like this is a case where it will only crash if we don't code this properly (and then it will always crash), so I'm on board with dropping the no-op. Let's go that way for now, but I'm happy to reconsider that decision later.

src/analytics/trackers.js
mchowning: > From my testing, sending the entirety of innerBlocks will likely not work, as it is an object with nested properties. I believe Tracks only receives strings, integers, and booleans.

Good point, it does have to be something like a string.

> I might propose we drop innerBlocks altogether for a first iteration. I'm not too sure how useful a string of inner block names will be for analysis. WDYT?

👍  Let's drop it for the first iteration.

Thinking beyond the first iteration... One easy option that comes to mind would be to just include an eventprop identifying the number of nested blocks (counted recursively), and nothing else. It's not much information, but it's more than nothing. A trickier, but more interesting possibility would be to fire off a different "inner_block_inserted" event for each inner block.

dcalhoun: > One easy option that comes to mind would be to just include an eventprop identifying the number of nested blocks (counted recursively), and nothing else. It's not much information, but it's more than nothing. 

If our goal is merely a _simple_ indicator, another easy option would be to include a `has_inner_blocks: boolean` event property. The unsupported blocks example I provided earlier includes a similar `has_unsupported_blocks` as well as a comma-separated list of block names under `unsupported_blocks`.

src/analytics/trackers.js
jhnstn: I agree, I let too much YAGNI slip in with this structure 😄 
I do prefer to  keep the concerns separated so I opted for a new structure:
- All the tracking related to redux is now in `src/analytics/redux`. ( after reading your comment I realized the namespaced redux events would make no sense to any future potential event source so there was no need to have them in a generalized location )
- `src/analytics/redux/index` is only concerned with the middleware. It learns what should be tracked but delegates how it's tracked. i.e. no event property transformation or how the event gets sent to the host.
- `src/analytics/redux/tracked_events` defines what and how an action is tracked.

I find this new structure makes it clear that we (presently) only tracking redux events but that `src/analytics` , as a module, is open for future event sources. Any additional event tracking would be independent from the redux tracking since all the redux stuff is encapsulated in `src/analytics/redux`

I also opted to keep all the tracking logic separate from the middleware logic since updating an analytic event should not require knowing about the middleware flow. I should only need review `src/analytics/redux/tracked_events` to understand which events are tracked and how the analytic event payload is prepared. 
I also hope this separation would discourage folks from changing the middleware to handle a specific event case. Any specific event handling should happen in the `tracked_events` file.

src/analytics/index.js
jhnstn: I get that, as is it's a tad ceremonial. I view this more from the outside. That is, looking at `src/analytics` _from the inside_ this looks redundant. But from the outside, say from `src/index` , all the caller should know what to do is initialize the analytics flow. So my intent here was more about what the module looks like from the outside.

That said my new proposed structure should ( I hope 😅  ) reduce the appearance of "modular frosting"  , I still have a thin index file that does some indirection. However the new structure is domain based, right now the only domain the analytics system can handle is redux. I did update the index file to reify the domain concept. it's still doing the same thing but less terse.

src/analytics/trackers.js
jhnstn: > In regards to the work managed by the tracker and wordpress-mobile/WordPress-Android#15406 (comment), I propose we handle all of the sanitization of events and properties within the tracker. 

good point. 👍 

> Additionally, I think we should send the entirety of the event name to the host, e.g. editor_block_inserted rather than insertBlock or block_inserted. 

👍 

> For example, we might rename Redux actions and convert property casing to match the patterns expected by the host app. WDYT?

I played around with this but after reviewing what Calypso is doing to process events I felt sanitizing the property keys was not all that helpful. There is a healthy amount of property filtering and transformation going on in Calypso, also a fair amount of derived properties. `innerBlocks` as we now know is a good example of that.

src/analytics/trackers.js
jhnstn: I went back and reviewed how Calypso handles inner blocks. Calypso recursively sends an individual event for each inner block. While doing so it derives a `parent_block_name` for relational analysis. 

I added a `trackBlocksHandler` function that is mostly copied from https://github.com/Automattic/wp-calypso/blob/trunk/apps/wpcom-block-editor/src/wpcom/features/tracking.js#L165

I should also note while going through the Calypso code I uncovered some gotchas that I overlooked on the first iteration of this change set. In particular I was assuming the redux action will always have a block object. That _might_ be the case with the `insertBlock` action but Calypso handles the case when only a block id is in the action. I added that in to this PR. 

Going forward I'd highly recommend checking out Calypso when ever we add more trackers.

src/analytics/trackers.js
dcalhoun: The new approach makes sense and looks good to me. Thank you for taking the time to explain. Much appreciated. 🙇🏻

src/analytics/index.js
dcalhoun: > That is, looking at `src/analytics` _from the inside_ this looks redundant. But from the outside, say from `src/index` , all the caller should know what to do is initialize the analytics flow. So my intent here was more about what the module looks like from the outside.

Ah, I see. Yes, that makes sense. Thank you for explaining.

src/analytics/trackers.js
dcalhoun: > I played around with this but after reviewing what Calypso is doing to process events I felt sanitizing the property keys was not all that helpful. There is a healthy amount of property filtering and transformation going on in Calypso, also a fair amount of derived properties. `innerBlocks` as we now know is a good example of that.

I am not sure I follow what you mean in the comparison to Calypso. To clarify, I largely used "sanitize" in reference to ensuring the casing matched the host app expectations. E.g. allow us to use the linter-enforced camelcase in JavaScript, while sending the host app enforced snakecase — without managing the rename ourselves within each event property object definition. 

That said, I think it is fine to not build a "robust" sanitization feature within the tracker. 👍🏻

jhnstn: I totally miss-read your comment about sanitization, @dcalhoun. I was thinking your suggestion was that we would move toward an auto processing flow not just a key transformation. My bad. 

Although what I found from looking at the Calypso code is that event properties names are rarely 1:1 . `block.name => block_name` for example. So we'll end up having to rename anything anyways, might as well just rename to what everyone down stream expects. 

We could also get around the lint rule by using strings as keys:
```
( { name } ) => ( { 'block_name' : name } )
```
Assuming we use anonymous functions to transform properties.

Thanks again for all the feedback! It's been a helpful conversation for me.

src/analytics/redux/tracked_events.js
```diff
@@ -0,0 +1,84 @@
...
+export const trackedEvents = {
+	'core/block-editor': {
+		insertBlock( blocks ) {
+			trackBlocksHandler(
+				blocks,
+				'editor_block_inserted',
+				( { name } ) => ( { block_name: name } )
+			);
+		},
+		insertBlocks( blocks ) {
```
mchowning, line 76: I can easily trigger `insertBlock`, but I'm not sure how to trigger `insertBlocks` from mobile. Any suggestions?

dcalhoun: From researching the usage of the `insertBlocks` action creator, it would appear [splitting at the end of some rich text blocks](https://github.com/WordPress/gutenberg/blob/59a9764dd09e4299bf5443b04a83c4dfd3abce9f/packages/block-library/src/pullquote/edit.native.js#L116-L120) causes a `insertBlocks` Redux action to trigger. E.g. placing a the cursor in a PullQuote block's citation input and pressing <kbd>Return</kbd> will trigger the action. 

That said, neither WPAndroid or WPiOS implementations currently have sibling events defined for `insertBlocks`, so the resulting analytics event will not be sent. I plan to add the required sibling events.

dcalhoun: Following up to share that my plan is to rename the `editor_blocks_inserted` event to `editor_block_inserted` rather than add sibling events to WPAndroid and WPiOS. This approach mimics what [Calypso does](https://github.com/Automattic/wp-calypso/blob/1355d7d559db87ab7de935a917dd810ee0f1bda6/apps/wpcom-block-editor/src/wpcom/features/tracking.js#L677-L678), as both `insertBlock` and `insertBlocks` [leverage the same event name](https://github.com/Automattic/wp-calypso/blob/1355d7d559db87ab7de935a917dd810ee0f1bda6/apps/wpcom-block-editor/src/wpcom/features/tracking.js#L299). 

To me, this approach makes more sense, as we our goal is tracking the insertion of a block. Whether the `insertBlock` or `insertBlocks` action creator is leveraged is implementation details. In fact, the `insertBlock` action [calls `insertBlocks` itself](https://github.com/WordPress/gutenberg/blob/59a9764dd09e4299bf5443b04a83c4dfd3abce9f/packages/block-editor/src/store/actions.js#L523-L530).

dcalhoun (approved):
🚀 LGTM. I verified functionality via https://github.com/wordpress-mobile/WordPress-Android/pull/15406#pullrequestreview-790003627.

PR #4105, opened 2021-10-12
Fix overflowing content for native Help screen

### LINKED ISSUE #4100: Help view erroneously reveals parent screen when navigating to child sheet
**Describe the bug**
While in landscape orientation or on a larger screen device, when navigating to a sub-sheet within the Help view the parent sheet is visible rather than being appropriately obscured. 

**To Reproduce**

1. Open the block editor in landscape orientation. 
2. Tap the three-dot menu in the top-right corner. 
3. Tap "Help & Support." 
4. Tap one of the help article topics to navigate down to the information.
5. 💥 The top-level parent Help screen remains visible after navigating to a child screen. The top-level Help screen appears to the left of the bottom sheet modal. 

**Expected behavior**
The top-level parent Help screen is not visible after navigating to a child screen. 

**Screenshots**
parent-visible-behind-child

https://user-images.githubusercontent.com/438664/136848461-a0eac6e1-f025-42a1-b708-e56ef683b7e0.MP4

 - Device: iPhone SE
 - OS: iOS 15.0
 - App version: 18.4

**Additional context**
The "default transition presets" [applied to the navigation](https://github.com/WordPress/gutenberg/blob/b1fd6b2ce6221f1297da78b6c48aeee56130501b/packages/editor/src/components/editor-help/index.native.js#L192) may be the culprit of this issue. It is the first usage of these presets in the project. Based on the presets, this only affects iOS.

dcalhoun:
## Related PRs

* https://github.com/WordPress/gutenberg/pull/35552

## Description
Fixes #4100. When navigating to a Help section child screen, the parent screen would remain visible outside of the container. Hiding the overflow of the containing element fixes this.

The reason this occurred for the first time now is likely because it is the first time we have utilized the default navigation transitions for the platforms, which is provided by React Navigation. This means the screen animates from right to left on iOS, rather than a cross-fade.

To test: See https://github.com/WordPress/gutenberg/pull/35552. 

PR submission checklist:

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -3,6 +3,7 @@ Unreleased
 * [*] [Unsupported Block Editor] Fix text selection bug for Android [https://github.com/wordpress-mobile/gutenberg-mobile/pull/3937]
 * [*] Embed block: Fix inline preview cut-off when editing URL [https://github.com/wordpress-mobile/gutenberg-mobile/pull/4072]
 * [**] Embed block: Include Jetpack embed variants [https://github.com/wordpress-mobile/gutenberg-mobile/pull/4008]
+* [*] Fixed erroneous overflow within editor Help screens. [https://github.com/wordpress-mobile/gutenberg-mobile/pull/4105]
 
 1.63.0
 ------


twstokes:
LGTM. Tests passed for me @dcalhoun!

twstokes (approved):

PR #4114, opened 2021-10-13
Try running UI tests on any updates to develop

### LINKED ISSUE #4101: Run our full automation test suite on all new changes to the develop branch

We found in the 1.63.0 release that Dependabot caught an incoming Gutenberg breakage after the fact (i.e. after a previous PR had update the Gutenberg submodule ref). Dependabot was able to catch the breakage while the previous PR did not, because Dependabot [runs optional tests by default](https://github.com/wordpress-mobile/gutenberg-mobile/blob/7909858ebef88b1581c8c19afaa848c87843f919/.circleci/config.yml#L359-L365), but regular PRs do not (they are optional tests that must be manually triggered).

Dependabot [runs daily](https://github.com/wordpress-mobile/gutenberg-mobile/blob/91cd7cbc143c664b3d7f4b3a730ae3b6d017ac8a/.github/dependabot.yml#L11), so to reduce the amount of time between a breakage being introduced and it potentially being caught by Dependabot, this issue proposes to have all updates to the Gutenberg Mobile `develop` branch trigger the full suite of tests. I think this is similar to the main apps which run optional tests after PRs have merged to `develop` (e.g. [here](https://github.com/wordpress-mobile/WordPress-iOS/blob/34378daba930fba6d431b654161ae19758854c4f/.circleci/config.yml#L314) for WPiOS, and [here](https://github.com/wordpress-mobile/WordPress-Android/blob/4ad8ab604d44eb14a2571355098199de63f3ea34/.circleci/config.yml#L361) for WPAndroid).

**Describe alternatives you've considered**

Alternatives include:
- making Gutenberg Mobile optional tests not optional anymore (would slow down our workflow for all developers)
- leaving things as-is since this proposal only shortens the time between an incoming Gutenberg breakage being introduced and it being detected, it doesn't avoid detection happening after the fact
- Moving optional tests to the Gutenberg repo and possibly running them there on all updates to `trunk` (would slow down the workflow for all mobile/web developers)

guarani:
Fixes https://github.com/wordpress-mobile/gutenberg-mobile/issues/4101

This PR follows up on the proposal in https://github.com/wordpress-mobile/gutenberg-mobile/issues/4101 to automatically run our full automated test suite on all changes to the `develop` branch. This is still just a proposal, so the idea of this PR is to try this idea out before we decide on it. If we're happy with it, we can merge this PR to put the proposal into practice.

⚠️ Before merging: If these changes work and are approved, we need to remove the testing branch (this PR's branch) from the config and set `post-to-slack` to `true`.

## To test

1. Ensure a new `Test iOS on Device - Full (On Updated Develop)` is seen below under CI checks
2. Ensure a new `Test Android on Device - Full (On Updated Develop)` is seen below under CI checks

Note: The two new workflows will only run on the `develop` branch once this PR is merged. They only appear on this PR because we're temporarily adding this PR's branch to the list of branches the workflow runs on.

PR submission checklist:


guarani:
👋 I'm removing the review request since I noticed how Optional UI Tests is still on hold:

| Screen Shot 2021-10-22 at 13 51 57 |
| - |

guarani:
👋 Hi @mchowning, this is ready for review again.

There's some repetition in the CircleCI config. For example, `ios-device-checks` is called three different times and the only difference between them is the name given to the job and the branch the job is triggered on. While it might be worthwhile finding a more concise syntax, CircleCI config will be replaced with Buildkite config in the future, so any optimizations might no longer be necessary.

.circleci/config.yml
```diff
@@ -386,6 +387,22 @@ workflows:
...
+          name: Test iOS on Device - Full (On Develop)
+          post-to-slack: false
+          filters:
+            branches:
+              only: 
+                - develop
+                - try/running-ui-tests-on-updated-develop
+      - android-device-checks:
+          name: Test Android on Device - Full (On Develop)
+          post-to-slack: false
```
mchowning, line 400: Is `post-to-slack` false just for testing, or is the plan to leave it as false when we merge this?

guarani: The plan was to set it to `true` before merging, but I forgot to mention that in the PR description. I'll update that now, thanks for noticing!

guarani:
Thanks for the review, @mchowning! I've addressed the feedback and this should be ready to review again.

mchowning (approved):
This looks good to me. Thanks @guarani ! 

> There's some repetition in the CircleCI config. For example, ios-device-checks is called three different times and the only difference between them is the name given to the job and the branch the job is triggered on. While it might be worthwhile finding a more concise syntax, CircleCI config will be replaced with Buildkite config in the future, so any optimizations might no longer be necessary.

For merging, would it be cleaner to just remove the two new jobs you created and instead add the `develop` branch to the "release" jobs and give those jobs a more generic name?

We probably should go ahead and switch from `develop` to `trunk` as well since the main apps are making that switch, but that's not an issue that needs to be addressed as part of this PR.

guarani:
> For merging, would it be cleaner to just remove the two new jobs you created and instead add the `develop` branch to the "release" jobs and give those jobs a more generic name?

That's a good point. I'd considered it [in a similar PR](https://github.com/wordpress-mobile/gutenberg-mobile/pull/4151#issue-1033679222) (for running these tests on release PRs) but forgot to mention the discussion here. In short, my position was that it would be valuable to have separate names for each. While the Slack notification doesn't show these custom names, clicking on the notification opens CircleCI which displays the job name prominently, which I think is useful. Otherwise you'd need to look at the branch name in the CircleCI job if you wanted to know what type of test was run. I have no strong preference here though, but figured this was the safer option for now. Let me know if you think we should change this in a follow-up PR.

Thanks for the review!

guarani:
> ⚠️ Before merging: If these changes work and are approved, we need to remove the testing branch (this PR's branch) from the config and set `post-to-slack` to `true`.

I didn't do this before merging, so expect a follow-up PR 🙈 
Update: Clean-up will be done in https://github.com/wordpress-mobile/gutenberg-mobile/pull/4224, will request a review on it once its checks have passed.

guarani:
👋 @mchowning, the [clean-up PR](https://github.com/wordpress-mobile/gutenberg-mobile/pull/4224) and I added you as a reviewer if you don't mind.
Update: Clean up PR merged.

PR #2860, opened 2020-12-03
Fix "curl: (52) Empty reply from server" error

ceyhun:
`curl: (52) Empty reply from server` errors are happening randomly during `Upload .app to sauce labs` step on macOS e2e test jobs:
- https://app.circleci.com/pipelines/github/wordpress-mobile/gutenberg-mobile/9216/workflows/d17f0b53-0e5e-4b73-9ef5-0c8baf0c036f/jobs/48999
- https://app.circleci.com/pipelines/github/wordpress-mobile/gutenberg-mobile/9209/workflows/ad1e9a04-60af-40ea-877a-5aad951d942f/jobs/48964
- https://app.circleci.com/pipelines/github/wordpress-mobile/gutenberg-mobile/7867/workflows/350f1dfa-fc55-4614-b972-1f5576fab8df/jobs/41769
- https://app.circleci.com/pipelines/github/wordpress-mobile/gutenberg-mobile/9384/workflows/9b7591b6-890b-4313-a2cf-553bb0d8378c/jobs/49830

One of the solutions suggested by Sauce Labs is to upgrade `curl` to a newer version:
https://support.saucelabs.com/hc/en-us/articles/360052420274--Empty-reply-from-server-when-uploading-application-to-Sauce-Storage-using-curl

And that's what is done in this PR.

To test:

Check if `Test iOS on Device` CI jobs are successful.

PR submission checklist:

- [x] I have considered adding unit tests where possible.
- [x] I have considered if this change warrants user-facing release notes [more info](https://github.com/wordpress-mobile/gutenberg-mobile/blob/develop/docs/Release-notes.md) and have added them to [RELEASE-NOTES.txt](https://github.com/wordpress-mobile/gutenberg-mobile/blob/develop/RELEASE-NOTES.txt) if necessary.

PR #4123, opened 2021-10-14
Fix connection errors causing failed jobs on CircleCI

guarani:
This aims to fix errors we're seeing similar to `Error: Failed to download resource "nghttp2"` and `ssh: connect to host github.com port 22: Operation timed out`.

Running `brew update` was the suggested fix [here](https://github.community/t/curl-7-failed-to-connect-to-api-github-com-port-443-connection-refused/14853/3) and [here](https://stackoverflow.com/a/67700844/1305067) for similar-sounding issues.

As @dcalhoun noted, https://github.com/wordpress-mobile/gutenberg-mobile/pull/2860 updated `curl` but also added `HOMEBREW_NO_AUTO_UPDATE=1` which prevents brew from updating when `curl` is installed. To try to determine if this change will cause a performance issue, the job was ran twice and both times it finished in a "normal" duration (20 to 30 minutes).

### To test

Check if all CI tests pass (especially `Test iOS on Device - Canaries` → `Upload .app to sauce labs`).

PR submission checklist:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -307,7 +307,7 @@ jobs:
         name: Upload .app to sauce labs
         command: |
           source bin/sauce-pre-upload.sh
-          HOMEBREW_NO_AUTO_UPDATE=1 brew install curl
+          brew install curl
           /usr/local/opt/curl/bin/curl --version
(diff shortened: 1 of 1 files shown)


dcalhoun (approved):
LGTM. We'll see if any degradations come after this change, but I think it is worth a try for now.

guarani:
> LGTM. We'll see if any degradations come after this change, but I think it is worth a try for now.

Thanks @dcalhoun! I'll try to keep an eye out too.

PR #4143, opened 2021-10-21
Jetpack Tiled Gallery Block : `save` functionality submodule hash update

illusaen:
Updating submodule hash for Jetpack Tiled Gallery block's `save` functionality.
See details in: [#21481](https://github.com/Automattic/jetpack/pull/21481)

PR submission checklist:


guarani (commented):
I see some failing tests – it looks like our integration tests for the Tiled Gallery caught an issue (or maybe they need to be updated).

illusaen:
@guarani Leaving a note here as a reminder to myself (and everyone):

Issue with this right now is that the unit test is failing with a max recursion hit error; somewhere the state is being set over and over (and thus calling render over and over) when run from the unit test but not when run manually.

As per our discussion on Slack, do we want to merge this to our feature branch? We can open a new PR for the unit test bugfix.

guarani:
👋 @illusaen Thanks for sharing the info here about the failing test.

> As per our discussion on Slack, do we want to merge this to our feature branch? We can open a new PR for the unit test bugfix.

If we merge as-is, would this failing test cause future PRs targeting the feature branch to fail? If so, would it make sense to disable the test in this PR and then merge it? We could create a new issue to fix the bug (and re-enable the test).

illusaen:
> If we merge as-is, would this failing test cause future PRs targeting the feature branch to fail? If so, would it make sense to disable the test in this PR and then merge it? We could create a new issue to fix the bug (and re-enable the test).

@guarani Yes that sounds good. Did some elimination and it's definitely got something to do with the `columns` attribute; I think the `useEffect` calls are too spread out and need to be consolidated together so we can really encapsulate what we want.

For example:
- On inserting a new tiled gallery block, our columns needs to be set to the default, i.e. 2.
- On loading a serialized block, our columns needs to be set to:
    * if there are no images in serialized block, set to default
    * if there are images, set to `Math.max(images.length, default)`
- When adding a photo to the tiled gallery block, we need to update the max columns that are allowed (and current # of columns to be under that max if necessary).
- When removing a photo from the tiled gallery block, we likewise need to update the max columns that are allowed (and current # of columns to be under that max if necessary).

illusaen:
Skipping unit test.

guarani:
Thanks for sharing those insights, @illusaen!

guarani:
I've just updated our feature branches with the latest from Gutenberg `trunk` and will update this branch in the hopes that doing so will fix the build failure.

PR #4176, opened 2021-10-28
[RNMobile] Refactor Fixed Height Styling for Gallery Block

SiobhyB:
`gutenberg`: https://github.com/WordPress/gutenberg/pull/36013

# Description

A fixed height is currently applied to the image block when _any_ parent block makes use of [block context](https://developer.wordpress.org/block-editor/reference-guides/block-api/block-context/). This has an unwanted effect on parent blocks that use context but aren't designed to have a fixed height.

To avoid parent blocks from being styled to have a fixed height by default, this PR introduces a specific `fixedHeight` context. This will allow parent blocks to explicitly declare when the image block should have a fixed height, as per their requirements.

A core block that this will impact is the Gallery block, which will now provide the `fixedHeight` context to the Image block.

# Testing

To test: Please refer to the Gutenberg PR as the central, most up-to-date place for testing instruction.

<hr />

PR submission checklist:


guarani (approved):
Approved via https://github.com/WordPress/gutenberg/pull/36013#pullrequestreview-791912106.

Issue #4168, opened 2021-10-27
Tiled Gallery `save` workaround: Unit test seeing infinite recursion when rendering the block


illusaen:
See details in the [gbm PR](https://github.com/wordpress-mobile/gutenberg-mobile/pull/4143). Note: remove the added `.skip` for the unit test when fixed!

guarani:
👋 @SiobhyB just a heads-up, I moved this from "In Progress" to "Todo" on our project board, so feel free to un-assign yourself from this issue.

SiobhyB:
Perfect, thank you @guarani!

PR #4255, opened 2021-11-18
Tiled Gallery: Register block before running test

SiobhyB:
### Description

With this PR, the Tiled Gallery, added at https://github.com/wordpress-mobile/gutenberg-mobile/blob/add/tiled-gallery-block/src/jetpack/test/tiled-gallery.js, is updated.

The test was previously not running at all due to a conflict caused by some recent file changes/refactors.

To get the test to run, it's necessary to first call `registerJetpackBlocks`, as otherwise the Tiled Gallery block won't be registered while they're being run.

Note, the test is currently failing due to the error outlined at https://github.com/wordpress-mobile/gutenberg-mobile/issues/4168. This PR doesn't address that error. Instead, the purpose of this PR is simply to ensure the test runs.

### Testing

To test, it'll be necessary to first remove the `.skip` command added [here](https://github.com/wordpress-mobile/gutenberg-mobile/blob/add/tiled-gallery-block/src/jetpack/test/tiled-gallery.js#L37) then run `npm run test src/jetpack/test/tiled-gallery.js` while `cd`'d into the `gutenberg-mobile` folder from the terminal.

`.skip` isn't removed with this PR as the test still fails, which will be addressed in a separate PR.

<hr />

PR submission checklist:

### DIFF
--- a/src/jetpack/test/tiled-gallery.js
+++ b/src/jetpack/test/tiled-gallery.js
@@ -18,13 +18,23 @@ import { registerCoreBlocks } from '@wordpress/block-library';
 /**
  * Internal dependencies
  */
-import setupJetpackEditor from '../../jetpack-editor-setup';
+import {
+	registerJetpackBlocks,
+	setupJetpackEditor,
+} from '../../jetpack-editor-setup';
+
+const defaultProps = {
+	capabilities: {
+		tiledGalleryBlock: true,
+	},
+};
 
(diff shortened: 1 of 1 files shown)


fluiddot (approved):
LGTM 🎊 !

I verified that the test is running but fails with the error mentioned in https://github.com/wordpress-mobile/gutenberg-mobile/issues/4168.

PR #4256, opened 2021-11-18
Jetpack Tiled Gallery: Enable child block caption field to be hidden by parent blocks

### LINKED ISSUE #4190: Jetpack Tiled Gallery: Remove/Hide Image Settings and Caption
**Describe the bug**

Images are pulled into the Jetpack Tiled Gallery block via [the inner blocks API](https://developer.wordpress.org/block-editor/how-to-guides/block-tutorial/nested-blocks-inner-blocks/). As a result of this implementation, all of the image block's settings are available by default, including its settings tab and editable caption field.

For the purpose of the Tiled Gallery block, it shouldn't be possible for users to edit an individual image's settings or caption. The reason for this is that this would add styles/attributes to the block's HTML and the web wouldn't know how to interpret this output.

**To Reproduce**

1. Add the Tiled Gallery block to the demo app and upload a few images to it. 
2. Tap on any individual image and observe how it's possible to edit it via the settings tab or the editable caption field.

**Expected behavior**

It shouldn't be possible to access settings for individual images within the Tiled Gallery. Instead, mobile should match up with the web's behaviour, where individual settings and caption fields aren't accessible.

 - Device: Pixel 5
 - OS: Android 12
 - Version: Demo app

**Additional context**

A solution to this is likely to require [block contexts](https://developer.wordpress.org/block-editor/reference-guides/block-api/block-context/). However, contexts for a third-party block have never been added before. If we do use contexts, we should ensure that they wouldn't _only_ be useful for the Tiled Gallery block. They'd need to have a clear use for other third-parties who may wish to create a block using images as inner blocks. 

In https://github.com/wordpress-mobile/gutenberg-mobile/pull/4099, a proposal for "context constants" has been made. The idea behind these "context constants" is to better indicate when contexts aren't being serialised, and are primarily used for the purpose of styling or changing specific aspects of an inner block. If approved, it might be more straightforward to add contexts the type of contexts we're looking to add to address this issue. 

A short-term workaround, in case the idea of contexts isn't approved, could be using CSS to hide these settings instead. This will need to be explored further.

SiobhyB:
Fixes: https://github.com/wordpress-mobile/gutenberg-mobile/issues/4190

`Gutenberg:` https://github.com/WordPress/gutenberg/pull/36618

This PR provides a way for parent blocks to hide the caption field (if used) in any inner child block.

To test: Please refer to [the Gutenberg PR](https://github.com/WordPress/gutenberg/pull/36618) as the "central PR" with the most up-to-date testing instructions and also for more detailed information.

<hr />

PR submission checklist:

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -1,6 +1,7 @@
 Unreleased
 ---
 * [*] [Embed block] Included Link in Block Settings [https://github.com/wordpress-mobile/gutenberg-mobile/pull/4189]
+* [**] Fix tab titles translation of inserter menu [https://github.com/wordpress-mobile/gutenberg-mobile/pull/4248]
 
 1.66.0
 ---


guarani (approved):
Approved via https://github.com/WordPress/gutenberg/pull/36618#pullrequestreview-810481766.

PR #4263, opened 2021-11-19
Jetpack Tiled Gallery: Fix infinite loop error in unit test

### LINKED ISSUE #4168: Tiled Gallery `save` workaround: Unit test seeing infinite recursion when rendering the block

SiobhyB:
Fixes https://github.com/wordpress-mobile/gutenberg-mobile/issues/4168

`Jetpack:` https://github.com/Automattic/jetpack/pull/21829

As described in the above issue, [the `inserts the block` test](https://github.com/wordpress-mobile/gutenberg-mobile/blob/a6155a45fc3d5d9417d36efa48b9824d73476108/src/jetpack/test/tiled-gallery.js#L48) was failing to run with the following error:

```
Error: Uncaught [Error: Maximum update depth exceeded. This can happen when a component repeatedly calls setState inside componentWillUpdate or componentDidUpdate. React limits the number of nested updates to prevent infinite loops.]
```

This PR proposes a fix for that error. 

To test: Please refer to [the Jetpack PR](https://github.com/Automattic/jetpack/pull/21829) as the "central PR" with the most up-to-date testing instructions.

<hr />

PR submission checklist:


guarani (approved):
Tested and working great!

Reading the PR description, it says that the infinite loop was caused by `setMaxWidth` setting the `sizes` variable, which triggered another call to `setMaxWidth` because `sizes` was in `useEffect`'s dependency array.
With the new solution, doesn't `setMaxWidth` modify width causing `setMaxWidth` to be called again since width is now in the dependency array? I feel like I'm missing something here, do you know what it could be?
Either way, it's fixed and the change looks safe, so approving now.

SiobhyB:
@guarani, thanks for testing this! I did some more digging following our Slack chat with @illusaen. 

> With the new solution, doesn't setMaxWidth modify width causing setMaxWidth to be called again since width is now in the dependency array? I feel like I'm missing something here, do you know what it could be?

I'd jumped to an assumption and misunderstood the root cause of the problem here. 🤦‍♀️ 

The problem wasn't with the `setMaxWidth` call, which actually doesn't change the `sizes` object (that object seems to only change based on the available screen `width` and `height`). 

Instead, I believe the root cause is with the `setAttributes( { columnWidths: [ columnWidths ] } )` call in the hook [here](https://github.com/Automattic/jetpack/blob/bd358d38f0e9524a096cf9e549fe90aa2eccbb36/projects/plugins/jetpack/extensions/blocks/tiled-gallery/edit.native.js#L60). A change to the column width causes a change to the height (as it generally changes the number of rows) which is what changes the `sizes` object. 

As the change in `height` was ultimately the issue, the changes in this PR still resolves the issue by destructuring `width`.

SiobhyB:
I see two possible solutions, and wanted to get thoughts:

**1. Stick with the current solution.**

There were some concerns around the way `width` is destructured, which I've asked more about [here](https://github.com/Automattic/jetpack/pull/21829/files#r757610189). If that isn't a blocker, the current solution in this PR seems to work as expected.

**2. Split the `useEffect` hook into two.**

A second solution is to split the `useEffect` hook into two, one with the `setMaxWidth` call and the second with the `setAttributes` call. The only thing that's stopped me from going with this second option is the concerns expressed at https://github.com/wordpress-mobile/gutenberg-mobile/pull/4143#issuecomment-952688322 about how spread out the existing `useEffect` calls are, so wanted to get thoughts first.

PR #4332, opened 2021-12-02
Fix missing translations by refactoring the native editor initialization

### LINKED ISSUE #4329: Text translations retrieved at global scope don't return translated value
# Description
Most of the strings marked as untranslated are actually retrieving the translated text, and their translations are included in the i18n cache files, however, the translation is not displayed.

**Example:**
- String: `Align left`
- [Reference in Gutenberg](https://github.com/WordPress/gutenberg/blob/e96d47b35d8933ed15b1e242b9c85b855707737b/packages/block-editor/src/components/block-alignment-control/ui.js#L38)
Screenshot 2021-12-02 at 17 18 17

- [Reference in Spanish translation file](https://github.com/wordpress-mobile/gutenberg-mobile/blob/develop/src/i18n-cache/gutenberg/data/es.json)
Screenshot 2021-12-02 at 17 17 00

As you can see, in this case, the translation of the string is retrieved at global scope in the `BlockAlignmentUI` component. Therefore, only if the locale data was set prior to the importation of this file, the translation won't be available. 

Specifically, the file of the `BlockAlignmentUI` component ([reference](https://github.com/WordPress/gutenberg/blob/trunk/packages/block-editor/src/components/block-alignment-control/ui.js)) is one of the files imported during initialization, here is the import trace:

1. On the native editor initialization, [packages/react-native-editor/src/index.js#gutenbergSetup](https://github.com/WordPress/gutenberg/blob/e96d47b35d8933ed15b1e242b9c85b855707737b/packages/react-native-editor/src/index.js#L47) imports `@wordpress/edit-post` package.
1. [packages/edit-post/src/index.native.js](https://github.com/WordPress/gutenberg/blob/e96d47b35d8933ed15b1e242b9c85b855707737b/packages/edit-post/src/index.native.js#L12) imports `Editor` component.
1. [packages/edit-post/src/editor.native.js](https://github.com/WordPress/gutenberg/blob/e96d47b35d8933ed15b1e242b9c85b855707737b/packages/edit-post/src/editor.native.js#L12) imports `EditorProvider` component from `@wordpress/editor` package.
1. [packages/editor/src/components/provider/index.js](https://github.com/WordPress/gutenberg/blob/e96d47b35d8933ed15b1e242b9c85b855707737b/packages/editor/src/components/provider/index.js#L8-L11) imports `BlockEditorProvider` component from `@wordpress/block-editor` package.
1. [packages/block-editor/src/index.js](https://github.com/WordPress/gutenberg/blob/e96d47b35d8933ed15b1e242b9c85b855707737b/packages/block-editor/src/index.js#L4) imports `hooks`.
**NOTE:** Instead of checking the `BlockEditorProvider` component file, we focus on the `@wordpress/block-editor` package index as it's the entry point when importing any module under the package.
1. [packages/block-editor/src/hooks/index.native.js](https://github.com/WordPress/gutenberg/blob/e96d47b35d8933ed15b1e242b9c85b855707737b/packages/block-editor/src/hooks/index.native.js#L5) imports `align` hook (native version).
1. [packages/block-editor/src/hooks/align.native.js](https://github.com/WordPress/gutenberg/blob/trunk/packages/block-editor/src/hooks/align.native.js#L15) imports `align` hook (web version).
1. [packages/block-editor/src/hooks/align.js](https://github.com/WordPress/gutenberg/blob/e96d47b35d8933ed15b1e242b9c85b855707737b/packages/block-editor/src/hooks/align.js#L21) imports `BlockAlignmentControl` component.
1. [packages/block-editor/src/components/block-alignment-control/index.js](https://github.com/WordPress/gutenberg/blob/e96d47b35d8933ed15b1e242b9c85b855707737b/packages/block-editor/src/components/block-alignment-control/index.js#L4) imports `BlockAlignmentUI` component.
1. [packages/block-editor/src/components/block-alignment-control/ui.js](https://github.com/WordPress/gutenberg/blob/e96d47b35d8933ed15b1e242b9c85b855707737b/packages/block-editor/src/components/block-alignment-control/ui.js#L38) retrieves the `Align left` translation.

This means that the above code is executed before setting the locale data in [this line](https://github.com/WordPress/gutenberg/blob/e96d47b35d8933ed15b1e242b9c85b855707737b/packages/react-native-editor/src/index.js#L144) hence, the translation for `Align left` is not available at that moment.

## Potential solutions

In the following sections, I describe different solutions that we could apply. The first ones involve isolating the imports required to initialize the editor, so we can guarantee that we can set the locale data before any code related to the editor is executed.

### 1 - Refactor initialization code

I noticed that this is not a problem in the web version because the locale data is set before the editor is initialized. I haven't checked further but my impression is that this is done by the backend part of WordPress.

Screenshot 2021-12-02 at 17 59 21

In our case, the backend is the native code but we can't call the Javascript side until it's initialized. Besides, we're already handling initialization code in the files located in the `react-native-editor` package, as well as in the Gutenberg Mobile files.

The entry point in the Javascript side for React Native is when the app is registered in the `AppRegistry` library ([reference](https://reactnative.dev/docs/appregistry#registercomponent)), this part is handled by the `render` function located in the `@wordpress/element` package ([reference](https://github.com/WordPress/gutenberg/blob/e96d47b35d8933ed15b1e242b9c85b855707737b/packages/element/src/react-platform.native.js#L17-L42)) and is called by the `initializeEditor` function ([reference](https://github.com/WordPress/gutenberg/blob/e96d47b35d8933ed15b1e242b9c85b855707737b/packages/edit-post/src/index.native.js#L31)). 

A potential approach would be to move the code related to the `AppRegistry` to the initialization code in the `react-native-editor` package, which at the same time, matches the expected behavior of being this package the contact point with React Native. This way we can import and set the locale data when React Native initializes Javascript and before importing the files for initializing the editor.

Nevertheless, this approach has the following critical downsides:
- Refactor the initialization code is not trivial and would require deep testing.

### 2 - Download translations from main apps GlotPress projects

This approach implies that we no longer rely on the `translations` object passed from the native side via the initial props ([reference](https://github.com/WordPress/gutenberg/blob/e96d47b35d8933ed15b1e242b9c85b855707737b/packages/react-native-editor/src/index.js#L58)), since we'll download it at bundle generation time as we do with Gutenberg translations. This way we can import the translations and set the locale data before importing the files to initialize the code, assuring that translations are ready when they're retrieved at global scope.

Nevertheless, this approach has the following critical downsides:
- When introducing new strings, we don't have a way to get their translations after cutting a release. Usually, we request translations for new strings via the localization strings files of the main apps, so unless we generate a new JS bundle at the end of the release cycle (once the translations are ready), we won't have the translations until the next release version.

### 3 - Modify source code files

As a last alternative, we could modify the files that retrieve the translation text at global scope, and move the code to functions or component instantiation to prevent this issue.

Nevertheless, this approach has the following critical downsides:
- We would require updating several files and most likely introducing a native version just for this case.
- This option won't prevent that in the future we implement new code that follows the same behavior.

# To Reproduce
1. Change the device's language to non-English.
2. Open the app.
3. Select a Paragraph block.
4. Tap on the alignment button located in the toolbar.
5. Observe that the alignment options are not translated.

# Expected behavior
All strings should be translated.

# Screenshots

 - Device: iPhone 12 Pro Max (simulator)
 - OS: iOS 15.0
 - Version N/A

fluiddot:
Gutenberg PR: https://github.com/WordPress/gutenberg/pull/37073

Fixes https://github.com/wordpress-mobile/gutenberg-mobile/issues/4329

To test:
Follow testing instructions from Gutenberg PR https://github.com/WordPress/gutenberg/pull/37073.

PR submission checklist:


src/setup.js
```diff
@@ -0,0 +1,72 @@
...
+		const capabilities = props.capabilities ?? {};
+		registerJetpackBlocks( props );
+		setupBlockExperiments( capabilities );
+	} );
+};
+
+export default () => {
+	initAnalytics();
+	setupHooks();
+};
```
fluiddot, line 50: This setup code covers the same functionality as we had previously ([reference](https://github.com/wordpress-mobile/gutenberg-mobile/blob/develop/src/index.js)).

src/index.js
```diff
@@ -1,65 +1,9 @@
...
-	registerJetpackEmbedVariations( props );
-} );
+import { registerGutenberg } from '@wordpress/react-native-editor';
 
-addAction( 'native.render', 'gutenberg-mobile', ( props ) => {
-	const capabilities = props.capabilities ?? {};
-	registerJetpackBlocks( props );
-	setupBlockExperiments( capabilities );
+registerGutenberg( () => {
+	const { default: gutenbergMobileSetup } = require( './setup' );
```
fluiddot: The setup code is lazy-loaded to guarantee that we can execute code, like setting the locale data, before any other code is imported or initialized. 

**NOTE:** Although this is not required yet, it will be needed when adding the translations of plugins like Jetpack or Layout Grid.

src/index.js
```diff
@@ -1,65 +1,13 @@
...
-	const capabilities = props.capabilities ?? {};
-	registerJetpackBlocks( props );
-	setupBlockExperiments( capabilities );
+import { registerGutenberg } from '@wordpress/react-native-editor';
+
+registerGutenberg( {
+	beforeInitCallback: () => {
+		// We have to lazy import the setup code to prevent executing any code located
+		// at global scope before the editor is initialized, like translations retrieval.
+		const { default: gutenbergMobileSetup } = require( './setup' );
```
dcalhoun: Leaving a link to the following comment, as I wondered about this line and the comment answered my question. I almost overlooked it as GitHub does not show "Outdated" comments when viewing the code changes themselves, only the _Coversation_ tab displayed "Outdated" comments. 

https://github.com/wordpress-mobile/gutenberg-mobile/pull/4332#discussion_r762042313

Is it accurate to state that all of the changes in this particular _Gutenberg Mobile_ PR (i.e. the lazy loading) are in service of future translation logic and not utilized today?

dcalhoun (commented):
Thanks for putting this work together. This is looking good. 

I left one comment for consideration as well as a further review in https://github.com/WordPress/gutenberg/pull/37073#pullrequestreview-836754169.

src/index.js
fluiddot: The GB-mobile setup code, specifically the Jetpack initialization, imports modules that might lead to code that retrieves translations at global scope, so these changes are related to fixing the issue https://github.com/wordpress-mobile/gutenberg-mobile/issues/4329. Although they will also help with future translation logic as it provides a place to execute code before initializing the editor.

src/block-experiments-setup.js
```diff
@@ -3,23 +3,9 @@
...
-
-export function setupBlockExperimentsLocale( locale, extraTranslations ) {
-	// Setup locale for Layout Grid
-	setupLocale( {
-		domain: LAYOUT_GRID_LOCALE_DOMAIN,
-		locale,
-		extraTranslations,
-		getTranslation: getLayoutGridTranslation,
-	} );
-}
```
fluiddot, line 25: The translations for the layout grid block are now included in the index file when registering Gutenberg ([reference](https://github.com/wordpress-mobile/gutenberg-mobile/blob/504270493676fe9c2a219519bb70a8d76106dee1/src/index.js#L12-L21)).

src/jetpack-editor-setup.js
```diff
@@ -66,15 +62,6 @@ export function setupJetpackEditor( jetpackState ) {
...
 }
 
-export function setupJetpackLocale( locale, extraTranslations ) {
-	setupLocale( {
-		domain: JETPACK_LOCALE_DOMAIN,
-		locale,
-		extraTranslations,
-		getTranslation: getJetpackTranslation,
-	} );
-}
```
fluiddot, line 76: The translations for Jetpack are now included in the index file when registering Gutenberg ([reference](https://github.com/wordpress-mobile/gutenberg-mobile/blob/504270493676fe9c2a219519bb70a8d76106dee1/src/index.js#L12-L21)).

src/setup-locale.js
```diff
@@ -1,28 +0,0 @@
...
-	console.log(
-		domain ? `${ domain } - locale` : 'locale',
-		locale,
-		allTranslations
-	);
-	// Only change the locale if it's supported by gutenberg
-	if ( translations || extraTranslations ) {
-		setLocaleData( allTranslations, domain );
-	}
-};
```
fluiddot, line 28: This file is no longer required as locale setup for plugins is now done via the plugin translations param when registering Gutenberg ([reference](https://github.com/wordpress-mobile/gutenberg-mobile/blob/504270493676fe9c2a219519bb70a8d76106dee1/src/index.js#L30)).

fluiddot:
Heads up that I've extracted the setup code of Jetpack and block experiments plugins to their own files in https://github.com/wordpress-mobile/gutenberg-mobile/pull/4332/commits/87ed8dfd704e32649c4df6bc284f447d5ecfe81c, the purpose of this change is to isolate the code of each plugin. In the future, this code might be provided by the plugin repository.

Additionally, I decided to add some tests to cover the changes in https://github.com/wordpress-mobile/gutenberg-mobile/pull/4332/commits/c48f90e9832e3353c472a14470fb968c2605367a.

.eslintrc.js
```diff
@@ -26,5 +26,25 @@ module.exports = {
...
+	},
+	overrides: [
+		{
+			// Workaround for addressing errors when importing react-native components.
+			// Related issue: https://git.io/JSKeJ
+			files: [ 'src/**/*.js' ],
+			settings: {
+				'import/ignore': [ 'react-native' ],
+			},
+		},
```
fluiddot, line 50: This setting override was taken from Gutenberg's ESLint configuration ([reference](https://github.com/WordPress/gutenberg/blob/43d950788629daaf588c80a2a56173124f112c60/.eslintrc.js#L186-L191)).

jest.config.js
```diff
@@ -54,7 +47,10 @@ module.exports = {
 		'\\.(scss)$': '<rootDir>/' + configPath + '/__mocks__/styleMock.js',
 		'\\.(jpg|jpeg|png|gif|eot|otf|webp|ttf|woff|woff2|mp4|webm|wav|mp3|m4a|aac|oga)$':
 			'<rootDir>/' + configPath + '/__mocks__/fileMock.js',
-		...transpiledPackages,
+		[ `@wordpress\\/(${ transpiledPackageNames.join(
+			'|'
+		) })$` ]: '<rootDir>/gutenberg/packages/$1/src',
+		'test/helpers$': '<rootDir>/' + configPath + '/helpers.js',
```
fluiddot, line 53: The logic for mapping the WordPress modules has been updated based on the Jest configuration we have in Gutenberg for unit tests:
- https://github.com/WordPress/gutenberg/blob/43d950788629daaf588c80a2a56173124f112c60/test/native/jest.config.js#L18-L20
- https://github.com/WordPress/gutenberg/blob/43d950788629daaf588c80a2a56173124f112c60/test/native/jest.config.js#L55-L58

.eslintrc.js
```diff
@@ -26,5 +26,25 @@ module.exports = {
...
+		{
+			// Ignore test helpers import solving as the module is only available when using Jest.
+			files: [ 'src/**/test/*.js' ],
+			rules: {
+				'import/no-unresolved': [
+					'error',
+					{ ignore: [ 'test/helpers' ] },
+				],
+			},
+		},
```
fluiddot: This rule has to be overridden for unit tests due to the fact that module "test/helpers" is mapped from Jest configuration ([reference](https://github.com/wordpress-mobile/gutenberg-mobile/blob/82227ef9da1e238b566e04a370e923f545cbc347/jest.config.js#L53)), so the ESLint resolver won't be able to solve it.

src/block-experiments-setup.js
```diff
@@ -1,8 +1,30 @@
...
+			if ( capabilities.layoutGridBlock ) {
+				registerBlock();
+			}
+		}
+	);
+};
+
+export default () => {
+	setupHooks();
+};
```
dcalhoun, line 30: Do we envision additional expansion of this default export or would it make sense to avoid defining and calling `setupHooks`? I wonder if the indirection is worthwhile at this point. 

Additionally, do you think it is worthwhile to avoid [anonymous exports](https://twitter.com/addyosmani/status/1411233253948747777)? 

```js
// Potentially used named `setUpBlockExperiments` to avoid anonymous export
export default function setUpBlockExperiments() {
  // ...inline logic instead of call to `setupHooks`...
}
```

dcalhoun: Further along in my review, I noticed this particular code structure is [already in place](https://github.com/wordpress-mobile/gutenberg-mobile/blob/82227ef9da1e238b566e04a370e923f545cbc347/src/setup.js#L49). Feel free to ignore my comment if the preference is to follow established structure.

.eslintrc.js
dcalhoun: Dropping a note that this relates to https://github.com/WordPress/gutenberg/pull/37073#discussion_r777730123. Why does this repository take a different approach of leveraging `overrides` in comparison to the inline ESLint comment found in Gutenberg?

dcalhoun (approved):
This looks great! Thank you for making the time to invest in testing. 👏🏻 I left a few comments to help me better understand, but I do not consider them blockers. Well done! 🚀

.eslintrc.js
fluiddot: Right, in this case, I wanted only to ignore the path `test/helpers` but I agree that we should apply the same overriding as we do in Gutenberg 👍 .

fluiddot: I've just applied the mentioned change in https://github.com/wordpress-mobile/gutenberg-mobile/pull/4332/commits/59deaf15cb741652db1bdacb39b80636a28bc027.

src/block-experiments-setup.js
fluiddot: I agree that we could simplify this file by unifying the logic into the default function or an exported named function, my idea was to apply the same structure for the plugin setup files as GB-mobile, especially in order to encourage structuring a bit more strictly how plugins arrange their setup code. Probably it's not worthwhile today but I think it would be great to start exploring and inducing structure improvements for future plugin additions, wdyt?

.eslintrc.js
```diff
@@ -26,5 +26,27 @@ module.exports = {
...
+			// Disable rules for tests as we do in Gutenberg.
+			// Reference: https://git.io/JSMAQ
+			files: [ 'src/**/test/**/*.js' ],
+			rules: {
+				'import/default': 'off',
+				'import/no-extraneous-dependencies': 'off',
+				'import/no-unresolved': 'off',
+				'import/named': 'off',
+				'@wordpress/data-no-store-string-literals': 'off',
+			},
```
fluiddot, line 41: This rules override was taken from Gutenberg's ESLint configuration ([reference](https://github.com/WordPress/gutenberg/blob/43d950788629daaf588c80a2a56173124f112c60/.eslintrc.js#L178-L184)).

src/block-experiments-setup.js
dcalhoun: @fluiddot I agree it makes most sense to postpone any code restructuring. Thanks!

dcalhoun (approved):
The last round of changes look good to me. 🚀

ttahmouch:
Hey, @fluiddot . We will cut the `1.69.0` release on `Thursday, January 6, 2022`. We plan to circle back and bump this PR to the next milestone then, but please let me know if you'd rather us work to include this PR in `1.69.0`.

fluiddot:
> Hey, @fluiddot . We will cut the `1.69.0` release on `Thursday, January 6, 2022`. We plan to circle back and bump this PR to the next milestone then, but please let me know if you'd rather us work to include this PR in `1.69.0`.

I'm about to merge this PR, so it will be included in `1.69.0`, thanks for the ping 🙇 .

ttahmouch:
> > Hey, @fluiddot . We will cut the `1.69.0` release on `Thursday, January 6, 2022`. We plan to circle back and bump this PR to the next milestone then, but please let me know if you'd rather us work to include this PR in `1.69.0`.
> 
> I'm about to merge this PR, so it will be included in `1.69.0`, thanks for the ping 🙇 .

Thanks for letting me know, @fluiddot . I appreciate it. :)

Would you happen to know if it's basically just going to be us running the release toolkit again to ensure your merge makes it into the releases?

fluiddot:
> Thanks for letting me know, @fluiddot . I appreciate it. :)
> 
> Would you happen to know if it's basically just going to be us running the release toolkit again to ensure your merge makes it into the releases?

Yep, from my side it's only a matter of merging the PR, the release script would do the rest once you run it next Thursday 👍 .

PR #4359, opened 2021-12-09
Add Jetpack and Layout Grid translations

fluiddot:
Fixes https://github.com/wordpress-mobile/gutenberg-mobile/issues/2615

## Description

This PR adds the translation files for localizing Jetpack blocks and the Layout Grid block (block experiments).

**I'd like to note that, although this PR adds the missing translations, it's only the first part of the process of the incorporation process, as these changes don't bring the logic for updating them.**

### Translation files
For this purpose, a new folder located at `src/i18n-translations` has been created and contains the translations for each project. These files have been generated as part of a different PR (https://github.com/wordpress-mobile/gutenberg-mobile/pull/4366), where optimized translation files will be produced including only the strings used by the native version of the editor.

**NOTE:** In the future, this folder won't be necessary as translations will be generated on the fly via a new command and when producing the JS bundle.

### POT files
For the strings that are not part of the web version and therefore not included in the plugin's GlotPress project, we have to include them in the localization strings files (`bundle/ios/GutenbergNativeTranslations.swift` / `bundle/android/strings.xml`), exactly the same as we do for the Gutenberg strings.

This implied adding new NPM commands to generate the required POT files and updating the scripts we use for generating the localization strings files (`po2swift.js` / `po2android.js`), so they support multiple POT files as now we'll have one file each Gutenberg, Jetpack, and Layout Grid projects.

**NOTE:** I'm working on a new command to unify the POT files generation into a single one, so eventually, we won't need to add extra commands when adding translations of new plugins. 

### Localization strings files

The strings added to `bundle/ios/GutenbergNativeTranslations.swift` and `bundle/android/strings.xml` files are strings coming from Jetpack and Layout Grid plugins that require translation as they're only referenced in `.native.js` files.

### Verify that localization strings files are not modified
- Run command: `npm run genstrings`
- Observe that neither `bundle/ios/GutenbergNativeTranslations.swift` and `bundle/android/strings.xml` are modified.

### Verify that Jetpack translations are present
1. Change the device's language to non-English.
1. Open the app.
1. Insert a Jetpack block (i.e.: Contact Info or Story blocks)
1. Observe that strings related to the block are translated to the selected language.
1. If any string remains untranslated, verify that is included in the localization strings files (`bundle/ios/GutenbergNativeTranslations.swift` / `bundle/android/strings.xml`). 

### Verify that Layout Grid translations are present
1. Change the device's language to non-English.
1. Open the app.
1. Insert the Layout Grid block
[...]

Issue #2162, opened 2020-04-16
Improve i18n process to support jetpack


Tug:
Currently we import many translations from gutenberg that we don't need.

Moreover, after adding jetpack as a submodule we will need translations for it as well and we should avoid distributing jetpack translations files as they are pretty big (>400KB per locale).

We could be a bit smarter:
1. Download all translation files from gutenberg and jetpack and merge them into `i18n-cache/temp/{locale}.json`
2. Extract all strings from `bundle/{platform}/App.js`
3. For the ones that have a translation entry in `i18n-cache/temp/{locale}.json` -> add them to `i18n-cache/data/{locale}.json`
4. For the others they will be generated as a `gutenberg-mobile.pot` file (we can update our pot generation to take `bundle/{platform}/App.js` as input and substract strings from `gutenberg.pot`)
5. We use the `bin/po2android.js` and `bin/po2swift.js` scripts to convert the gutenberg-mobile.pot file to bundle/android/strings.xml and `bundle/ios/GutenbergNativeTranslations.swift` (same as today)
6. We receive translations for those strings from the main apps at runtime (same as today) (edited)

fluiddot:
Following with the idea of avoiding distributing the entire jetpack translation files, I've benchmarked the size of the JS bundle in relation to the translations, here are the results:

### Only include Gutenberg translations
- Bundle including all translations: 12.8 MB _(current size)_
- Bundle without translations: 5MB
- **Translations are increasing by 7.8 MB the bundle size.**

I was surprised that the translations mean 61% of the bundle size 😨 .

### Include Jetpack and Layout Grid translations

**All translations:**
- Bundle including all translations: 28.3 MB
- Bundle without translations: 5 MB

**Translations are increasing by 23.3 MB the bundle size.**

As you can see, if we don't optimize the translation files, we would almost be doubling the current bundle size and translations would mean 82% of the total size.

**Optimized translations (filtering out unused strings):**
- Bundle including optimized translations: 7MB
- Bundle without translations: 5 MB

**Translations are increasing by 2 MB the bundle size.**

On the other side, by optimizing the translations files we would save more than 89% of the total size that translations introduce to the bundle 🏋️ .

<hr>

With these results looks clear that optimizing the translations is mandatory unless we want to significantly impact the bundle size and therefore the final app size.

fluiddot:
This issue has been addressed in https://github.com/wordpress-mobile/gutenberg-mobile/pull/4366.

PR #4366, opened 2021-12-10
Add i18n update script

fluiddot:
Follows up https://github.com/wordpress-mobile/gutenberg-mobile/pull/4359 and related to https://github.com/wordpress-mobile/gutenberg-mobile/issues/2162.

**Installable build can be obtained from the following PRs:**
- [Android](https://github.com/wordpress-mobile/WordPress-Android/pull/15778)
- [iOS](https://github.com/wordpress-mobile/WordPress-iOS/pull/17705)

## Description

This PR incorporates a new script to handle the process of updating the i18n localizations including improvements, the process consists of the following steps:
1. Extract used strings of plugins (including Gutenberg) from source-map files.
2. Download translations and optimize them for each plugin.
3. Override the i18n cache of `react-native-editor` package, as we're currently doing ([reference](https://github.com/wordpress-mobile/gutenberg-mobile/blob/29fb512d13ebb96381029537ae9f2c595c7b5ed8/bin/i18n-download.sh#L9-L11)).
4. Generate localization strings files to incorporate translations only referenced by native-only JS files.

More information about the scripts involved in the process can be found in the [Gutenberg PR](https://github.com/WordPress/gutenberg/pull/37482).

### Verify i18n update command
1. Execute `npm run i18n:update`.
2. Observe that the content of folder `src/i18n-cache` contains the optimized translations of Gutenberg, Jetpack and Layout Grid.
3. Observe that localization strings files remain unmodified.

**NOTE:** I spotted the following warnings that might require further investigation, however, I think it shouldn't block the PR as we're already missing them and can be addressed a new PR:

**Gutenberg:**
```
WARNING: The following strings are missing from translations:
[
  'Pixels (px)',
  'Percentage (%)',
  'Percent (%)',
  'Relative to parent font size (em)',
  'Relative to root font size (rem)',
  'Viewport width (vw)',
  'Viewport height (vh)',
  'Viewport smallest dimension (vmin)',
  'Viewport largest dimension (vmax)',
  'Width of the zero (0) character (ch)',
  'x-height of the font (ex)',
  'Centimeters (cm)',
  'Millimeters (mm)',
  'Inches (in)',
  'Picas (pc)',
  'Points (pt)',
  'Linear',
  'Radial',
  'Angle',
  'Newest to oldest',
  'Oldest to newest',
  'A → Z',
  'Z → A',
  'Order by',
  'Relative to parent font size (em)\x04ems',
  'Relative to root font size (rem)\x04rems'
]

WARNING: The following native-only strings are already being included in translations:
[ 'Featured' ]
```

**Layout Grid:**
```
WARNING: The following strings are missing from translations:
[ 'One column', 'Two columns', 'Three columns', 'Four columns' ]
```

### ~~Verify that bundle command executes i18n update command~~
~~1. Execute `npm run bundle`
2. Observe that the `i18n:update` command is executed.~~

~~**NOTE:** It's expected to see modifications on files under `bundle/ios/assets/gutenberg/packages/react-native-editor/i18n-cache/data`. Once the PR is ready to merge, we could update the bundle and push these changes.~~

The i18n update command has been unlinked from the bundle generation ([commit reference 1](https://github.com/wordpress-mobile/gutenberg-mobile/pull/4366/commits/ac35a6ff93e09862e34c55e743e956fcd9b596c8)) because it was making some CI jobs fail ([reference](https://buildkite.com/wordpress-mobile/gutenberg-mobile/builds/1719#34a373d2-b44b-4186-bcfe-219eba666b6c/170-785), [commit reference 2](https://github.com/wordpress-mobile/gutenberg-mobile/pull/4366/commits/0c54bd690f29e3a65218eb66f5dbbc3ab1b24303)) due to the use of PHP. It shouldn't block the PR as this can be addressed later on.

**NOTE:** This issue will be addressed in a [different PR](https://github.com/wordpress-mobile/gutenberg-mobile/pull/4411) once the Docker image is updated in https://github.com/wordpress-mobile/docker-gb-mobile-image/pull/3. In the meantime, we'll keep the same behavior as we currently have in `develop` when generating the JS bundle.

PR submission checklist:


src/i18n-cache/index.js
```diff
@@ -1,172 +0,0 @@
-/* eslint-disable no-console */
```
fluiddot: This logic is almost identical to the same file we have in Gutenberg ([reference](https://github.com/WordPress/gutenberg/blob/trunk/packages/react-native-editor/i18n-cache/index.js)), so it has been deleted and a new file to unify the logic was created [here](https://github.com/WordPress/gutenberg/blob/e47a69fa45793847a7c91a06b9f84ad996db5bef/packages/react-native-editor/bin/i18n-translations-download.js).

bin/po2android.js
```diff
@@ -63,79 +76,101 @@ const getUniqueName = ( function() {
...
-		const comment = translation.comments.extracted || '';
-		let localizedEntry = '';
-		if ( comment ) {
-			localizedEntry += `${ indent }<!-- ${ comment.replace( '--', '—' ) } -->\n`;
-		}
-		if ( translation.msgid_plural ) {
-			localizedEntry += `${ indent }<string-array name="${ uniqueName }" tools:ignore="UnusedResources">
+function po2Android( strings ) {
+	const androidResourcesMap = Object.values( strings ).reduce(
+		( result, { string, stringPlural, comments } ) => {
```
fluiddot: The main change here is that the strings are now extracted from the used strings object (generated by [this script](https://github.com/WordPress/gutenberg/blob/e47a69fa45793847a7c91a06b9f84ad996db5bef/packages/react-native-editor/bin/extract-used-strings.js)) instead of the POT files.

bin/po2swift.js
```diff
@@ -1,47 +1,55 @@
...
+	path = require( 'path' );
 
-function po2Swift( poInput ) {
-	const po = gettextParser.po.parse( poInput );
-	const translations = po.translations[ '' ];
-	const swiftStringsMap = Object.values( translations ).reduce( ( result, translation ) => {
-		if ( ! translation.msgid ) {
+function po2Swift( strings ) {
+	const swiftStringsMap = strings.reduce(
+		( result, { string, comments } ) => {
```
fluiddot: The main change here is that the strings are now extracted from the used strings object (generated by [this script](https://github.com/WordPress/gutenberg/blob/e47a69fa45793847a7c91a06b9f84ad996db5bef/packages/react-native-editor/bin/extract-used-strings.js)) instead of the POT files.

bin/po2swift.js
```diff
@@ -1,47 +1,55 @@
...
+	const destination = process.argv[ 2 ];
+	const usedStringsFile = process.argv[ 3 ];
+
+	const usedStrings = getUsedStrings( usedStringsFile );
+	const flattenUsedStrings = Object.entries( usedStrings ).reduce(
+		( result, [ domain, strings ] ) => {
+			return [ ...result, ...Object.values( strings ) ];
+		},
+		[]
+	);
```
fluiddot: Used strings are grouped by domain, since we only provide one localization strings files, we unify them here and their translations will be requested as part of the main apps' strings.

bin/po2swift.js
```diff
@@ -1,47 +1,55 @@
...
+	let onlyNativeStrings = flattenUsedStrings.filter( ( { platforms } ) => {
+		return ! platforms.includes( 'web' );
+	}, [] );
+	onlyNativeStrings = [ ...new Set( onlyNativeStrings ) ];
```
fluiddot: We only add the localization strings files the strings referenced in `.native.js` and `.ios.js` files.

bin/po2swift.js
```diff
@@ -1,47 +1,55 @@
 #!/usr/bin/env node
```
fluiddot: Most of the changes in this file are produced by the fact that it wasn't properly formatted.

bin/po2android.js
```diff
@@ -2,7 +2,8 @@
```
fluiddot: Most of the changes in this file are produced by the fact that it wasn't properly formatted.

bin/po2android.js
```diff
@@ -63,79 +76,101 @@ const getUniqueName = ( function() {
...
+	const destination = process.argv[ 2 ];
+	const usedStringsFile = process.argv[ 3 ];
+
+	const usedStrings = getUsedStrings( usedStringsFile );
+	const flattenUsedStrings = Object.entries( usedStrings ).reduce(
+		( result, [ domain, strings ] ) => {
+			return [ ...result, ...Object.values( strings ) ];
+		},
+		[]
+	);
```
fluiddot: Used strings are grouped by domain, since we only provide one localization strings files, we unify them here and their translations will be requested as part of the main apps' strings.

bin/po2android.js
```diff
@@ -63,79 +76,101 @@ const getUniqueName = ( function() {
...
+	let onlyNativeStrings = flattenUsedStrings.filter( ( { platforms } ) => {
+		return ! platforms.includes( 'web' );
+	}, [] );
+	onlyNativeStrings = [ ...new Set( onlyNativeStrings ) ];
```
fluiddot: We only add the localization strings files the strings referenced in `.native.js` and `.android.js` files.

bin/i18n-download.sh
```diff
@@ -1,11 +0,0 @@
...
-
-# Exit if any command fails.
-set -e
-
-if [ -n "${REFRESH_I18N}" ] ; then
-  node src/i18n-cache/index.js $1 || pFail
-fi
-# Overwrite the i18n cache of react-native-editor package
-cp -r src/i18n-cache/gutenberg/data gutenberg/packages/react-native-editor/i18n-cache
-cp -r src/i18n-cache/gutenberg/index.native.js gutenberg/packages/react-native-editor/i18n-cache
```
fluiddot: This part is now done in the `i18n-update` script.

bin/i18n-update.sh
```diff
@@ -0,0 +1,120 @@
...
+  exit 1
+}
+
+function fetch_translations() {
+  local PLUGIN_NAME=$1
+  local OUTPUT_PATH=$2
+  local USED_STRINGS_FILE=$3
+
+  echo -e "\n\033[1mDownload I18n translations for \"$PLUGIN_NAME\" plugin\033[0m"
+  node gutenberg/packages/react-native-editor/bin/i18n-translations-download $PLUGIN_NAME $OUTPUT_PATH $USED_STRINGS_FILE
```
fluiddot: We'll use the new script created in [this PR](https://github.com/WordPress/gutenberg/pull/37482) for downloading translations.

bin/i18n-update.sh
```diff
@@ -0,0 +1,120 @@
...
+
+  if [[ "$PLUGIN_NAME" == "gutenberg" ]]; then
+    echo "Update \"react-native-editor\" package i18n cache"
+    cp -r $OUTPUT_PATH/gutenberg/data gutenberg/packages/react-native-editor/i18n-cache
+    cp $OUTPUT_PATH/gutenberg/index.js gutenberg/packages/react-native-editor/i18n-cache/index.native.js
+  fi
```
fluiddot: The translations of Gutenberg are imported from a file in Gutenberg ([reference](https://github.com/WordPress/gutenberg/blob/78c7f0c7a9c3d69979d0633375325104fb03b86d/packages/react-native-editor/src/index.js#L10)), so we overwrite that content with was produced by this script.

bin/i18n-update.sh
```diff
@@ -0,0 +1,120 @@
...
+}
+
+# Set target path
+if [[ -n $LOCAL_PATH ]]; then
+  TARGET_PATH=$LOCAL_PATH
+else
+  TARGET_PATH=$(mktemp -d)
```
fluiddot, line 76: By default, the intermediate files produced by the script will be held in a temporary directory.

bin/i18n-update.sh
```diff
@@ -0,0 +1,120 @@
...
+  PLUGIN_FOLDER=${PLUGINS[index+1]}
+
+  if [[ ! -d $PLUGIN_FOLDER ]]; then
+    NOT_FOUND_PLUGIN_FOLDERS+=( $PLUGIN_FOLDER )
+    echo -e "\033[0;31mPlugin folder \"$PLUGIN_FOLDER\" doesn't exist.\033[0m"
+  fi
+done
+if [[ -n $NOT_FOUND_PLUGIN_FOLDERS ]]; then
+  exit 1
+fi
```
fluiddot: Here we validate that the plugin folders provided via the arguments exist before executing the script.

bin/i18n-update.sh
```diff
@@ -0,0 +1,120 @@
...
+
+# Set used strings target
+USED_STRINGS_PATH="$TARGET_PATH/used-strings.json"
+
+# Extract used strings for plugins
+node gutenberg/packages/react-native-editor/bin/extract-used-strings $USED_STRINGS_PATH "${PLUGINS[@]}"
```
fluiddot: Similar to the translations download, we'll use the new script created in [this PR](https://github.com/WordPress/gutenberg/pull/37482) for extracting the used strings.

bin/i18n-update.sh
```diff
@@ -0,0 +1,120 @@
...
+done
+
+# Download translations of Gutenberg
+fetch_translations "gutenberg" $TRANSLATIONS_OUTPUT_PATH $USED_STRINGS_PATH
+
+echo -e "\n\033[1mGenerating localization strings files\033[0m"
+
+# Generate localization strings files
+./bin/po2android.js bundle/android/strings.xml $USED_STRINGS_PATH
+./bin/po2swift.js bundle/ios/GutenbergNativeTranslations.swift $USED_STRINGS_PATH
```
fluiddot: As mentioned in other comments, now the scripts in charge of generating the localization strings files will use the used strings JSON generated earlier instead of the POT files ([reference](https://github.com/wordpress-mobile/gutenberg-mobile/blob/29fb512d13ebb96381029537ae9f2c595c7b5ed8/package.json#L84-L85)).

bin/po2android.js
```diff
@@ -63,79 +76,120 @@ const getUniqueName = ( function() {
...
+	const usedStringsFile = process.argv[ 3 ];
+
+	const usedStrings = getUsedStrings( usedStringsFile );
+	let flattenUsedStrings = Object.entries( usedStrings ).reduce(
+		( result, [ domain, strings ] ) => {
+			return [ ...result, ...Object.values( strings ) ];
+		},
+		[]
+	);
+	flattenUsedStrings = [ ...new Set( flattenUsedStrings ) ];
```
fluiddot: Used strings are grouped by domain, since we only provide one localization strings files, we unify them here and their translations will be requested as part of the main apps' strings.

bin/po2swift.js
```diff
@@ -1,47 +1,74 @@
...
+	const usedStringsFile = process.argv[ 3 ];
+
+	const usedStrings = getUsedStrings( usedStringsFile );
+	let flattenUsedStrings = Object.entries( usedStrings ).reduce(
+		( result, [ domain, strings ] ) => {
+			return [ ...result, ...Object.values( strings ) ];
+		},
+		[]
+	);
+	flattenUsedStrings = [ ...new Set( flattenUsedStrings ) ];
```
fluiddot: Used strings are grouped by domain, since we only provide one localization strings files, we unify them here and their translations will be requested as part of the main apps' strings.

bin/po2android.js
```diff
@@ -63,79 +76,120 @@ const getUniqueName = ( function() {
...
+
+		const isNative =
+			! platforms.includes( 'web' ) && platforms.includes( 'android' );
+		const hasContext = !! item.context;
+		if ( isNative && hasContext ) {
+			nativeStringsWithContext.push( string );
+		}
+
+		return isNative && ! hasContext;
+	}, [] );
```
fluiddot: We only add the localization strings files of strings referenced in `.native.js` and `.android.js files`, and that doesn't have context.

bin/po2swift.js
```diff
@@ -1,47 +1,74 @@
...
+
+		const isNative =
+			! platforms.includes( 'web' ) && platforms.includes( 'ios' );
+		const hasContext = !! item.context;
+		if ( isNative && hasContext ) {
+			nativeStringsWithContext.push( string );
+		}
+
+		return isNative && ! hasContext;
+	}, [] );
```
fluiddot: We only add the localization strings files of strings referenced in `.native.js` and `.ios.js files`, and that doesn't have context.

bin/po2android.js
fluiddot: The current list of strings that we won't be able to translate are:
```
[
  'Link rel attribute value placeholder\x04None',
  'inline textinput cell\x04%s. Empty',
  'inline textinput cell\x04%1$s, %2$s. Empty',
  'inline textinput cell\x04%1$s, %2$s',
  'inline textinput cell\x04%1$s, %2$s, %3$s',
  'switch control\x04%s. On',
  'switch control\x04%s. Off',
  'switch control\x04%1$s, %2$s. On',
  'switch control\x04%1$s, %2$s. Off'
]
```

**NOTE:** The strings contain both the value and context, it has the following format: `<CONTEXT>\x04<STRING>`.

package.json
```diff
@@ -63,6 +63,8 @@
 		"postrn-bundle": "patch-package --reverse --patch-dir gutenberg/packages/react-native-editor/metro-patch",
 		"i18n:gutenberg": "./bin/i18n-download.sh gutenberg",
 		"i18n:gutenberg:force": "cross-env REFRESH_I18N=1 ./bin/i18n-download.sh gutenberg",
+		"i18n:update": "./bin/i18n-update.sh jetpack ./jetpack/projects/plugins/jetpack/extensions layout-grid ./block-experiments/blocks/layout-grid",
```
fluiddot, line 64: Eventually, once this script is ready to merge, we could remove all the `makepot` and `wpcli` commands and attach this one to `prebundle:js`.

package.json
```diff
@@ -61,28 +61,15 @@
...
-		"makepot:ios": "npm run wp -- i18n make-pot ./gutenberg/packages --include=*.native.js,*.ios.js --exclude=test/*,e2e-tests/*,build/*,build-module/*,build-style/* --skip-php --subtract=./gutenberg.pot --ignore-domain gutenberg-ios.pot",
-		"premakepot:gutenberg": "npm run clean:gutenberg && npm run build:gutenberg",
-		"makepot:gutenberg": "npm run wp -- i18n make-pot ./gutenberg/packages --exclude=test/*,e2e-tests/*,build/*,build-module/*,build-style/*,*.native.js,*.ios.js,*.android.js,bundle/* --ignore-domain gutenberg.pot",
-		"postmakepot:gutenberg": "npm run clean:gutenberg",
-		"pregenstrings": "npm run makepot:gutenberg",
-		"genstrings": "npm run genstrings:android && npm run genstrings:ios",
-		"genstrings:android": "npm run makepot:android && ./bin/po2android.js gutenberg-android.pot bundle/android/strings.xml",
-		"genstrings:ios": "npm run makepot:ios && ./bin/po2swift.js gutenberg-ios.pot bundle/ios/GutenbergNativeTranslations.swift",
```
fluiddot: All these commands are no longer necessary as it's already covered by the i18n update script.

package.json
```diff
@@ -61,28 +61,15 @@
...
 		"rn-bundle": "react-native bundle",
 		"postrn-bundle": "patch-package --reverse --patch-dir gutenberg/packages/react-native-editor/metro-patch",
-		"i18n:gutenberg": "./bin/i18n-download.sh gutenberg",
-		"i18n:gutenberg:force": "cross-env REFRESH_I18N=1 ./bin/i18n-download.sh gutenberg",
-		"bundle": "npm run clean; npm run bundle:js && npm run genstrings",
-		"prebundle:js": "npm run i18n:gutenberg:force",
+		"i18n:update": "./bin/i18n-update.sh jetpack ./jetpack/projects/plugins/jetpack/extensions layout-grid ./block-experiments/blocks/layout-grid",
+		"i18n:update:debug": "./bin/i18n-update.sh --debug jetpack ./jetpack/projects/plugins/jetpack/extensions layout-grid ./block-experiments/blocks/layout-grid",
+		"bundle": "npm run clean; npm run bundle:js",
+		"prebundle:js": "npm run i18n:update",
```
fluiddot: Now when generating a bundle, we'll always update the translations and produce the localization strings files.

.buildkite/pipeline.yml
```diff
@@ -23,7 +23,7 @@ steps:
         nvm install && nvm use
         npm ci --unsafe-perm --prefer-offline --no-audit --no-progress
 
-        npm run prebundle:js
+        # npm run prebundle:js
```
fluiddot: The command that updates the i18n localizations uses WP-CLI that requires PHP, so we would need to update the [`gb-mobile-image` docker image](https://github.com/wordpress-mobile/docker-gb-mobile-image) and include PHP as part of the installed tools.

bin/i18n-update.sh
```diff
@@ -0,0 +1,124 @@
...
+  esac
+done
+
+function error() {
+  echo -e "\033[0;31m$1\033[0m"
+  exit 1
+}
+
+function fetch_translations() {
+  local PLUGIN_NAME=$1
```
jhnstn: minor nitpick: local variables should be lower case. ALL_UPPERCASE is generally used for exported variables only.

bin/i18n-update.sh
```diff
@@ -0,0 +1,124 @@
...
+  local OUTPUT_PATH=$2
+  local USED_STRINGS_FILE=$3
+
+  echo -e "\n\033[1mDownload I18n translations for \"$PLUGIN_NAME\" plugin\033[0m"
+  node gutenberg/packages/react-native-editor/bin/i18n-translations-download $PLUGIN_NAME $OUTPUT_PATH $USED_STRINGS_FILE
```
jhnstn: All these variables should be wrapped in strings to avoid string splitting
see https://github.com/koalaman/shellcheck/wiki/SC2086

bin/i18n-update.sh
```diff
@@ -0,0 +1,124 @@
...
+#      source code files ("native.js", "ios.js or "android.js"). The translations of these
+#      strings are not included in the GlotPress projects of their plugins hence, they require
+#      to be requested as part of the main apps’ strings.
+#
+#      This step produces the following output files:
+#      - bundle/android/strings.xml                     [Localization strings files for Android platform]
+#      - bundle/ios/GutenbergNativeTranslations.swift   [Localization strings files for iOS platform]
+
+# Exit if any command fails
+set -e
```
jhnstn: Nitpick: might want to consider using `set -euxo pipefail` see https://vaneyckt.io/posts/safer_bash_scripts_with_set_euxo_pipefail/

`set -u` is a bit tricky. just need to use the `INPUT=${VAR:-"some default"}` form to make sure everything is set.

bin/i18n-update.sh
```diff
@@ -0,0 +1,124 @@
...
+PLUGINS=( "$@" )
+
+echo -e "\n\033[1m== Updating i18n localizations ==\033[0m"
+
+# Validate parameters
+if [[ $((${#PLUGINS[@]}%2)) -ne 0 ]]; then
+  error "Plugin arguments must be be even."
+fi
+
+for (( index=0; index<${#PLUGINS[@]}; index+=2 )); do
```
jhnstn, line 90: It would be useful to note that the plugin arguments are tuples  i.e. ` domain path/to/plugin` 
That will make it easier to grok why the loop is indexing by 2

bin/i18n-update.sh
```diff
@@ -0,0 +1,124 @@
...
+    cp -r $OUTPUT_PATH/gutenberg/data gutenberg/packages/react-native-editor/i18n-cache
+    cp $OUTPUT_PATH/gutenberg/index.js gutenberg/packages/react-native-editor/i18n-cache/index.native.js
+  fi
+}
+
+# Set target path
+if [[ -n $LOCAL_PATH ]]; then
+  TARGET_PATH=$LOCAL_PATH
+else
+  TARGET_PATH=$(mktemp -d)
```
jhnstn, line 76: Probably should add a trap to clean up the temp file :  
something like `trap 'rm -f "$TARGET_PATH"' EXIT`. 
That will run when the script exits ( well except on `kill -9` )

bin/i18n-update.sh
```diff
@@ -0,0 +1,124 @@
...
+
+for (( index=0; index<${#PLUGINS[@]}; index+=2 )); do
+  PLUGIN_FOLDER=${PLUGINS[index+1]}
+
+  if [[ ! -d $PLUGIN_FOLDER ]]; then
+    NOT_FOUND_PLUGIN_FOLDERS+=( $PLUGIN_FOLDER )
+    echo -e "\033[0;31mPlugin folder \"$PLUGIN_FOLDER\" doesn't exist.\033[0m"
+  fi
+done
+if [[ -n $NOT_FOUND_PLUGIN_FOLDERS ]]; then
```
jhnstn: It probably doesn't matter here but _might_ want to check the length of `$NOT_FOUND_PLUGIN_FOLDERS` since this `if` is just checking the first element. something like
`if [[ ${#NOT_FOUND_PLUGIN_FOLDERS[@]} -gt 0 ]]; then ...`

jhnstn:
I ran `npm run i18n:update` and got the expected changes in `src/i18n-cache`. I did get slightly different warnings.
for Jetpack : 
```
WARNING: The following native-only strings are already being included in translations:
[ 'Featured', 'Text color' ]
```
I also noticed that `Text color` was added to the android and ios string files.

I should also mention that I ran into a problem with building the TypeScript files (specifically `npx tsc --build` was failing when ran in `./gutenberg`). I had to clean out the node modules in Gutenberg to get past the errors. I'm assuming that is something that would need to be addressed in Gutenberg but figure I'd mention it here if anyone else runs into it.

Overall the changes look good. I'll leave it up to you @fluiddot to pick up the bash suggestions I left. None of them are blockers however. Let me know when you get back if you want merge as is or make those changes.

bin/i18n-update.sh
jhnstn: It's been awhile since I've read that article 😊  Probably don't want `set -x` in there once this is ready to ship but it helps for debugging.

fluiddot: Yep, I wouldn't add the `-x` option as it will print too much debug info.

fluiddot: I applied the suggestion in https://github.com/wordpress-mobile/gutenberg-mobile/pull/4366/commits/8bc16298d8302cd8052e1cb59333def57759d97a.

bin/i18n-update.sh
fluiddot: I've renamed all local variables in https://github.com/wordpress-mobile/gutenberg-mobile/pull/4366/commits/9274f89558f303ea360545d9e170ba0c53489a20. Additionally, I updated the script `generate-pot-files.sh` in Gutenberg with the same format ([reference](https://github.com/WordPress/gutenberg/pull/37482/commits/f60f993c4476eb5174eb4525f3ae0e7cd7723af3)).

bin/i18n-update.sh
fluiddot: Good catch, I've applied this in https://github.com/wordpress-mobile/gutenberg-mobile/pull/4366/commits/8a7dd7aadd07574c5051f2f5a78aa24dbd438f1d. And similar to local variable naming, I also updated the script `generate-pot-files.sh in Gutenberg ([reference](https://github.com/WordPress/gutenberg/pull/37482/commits/f60f993c4476eb5174eb4525f3ae0e7cd7723af3)).

bin/i18n-update.sh
fluiddot: Oh I didn't know about that command 🤩 , I added this in https://github.com/wordpress-mobile/gutenberg-mobile/pull/4366/commits/a9cc79e9051e094d54dd161d2d576c9513311cb3.

> That will run when the script exits ( well except on kill -9 )
I don't think this should be a major issue as it's a temporary folder and eventually it will be deleted by the OS.

bin/i18n-update.sh
fluiddot: Right, I updated the error message in https://github.com/wordpress-mobile/gutenberg-mobile/pull/4366/commits/c20f4d6603805f8db499bd723ab8e7e44ca7c2d0.

bin/i18n-update.sh
fluiddot: As far as I checked both options (non-zero vs array length) work but since the value is an array I agree it's more consistent checking its length 👍 .

I modified it in https://github.com/wordpress-mobile/gutenberg-mobile/pull/4366/commits/7edfa4d026c74961768008a670e7f56e6294d9af. Note that due to the use of option `-x` (treat unset variables as an error), the way to obtain the array length is a bit different. I followed the recommendation from [this StackOverflow article](https://stackoverflow.com/a/61551944) related to the safe idiom for this calculation.

fluiddot (commented):
Thanks @jhnstn for reviewing the changes 🙇 , I've replied to your comments and applied the suggestions.

Regarding the feedback from https://github.com/wordpress-mobile/gutenberg-mobile/pull/4366#issuecomment-1000542029:

> I did get slightly different warnings.
> for Jetpack :
> 
> WARNING: The following native-only strings are already being included in translations:
> [ 'Featured', 'Text color' ]
> I also noticed that Text color was added to the android and ios string files.

Looks like this string was recently added ([PR reference](https://github.com/WordPress/gutenberg/pull/36028)) so when I generated the strings during testing, I didn't have this change.

It's interesting because this is a case we might run into in the future, as far as I checked, the string `Text color` is included in the Gutenberg GlotPress project ([reference](https://translate.wordpress.org/projects/wp-plugins/gutenberg/dev/en-gb/default/?filters%5Bterm%5D=Text+color&filters%5Bterm_scope%5D=scope_any&filters%5Bstatus%5D=current_or_waiting_or_fuzzy_or_untranslated&filters%5Bcase_sensitive%5D=yes&filters%5Buser_login%5D=&filter=Apply+Filters&sort%5Bby%5D=priority&sort%5Bhow%5D=desc)) because it was referenced in the web version but the value was changed months ago ([reference](https://github.com/WordPress/gutenberg/pull/34680/files#diff-6ef6519967765b9d4a49b27938bf474f58270c38d8f5358cf4c878c97e2b4276L21-R21)). Not sure why it's still referenced in the GlotPress project but I'd expect that in next Gutenberg plugin releases it's removed, so I think it should be safe to consider the string as part of the native version even though it will be eventually duplicated.

> I should also mention that I ran into a problem with building the TypeScript files (specifically npx tsc --build was failing when ran in ./gutenberg). I had to clean out the node modules in Gutenberg to get past the errors. I'm assuming that is something that would need to be addressed in Gutenberg but figure I'd mention it here if anyone else runs into it.

Sorry to hear you had problems when compiling the Typescript files, did you experience it when running the command `npm run build:gutenberg`?

> Overall the changes look good. I'll leave it up to you @fluiddot to pick up the bash suggestions I left. None of them are blockers however. Let me know when you get back if you want merge as is or make those changes.

Awesome, I already included the suggestions so I'd appreciate it if you could take another look to the changes, thanks 🙇 !

.buildkite/pipeline.yml
fluiddot: This will be solved in https://github.com/wordpress-mobile/docker-gb-mobile-image/pull/3.

jhnstn:
> Sorry to hear you had problems when compiling the Typescript files, did you experience it when running the command `npm run build:gutenberg`?

I believe that was the parent command but it was for sure an issue trying to transpile the Typescript. I didn't trace back the root issue since it was complaining about something in `gutenberg/node_modules`. I would say it's a an unlikely edge case we might not see again. Just wanted to note that reinstalling the node modules fixed the issue.

jhnstn (approved):
All the changes look good. Excellent work @fluiddot

fluiddot:
@jhnstn thank you very much for the review ❤️  ! I'll mark your last comments as resolved ✅ .

Regarding this statement from the PR's description:
> The i18n update command has been unlinked from the bundle generation ([commit reference 1](https://github.com/wordpress-mobile/gutenberg-mobile/pull/4366/commits/ac35a6ff93e09862e34c55e743e956fcd9b596c8)) because it was making some CI jobs fail ([reference](https://buildkite.com/wordpress-mobile/gutenberg-mobile/builds/1719#34a373d2-b44b-4186-bcfe-219eba666b6c/170-785), [commit reference 2](https://github.com/wordpress-mobile/gutenberg-mobile/pull/4366/commits/0c54bd690f29e3a65218eb66f5dbbc3ab1b24303)) due to the use of PHP. It shouldn't block the PR as this can be addressed later on.

I realized that the generated bundle for Android actually requires to include the localizations, otherwise, we would miss all translations. Therefore, we need to re-enable the `prebundle:js` command and provide either the original logic related to downloading the translations or use the new i18n update command, however for the latter, the PR would be blocked until the Docker image is updated ([reference](https://github.com/wordpress-mobile/docker-gb-mobile-image/pull/3)).

At this point and in order to prevent blocking the PR, I'm leaning towards providing the original logic and attaching the i18n update command in a new PR. This would imply making some small changes so I'd appreciate it if you could take another look once I push them, thanks 🙇

bin/i18n-update.sh
```diff
@@ -0,0 +1,127 @@
...
+done
+
+# Download translations of Gutenberg
+fetch_translations "gutenberg" "$TRANSLATIONS_OUTPUT_PATH" "$USED_STRINGS_PATH"
+
+echo -e "\n\033[1mGenerating localization strings files\033[0m"
+
+# Generate localization strings files
+./bin/strings2android.js bundle/android/strings.xml "$USED_STRINGS_PATH"
+./bin/strings2swift.js bundle/ios/GutenbergNativeTranslations.swift "$USED_STRINGS_PATH"
```
fluiddot, line 127: The previous scripts used here`po2android`/`po2swift` have been renamed to `strings2android`/`strings2swift` which matches the file we use for conversion (used strings JSON files).

jhnstn (approved):
>  I'm leaning towards providing the original logic and attaching the i18n update command in a new PR.

Sounds good to me, :shipit:

PR #4383, opened 2021-12-16
Update submodule dependencies

dcalhoun:
Update Gutenberg version and capture lock file changes originating from dependency updates within the `gutenberg` submodule. 

Additionally, the Jest test runner was resolving different version of Jest sub-dependencies, which resulted in the following error.

```
TypeError: snapshotResolver.resolveSnapshotPath is not a function

  at Object._default [as default] (node_modules/jest-jasmine2/build/setup_jest_globals.js:101:41)
```

Specifically, the test runner resolved `node_modules/jest-jasmine2@26.6.3 > gutenberg/node_modules/jest-snapshot@27.4.5`. `jest-snapshot@27` includes a breaking change of a newly returned `Promise`: facebook/jest#8829

This reordering prioritizes the dependencies found in the top-level `node_modules` directory over the directory found within the `gutenberg` submodule. The origin of this incorrect module resolution is a combination of https://github.com/WordPress/gutenberg/pull/37396 and https://github.com/WordPress/gutenberg/pull/33448.

To test: Passing CI checks should suffice.

PR submission checklist:


jest.config.js
```diff
@@ -59,7 +59,6 @@ module.exports = {
 	haste: {
 		defaultPlatform: rnPlatform,
 		platforms: [ 'android', 'ios', 'native' ],
-		providesModuleNodeModules: [ 'react-native', 'react-native-svg' ],
```
dcalhoun: This configuration is no longer required per https://github.com/facebook/jest/pull/8535.

package.json
```diff
@@ -38,8 +38,9 @@
 		"gettext-parser": "^4.0.3",
 		"husky": "^3.0.2",
 		"identity-obj-proxy": "^3.0.0",
-		"jest": "^24.9.0",
+		"jest": "26.6.3",
 		"jest-junit": "^6.3.0",
+		"jest-snapshot": "^27.4.5",
```
dcalhoun: These changes mirror Gutenberg updates in https://github.com/WordPress/gutenberg/pull/37396 and are required to avoid a breaking change within https://github.com/facebook/jest/pull/8829.

fluiddot: I see that this dependency was added to `e2e-tests` package ([reference](https://github.com/WordPress/gutenberg/pull/37396/files#diff-73662dce866d908da409d58c0dd1b3fcef7016ad2125dbb7b01a4172249f5280R35)) that AFAIK we're not using it in GB-mobile, do we need it even though?

package.json
```diff
@@ -38,8 +38,9 @@
 		"gettext-parser": "^4.0.3",
 		"husky": "^3.0.2",
 		"identity-obj-proxy": "^3.0.0",
-		"jest": "^24.9.0",
+		"jest": "26.6.3",
```
fluiddot: I'm wondering if we should use the `^` in the version as we have in the rest of dependencies, wdyt?

fluiddot (dismissed):
I've just added a couple of minor comments, the PR LGTM 🎊 !

- Confirmed that CI checks are passing ✅ 
- Run `npm install` locally and verify that `package-lock.json` is not modified ✅

package.json
dcalhoun: > I see that this dependency was added to `e2e-tests` package ([reference](https://github.com/WordPress/gutenberg/pull/37396/files#diff-73662dce866d908da409d58c0dd1b3fcef7016ad2125dbb7b01a4172249f5280R35)) that AFAIK we're not using it in GB-mobile, do we need it even though?

@fluiddot I agree that we do not leverage `jest-snapshot` directly and it is odd that we would need to add it as a direct dependency. However, without adding `"jest-snapshot": "^27.4.5"`, the CI tasks currently fail. 

The failure without the addition seems to be related to dependency resolution. For example, without this addition, the test scripts resolve `node_modules/jest-jasmine2@26.6.3 > gutenberg/node_modules/jest-snapshot@27.4.5`. This results in the error captured in the PR description, as `jest-snapshot@27` includes a breaking change of a newly returned `Promise`: https://github.com/facebook/jest/pull/8829

Adding `"jest-snapshot": "^27.4.5"` resolved the error, but it may be an inappropriate solution. It would appear that the resolution path is now `node_modules/jest-jasmine2@26.6.3 > node_modules/jest-snapshot@27.4.5/node_modules/jest-snapshot@26.6.3`, meaning that `node_modules/jest-snapshot@27.4.5` is unused. 

It is all quite strange. I am researching this further.

package.json
dcalhoun: The explicit version was set to mirror [Gutenberg's approach](https://github.com/WordPress/gutenberg/blob/78b235a583f3ab67f82301eff635e1ff95266db5/package.json#L178), but I am unaware of a reason to avoid `^` in this circumstance. I will add the `^` to the version range.

package.json
fluiddot: > @fluiddot I agree that we do not leverage jest-snapshot directly and it is odd that we would need to add it as a direct dependency. However, without adding "jest-snapshot": "^27.4.5", the CI tasks currently fail.

Oh I see, in that case, let's stick with this approach until we found a better alternative 👍 .

guarani (approved):
The changes look good!
In the `package-lock.json` file I see only changes to `@types/react` and `@types/react-dom` which come from the update to the Gutenberg submodule (I see the same version updates there).
Thanks so much @dcalhoun!

PR #4443, opened 2022-01-07
Fix LinkPicker freeze when virtual keyboard is hidden

dcalhoun:
This brings in https://github.com/WordPress/gutenberg/pull/37782 to fix an app freeze after closing the link picker while a device's virtual keyboard is hidden. 

To test: See https://github.com/WordPress/gutenberg/pull/37782. 

PR submission checklist:

- [x] I have considered adding unit tests where possible.
- [x] I have considered if this change warrants user-facing release notes [more info](https://github.com/wordpress-mobile/gutenberg-mobile/blob/develop/Release-notes.md) and have added them to [RELEASE-NOTES.txt](https://github.com/wordpress-mobile/gutenberg-mobile/blob/develop/RELEASE-NOTES.txt) if necessary.

PR #4438, opened 2022-01-07
Release 1.69.0

mkevins:
Release for Gutenberg Mobile 1.69.0

## Related PRs

- Gutenberg: https://github.com/WordPress/gutenberg/pull/37767
- WPAndroid: https://github.com/wordpress-mobile/WordPress-Android/pull/15805
- WPiOS: https://github.com/wordpress-mobile/WordPress-iOS/pull/17726

- Aztec-iOS: N/A
- Aztec-Android: N/A

## Extra PRs that Landed After the Release Was Cut

No extra PRs yet. 🎉

### Give multi-line block names central alignment in inserter
* https://github.com/wordpress-mobile/gutenberg-mobile/pull/4343
* https://github.com/WordPress/gutenberg/issues/34748

### Fix missing translations by refactoring the editor initialization code
* https://github.com/wordpress-mobile/gutenberg-mobile/pull/4332
* https://github.com/wordpress-mobile/gutenberg-mobile/issues/4329

### Add Jetpack and Layout Grid translations
* https://github.com/wordpress-mobile/gutenberg-mobile/pull/4359
* https://github.com/wordpress-mobile/gutenberg-mobile/issues/2615

### Fix text formatting mode lost after backspace is used
* https://github.com/wordpress-mobile/gutenberg-mobile/pull/4423
* https://github.com/WordPress/gutenberg/issues/34938

### Add missing translations of unsupported block editor modal
* https://github.com/wordpress-mobile/gutenberg-mobile/pull/4410

## Test plan

Once the installable builds of the main apps are ready, perform a quick smoke test of the editor on both iOS and Android to verify it launches without crashing. We will perform additional testing after the main apps cut their releases.

## Release Submission Checklist

- [ ] Verify Items from test plan have been completed
- [x] Check if `RELEASE-NOTES.txt` is updated with all the changes that made it to the release. Replace `Unreleased` section with the release version and create a new `Unreleased` section.
- [x] Check if `gutenberg/packages/react-native-editor/CHANGELOG.md` is updated with all the changes that made it to the release. Replace `## Unreleased` with the release version and create a new `## Unreleased`.
- [x] Bundle package of the release is updated.

PR #4448, opened 2022-01-09
Merge release 1.69.0 to develop

mkevins:
This PR merges the 1.69.0 release back to `develop`.


ttahmouch (approved):
These changes look good to me. 👍

RELEASE-NOTES.txt
```diff
@@ -1,5 +1,8 @@
 Unreleased
 ---
+
+1.69.0
```
fluiddot, line 4: I noticed that the entry "Fix app freeze when closing link picker while virtual keyboard is hidden" has been included in `1.69.0` but [the PR](https://github.com/wordpress-mobile/gutenberg-mobile/pull/4443) was merged after the release was cut ([here](https://github.com/wordpress-mobile/gutenberg-mobile/pull/4438/files?authenticity_token=oDLzV1onSLSEyk638jPZOgjByqbOLDtLeJJPLELWhIerXgUhCwT2CPWcmgTuz7mnPvyrmVyiweIdvwom30k_xg&hide-deleted-files=true&file-filters%5B%5D=#diff-558ce8820a544da19f7fccdb2d1d661e46d68c130a1a39abdd222aa3484a5873) you can check the release notes of the release branch). I'll update the release notes to reflect this change.

fluiddot: Similarly, we could also update the `react-native-editor` changelog file in Gutenberg, which also included this entry under `1.69.0`.

fluiddot: I'm thinking that since this change will be released as a betafix, I think it makes more sense to update both release notes and changelog files there.

mkevins: Good catch Carlos! The good thing is, this was _not_ included in the actual release (nor release notes), but was only included in this merge back to `develop` (as an auto-resolved conflict when merging `develop` in [this commit](https://github.com/wordpress-mobile/gutenberg-mobile/pull/4448/commits/dcae8cfacd50f126fa13dc9a6712b32d38c09223)).

> I'm thinking that since this change will be released as a betafix, I think it makes more sense to update both release notes and changelog files there.

I agree, that's the correct approach, and actually, can be under it's own section `1.69.1`. Thanks for pointing this out!

PR #4463, opened 2022-01-13
RichText - Use parsed font size values when comparing new changes

geriux:
- Gutenberg PR: https://github.com/WordPress/gutenberg/pull/37951

This PR fixes a bug within `RichText`, where in some cases it would generate a render loop when setting a new font-size value and this didn't match the parsed value.

PR submission checklist:


fluiddot (approved):
Approved via https://github.com/WordPress/gutenberg/pull/37951#pullrequestreview-852909416.

**NOTE:** We might consider adding an entry to the release notes as this bugfix looks like a user-facing change.

geriux:
> **NOTE:** We might consider adding an entry to the release notes as this bugfix looks like a user-facing change.

I didn't add one since this will be included in a betafix, if I'm not mistaken, we don't add new entries for these cases right?

geriux:
Thinking about it, I think that's the case for the main apps right? So we do include them in Gutenberg and Gutenberg Mobile but not in the main apps beta fixes since the changelog usually is already edited and translated, is that correct?

fluiddot:
> Thinking about it, I think that's the case for the main apps right? So we do include them in Gutenberg and Gutenberg Mobile but not in the main apps beta fixes since the changelog usually is already edited and translated, is that correct?

Right, I guess that when the beta fix is cut, this change will be included under a new section for that version 👍 .

Regarding the changelog of the main apps, I think that's correct, we don't update it because usually the release notes had already been sent for translations before the beta fix is merged.

Issue #4432, opened 2022-01-05
Remaining untranslated strings after improving the translation pipeline


fluiddot:
Although most of the strings are now translated after introducing improvements in the translation pipeline, we identified the following strings that remain untranslated and would require further investigation:

- ~~`Order by` (only referenced in the Search block)~~ _Already translated._
- ~~`Newest to oldest` (only referenced in the Search block)~~ _Already translated._
- ~~`Oldest to newest` (only referenced in the Search block)~~ _Already translated._
- ~~`One` (only referenced in the Layout Grid block)~~ _Will be translated as a native-only string._
- ~~`Two` (only referenced in the Layout Grid block)~~ _Will be translated as a native-only string._
- ~~`Three` (only referenced in the Layout Grid block)~~ _Will be translated as a native-only string._
- ~~`Four` (only referenced in the Layout Grid block)~~ _Will be translated as a native-only string._
- ~~`Angle`~~ _Already translated._
- ~~`Solid`~~ _Fixed in https://github.com/wordpress-mobile/gutenberg-mobile/pull/4479._
- ~~`Gradient`~~ _Fixed in https://github.com/wordpress-mobile/gutenberg-mobile/pull/4479._
- ~~`Linear`~~ _Fixed in https://github.com/wordpress-mobile/gutenberg-mobile/pull/4479._
- ~~`Radial`~~ _Fixed in https://github.com/wordpress-mobile/gutenberg-mobile/pull/4479._

### Only missing on Android
- ~~`Set as Featured Image`~~ _Fixed in https://github.com/wordpress-mobile/gutenberg-mobile/pull/4464_.
- ~~`Remove as Featured Image`~~ _Fixed in https://github.com/wordpress-mobile/gutenberg-mobile/pull/4464_.
- ~~`X-Axis Position`~~ _Fixed in https://github.com/wordpress-mobile/gutenberg-mobile/pull/4484._
- ~~`Y-Axis Position`~~ _Fixed in https://github.com/wordpress-mobile/gutenberg-mobile/pull/4484._
- ~~`Describe the purpose of the image. Leave empty if the image is purely decorative.`~~ _Fixed in https://github.com/WordPress/gutenberg/pull/38115._
- ~~`Add a new block at any time by tapping on the + icon in the toolbar on the bottom left.`~~ _Fixed in https://github.com/WordPress/gutenberg/pull/38115._

**Replace featured image popup:**
- `Featured Image Already Set`
[...]

PR #4464, opened 2022-01-13
Fix missing Featured Image translations

jhnstn:
Fix missing translations on Android
See GB : https://github.com/WordPress/gutenberg/pull/37956

Partial Fix for #4432

To test:
See test step on https://github.com/WordPress/gutenberg/pull/37956

**NOTE:** This issue only affects the Android platform.
_(Edited by @fluiddot)_

PR submission checklist:


fluiddot (changes requested):
I think the localization strings file for iOS should be also updated as the string now has a different value ([reference](https://github.com/wordpress-mobile/gutenberg-mobile/blob/develop/bundle/ios/GutenbergNativeTranslations.swift)). You could update both by running the command `npm run i18n:update`.

bundle/android/strings.xml
```diff
@@ -142,7 +142,7 @@
     <string name="gutenberg_native_duplicate_block" tools:ignore="UnusedResources">Duplicate block</string>
     <string name="gutenberg_native_each_block_has_its_own_settings_to_find_them_tap_on_a_block_its_s" tools:ignore="UnusedResources">Each block has its own settings. To find them, tap on a block. Its settings will appear on the toolbar at the bottom of the screen.</string>
     <string name="gutenberg_native_edit_cover_media" tools:ignore="UnusedResources">Edit cover media</string>
-    <string name="gutenberg_native_edit_file" tools:ignore="UnusedResources">Edit file</string>
+    <string name="gutenberg_native_edit_file" tools:ignore="UnusedResources">Edit file </string>
```
fluiddot: Not sure if this is accurate, this string is referenced [here](https://github.com/WordPress/gutenberg/blob/1f1bccf711a9f783881ea51b3f0bf6b32c4ebc18/packages/block-library/src/file/edit.native.js#L232) but it doesn't include the extra space 🤔 .

bundle/android/strings.xml
```diff
@@ -311,7 +311,7 @@ translators: Block name. %s: The localized block name -->
     <!-- translators: %s: Select font size option value e.g: "Selected: Large".
 translators: %s: Select control option value e.g: "Auto, 25%". -->
     <string name="gutenberg_native_selected_s" tools:ignore="UnusedResources">Selected: %s</string>
-    <string name="gutenberg_native_set_as_featured_image" tools:ignore="UnusedResources">Set as Featured Image </string>
+    <string name="gutenberg_native_set_as_featured_images" tools:ignore="UnusedResources">Set as Featured Images </string>
```
fluiddot: I couldn't find a commit where this string was changed, do you have any clue why it changed?

jhnstn: good catch, I _think_ I might have changed this and the other change above while trying to figure out the issue.

jhnstn: I know what I did 🤦 , I ran the `i18n:update` command while on the wrong gutenberg ref.

fluiddot: Oh, that explains the changes 😅.

jhnstn: Fixed in https://github.com/wordpress-mobile/gutenberg-mobile/pull/4464/commits/5821350c3c9bd9d127fe04e33fba2115c61bccdf

fluiddot (approved):
LGTM 🎊 !

Approved via https://github.com/WordPress/gutenberg/pull/37956#pullrequestreview-852844649.

**NOTE:** I see that the Gutenberg ref is not pointing to the latest commit of https://github.com/WordPress/gutenberg/pull/37956, we should update it once the Gutenberg PR is merged.

fluiddot:
@jhnstn I'm wondering if we should update the PR's title in order to describe more accurately the introduced change, wdyt?
On the other hand, since the Gutenberg PR added an entry to the changelog file, probably we should do the same in the release notes.

fluiddot:
I noticed while testing the iOS beta version `19.1 (19.1.0.0)` that the strings referenced in the [Gutenberg PR](https://github.com/WordPress/gutenberg/pull/37956) are not translated. I guess this is caused by the fact that the strings have changed and localization files in WordPress-iOS haven't been updated yet ([reference](https://github.com/wordpress-mobile/WordPress-iOS/blob/1a91e9c6ec630abd6e595fc5b3c9c01b75691d0d/WordPress/Resources/es.lproj/Localizable.strings#L6361)).

**Not translated strings:**
- `Set as Featured Image`
- `Remove as Featured Image`

PR #4469, opened 2022-01-14
Add clean i18n cache command

fluiddot:
Adds a clean command for removing the i18n cache.

**To test:**
1. Run command `npm run clean`.
2. Observe that the folder `src/i18n-cache` is removed.
3. Run command `npm run ci`.
4. Observe that the folder `src/i18n-cache` is created and contains translation files for supported plugins: Gutenberg (`gutenberg` folder) , Jetpack (`jetpack` folder) and Layout Grid (`layout-grid` folder).

PR submission checklist:

- [x] I have considered adding unit tests where possible.
- [x] I have considered if this change warrants user-facing release notes [more info](https://github.com/wordpress-mobile/gutenberg-mobile/blob/develop/Release-notes.md) and have added them to [RELEASE-NOTES.txt](https://github.com/wordpress-mobile/gutenberg-mobile/blob/develop/RELEASE-NOTES.txt) if necessary.

PR #4474, opened 2022-01-14
Force check i18n cache when cleaning project

fluiddot:
Following up https://github.com/wordpress-mobile/gutenberg-mobile/pull/4469, this PR includes the command `npm run i18n:check-cache` in i18n cache cleaning command (`npm run clean:i18n`). This way we assure that the i18n cache is always present, which is required for generating the JS bundle. 

Thanks @guarani for spotting this and suggestion the change in https://github.com/wordpress-mobile/gutenberg-mobile/pull/4469#discussion_r785089429 🙇 .

### Clean command recreates i18n cache
1. Run command `npm run clean`.
2. Observe that the folder `src/i18n-cache` is removed when the command is being executed.
3. Observe that the folder `src/i18n-cache` is created again.

### JS bundle is generated properly
1. Run command `npm run bundle`.
2. Observe that the command is executed properly without errors.

PR submission checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -89,7 +89,7 @@
 		"build:gutenberg": "cd gutenberg && npm run build",
 		"clean": "npm run core clean; npm run clean:gutenberg; npm run clean:i18n",
 		"clean:gutenberg": "cd gutenberg && npm run clean:packages",
-		"clean:i18n": "rm -rf src/i18n-cache",
+		"clean:i18n": "rm -rf src/i18n-cache && npm run i18n:check-cache",
 		"lint": "eslint . --ext .js",
 		"lint:fix": "npm run lint -- --fix",
 		"version": "npm run bundle && git add -A bundle"


guarani (approved):
Thanks for the fix here, @fluiddot! 

The test cases worked properly on my machine; however, the `npm run bundle` command fails when the bundle is not present. For example:
```
rm -rf bundle
npm run bundle
```
fails (I unfortunately didn't capture the error, but will add it here if I get the chance to re-run today).

Since this fixes the most common case where a bundle exists and it is re-generated, I think it's good to approve and merge, which I'll do now.

fluiddot:
> The test cases worked properly on my machine; however, the `npm run bundle` command fails when the bundle is not present. For example:
> 
> ```
> rm -rf bundle
> npm run bundle
> ```
> 
> fails (I unfortunately didn't capture the error, but will add it here if I get the chance to re-run today).

@guarani I think removing the `bundle` folder is not expected nor required when generating a bundle, as that folder is currently versioned. In what case do you think we would need to remove it?

guarani:
> @guarani I think removing the `bundle` folder is not expected nor required when generating a bundle, as that folder is currently versioned. In what case do you think we would need to remove it?

It's my understanding that `npm run bundle` – which updates (or creates) the bundle - used to handled the scenario where the bundle is not present (please correct me if I'm wrong here) and this changed recently.

It's often the case when debugging an issue that we delete build artifacts, derived data, etc and I think that includes the bundle. (The only reason I'm aware of that we check-in the bundle is because we don't want the main apps to have the tooling needed to build it.)

If for example, I had an issue where the main apps didn't seem to pick up an updated bundle, personally I probably would delete the bundle and generate it again to make sure nothing was being cached. If I was unaware of how `npm run bundle` generates an error when a bundle isn't present, I might be slowed down looking into the error.

guarani:
Do you happen to know why the bundle command would require an existing bundle?

fluiddot:
Ok, I see your point and totally agree with the fact that deleting the bundle shouldn't produce issues on the generation command, as it's an artifact only required by the integration with the main apps 👍 . In fact, I don't think that the bundle or any asset file contained within that folder (i.e. `bundle`) is required by any command 🤔 , I'll do a quick check to identify the script that is failing.

fluiddot:
> Do you happen to know why the bundle command would require an existing bundle?

Looks like it's failing in the scripts that generate the localization strings files ([reference](https://github.com/wordpress-mobile/gutenberg-mobile/blob/628e86fb4cb3ec3d15d0dc2edd75ce834b2fca0a/bin/strings2android.js#L191)). However, the cause is just the fact they expect the `bundle` folder to present, so this could be easily addressed by assuring that the folder is created in case it doesn't exist, I'll prepare a PR to correct this 🔧 .

fluiddot:
@guarani I've just created [this PR](https://github.com/wordpress-mobile/gutenberg-mobile/pull/4478), in order to address this issue, and assigned you as a reviewer. Let me know if you could take a look, thanks 🙇 !

guarani:
> Looks like it's failing in the scripts that generate the localization strings files ([reference](https://github.com/wordpress-mobile/gutenberg-mobile/blob/628e86fb4cb3ec3d15d0dc2edd75ce834b2fca0a/bin/strings2android.js#L191)). However, the cause is just the fact they expect the `bundle` folder to present, so this could be easily addressed by assuring that the folder is created in case it doesn't exist, I'll prepare a PR to correct this 🔧 .

Thanks for pinpointing the cause there 🙇 




> @guarani I've just created [this PR](https://github.com/wordpress-mobile/gutenberg-mobile/pull/4478), in order to address this issue, and assigned you as a reviewer. Let me know if you could take a look, thanks 🙇 !

Sure, will take a look, thanks!

PR #4557, opened 2022-02-08
Prevent bold headings after autocorrection

twstokes:
Fixes #4063

## Related PRs:
- https://github.com/WordPress/gutenberg/pull/38633
- https://github.com/wordpress-mobile/AztecEditor-iOS/pull/1334
- https://github.com/wordpress-mobile/WordPress-iOS/pull/17844
- https://github.com/wordpress-mobile/test-cases/pull/111

### Heading with no formatting

1. In the demo app, add a new heading. Ensure no formatting is applied (the default settings).
2. Type a sentence that will be autocorrected, such as: "Hello i'm testing"

**Expectations:**
-  The sentence is autocorrected (e.g. "Hello I'm testing").
- The autocorrected word (e.g. "I'm") does not have bold formatting enabled (the B in the formatting toolbar is not depressed).
- When placing the cursor inside the autocorrected word it does not have bold formatting enabled.
- In HTML mode the autocorrected word is not wrapped in `<strong>` tags:

```
<!-- wp:heading -->
<h2>Hello I’m testing</h2>
<!-- /wp:heading -->
```

### Heading with bold formatting

1. In the demo app, add a new heading and apply bold formatting by pressing the **B** on the formatting toolbar.
2. Type a sentence that will be autocorrected, such as: "Hello i'm testing"

**Expectations:**
-  The sentence is autocorrected (e.g. "Hello I'm testing").
- The autocorrected word (e.g. "I'm") has bold formatting enabled (the B in the formatting toolbar is depressed).
- When placing the cursor inside the autocorrected word it has bold formatting enabled.
- In HTML mode the heading's text content is wrapped in `<strong>` tags:

```
<!-- wp:heading -->
<h2><strong>Hello I’m testing a heading</strong></h2>
<!-- /wp:heading -->
```

PR submission checklist:

- [x] I have considered adding unit tests where possible.
- [x] I have considered if this change warrants user-facing release notes [more info](https://github.com/wordpress-mobile/gutenberg-mobile/blob/develop/Release-notes.md) and have added them to [RELEASE-NOTES.txt](https://github.com/wordpress-mobile/gutenberg-mobile/blob/develop/RELEASE-NOTES.txt) if necessary.

PR #4585, opened 2022-02-15
Generate bundle to apply fix from "Prevent headings from becoming bold after words are autocorrected"

twstokes:
This PR contains the result of running `npm run bundle` to generate the latest bundle from `develop` due to the following PR:
- https://github.com/wordpress-mobile/gutenberg-mobile/pull/4557

This is needed so we can [bump the GB Mobile version](https://github.com/wordpress-mobile/WordPress-iOS/pull/17844) for WPiOS.


twstokes:
After this is merged we'll need to create a tag for GB Mobile so that WPiOS can point to it.

fluiddot (approved):
Approved via https://github.com/wordpress-mobile/WordPress-iOS/pull/17844#pullrequestreview-884519944.

Nitpick: I'm wondering if we should update the PR's title to mention that the bundle was generated to apply the fix "Prevent headings from becoming bold after words are autocorrected", wdyt?

twstokes:
> Nitpick: I'm wondering if we should update the PR's title to mention that the bundle was generated to apply the fix "Prevent headings from becoming bold after words are autocorrected", wdyt?

Sure! I'd kept it generic.

PR #4645, opened 2022-03-11
Note to run the full local testsuite at least once

hypest:
If one tries to run specific Appium tests locally without having run the full testsuite first, the app binary is not built and the run fails with:

```
An unknown server-side error occurred while processing the command. Original error: Bad app: .../gutenberg/packages/react-native-editor/ios/build/GutenbergDemo/Build/Products/Release-iphonesimulator/GutenbergDemo.app. App paths need to be absolute or an URL to a compressed app file: The application at '.../gutenberg/packages/react-native-editor/ios/build/GutenbergDemo/Build/Products/Release-iphonesimulator/GutenbergDemo.app' does not exist or is not accessible
```

This PR adds a small note to the README to nudge to run the full testsuite first since those npm scripts will also build the app first.

PR submission checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,6 +136,8 @@ and for Android:
 
 `npm run test:e2e:android:local`
 
+**Note:** Make sure you've run the above commands at least once, so the demo app binaries are built before running individual tests below.
+
 To run a single test instead of the entire suite, use `npm run device-tests:local`. Here's an example that runs only `gutenberg-editor-paragraph.test`:
 
 `TEST_RN_PLATFORM=ios npm run device-tests:local gutenberg-editor-paragraph.test`


README.md
```diff
@@ -136,6 +136,8 @@ and for Android:
 
 `npm run test:e2e:android:local`
 
+**Note:** Make sure you've run the above commands at least once, so the demo app binaries are built before running individual tests below.
+
```
fluiddot, line 140: Alternatively, we could provide the commands to generate the demo app binaries:
- Android: `npm run test:e2e:bundle:android && npm run core test:e2e:build-app:android`
- iOS: `npm run test:e2e:bundle:ios && npm run core test:e2e:build-app:ios`

fluiddot (approved):
LGTM 🎊 , thanks @hypest for adding the helpful note.

README.md
hypest: Thought about that but felt it's becoming a bit verbose since we need to mention both platforms and the commands are currently lengthy. And to be honest, I feel that the build scripts should be run automatically before the test run (but I think we were not doing that originally because it takes time).

No strong feelings on this one though, perhaps it's still best to add the commands to the note, wdyt?

fluiddot: No strong feelings either from my side, and I agree about preventing making it too verbose. Let's stick with the note without the commands.

PR #4352, opened 2021-12-08
Hide help button from UBE

SiobhyB:
Addresses https://github.com/wordpress-mobile/gutenberg-mobile/issues/4339

## Related PRs and Installable Builds

* `Gutenberg`: https://github.com/WordPress/gutenberg/pull/37221
* `WPiOS`: https://github.com/wordpress-mobile/WordPress-iOS/pull/17644
* `WPAndroid`: https://github.com/wordpress-mobile/WordPress-Android/pull/15674

## Description

As described in https://github.com/wordpress-mobile/gutenberg-mobile/issues/4339, a help button is currently displaying in the bottom right-hand corner of the unsupported block editor (UBE) in both the Android and iOS apps. The button can be seen while logged into WordPress.com. When tapped on, the button isn't working as expected on iOS and prompts users to navigate away from the editor on Android, creating the potential for edits to be lost.

With this PR, the button is hidden from view in order to avoid confusion and potential loss of edits. A high-level overview of the code changes can be found in the `Types of changes` section of [the Gutenberg PR](https://github.com/WordPress/gutenberg/pull/37221).

## Testing

Please refer to [the Gutenberg PR](https://github.com/WordPress/gutenberg/pull/37221) as the "central PR" with the most up-to-date testing instructions. 

<hr />

PR submission checklist:

- [ ] I have considered adding unit tests where possible.
- [ ] I have considered if this change warrants user-facing release notes [more info](https://github.com/wordpress-mobile/gutenberg-mobile/blob/develop/Release-notes.md) and have added them to [RELEASE-NOTES.txt](https://github.com/wordpress-mobile/gutenberg-mobile/blob/develop/RELEASE-NOTES.txt) if necessary.

PR #4820, opened 2022-05-03
Remove WordPress.com tour banner from view in UBE

### LINKED ISSUE #4079: Unwanted popup shown in Unsupported Block Editor
**Describe the bug**
When editing an unsupported block, a "Welcome to WordPress!" popup is shown, which shouldn't appear.

**To Reproduce**
_This may be only reproducible on a new site, since the popup might only be shown once._

1. Locate a post that contains an unsupported block (e.g. Calendar at time of writing)
2. Open the post and expect to see the block rendered as a placeholder with the text "Unsupported"
3. Tap the `(?)` icon and expect the option to edit the block in a web browser to be shown
4. Tap the edit in a web browser button and expect the block to be shown, but notice that a "Welcome to WordPress!" popup is visible

**Expected behavior**
At step 4 above, the popup should not be there

**Screenshots**

Screenshot showing unwanted popup visible in the Unsupported Block Editor 

 - Device: Samsung Galaxy S10
 - OS: Android 11
 - WPAndroid 18.4

**Additional context**

I haven't tested WPiOS, but I'm very sure the same issue would be present there.

SiobhyB:
Fixes https://github.com/wordpress-mobile/gutenberg-mobile/issues/4079

## What?

This PR removes the WordPress.com tour banner that sometimes pops up when opening the unsupported block editor (UBE).

| Before  | After |
| ------------- | ------------- |
|  |  |

_Note, the cookie banner remains as I'm assuming it'd be necessary to get the thumbs up from legal before hiding it and it's not as disruptive as the tour banner._

## Why?

The tour banner is specific to the web version of the editor, it's therefore confusing and distracting within the UBE.

## How?

Previously, as part of the changes in https://github.com/wordpress-mobile/gutenberg-mobile/pull/4352, [the `external-style-overrides.css` file](https://github.com/wordpress-mobile/gutenberg-mobile/blob/3fbea3f96c8fdf83cea89cae08c088d687ab5023/resources/unsupported-block-editor/external-style-overrides.css) was created for the purpose of "housing" CSS that didn't belong in the Gutenberg repository. It has a Gutenberg equivalent [here](https://github.com/WordPress/gutenberg/blob/da176fa9cbd19c3ed04be993563a3229063442e7/packages/react-native-bridge/common/gutenberg-web-single-block/editor-style-overrides.css).

Following the approach we've taken to hide other elements in the UBE, this PR adds CSS to the `external-style-overrides.css` that targets the tour banner and hides it from view.

## Testing Instructions

_This may be only reproducible on a new WordPress.com site, since the popup might only be shown once._

1. Locate a post that contains an unsupported block (e.g. Calendar at time of writing)
2. Open the post and expect to see the block rendered as a placeholder with the text "Unsupported"
3. Tap the (?) icon and expect the option to edit the block in a web browser to be shown
4. Tap the edit in a web browser button and expect the block to be shown, confirm that the "Welcome to WordPress!" banner is no longer visible

<hr />

PR submission checklist:

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -2,7 +2,6 @@ Unreleased
 ---
 * [**] [Cover block] Improve color contrast between background and text [https://github.com/wordpress-mobile/gutenberg-mobile/pull/4808]
 * [**] [Buttons block] Fix Android-only issue related to displaying formatting buttons after closing the block settings [https://github.com/wordpress-mobile/gutenberg-mobile/pull/4814]
-
 * [*] [Unsupported Block Editor] Prevent WordPress.com tour banner from displaying. [https://github.com/wordpress-mobile/gutenberg-mobile/pull/4820]
 
 ## 1.75.0
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -1,6 +1,8 @@
 Unreleased
 ---
 
+* [*] [Unsupported Block Editor] Prevent WordPress.com tour banner from displaying. [https://github.com/wordpress-mobile/gutenberg-mobile/pull/4820]
+
 ## 1.75.0
 ---
 * [*] [Latest Posts block] Add featured image settings [https://github.com/WordPress/gutenberg/pull/39257]
--- a/resources/unsupported-block-editor/external-style-overrides.css
+++ b/resources/unsupported-block-editor/external-style-overrides.css
@@ -4,6 +4,6 @@
 }
 
 /* Remove WordPress.com "tour" pop-up. */
-.wpcom-tour-kit .tour-kit-frame__container {
+.tour-kit-frame__container {
 	display: none;
 }
--- a/resources/unsupported-block-editor/external-style-overrides.css
+++ b/resources/unsupported-block-editor/external-style-overrides.css
@@ -2,3 +2,8 @@
 .a8c-faux-inline-help {
 	display: none;
 }
+
+/* Remove WordPress.com "tour" pop-up. */
+.wpcom-tour-kit .tour-kit-frame__container {
+	display: none;
+}


mchowning:
👋  @SiobhyB ! I'm having some trouble testing this. When I create a new free WordPress.com site, I'm not seeing any "Welcome to WordPress" banner either on the web when I start a new post or in the app (without the changes from this PR applied) when I edit an unsupported block.  I know I've seen it before, but I don't know if there's something I'm doing wrong now to keep it from showing up. Any ideas?

SiobhyB:
Thanks for taking a look at this, @mchowning! I did some digging and found there's an option to enable the Welcome Guide under the editor's settings on the web:

Screenshot 2022-05-06 at 11 30 16

I think if you enable it there, you should then be able to see it within the UBE on the app. Would you be able to give that a try to see if it works for you?

mchowning (approved):
Thanks for this fix @SiobhyB . I updated the gb-mobile reference in WPAndroid and saw that this change works perfectly.

If I use metro though (instead of the bundle), then the display property doesn't get applied and the popup still appears. I'm not sure exactly why that is, but I noticed you had to make a tweak to get this to work in the bundle—maybe that made it so that it didn't work in metro? Anyway, I'll go ahead and approve since it is the bundle that is the most important. If there were a way to get this to work both with metro and the bundle that would be great though.

SiobhyB:
Thanks @mchowning! 🙌 

> If I use metro though (instead of the bundle), then the display property doesn't get applied and the popup still appears.

I only able to get the fix working with metro if I restarted the activity and re-ran the app in Android, it didn't automatically update like other changes to Gutenberg do (I think that's expected with UBE changes). Perhaps that's related? I'll go ahead to merge the PR as I believe it will work as expected with metro with a fresh install/restart. Let me know if there's still problems that I'm missing, though.

PR #4821, opened 2022-05-03
Fix pnpm version issue

jhnstn:
Fixes npm post install script by forcing `npx` to use `pnpm` at `v6`.

Jetpack currently uses `pnpm` at `^6.23.6` and `pnpm` recently released `7.0.0` which `npx` will install.  

To test:
- Verify that `npm install` succeeds
- CI "Check Correctness" step succeeds

PR submission checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,7 +53,7 @@
 		"wd": "^1.11.1"
 	},
 	"scripts": {
-		"postinstall": "patch-package && npm ci --prefix gutenberg && npm run i18n:check-cache && cd jetpack/projects/plugins/jetpack && npx pnpm install",
+		"postinstall": "patch-package && npm ci --prefix gutenberg && npm run i18n:check-cache && cd jetpack/projects/plugins/jetpack && npx pnpm@6 install",
 		"start": "echo \"\\x1b[33mThe start command is not available in this project. It is strongly recommended to use \\x1b[1:33mstart:reset\\x1b[0m\\x1b[33m to perform some cleanup when starting the metro bundler.\nOr you may use \\x1b[1:33mstart:quick\\x1b[0m\\x1b[33m for a quicker startup, but this may lead to unexpected javascript errors when running the app.\\x1b[0m\"",
 		"start:reset": "npm run core clean:runtime && npm run start:quick -- --reset-cache",
 		"start:quick": "react-native start --config ./metro.config.js",


mchowning (approved):
Looks good and is working well. Thanks for the quick fix on this @jhnstn ! 🙇

package.json
```diff
@@ -53,7 +53,7 @@
 		"wd": "^1.11.1"
 	},
 	"scripts": {
-		"postinstall": "patch-package && npm ci --prefix gutenberg && npm run i18n:check-cache && cd jetpack/projects/plugins/jetpack && npx pnpm install",
+		"postinstall": "patch-package && npm ci --prefix gutenberg && npm run i18n:check-cache && cd jetpack/projects/plugins/jetpack && npx pnpm@6 install",
```
hypest, line 56: I wonder if we can automate this a bit (in a follow up PR) by reading the pnpm version from the Jetpack repo:

```
npx pnpm@`node -e "console.log(require('./jetpack/package.json').engines.pnpm);"`
```

fluiddot: I think the version calculation might be a bit more complex because the value can be a range ([reference](https://docs.npmjs.com/cli/v7/configuring-npm/package-json#engines)). In fact, I notice that the current value is `^6.32.3` so we'd need to parse it. In relation to this, I think the [`semver` package](https://www.npmjs.com/package/semver) might help us with this. Specifically, the function `minVersion` returns the minimum version of a range, hence we could use it to calculate the version.

jhnstn: I had a similar concern about automating the version but turns out `npx` can handle the ranges (TIL) :

```
> npx pnpm@"^6.32.3 || ~7.0.1" --version
6.32.11
```
seems to handle anything that is allowed in `package.json` configs

I also found a slightly cleaner way to get the pnpm verison:

```
> npx -c 'echo "$npm_package_engines_pnpm"'
```

fluiddot: Oh, that's very interesting, I didn't know it accepted ranges (TIL) 😅 . In that case, it really simplifies this approach as we can use the raw value from `package.json`.

> I also found a slightly cleaner way to get the pnpm verison:
> 
> ```
> > npx -c 'echo "$npm_package_engines_pnpm"'
> ```

Awesome 👏.

hypest: 👍

PR #4830, opened 2022-05-05
Add Jetpack install script

jhnstn:
Fixes changes to Jetpack package hoisting introduced in https://github.com/Automattic/jetpack/commit/ef3305b95be354cb1e1a27327309483d354a4852

Setting up hoisting patterns in the project `.npmrc` file prevents CI from succeeding in this repo. This removes the hosting configuration from `.npmrc` when installing Jetpack on the CI. 

This also forces `npx` to use the same version of `pnpm` Jetpack is using (h/t @hypest )

To test:
- Verify CI passes 
- Verify `npm install` succeeds locally

PR submission checklist:


bin/install-jetpack.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
```
jhnstn, line 1: I'm not convinced about the name `install-jetpack` , perhaps something like `run-jetpack-install` would be better.

bin/install-jetpack.sh
```diff
@@ -0,0 +1,22 @@
...
+set -Eeuo pipefail
+
+pushd jetpack
+
+pnpm_version=$(npx -c 'echo "$npm_package_engines_pnpm"')
+
+# Remove pnpm hoisting config if on CI
+if [[ -n "${CI:-}" ]]; then
+  sed -i.bk '/hoist/d' .npmrc
```
jhnstn: This is the extreme option: reset the hoisting configuration back to the pnpm defaults. We might be able to go lighter and keep the Jetpack configuration but add what ever we need to build, that is we might only need to add a few `hoist-pattern[]=*@something/something*` calls.

jhnstn: Also, I tried overriding the npm configuration via the command line but from what I found the project level configs in `.npmrc` are not overridable from global or user npm configs. The only option is to modify the .npmrc file.

bin/install-jetpack.sh
```diff
@@ -0,0 +1,22 @@
...
+fi
+
+cd projects/plugins/jetpack
+npx pnpm@"$pnpm_version" install
+cd -
+
+# Retore .npmrc if on CI
+if [[ -n "${CI:-}" ]]; then
+  mv -f .npmrc.bk .npmrc
```
jhnstn: We need to restore the .npmrc file back to HEAD or the correctness step fails due to the Jetpack repo being "dirty"

bin/install-jetpack.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+set -Eeuo pipefail
+
+pushd jetpack
+
+pnpm_version=$(npx -c 'echo "$npm_package_engines_pnpm"')
+
+# Remove pnpm hoisting config if on CI
+if [[ -n "${CI:-}" ]]; then
```
jhnstn: Both [circleci](https://circleci.com/docs/2.0/env-vars/#built-in-environment-variables) and[ buildkite](https://buildkite.com/docs/pipelines/environment-variables#bk-env-vars-ci) export a `CI` environment variable.
As far as I can tell this is only a problem on CI, I've tried re-initializing the `jetpack` submodule and ran `npm install ` from the `./gutenberg-mobile` and everything seems ok. This part needs a bit more local testing to make sure it's just a CI issue.

PR #4833, opened 2022-05-05
Track block movement actions

### LINKED ISSUE #4831: [Drag & Drop blocks] Track block movement actions
- [x] Add an event related to dropping a block
- [x] Add events related to moving up/down a block

geriux:
Fixes https://github.com/wordpress-mobile/gutenberg-mobile/issues/4831

- `WordPress iOS PR` -> https://github.com/wordpress-mobile/WordPress-iOS/pull/18530
- `WordPress Android PR` -> https://github.com/wordpress-mobile/WordPress-Android/pull/16485

**Note**: The bundle and Gutenberg reference changes are included to be able to test the drag & drop event, those will not be included once this is approved and ready to be merged.

This PR adds a new event `editor_block_moved` that is triggered when a block is moved by either:
- Tapping on the up/down arrow button in the block toolbar.
- Long-pressing on the up/down arrow button and moving a block to the top or bottom.
- Drag & drop

The event has the following properties:
- `action_source` which can be:
  - `move_arrows_up` when a block is moved using the up arrow button.
  - `move_arrows_down` when a block is moved using the down arrow button.
  - `move_arrows_to_top` when a block is moved to the top by long-pressing on the up arrow button.
  - `move_arrows_to_bottom` when a block is moved to the bottom by long-pressing on the down arrow button.
  - `drag_and_drop` when a block is moved using the drag & drop blocks feature.
- `block_name` name of the block that is being moved.
- `inner_block` if it has inner blocks or not.

Once this is merged, it will only track when a block is moved using the up/down arrow buttons, since the drag & drop blocks feature is not merged yet.

## To test

**Builds**:
- [iOS](https://github.com/wordpress-mobile/WordPress-iOS/pull/18530#issuecomment-1119701117)
- [Android](https://github.com/wordpress-mobile/WordPress-Android/pull/16485#issuecomment-1119691299)

### Move a block by tapping on the up/down arrow buttons

- Open the app using the builds linked above.
- Create a new post/page
- Add some blocks
- Select a block and tap the up/down arrow buttons within the block toolbar.
- Using the live event page, **expect** the event `wp(ios|android)_editor_block_moved` to arrive successfully with the properties `action_source` containing: `move_arrows_(up|down)`, `inner_block` and, `block_name`.

### Move a block by long-pressing on the up/down arrow buttons

- Open the app using the builds linked above.
- Create a new post/page
- Add some blocks
- Select a block and long-press on the up/down arrow buttons within the block toolbar.
- Select either `Move to the top` or `Move to the bottom`.
- Using the live event page, **expect** the event `wp(ios|android)_editor_block_moved` to arrive successfully with the properties `action_source` containing: `move_arrows_to_(top|bottom)`, `inner_block` and, `block_name`.

### Move a block using the drag & drop blocks feature

- Open the app using the builds linked above.
- Create a new post/page
- Add some blocks
- Drag & drop a block
- Using the live event page, **expect** the event `wp(ios|android)_editor_block_moved` to arrive successfully with the properties `action_source` containing: `drag_and_drop, `inner_block` and, `block_name`.

PR submission checklist:

### DIFF
--- a/src/analytics/redux/tracked_events.js
+++ b/src/analytics/redux/tracked_events.js
@@ -64,6 +64,51 @@ function trackBlocksHandler(
 	} );
 }
 
+/**
+ * Helper function to track block movement events.
+ *
+ * @param {string[]} clientIds The client IDs of the blocks being moved
+ * @param {string}   action    Type of movement (arrows/drag & drop)
+ * @return {void}
+ */
+function trackBlockMoved( clientIds, action ) {
+	const block = select( 'core/block-editor' ).getBlock( clientIds?.[ 0 ] );
+
+	if ( block ) {
+		const eventProperties = {
+			action_source: action,
+			inner_block: !! block.innerBlocks.length,
+			block_name: block.name,
+		};
+
+		sendEventToHost( 'editor_block_moved', eventProperties );
+	}
+}
+
+/**
+ * Helper function to handle when a block is moved by an index position
+ *
+ * @param {string[]} clientIds The client IDs of the blocks being moved
+ * @param {number}   toIndex   Index of the new position of the block
+ * @return {void}
+ */
+function handleBlockMovedByPosition( clientIds, toIndex ) {
+	const lastBlockIndex = select( 'core/block-editor' ).getBlockCount() - 1;
+	const currentBlockIndex = select( 'core/block-editor' ).getBlockIndex(
+		clientIds?.[ 0 ]
+	);
+
+	if ( currentBlockIndex >= 0 ) {
+		if ( toIndex > currentBlockIndex && toIndex === lastBlockIndex ) {
+			// The block was moved to the bottom of the editor
+			trackBlockMoved( clientIds, 'move_arrows_to_bottom' );
+		} else if ( toIndex < currentBlockIndex && toIndex === 0 ) {
+			// The block was moved to the top of the editor
+			trackBlockMoved( clientIds, 'move_arrows_to_top' );
+		}
+	}
+}
+
 export const trackedEvents = {
 	'core/block-editor': {
 		insertBlock( blocks ) {
@@ -79,6 +124,28 @@ export const trackedEvents = {
 				'editor_block_inserted',
 				( { name } ) => ( { block_name: name } )
 			);
+		},
+		moveBlocksUp( clientIds ) {
+			trackBlockMoved( clientIds, 'move_arrows_up' );
+		},
+		moveBlocksDown( clientIds ) {
+			trackBlockMoved( clientIds, 'move_arrows_down' );
+		},
+		moveBlocksToPosition(
+			clientIds,
+			_fromRootClientId,
+			_toRootClientId,
+			toIndex
+		) {
+			const isDraggingBlock = select(
+				'core/block-editor'
+			).isDraggingBlocks();
+
+			if ( isDraggingBlock ) {
+				trackBlockMoved( clientIds, 'drag_and_drop' );
+			} else {
+				handleBlockMovedByPosition( clientIds, toIndex );
+			}
 		},
 	},
 };


src/analytics/redux/tracked_events.js
```diff
@@ -64,6 +64,20 @@ function trackBlocksHandler(
...
+ * @param {string} action Type of movement (arrows/drag & drop)
+ * @return {void}
+ */
+function trackBlockMoved( action ) {
+	const eventProperties = {
+		action_source: action,
+	};
+
+	sendEventToHost( 'editor_block_moved', eventProperties );
+}
```
fluiddot: It would be interesting to collect more information like the type of block being moved or if it has inner blocks, similar to what we do in the above `trackBlocksHandler` function. For the purpose of measuring the drag & drop blocks feature this information is enough, so we could apply this suggestion in the future and create a follow-up issue for this.

fluiddot (commented):
Amazing job @geriux 🏅 !

I confirm that the editor is sending `editor_block_moved` events for both actions of tapping on up/down arrows and drag & drop 🎊 .

However, I noticed that moving a block by long-pressing over the up/down arrows and tapping on "Move to Top" or "Move to bottom" is not generating an event. Not sure if these options are heavily used by users, but it might lead to receiving fewer moving block events. Probably not a critical issue but before approving the PR I'd like to double-check this with you @geriux.

geriux:
> However, I noticed that moving a block by long-pressing over the up/down arrows and tapping on "Move to Top" or "Move to bottom" is not generating an event. Not sure if these options are heavily used by users, but it might lead to receiving fewer moving block events. Probably not a critical issue but before approving the PR I'd like to double-check this with you

Nice catch! I think we should track those as well, should we include them in the action `move_arrows` or maybe it'd be interesting to track them separately? like `move_arrows_to_top`, `move_arrows_to_bottom`. What do you think? @fluiddot

fluiddot:
> > However, I noticed that moving a block by long-pressing over the up/down arrows and tapping on "Move to Top" or "Move to bottom" is not generating an event. Not sure if these options are heavily used by users, but it might lead to receiving fewer moving block events. Probably not a critical issue but before approving the PR I'd like to double-check this with you
> 
> Nice catch! I think we should track those as well, should we include them in the action `move_arrows` or maybe it'd be interesting to track them separately? like `move_arrows_to_top`, `move_arrows_to_bottom`. What do you think? @fluiddot

Good question, I'd lean towards the option that collects more information (i.e. tracking them separately). However, this makes me think if we should track the up and down movements separately too, wdyt?

**Potential sources of block movement action:**
- `drag_and_drop`
- `move_arrows_up`
- `move_arrows_down`
- `move_arrows_to_top`
- `move_arrows_to_bottom`

geriux:
> However, this makes me think if we should track the up and down movements separately too, wdyt?

I agree! I'll update them. Thank you!

src/analytics/redux/tracked_events.js
geriux: Nice suggestion, I ended up adding it here https://github.com/wordpress-mobile/gutenberg-mobile/pull/4833/commits/c21339900b1e81f58db3f693dd0e5a8d554cc758#diff-b264ddd07fecd2f9655557832fb8f89baa62f7f89c387e80df664c26ceb36084R98-R99

geriux:
Hey @fluiddot 👋 this is ready for another review 😃, the Android build is ready. The iOS build is still in progress.

src/analytics/redux/tracked_events.js
```diff
@@ -94,14 +142,29 @@ export const trackedEvents = {
...
+			clientIds,
+			_fromRootClientId,
+			_toRootClientId,
+			toIndex
+		) {
+			if ( isDraggingBlock ) {
+				trackBlockMoved( clientIds, 'drag_and_drop' );
+			} else {
+				handleBlockMovedByPosition( clientIds, toIndex );
+			}
```
fluiddot, line 148: Alternatively, we could check if any block is being dragged via [the `isDraggingBlocks` selector](https://github.com/WordPress/gutenberg/blob/681cde287a2c4907fc74e0347ccfb4b55d0f8ff1/packages/block-editor/src/store/selectors.js#L1284-L1293). This way we wouldn't need to keep the `isDraggingBlock` value and update it by listening to `startDraggingBlocks` and `stopDraggingBlocks`.

src/analytics/redux/tracked_events.js
```diff
@@ -64,20 +71,61 @@ function trackBlocksHandler(
...
 
+/**
+ * Helper function to handle when a block is moved by an index position
+ *
+ * @param {string[]} clientIds The client IDs of the blocks being moved
+ * @param {number}   toIndex   Index of the new position of the block
+ * @return {void}
+ */
+function handleBlockMovedByPosition( clientIds, toIndex ) {
+	const numberOfBlocks = select( 'core/block-editor' ).getBlockCount() - 1;
```
fluiddot: Nitpick: I understand that this constant holds the last block index instead of the number of blocks, so I'm wondering if we should rename it to something like `lastBlockIndex`, wdyt?

geriux:
> The iOS build is still in progress

Unfortunately, there are some issues with the CI job that creates an installable build, since it might take a while, would you be able to test it using a local build?

fluiddot:
> > The iOS build is still in progress
> 
> Unfortunately, there are some issues with the CI job that creates an installable build, since it might take a while, would you be able to test it using a local build?

Sure thing, I'll test it locally 👍.

src/analytics/redux/tracked_events.js
```diff
@@ -64,20 +71,61 @@ function trackBlocksHandler(
...
- * @param {string} action Type of movement (arrows/drag & drop)
+ * @param {string[]} clientIds The client IDs of the blocks being moved
+ * @param {string}   action    Type of movement (arrows/drag & drop)
  * @return {void}
  */
-function trackBlockMoved( action ) {
+function trackBlockMoved( clientIds, action ) {
+	const block = select( 'core/block-editor' ).getBlocksByClientId(
+		clientIds?.[ 0 ]
+	)?.[ 0 ];
```
fluiddot: If we only want to retrieve the data of the first block, we might consider using [the `getBlock` selector](https://github.com/WordPress/gutenberg/blob/9e1ad7c15c43c7d4aa7257cf6fdeaad15431bc74/packages/block-editor/src/store/selectors.js#L124-L152) and pass the first index of `clientIds`. In fact, that selector is the one used within the `getBlocksByClientId` selector ([reference](https://github.com/WordPress/gutenberg/blob/9e1ad7c15c43c7d4aa7257cf6fdeaad15431bc74/packages/block-editor/src/store/selectors.js#L333-L336)).

fluiddot (commented):
I tested the latest changes and look great to me, however, I noticed that the `drag_and_drop` event is being only sent in some cases when dragging and dropping the block in its original location (see attached video capture). Upon further investigation, I see that depending on if the block is dropped before and after, it sends the event:
- **Dropped before - Doesn't send the event.**
- **Dropped after - Sends the event.**

https://user-images.githubusercontent.com/14905380/167656494-f6d6c364-af5a-4ee5-84ab-a49e60dab826.mp4

src/analytics/redux/tracked_events.js
geriux: Nice suggestion! I'll update it, thanks!

src/analytics/redux/tracked_events.js
geriux: It actually stores the number of blocks to check if the `toIndex` value matches with it, that way we know if it was moved to the very end.

fluiddot: Yep, I meant that the name `numberOfBlocks` might not be accurate as it doesn't hold the number of blocks because it's subtracting 1 to the result of `getBlockCount`.

src/analytics/redux/tracked_events.js
geriux: Updated in https://github.com/wordpress-mobile/gutenberg-mobile/pull/4833/commits/2037f66df75d33a68292e074e30fe4d1b053e7f6

src/analytics/redux/tracked_events.js
geriux: Thanks for the suggestion! I updated in https://github.com/wordpress-mobile/gutenberg-mobile/pull/4833/commits/c8b88ce25a9f738d52f5c2e955983fb0c42415e1

geriux:
> Upon further investigation, I see that depending on if the block is dropped before and after, it sends the event:

Thanks for spotting the issue, when dragging the block right above and staying in the same position doesn't call `moveBlocksToPosition` when dropping because the index value is the same, when dragging right after it does since the index value is `currentIndex + 1`.

I feel it'd be an issue if we want to know if drag & drop was enabled in all cases, regardless if a block was actually moved. With the current approach, it will only be tracked if the block was actually moved. I think in this case, a "bug" would make sense but let me know if we should still track it anyways 😄

src/analytics/redux/tracked_events.js
geriux: Ahhh I see now 😅 I'll update it 👍

geriux: Updated in https://github.com/wordpress-mobile/gutenberg-mobile/pull/4833/commits/fbd41a26288f97e04093b9dea15cee822d075f7c

src/analytics/redux/tracked_events.js
```diff
@@ -89,9 +72,7 @@ function setIsDraggingBlock( isDragging ) {
  * @return {void}
  */
 function trackBlockMoved( clientIds, action ) {
-	const block = select( 'core/block-editor' ).getBlocksByClientId(
-		clientIds?.[ 0 ]
-	)?.[ 0 ];
+	const block = select( 'core/block-editor' ).getBlock( clientIds?.[ 0 ] );
```
fluiddot, line 75: If we expect that in some cases the `clientIds` is not defined (in relation to the statement `clientIds?.[0]`), probably we should check if the block returned is defined and prevent sending the event, wdyt?

fluiddot:
> > Upon further investigation, I see that depending on if the block is dropped before and after, it sends the event:
> 
> Thanks for spotting the issue, when dragging the block right above and staying in the same position doesn't call `moveBlocksToPosition` when dropping because the index value is the same, when dragging right after it does since the index value is `currentIndex + 1`.

Ok, I see, so in this case,  the block is actually being moved but to the same origin position.

> I feel it'd be an issue if we want to know if drag & drop was enabled in all cases, regardless if a block was actually moved. With the current approach, it will only be tracked if the block was actually moved. I think in this case, a "bug" would make sense but let me know if we should still track it anyways 😄

Right, probably the issue here is more related to the fact that the block is being moved than actually sending the event. Hence, the event logic is behaving as expected.

Regarding the issue related to preventing the block to be moved if the target location is the same as the origin, I think we could open a new issue as a follow-up.

src/analytics/redux/tracked_events.js
geriux: Updated in https://github.com/wordpress-mobile/gutenberg-mobile/pull/4833/commits/64337fa237e45ecd20ad4c16bb2020876f3f5dbb

geriux:
> Regarding the issue related to preventing the block to be moved if the target location is the same as the origin, I think we could open a new issue as a follow-up.

Well, currently it is intended to not move a block if it's in the same position, as you can see [here](https://github.com/WordPress/gutenberg/blob/f7a9e516eb43170a2d3723c633b39cd6dc93e16c/packages/block-editor/src/components/use-on-block-drop/index.native.js#L56-L62).

fluiddot (approved):
LGTM 🎊 ! Awesome work @geriux 🙇 !

**NOTE:** Following this note from the PR's description, before merging we should revert the Gutenberg reference update and JS bundle.

> The bundle and Gutenberg reference changes are included to be able to test the drag & drop event, those will not be included once this is approved and ready to be merged.

PR #4871, opened 2022-05-19
Use nvm to install Node js version required by Jetpack

jhnstn:
Fixes installing Jetpack after https://github.com/Automattic/jetpack/pull/24355 which introduces an operator that is not supported by Node v14.

This adds a new bin script to install Jetpack and it's dependencies.
The primary concerns of the script are:
1. use/install the same version of Node that Jetpack requires (via `./jetpack/.nvmrc`) 
2. Ensure the correct version of pnpm is used to install the JS packages

The script leverages `nvm` to manage  the Node install  required by Jetpack. `nvm` is available on Circle CI and is listed as a  dev dependency for Gutenberg Mobile. This change does make the former more of a hard requirement.

To test:

### Without the required Node version installed
0. Check to see if you have the Node version `16.13.2` installed,  ( `nvm ls` ).  If installed remove it with `nvm uninstall 16.13.2`. Optional: `nvm` will cache node versions which shouldn't be an issue for testing but if you want to try from a cold start try `rm -rf $NVM_DIR/.cache/bin/node-v16.13.2*`
1. Clear out any `node_module` directories in `./jetpack/*` e.g. `find ./jetpack -type d -name node_modules -prune -exec rm -rf {} \;` 
2. Run `$ npm install;  echo $?` and verify it outputs `0`
3. Verify that the bundle is ok and the metro server can resolve all of the dependencies.
4. Smoke test the app for any Jetpack issues e.g. Verify that the story block works as expected.

### With the required Node version installed
1. Run `$ nvm use` in the project root directory to make sure you are using the GBM required node version ( should be `14.*.*`)
2. Run `$ npm install;  echo $?` and verify it outputs `0`
3. Smoke test the app for any Jetpack issues

### Dependency check (optional)
On MacOS `nvm` will add the `nvm` setup to `~/.zshrc` which will not be available to the new `install-jetpack.sh` script since it runs in bash. The script should however handle issues with the nvm install gracefully. The following steps are valid if `nvm` is not found from bash. To check that, see if `$  bash -c "command -v nvm" ; echo $?` returns `1` 
1. Run `NVM_DIR=/dev/null ./bin/install-jetpack.sh`
2. Expect to see the message "nvm is not installed or cannot be sourced from /dev/null/nvm.sh. Please verify that $NVM_DIR points to the .nvm directory."
3. Verify that the command exits with `1` 

PR submission checklist:


bin/install-jetpack.sh
```diff
@@ -0,0 +1,27 @@
...
+    echo "nvm is not installed or cannot be sourced from ~/.nvm/nvm.sh. Please install nvm and run this script again."
+    exit 1
+  fi
+fi
+
+# Set up node requirement for Jetpack
+pushd jetpack
+
+node_version=$(cat .nvmrc)
+nvm install "$node_version"
```
jhnstn: Looks like we might not need the node version here. see https://github.com/nvm-sh/nvm/issues/318

jhnstn: Also don't need `nvm use`, `nvm install` will switch to the specified version.

bin/install-jetpack.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
+set -Eeuo pipefail
+
+# Check if nvm is installed
+[ -z "$NVM_DIR" ] && NVM_DIR="$HOME/.nvm"
+[ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"
```
jhnstn, line 6: I considered checking to see if nvm was installed via homebrew. In which case we could add a step that sources nvm.sh from `$ brew --prefix nvm`.  
Homebrew however does prints out a [significant caveat](https://github.com/Homebrew/homebrew-core/blob/HEAD/Formula/nvm.rb#L29) when installing nvm via brew and mentions the same issues on the [package page](https://formulae.brew.sh/formula/nvm). NVM also [warns about installing from homebrew as wel](https://github.com/nvm-sh/nvm#important-notes)l.

in general, if nvm is installed properly, `NVM_DIR` should point to the nvm directory.

anomiex:
If this gets merged, does that mean we could undo Automattic/jetpack#20990? As you found, we're not really testing for that declared Node 14 compatibility.

jhnstn:
> If this gets merged, does that mean we could undo https://github.com/Automattic/jetpack/pull/20990? As you found, we're not really testing for that declared Node 14 compatibility.

Seem appropriate to roll back those changes if we continue with this direction.

SiobhyB (approved):
I followed the steps in the PR description and confirmed all works as expected for me. In addition, I have the following PR that requires an update to the Jetpack reference: https://github.com/wordpress-mobile/gutenberg-mobile/pull/4899. Basing that PR off this branch, `jetpack/fix/node-version`, resolved all errors with the tests for me. 🙌 🎉

PR #5331, opened 2022-12-16
Remove `ui-tests-full-scheduled` job

geriux:
This PR removes the `ui-tests-full-scheduled` job that was currently running through a cron every day.

After some investigation, we could see we have other jobs that cover incoming changes very well by running the full tests suites so there is no need to have also a daily scheduled job.

We will be running the full tests in:

- Gutenberg via dependabot PRs (daily)
- Jetpack via dependabot PRs (daily)
- Any PR that gets merged into `trunk`
- Release branches

Kudos to @fluiddot for investigating the above list.

To test:
N/A

PR submission checklist:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -415,17 +415,3 @@ workflows:
             branches:
               only:
                 - /^release.*/
-  ui-tests-full-scheduled:
-    jobs:
-      - ios-device-checks:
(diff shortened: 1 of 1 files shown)


fluiddot (approved):
LGTM 🎊 !

PR #5456, opened 2023-02-07
Add visual tests support

geriux:
**Related PRs:**
- https://github.com/WordPress/gutenberg/pull/47845

## Description 

This PR adds the setup to be able to have visual tests, the idea is to start migrating the manual test we currently do each release and to also start expanding our test suites, like testing themes and important components like `RichText`.

We will be using two new libraries, [jest-image-snapshot](https://github.com/americanexpress/jest-image-snapshot) which enables us to compare two images using Jest. It compares the baseline image against the new one generated on CI, if the difference is over the threshold it will mark the test as failed and it will return the difference output as base64 in the console, this will be handy to check quickly what's breaking.

The other library is [jimp](https://github.com/jimp-dev/jimp#readme) which will be used to edit the screenshots taken using Appium, mostly to remove the status bar, resize and reduce quality.

As you'll see in the files changed and in the new tests, we just need to write them as we were doing for the other E2E tests, to match a visual change we'll just need to do:

```javascript
const screenshot = await takeScreenshot();
expect( screenshot ).toMatchImageSnapshot();
```

PR submission checklist:


fluiddot:
@geriux Seems we'd need to tweak the CI job configuration to address the following error in `Test Android on Device - Full`:
```
EACCES: permission denied, mkdir '/home/circleci/project/__device-tests__/__image_snapshots__/__diff_output__'
```

__device-tests__/utils.js
```diff
@@ -0,0 +1,22 @@
+/**
+ * External dependencies
+ */
+import jimp from 'jimp';
+
+const { isAndroid } = E2EUtils;
```
fluiddot: I couldn't find where `E2EUtils` is defined globally, @geriux could you share where's defined? Thanks!

fluiddot: Oh, I've just noticed this is defined [here](https://github.com/WordPress/gutenberg/pull/47845/files#diff-1bd8da7eecc6433e08ab95506f393935c137ac2806b7e3ec5c09796448ccb41bR7).

geriux:
> @geriux Seems we'd need to tweak the CI job configuration to address the following error in `Test Android on Device - Full`:

Yup we needed to create the directory for the snapshot difference output, I included it in https://github.com/wordpress-mobile/gutenberg-mobile/pull/5456/commits/a2ab57c2a29e80176b522795191fa13e9ce1f0d1

.circleci/config.yml
```diff
@@ -109,6 +109,11 @@ commands:
         - run:
             name: Create reports directory
             command: mkdir reports && mkdir reports/test-results
+  add-jest-snapshot-dir:
+      steps:
+        - run:
+            name: Create snapshots directory
+            command: mkdir __device-tests__/__image_snapshots__/__diff_output__
```
dcalhoun: I've noted the wrapping, double underscores in `__device-tests__` before. Do you know the reason for that pattern? Is there a reason to apply it recursively in children directories? Also, `device-tests` is hyphenated, while the new directories rely upon snake case.

dcalhoun (commented):
This looks great!

.circleci/config.yml
geriux: That is a good question, I'm not sure why `__device-tests__` has that name, for the image snapshots folders it is how the library generates them but we could customize the folder names according to the `jest-image-snapshot` [docs](https://github.com/americanexpress/jest-image-snapshot#%EF%B8%8F-api). Maybe we could rename them like:

- `__image_snapshots__` -> `image-snapshots`
- `__diff_output__` -> `diff`

What do you think?

dcalhoun: Ah, OK. Thanks for sharing that context via the docs. Maybe the double underscore pattern is merely a way to have generated directories grouped atop file lists with default name sort. 🤷🏻‍♂️ 

I do not have a strong opinion. The names you propose are clear and consistent. Those look good to me. 

If we'd prefer to avoid additional configuration, I think it is fine to retain the default names.

geriux: Since we are already updating some settings in https://github.com/wordpress-mobile/gutenberg-mobile/pull/5456/files#diff-ffabdf74f7609e8b6a7e64b0ddceff3ea0457ac950ada6ba1c7e39cd3bfe08c3R11 we can include these name customizations as well. I'll update the PR 😃

geriux: Updated in https://github.com/wordpress-mobile/gutenberg-mobile/pull/5456/commits/86a2046f3c0ef3b756916bd6509dc90faaad297d

geriux:
Hey there 👋 I've just updated this PR as there were some conflicts. I think I've addressed all comments but please let me know if there are any other questions, thank you!

__device-tests__/gutenberg-editor-gallery-visual.test.js
```diff
@@ -0,0 +1,35 @@
...
+			blockNames.gallery
+		);
+		expect( galleryBlock ).toBeTruthy();
+
+		// Wait for images to load
+		await editorPage.driver.sleep( 5000 );
+
+		// Visual test check
+		const screenshot = await takeScreenshot();
+		expect( screenshot ).toMatchImageSnapshot();
```
dcalhoun, line 33: I noticed the previous test has what appears to be clean up code removing the block. Should all tests including this one have that as well? This question applies to `__device-tests__/gutenberg-editor-group-visual.test.js` as well. 
```suggestion
		expect( screenshot ).toMatchImageSnapshot();
		
		await editorPage.removeBlockAtPosition( blockNames.gallery );
```

dcalhoun (approved):
Looking forward to merging this. I left one question related to test clean up. 

Additionally, what is the command used to run these tests locally?

dcalhoun:
> Additionally, what is the command used to run these tests locally?

Never mind. I was running the correct command locally: `npm run device-tests:local -- gutenberg-editor-gallery-visual.test.js`. 

The errors I encountered locally related to the e2e tests' requirement of Xcode 13. I only had Xcode 14 installed. Installing Xcode 13 and `sudo xcode-select -s /Applications/Xcode\ 13.app` allowed me to run the new visual tests successfully. 🎉

__device-tests__/gutenberg-editor-gallery-visual.test.js
geriux: It will depend on the test, for this case, to start with a "blank" canvas in the second test you need to remove the blocks you created before, as the other tests in the file will run in the same session.

geriux:
> > Additionally, what is the command used to run these tests locally?
> 
> Never mind. I was running the correct command locally: `npm run device-tests:local -- gutenberg-editor-gallery-visual.test.js`.
> 
> The errors I encountered locally related to the e2e tests' requirement of Xcode 13. I only had Xcode 14 installed. Installing Xcode 13 and `sudo xcode-select -s /Applications/Xcode\ 13.app` allowed me to run the new visual tests successfully. 🎉

Glad you got it running! Yeah until we upgrade to Appium 2, we are stuck with the Xcode 13 issue, becuase SauceLabs currently doesn't support Appium 2 for simulators on Android.

PR #5459, opened 2023-02-08
Register VideoPress block behind a DEV flag

SiobhyB:
With this PR, the VideoPress block has been registered behind a DEV flag. The block itself is not complete or fully functional, but is intended to serve as a useful placeholder that to be iterated on.

## Related PRs

* `Gutenberg`: https://github.com/WordPress/gutenberg/pull/47836
* `Jetpack`: https://github.com/Automattic/jetpack/pull/28812
* `iOS`: https://github.com/wordpress-mobile/WordPress-iOS/pull/20100
* `Android`: https://github.com/wordpress-mobile/WordPress-Android/pull/17928

## To Test

Run the editor locally with both Android and iOS, then go through the following checks:

- Verify that the VideoPress block can be inserted into a post via the block inserter, the block inserter search, and the slash inserter. 
- Verify that the block is only visible when running via your local development environment, it shouldn't be visible with the PR's associated installable build.
- Verify that all automated tests pass and that there are no obvious errors associated with the addition of the block.

<hr /> 

PR submission checklist:


renatoagds (approved):
LGTM

src/block-support/supported-blocks.json
```diff
@@ -33,7 +33,8 @@
 		"jetpack/email",
 		"jetpack/phone",
 		"jetpack/address",
-		"jetpack/tiled-gallery"
+		"jetpack/tiled-gallery",
+		"videopress/video"
```
fluiddot: I think these two blocks should be under the `devOnly` section as they won't be available to users.

fluiddot (approved):
LGTM 🎊 !

**NOTE:** I've restarted the CI jobs because the CircleCI ones didn't start in the last commit.

src/block-support/supported-blocks.json
SiobhyB: Ah, good catch! Moved in https://github.com/wordpress-mobile/gutenberg-mobile/pull/5459/commits/e4890c9a67dca19758aa92e6c18912dc0187d428.

src/jetpack-editor-setup.js
```diff
@@ -85,9 +85,16 @@ export function registerJetpackBlocks( { capabilities } ) {
 		capabilities.tiledGalleryBlock,
 		'jetpack/tiled-gallery'
 	);
+	hideBlockByCapability(
+		__DEV__ && capabilities.videoPressBlock,
```
SiobhyB: Adding the `__DEV__` flag here (vs. only as part of the `supportedJetpackBlocks` object, [like the Tiled Gallery block](https://github.com/wordpress-mobile/gutenberg-mobile/blob/e4890c9a67dca19758aa92e6c18912dc0187d428/src/jetpack-editor-setup.js#L30-L32)) because the VideoPress block lives in a separate package and is never passed to [this function](https://github.com/Automattic/jetpack/blob/b6c867d80387206e11eb9f8bb9d987ebc3b6187c/projects/plugins/jetpack/extensions/shared/register-jetpack-block.native.js#L22) for checking availability.

fluiddot: Good catch! I checked this change using the iOS installable build and although I can't insert the VideoPress block, it's still being rendered. The main reason for this is that `hideBlockByCapability` function hides the block but doesn't prevent being registered, hence being rendered as a regular block instead of unsupported.

As a workaround, we could implement a native version of `videopress/src/client/block-editor/blocks/video/index.ts` file ([where the block is registered](https://github.com/Automattic/jetpack/blob/2c1ee9cbd75f34375c1cb93a23dab9654e0bebdd/projects/packages/videopress/src/client/block-editor/blocks/video/index.ts#L19-L30)), and either:
1. Expose a `registerVideoPressBlock` function that registers the block in order to be used in Gutenberg Mobile.
2. Add the needed checks to conditionally register the block.

fluiddot (changes requested):

src/jetpack-editor-setup.js
SiobhyB: Ah, thank you for spotting that, I'll work on a `registerVideoPressBlock` function.

fluiddot (approved):
LGTM 🎊 !

- VideoPress block is displayed as unsupported in the installable builds (i.e. production version) ✅ .
- VideoPress block is rendered when using local development environment ✅ .

PR #5462, opened 2023-02-09
CircleCI - Update configuration to make builds once and share between jobs

geriux:
**Related PRs:**
- https://github.com/WordPress/gutenberg/pull/48026

This PR updates the CircleCI configuration to add two new steps to create the builds separately so they can be shared between jobs. The reasoning for this apart from optimizing the time it takes to run and credits usage, it's that we were reaching the limit amount to upload artifacts in the 15 mins timeframe, which is limited to 5.

There's no need to upload the build for each platform in each job (Canaries, Full tests) so now it will just do it once per platform.

It also adds `requires` steps for some jobs, now that the builds are done separately we can't run tests until those are uploaded.

This is also a pre-work for the visual regression tests where we will be introducing a new job for them so they run in parallel as the other E2E tests to save up running time.

Besides that, it also adds `npm-install-without-scripts` which will install the packages without the `postinstall` scripts to save quite a bit of time since we will only need some packages to run the tests like appium, etc.

The key names for the `NPM cache` jobs are also updated to just use the `arch` value since we use two different images, we don't need to store this per job name.

Finally, it updates the `nvm` command to use the latest version `v0.39.3`.

## Our current workflow in `trunk` is:
<kbd>Screen Shot 2023-02-13 at 15 12 34</kbd>

## This is how the test will run with this PR:
<kbd>Screen Shot 2023-02-13 at 15 11 07</kbd>

PR submission checklist:


.circleci/config.yml
```diff
@@ -137,7 +160,7 @@ parameters:
   linux-machine-image:
     type: string
     # Latest supported ubuntu image from https://circleci.com/docs/2.0/configuration-reference/#available-machine-images
-    default: "ubuntu-2004:202010-01"
+    default: "ubuntu-2004:2022.10.1"
```
geriux, line 163: I saw we were using a pretty old image so I updated it to the last one. I haven't seen any issues with the update.

.circleci/config.yml
```diff
@@ -2,41 +2,64 @@ version: 2.1
...
+        type: string
+        default: ""
     steps:
       - restore_cache:
           name: Restore NPM Cache
           keys:
-            - npm-i18n-v5-cache-v{{ .Environment.CACHE_TRIGGER_VERSION }}-job-{{ .Environment.CIRCLE_JOB }}-{{ checksum "gutenberg/package-lock.json" }}
-            - npm-i18n-v5-cache-v{{ .Environment.CACHE_TRIGGER_VERSION }}-job-{{ .Environment.CIRCLE_JOB }}-
+            - npm-i18n-v6-cache-{{ .Environment.CACHE_TRIGGER_VERSION }}-full-arch-{{ arch }}-{{ checksum "gutenberg/package-lock.json" }}
+            - npm-i18n-v6-cache-{{ .Environment.CACHE_TRIGGER_VERSION }}-full-arch-{{ arch }}
```
fluiddot: Now that we're updating the cache keys, I'm wondering if we should rethink its format:
- Prefix: We could use the job command for clarity, i.e.: `npm-install-full-cache-...`. This way we won't need to add the `full` word before the `-arch-` part.
- `v6`: We already set a cache version using the environment variable `CACHE_TRIGGER_VERSION`, so probably we don't need to set another one in the command. It's true that updating the variable will purge the cache of all jobs but I understand that's the purpose of the variable.

.circleci/config.yml
```diff
@@ -2,41 +2,64 @@ version: 2.1
...
       - save_cache:
           name: Save NPM Cache
-          key: npm-i18n-v5-cache-v{{ .Environment.CACHE_TRIGGER_VERSION }}-job-{{ .Environment.CIRCLE_JOB }}-{{ checksum "gutenberg/package-lock.json" }}
+          key: npm-i18n-v6-cache-{{ .Environment.CACHE_TRIGGER_VERSION }}-full-arch-{{ arch }}-{{ checksum "gutenberg/package-lock.json" }}
           paths:
             - ~/.npm
             - src/i18n-cache
+            - ~/.local/share/pnpm/store/v3
+            - ~/Library/pnpm/store/v3
+  npm-install-without-scripts:
```
fluiddot: If this job command will be only used for running E2E tests maybe we could name it differently, wdyt?

fluiddot (approved):
LGTM 🎊 ! I checked the new duration of E2E tests and this is a huge improvement, amazing work @geriux 🏅  ! 

As shared in https://github.com/WordPress/gutenberg/pull/48026#pullrequestreview-1297303160, I checked the E2E tests and all succeeded. I added some minor comments/suggestions but shouldn't block the merge of this PR.

.circleci/config.yml
geriux: Good idea! Let's take this opportunity to update them, I'll make the change 👍

.circleci/config.yml
geriux: Yeah, that's probably better so there's no confusion and it's easier to differentiate. I'll update it.

geriux:
FYI I've updated `nvm` in https://github.com/wordpress-mobile/gutenberg-mobile/pull/5462/commits/7b6d12701163b2f01f5d71f06261e5eee10b6fba

PR #5478, opened 2023-02-15
CircleCI checks - Fix running Full tests twice in certain branches

geriux:
This is a follow-up of https://github.com/wordpress-mobile/gutenberg-mobile/pull/5462 where it was causing an issue where the `Test iOS on Device - Full` and `Test Android on Device - Full` steps would get triggered for `release`, `dependabot`, and `trunk` branches when it shouldn't, as there are steps specifically for those branches.

Making the `Optional UI Tests` as required and that requirement being ignored for those branches would trigger full tests to run twice. This PR addresses that issue.

You could [see this happening](https://app.circleci.com/pipelines/github/wordpress-mobile/gutenberg-mobile?branch=trunk) in the `trunk` branch after the recent merge.

**To test**
CI checks should pass.

PR submission checklist:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -408,9 +408,21 @@ workflows:
       - ios-device-checks:
           name: Test iOS on Device - Full
           requires: [ "Optional UI Tests", "Test iOS", "iOS Build" ]
+          filters:
+            branches:
+              ignore:
+                - trunk
+                - /^dependabot/submodules/.*/
+                - /^release.*/
       - android-device-checks:
           name: Test Android on Device - Full
           requires: [ "Optional UI Tests", "Test Android", "Android Native Unit Tests", "Android Build" ]
+          filters:
+            branches:
+              ignore:
+                - trunk
+                - /^dependabot/submodules/.*/
+                - /^release.*/
       - android-native-unit-tests:
           name: Android Native Unit Tests
       - ios-device-checks:


fluiddot (approved):
LGTM 🎊 !

Ah, good catch. I completely overlooked this issue when reviewing https://github.com/wordpress-mobile/gutenberg-mobile/pull/5462. Thanks for fixing it @geriux  🙇 !

On a different note, I noticed that we have a bit of redundancy on the jobs for `trunk`, `dependabot`, and `release` branches ([reference](https://github.com/wordpress-mobile/gutenberg-mobile/blob/877f520aef4ff545b0719f8ec54038aa8d0b5fd9/.circleci/config.yml#L426-L473)). Not sure why we have one job for each one but seems we could unify them into one using the `only` filter, I'll create a quick PR as a follow-up on this.

geriux:
> Ah, good catch. I completely overlooked this issue when reviewing https://github.com/wordpress-mobile/gutenberg-mobile/pull/5462. Thanks for fixing it @geriux 🙇 !

No worries I thought that by having the requirement was enough 😅 

> On a different note, I noticed that we have a bit of redundancy on the jobs for `trunk`, `dependabot`, and `release` branches ([reference](https://github.com/wordpress-mobile/gutenberg-mobile/blob/877f520aef4ff545b0719f8ec54038aa8d0b5fd9/.circleci/config.yml#L426-L473)). Not sure why we have one job for each one but seems we could unify them into one using the `only` filter, I'll create a quick PR as a follow-up on this.

Yeah, I agree! Thanks for working on that!

fluiddot:
> On a different note, I noticed that we have a bit of redundancy on the jobs for `trunk`, `dependabot`, and `release` branches ([reference](https://github.com/wordpress-mobile/gutenberg-mobile/blob/877f520aef4ff545b0719f8ec54038aa8d0b5fd9/.circleci/config.yml#L426-L473)). Not sure why we have one job for each one but seems we could unify them into one using the `only` filter, I'll create a quick PR as a follow-up on this.

[Here is the PR](https://github.com/wordpress-mobile/gutenberg-mobile/pull/5482). @geriux I assigned you as a reviewer, the change is an enhancement so please don't prioritize the review. Thanks 🙇  !

PR #5501, opened 2023-02-23
Visual testing - Adds Sanity Test 4 Suite

geriux:
**Related PRs:**
- https://github.com/WordPress/gutenberg/pull/48358
- https://github.com/wordpress-mobile/test-cases/pull/131

This PR adds the first steps for the Sanity Test Suite 4:

**Spacer block**
  - Spacer block - Spacer in horizontal layout works as expected

**Buttons block**
  - Render custom text and background color
  - Render gradient background color
  - Check if selection/caret color matches font color
  - Edit text styles
  - Link from the clipboard is presented as an option in the link picker

It also updates the `takeScreenshot` util to take into account if the device is in landscape mode, if you want to avoid including the keyboard in the screenshot, or if you want to pass a custom height value for the screenshot to be cropped.

After several test failures, I realized that including the keyboard will generate false positives, as the keyboard could look different in our local environments, like having other languages, and the emoji keyboard. I also encountered issues where the Keyboard would have a totally different appearance on CI.

Finally, it changes the `comparisonMethod` to compare the screenshots to `ssim`. You can find a nice explanation on the [docummentation](https://github.com/americanexpress/jest-image-snapshot#recommendations-when-using-ssim-comparison), as the pixel by pixel comparison wasn't working right.

**Note**: Since this PR updates how it takes the screenshots, I'm updating all visual snapshots.

To test CI checks should pass on this PR.

PR submission checklist:


__device-tests__/utils.js
```diff
@@ -5,18 +5,59 @@ import jimp from 'jimp';
 
 const { isAndroid } = e2eUtils;
 
-export async function takeScreenshot() {
+/**
+ * Helper to take a screenshot and manipulate it.
+ *
+ * @param {Object}  options                   Options
+ * @param {boolean} [options.withoutKeyboard] Prevents showing the keyboard in the screenshot.
+ * @param {number}  [options.customHeight]    Specify a custom height in percentage.
```
dcalhoun: `custom` may be a superfluous descriptor. WDYT?

Given the unitless value, we might more clearly communicate the intent if we used decimals, similar to CSS properties like [`opacity`](https://developer.mozilla.org/en-US/docs/Web/CSS/opacity), or rename to set the percentage expectation. WDYT? 

```js
takeScreenshot({ height: 0.5 });
takeScreenshot({ heightPercentage: 50 });
```

__device-tests__/utils.js
```diff
@@ -5,18 +5,59 @@ import jimp from 'jimp';
...
+				? windowSize.height
+				: image.getHeight();
+			const newHeight = ( imageHeight * offset ) / windowSize.height;
+
+			image.crop( 0, 0, image.getWidth(), imageHeight - newHeight );
+		}
+	}
+
+	// Custom height in percentage
+	if ( customHeight ) {
```
dcalhoun: We could avoid the conditional logic with a default value for `customHeight`. Purely a preference comment, so feel free to ignore this. 

```js
const { withoutKeyboard, customHeight = 100 } = options;
```

__device-tests__/gutenberg-editor-sanity-test-4-visual.test.js
```diff
@@ -0,0 +1,235 @@
+/**
+ * Internal dependencies
+ */
+const { blockNames } = editorPage;
+const { toggleOrientation, selectTextFromTextInput, setClipboard } = e2eUtils;
+import { takeScreenshot } from './utils';
+
+describe( 'Gutenberg Editor - Test Suite 4', () => {
```
dcalhoun, line 8: I wonder if we should eventually stop grouping the suite numbers and organize them by subject matter instead, e.g. by block type. I presume the suite numbers were originally used to delegate relatively balanced batches of tests. 

WDYT?

__device-tests__/gutenberg-editor-sanity-test-4-visual.test.js
```diff
@@ -0,0 +1,235 @@
...
+	it( 'Spacer block - Spacer in horizontal layout works as expected', async () => {
+		await editorPage.addNewBlock( blockNames.spacer );
+
+		await toggleOrientation( editorPage.driver );
+
+		// Wait for the device to finish rotating
+		await editorPage.driver.sleep( 3000 );
```
dcalhoun, line 15: Would it make sense to poll `getOrientation` on an interval or does that method prematurely return a value that does not match reality?

__device-tests__/gutenberg-editor-sanity-test-4-visual.test.js
```diff
@@ -0,0 +1,235 @@
...
+<!-- /wp:button --></div>
+<!-- /wp:buttons -->`;
+
+		it( 'Render custom text and background color', async () => {
+			await editorPage.setHtmlContent( buttonCustomColors );
+
+			const buttonsBlock = await editorPage.getBlockAtPosition(
+				blockNames.buttons
+			);
+			expect( buttonsBlock ).toBeTruthy();
```
dcalhoun, line 57: Are the `toBeTruthy` assertions in this file testing for something specific that is separate from `toMatchImageSnapshot`?

__device-tests__/gutenberg-editor-sanity-test-4-visual.test.js
```diff
@@ -0,0 +1,235 @@
...
+
+			const link = 'https://wordpress.org/';
+
+			await setClipboard( editorPage.driver, link );
+
+			await editorPage.toggleFormatting( 'Edit link' );
+			await editorPage.openLinkToSettings();
+
+			// Wait for the modal to open
+			await editorPage.driver.sleep( 3000 );
```
dcalhoun, line 203: Not necessarily required right now — maybe it is premature — but I wonder if we should abstract the modal animation times so that it is (1) clear what is occurring and (2) could be modified across tests as necessary. 

```js
const MODAL_OPEN_DURATION = 3000;
const MODAL_CLOSE_DURATION = 2000;
```

dcalhoun (commented):
Great work! 👏🏻 

All of my comments are merely thoughts, not blockers by any means. 

Also, I noticed the title selection changed in [`__device-tests__/image-snapshots/gutenberg-editor-group-visual-test-js-gutenberg-editor-visual-test-for-group-block-should-show-the-empty-placeholder-for-the-unselected-state-1-android.png`](https://github.com/wordpress-mobile/gutenberg-mobile/pull/5501/files#diff-174172b2c49804f90bb18bfc758138fa0750cd98f967ded57208801703e24e4d). Was that intended?

__device-tests__/gutenberg-editor-sanity-test-4-visual.test.js
geriux: >I presume the suite numbers were originally used to delegate relatively balanced batches of tests.

Yeah, I think that was the main idea back then.

Currently, on CI, each test file takes quite a bit to get the environment ready to run the tests, I'd say around 2 minutes per file (on iOS at least) since the machines we use are not that powerful. I think we should avoid having too many files for now so the running time doesn't go too high.

__device-tests__/gutenberg-editor-sanity-test-4-visual.test.js
geriux: It returns the return value quite fast since the rotation doesn't take much time but in some cases, some UI elements haven't finished getting to their positions resulting in having them in the wrong places when the screenshot is taken 😅

__device-tests__/gutenberg-editor-sanity-test-4-visual.test.js
geriux: Since we are adding this block from the HTML view and then toggling back, to have a way of waiting for elements to render correctly before taking a screenshot I used the `toBeTruthy` assertion to make sure the block is there before the next step.

__device-tests__/gutenberg-editor-sanity-test-4-visual.test.js
geriux: Good idea, maybe even having a `waitForModalToOpen` / `waitForModalToClose` utils. I can create a follow-up PR to introduce either that or just the constants.

__device-tests__/utils.js
geriux: True! I like the idea of having the expectation in the name like you suggested `heightPercentage` I'll update it!

__device-tests__/utils.js
geriux: We could but I wonder if it'd make sense to run `crop` on the image unnecessarily if we remove the `if` condition 🤔

geriux:
> Great work! 👏🏻
> 
> All of my comments are merely thoughts, not blockers by any means.

Thanks for reviewing!

> Also, I noticed the title selection changed in [`__device-tests__/image-snapshots/gutenberg-editor-group-visual-test-js-gutenberg-editor-visual-test-for-group-block-should-show-the-empty-placeholder-for-the-unselected-state-1-android.png`](https://github.com/wordpress-mobile/gutenberg-mobile/pull/5501/files#diff-174172b2c49804f90bb18bfc758138fa0750cd98f967ded57208801703e24e4d). Was that intended?

Yes! I actually realized we needed to update that screenshot when I changed the comparison method, now it matches the iOS screenshot.

__device-tests__/utils.js
geriux: Updated in https://github.com/wordpress-mobile/gutenberg-mobile/pull/5501/commits/2533b8497ac4faf3851b3ce4f127dbf4eb5d6d1c

__device-tests__/utils.js
dcalhoun: Ah, great point. Ignore my previous suggestion.

__device-tests__/gutenberg-editor-sanity-test-4-visual.test.js
dcalhoun: The cost of each test file is a great point. 

There may be a more logical, coherent grouping, but I do not have one to share at this time. I think it makes sense to keep them structured in the "sanity" groups for the time being. 

Maybe in the future, if we identify categories or, possibly, levels of flakiness (i.e. requiring a different threshold of precision), we could reorganize the tests to optimize for stability, speed, and comprehensibility.

__device-tests__/gutenberg-editor-sanity-test-4-visual.test.js
dcalhoun: 👍🏻 I figured that was the case. Thank you for confirming.

__device-tests__/gutenberg-editor-sanity-test-4-visual.test.js
dcalhoun: Ah, I see. The `getBlockAtPosition` is the key line awaiting the required render. The assertion is an attempt at communicating the intent. 👍🏻

__device-tests__/gutenberg-editor-sanity-test-4-visual.test.js
dcalhoun: While I still think naming the sleep durations may help communicate intent, I may "walk back" on my suggestion after spending time in the e2e tests yesterday. My hot take is that we have _a lot_ of test helpers, to the quantity that I find it difficult to comprehend each of them and their intent. I share that to say I now have reservations in regards to expanding test helpers further. 

As a tangential thought to consider outside the scope of this PR, I wonder if we would benefit from evaluating if there any helpers we might should remove and replace with composing other existing, lower-level helpers. My perception is that we have some helpers that are very specific and niche. Obviously, a subject to discuss in a different place. Just sharing my thoughts.

__device-tests__/gutenberg-editor-sanity-test-4-visual.test.js
```diff
@@ -0,0 +1,235 @@
...
+
+			// Visual test check for unselected state
+			const screenshot = await takeScreenshot();
+			expect( screenshot ).toMatchImageSnapshot();
+
+			await buttonsBlock.click();
+			await editorPage.removeBlockAtPosition( blockNames.buttons );
+		} );
+
+		it( 'Link from the clipboard is presented as an option in the link picker', async () => {
```
dcalhoun, line 190: My perception is that all tests in this PR besides this one are very visual and likely benefit from visual regression testing snapshots. However, it seems we could likely test the spirit of this link picker test with a (possibly faster or simpler) integration test. 

This is more a question regarding our overarching approach to testing, not necessarily directly related to the work of this PR. Not looking to block this PR, but seeking your thoughts. WDYT?

dcalhoun (approved):
Happy to have this coverage. Thank you!

__device-tests__/gutenberg-editor-sanity-test-4-visual.test.js
geriux: > My hot take is that we have a lot of test helpers, to the quantity that I find it difficult to comprehend each of them and their intent.

I share the same feeling 😅 

> I wonder if we would benefit from evaluating if there any helpers we might should remove and replace with composing other existing, lower-level helpers. My perception is that we have some helpers that are very specific and niche. Obviously, a subject to discuss in a different place. 

I agree, I think it's good you shared this, working on adding new tests always brings new perspectives and all the feedback will improve our current workflows and the way we do things.

__device-tests__/gutenberg-editor-sanity-test-4-visual.test.js
geriux: For this case, since we are testing copying a link from the clipboard and expecting it to show in the link picker it is very attached to the device imagine something fails like a new OS permission is added and the link can't be pasted, and having a visual test of the whole workflow would allow us not to have to test this manually.

geriux:
> Happy to have this coverage. Thank you!

Thank you for the review! 🚀

__device-tests__/gutenberg-editor-sanity-test-4-visual.test.js
dcalhoun: Fair point. Avoiding a clipboard API mock provides additional coverage (e.g. OS permissions). There a few approaches to testing the spirit of this test case we could consider: 

1. Capture visual images within an e2e tests. 
2. Assert element presence within an e2e tests. 
3. Mock the clipboard API and assert element presence within an integration tests. 

It seems you prefer the first two for this specific test case, which I think is reasonable. 

Additionally, I do not know if there is a meaningful "cost" between the first two. The first may "cost" more in regards to test speed or repository size for stored images. While the test case description doesn't necessarily _require to visual testing_ (we could just assert the element presence), it is additional coverage. I suppose it is a judgement call as to whether the visual coverage is worth it.

PR #5544, opened 2023-03-09
Update to Jest 29

geriux:
This PR updates the Jest version to `29` to match Gutenberg's current version introduced [in this PR.](https://github.com/WordPress/gutenberg/pull/47388)

PR submission checklist:


jest.config.js
```diff
@@ -34,7 +34,9 @@ module.exports = {
 		'<rootDir>/jetpack/',
 		'/__device-tests__/',
 	],
-	testURL: 'http://localhost/',
+	testEnvironmentOptions: {
```
geriux, line 37: `testURL` is deprecated so we needed to update it to the new format.

jest_setup.js
```diff
@@ -0,0 +1,17 @@
+// Jetpack mocks
+global.window.JP_CONNECTION_INITIAL_STATE = {};
```
geriux: This was added to remove the console error: 

`Jetpack Connection package: Initial state is missing. `

This was happening before the Jest upgrade.

jest_setup.js
```diff
@@ -0,0 +1,17 @@
...
+
+jest.mock( './jetpack/projects/js-packages/config/src', () => ( {
```
geriux: Added to remove the console error:

`jetpackConfig is missing in your webpack config file. See @**********/jetpack-config`

jest_setup.js
```diff
@@ -0,0 +1,17 @@
...
+	__esModule: true,
+} ) );
+
+jest.mock(
```
geriux: After the Jest 29 update, there is an issue where it can't find the native export of this file so I added a mock.

geriux:
FYI there are a few CI failures for this PR due to other reasons like `node` failing to download.

package.json
```diff
@@ -39,8 +39,9 @@
 		"gettext-parser": "^4.0.3",
 		"husky": "^3.0.2",
 		"identity-obj-proxy": "^3.0.0",
-		"jest": "^27.4.5",
+		"jest": "29.5.0",
 		"jest-image-snapshot": "^6.1.0",
+		"jest-jasmine2": "29.5.0",
```
geriux, line 44: `jest-jasmine2` is no longer included with Jest and it has [to be installed separately](https://jestjs.io/docs/28.x/upgrading-to-jest28#test-runner).

jest.config.js
```diff
@@ -25,7 +25,7 @@ module.exports = {
 	// Automatically clear mock calls and instances between every test
 	clearMocks: true,
 	preset: './gutenberg/node_modules/react-native/jest-preset.js',
-	setupFiles: [ '<rootDir>/' + configPath + '/setup.js' ],
+	setupFiles: [ '<rootDir>/' + configPath + '/setup.js', './jest_setup.js' ],
```
fluiddot: Nitpick: I'm wondering if we could use a similar name format for `jest_setup.js` as this file, i.e. `jest.setup.js`, WDYT?

jest_setup.js
fluiddot: I saw this approach is also used in the Jetpack unit tests: https://github.com/Automattic/jetpack/blob/trunk/projects/js-packages/connection/jest.setup.js

jest_setup.js
fluiddot: I checked how this is mocked in the Jetpack unit tests and found [this configuration](https://github.com/Automattic/jetpack/blob/2a306ced82a35a6a4bf199edc8af77494584b53d/tools/js-tools/jest/config.base.js#L27-L29). I tested locally and seems to work by adding the following to `jest.config.js`:
```patch
diff --git forkSrcPrefix/jest.config.js forkDstPrefix/jest.config.js
index 81bbacc988d81e281a87e34086a20bf93436e847..4846110d6ae9b9f0b206d44fc8510a1c3f24bf55 100644
--- forkSrcPrefix/jest.config.js
+++ forkDstPrefix/jest.config.js
@@ -51,6 +51,8 @@ module.exports = {
 			'|'
 		) })$` ]: '<rootDir>/gutenberg/packages/$1/src',
 		'test/helpers$': '<rootDir>/' + configPath + '/helpers.js',
+		jetpackConfig:
+			'<rootDir>/jetpack/tools/js-tools/jest/jest-jetpack-config.js',
 	},
 	haste: {
 		defaultPlatform: rnPlatform,

```

jest.config.js
geriux: Sure! I'll update it 👍

jest_setup.js
geriux: Nice suggestion! I'll apply this change.

jest_setup.js
geriux: I tried to use this setup but it looks like it is not compatible with the environment we use:

```javascript
TypeError: Cannot convert undefined or null to object
        at Function.getOwnPropertyDescriptors (<anonymous>)

      window.location = Object.defineProperties(
```

What do you think if we copy these:

```javascript
window.JP_CONNECTION_INITIAL_STATE = {
	userConnectionData: {
		currentUser: {
			wpcomUser: { Id: 99999, login: 'bobsacramento', display_name: 'Bob Sacrmaneto' },
		},
	},
};
```

Into our `global.window.JP_CONNECTION_INITIAL_STATE` mock?

jest.config.js
geriux: Updated in https://github.com/wordpress-mobile/gutenberg-mobile/pull/5544/commits/9278461c173409d342520b8d7884c7f27b8f04cd and https://github.com/wordpress-mobile/gutenberg-mobile/pull/5544/commits/be3b94f1f4f716e706cee3dbbedb57d9647b5335

jest_setup.js
geriux: Something like this https://github.com/wordpress-mobile/gutenberg-mobile/pull/5544/commits/29521810d3368f3f0a6b45f07a66cf8d42d78412

fluiddot: Yeah, that approach looks good to me, thanks! In the future, we'd need to revisit how to manage the Jetpack connection in GB-mobile and populate with data coming form the host (i.e. the app).

fluiddot (approved):
LGTM 🎊 ! Thanks @geriux for fixing this 🙇 !

I tested locally by running `npm ci && npm run test` and all tests passed ✅ .

PR #5610, opened 2023-03-30
Add Rich Text Writing Flow E2E tests for cases TC007, TC009, TC010

derekblank:
Adds Rich Text [Writing Flow](https://github.com/wordpress-mobile/test-cases/blob/trunk/test-cases/gutenberg/writing-flow/rich-text-formatting.md) E2E test cases for [TC007](https://github.com/wordpress-mobile/test-cases/blob/trunk/test-cases/gutenberg/writing-flow/rich-text-formatting.md#tc007), [TC009](https://github.com/wordpress-mobile/test-cases/blob/trunk/test-cases/gutenberg/writing-flow/rich-text-formatting.md#tc009), and [TC010](https://github.com/wordpress-mobile/test-cases/blob/trunk/test-cases/gutenberg/writing-flow/rich-text-formatting.md#tc010).

To test:
- Android: Run `TEST_RN_PLATFORM=android npm run device-tests:local gutenberg-editor-writing-flow.test`
- iOS: Run `TEST_RN_PLATFORM=ios npm run device-tests:local gutenberg-editor-writing-flow.test`

PR submission checklist:

**Todo Checklist**
TC007:
- [x] [Verify best method to select text formatting](https://github.com/wordpress-mobile/gutenberg-mobile/compare/test/rich-text-writing-flow?expand=1#diff-bbf7f3f59de95b206e4b35babff68b7302411774427c2862ff9ff47e648c4b71R27-R28)
- [x] [Identify best method to identify format buttons](https://github.com/wordpress-mobile/gutenberg-mobile/compare/test/rich-text-writing-flow?expand=1#diff-bbf7f3f59de95b206e4b35babff68b7302411774427c2862ff9ff47e648c4b71R31)
- [ ] [Identify if snapshot should be used in place of assertion (and implement, if so)](https://github.com/wordpress-mobile/gutenberg-mobile/compare/test/rich-text-writing-flow?expand=1#diff-bbf7f3f59de95b206e4b35babff68b7302411774427c2862ff9ff47e648c4b71R33)
- [ ] [Repeat for all formatting types](https://github.com/wordpress-mobile/gutenberg-mobile/compare/test/rich-text-writing-flow?expand=1#diff-bbf7f3f59de95b206e4b35babff68b7302411774427c2862ff9ff47e648c4b71R35)
- [ ] Verify if markup is correct with assertion

TC009:

- [x] [Identify if snapshot should be used in place of assertion (and implement, if so)](https://github.com/wordpress-mobile/gutenberg-mobile/compare/test/rich-text-writing-flow?expand=1#diff-bbf7f3f59de95b206e4b35babff68b7302411774427c2862ff9ff47e648c4b71R66)
- [ ] Verify if markup is correct with assertion

TC010:

- [x] [Identify if snapshot should be used in place of assertion (and implement, if so)](https://github.com/wordpress-mobile/gutenberg-mobile/compare/test/rich-text-writing-flow?expand=1#diff-bbf7f3f59de95b206e4b35babff68b7302411774427c2862ff9ff47e648c4b71R98)
- [ ] Verify if markup is correct with assertion

### DIFF
--- a/__device-tests__/gutenberg-editor-writing-flow.test.js
+++ b/__device-tests__/gutenberg-editor-writing-flow.test.js
@@ -1,0 +1,104 @@
+/**
+ * Internal dependencies
+ */
+const { blockNames } = editorPage;
+
+describe( 'Gutenberg Editor - Writing Flow', () => {
+	describe( 'TC007 - Test format detection under the cursor', () => {
+		it( 'checks that the proper format buttons are selected when the cursor is under', async () => {
+			const paragraphHTML = `<!-- wp:paragraph -->
+            <p><span accessibility-id="bold"><strong>bold</strong></span>
+            <span accessibility-id="bold-italic"><strong><i>bold-italic</i></strong>.</span>
+            <span accessibility-id-"strikethrough"><s>strikethrough</s></span></p>
+            <!-- /wp:paragraph -->`;
+
+			// On a rich-text based component, add bold, italic, strikethrough and link formatted text
+			await editorPage.addNewBlock( blockNames.paragraph );
+			const paragraphBlockElement = await editorPage.getTextBlockAtPosition(
+				blockNames.paragraph
+			);
+
+			await editorPage.typeTextToTextBlock(
+				paragraphBlockElement,
+				paragraphHTML
+			);
+
+			// Select bold text
+			// TODO: verify if this is the best method to select text formatting
+			await editorPage.elementsByAccessibilityId( 'bold' );
+
+			// Check that the proper format buttons get selected
+			// TODO: Identify best method to identify format buttons
+
+			// TODO: Invoke assertion or snapshot here?
+
+			// TODO: Repeat for bold-italic and strikethrough
+
+			await editorPage.removeBlock();
+		} );
+	} );
+
+	describe( `TC009 - Test autocorrection doesn't apply formatting to Heading`, () => {
+		it( 'checks that formatting is not applied to autocorrected text', async () => {
+			const headingText = `<p>Mispelled</p>`;
+
+			// Add a Heading block without any formatting applied.
+			await editorPage.addNewBlock( blockNames.heading );
+			const headingBlockElement = await editorPage.getTextBlockAtPosition(
+				blockNames.heading
+			);
+
+			// Type a sentence with a word misspelled that will be autocorrected by the editor
+			await editorPage.typeTextToTextBlock(
+				headingBlockElement,
+				headingText
+			);
+
+			// Highlight the word that was autocorrected
+			// Simulated double tap. TODO: Verify if this is the best method to highlight text
+			await editorPage.tapElement( headingBlockElement );
+			await editorPage.tapElement( headingBlockElement );
+
+			// Check that the word has no formatting applied
+			const selectedText = await editorPage.getTextForParagraphBlockAtPosition(
+				1
+			);
+			// TODO: verify if a snapshot should be used in place of an assertion, or both
+			expect( selectedText ).toMatch( headingText );
+
+			await editorPage.removeBlock();
+		} );
+	} );
+
+	describe( `TC010 - Test autocorrection doesn't remove formatting from Heading`, () => {
+		it( 'checks that formatting is not removed from autocorrected text', async () => {
+			const headingText = '<p><strong>Mispelled</strong></p>';
+
+			// Add a Heading block
+			await editorPage.addNewBlock( blockNames.heading );
+			const headingBlockElement = await editorPage.getTextBlockAtPosition(
+				blockNames.heading
+			);
+
+			// Type a sentence with a word misspelled that will be autocorrected by the editor
+			await editorPage.typeTextToTextBlock(
+				headingBlockElement,
+				headingText
+			);
+
+			// Highlight the word that was autocorrected
+			// Simulated double tap. TODO: Verify if this is the best method to highlight text
+			await editorPage.tapElement( headingBlockElement );
+			await editorPage.tapElement( headingBlockElement );
+
+			// Check that the word has formatting applied
+			const selectedText = await editorPage.getTextForParagraphBlockAtPosition(
+				1
+			);
+			// TODO: verify if a snapshot should be used in place of an assertion, or both
+			expect( selectedText ).toMatch( headingText );
+
+			await editorPage.removeBlock();
+		} );
+	} );
+} );
--- a/__device-tests__/gutenberg-editor-writing-flow.test.js
+++ b/__device-tests__/gutenberg-editor-writing-flow.test.js
@@ -58,7 +58,7 @@ describe( 'Gutenberg Editor - Writing Flow', () => {
 
 	describe( `TC009 - Test autocorrection doesn't apply formatting to Heading`, () => {
 		it( 'checks that formatting is not applied to autocorrected text', async () => {
-			const headingText = `<p>Mispelled</p>`;
+			const headingText = `Mispelled`;
 
 			// Add a Heading block without any formatting applied.
 			await editorPage.addNewBlock( blockNames.heading );
@@ -90,7 +90,7 @@ describe( 'Gutenberg Editor - Writing Flow', () => {
 
 	describe( `TC010 - Test autocorrection doesn't remove formatting from Heading`, () => {
 		it( 'checks that formatting is not removed from autocorrected text', async () => {
-			const headingText = '<p><strong>Mispelled</strong></p>';
+			const headingText = 'Mispelled';
 
 			// Add a Heading block
 			await editorPage.addNewBlock( blockNames.heading );
@@ -98,7 +98,10 @@ describe( 'Gutenberg Editor - Writing Flow', () => {
 				blockNames.heading
 			);
 
+			// Toggle bold formatting
 			// Type a sentence with a word misspelled that will be autocorrected by the editor
+
+			await editorPage.toggleFormatting( 'Bold' );
 			await editorPage.typeTextToTextBlock(
 				headingBlockElement,
 				headingText
@@ -115,6 +118,8 @@ describe( 'Gutenberg Editor - Writing Flow', () => {
 				1
 			);
 			expect( selectedText ).toMatch( headingText );
+
+			// TODO: Add assertion or visual snapshot that bold formatting is retained
 
 			await editorPage.removeBlock();
 		} );
--- a/__device-tests__/gutenberg-editor-writing-flow.test.js
+++ b/__device-tests__/gutenberg-editor-writing-flow.test.js
@@ -1,6 +1,7 @@
 /**
  * Internal dependencies
  */
+import { tapSelectAllAboveElement } from './helpers/utils';
 const { blockNames } = editorPage;
 
 describe( 'Gutenberg Editor - Writing Flow', () => {
@@ -55,9 +56,10 @@ describe( 'Gutenberg Editor - Writing Flow', () => {
 			);
 
 			// Highlight the word that was autocorrected
-			// Simulated double tap. TODO: Verify if this is the best method to highlight text
-			await editorPage.tapElement( headingBlockElement );
-			await editorPage.tapElement( headingBlockElement );
+			await tapSelectAllAboveElement(
+				editorPage.driver,
+				headingBlockElement
+			);
 
 			// Check that the word has no formatting applied
 			const selectedText = await editorPage.getTextForParagraphBlockAtPosition(
@@ -87,9 +89,10 @@ describe( 'Gutenberg Editor - Writing Flow', () => {
 			);
 
 			// Highlight the word that was autocorrected
-			// Simulated double tap. TODO: Verify if this is the best method to highlight text
-			await editorPage.tapElement( headingBlockElement );
-			await editorPage.tapElement( headingBlockElement );
+			await tapSelectAllAboveElement(
+				editorPage.driver,
+				headingBlockElement
+			);
 
 			// Check that the word has formatting applied
 			const selectedText = await editorPage.getTextForParagraphBlockAtPosition(
--- a/__device-tests__/gutenberg-editor-writing-flow.test.js
+++ b/__device-tests__/gutenberg-editor-writing-flow.test.js
@@ -2,6 +2,7 @@
  * Internal dependencies
  */
 import { tapSelectAllAboveElement } from './helpers/utils';
+import { takeScreenshot } from './utils';
 const { blockNames } = editorPage;
 
 describe( 'Gutenberg Editor - Writing Flow', () => {
@@ -28,12 +29,18 @@ describe( 'Gutenberg Editor - Writing Flow', () => {
 			// Toggle various formatting options, then proceed with typing more text
 			await editorPage.toggleFormatting( 'Bold' );
 
+			const boldScreenshot = await takeScreenshot();
+			expect( boldScreenshot ).toMatchImageSnapshot();
+
 			await editorPage.typeTextToTextBlock(
 				paragraphBlockElement,
 				paragraphText[ 1 ]
 			);
 
 			await editorPage.toggleFormatting( 'Italic' );
+
+			const boldItalicScreenshot = await takeScreenshot();
+			expect( boldItalicScreenshot ).toMatchImageSnapshot();
 
 			await editorPage.typeTextToTextBlock(
 				paragraphBlockElement,
@@ -50,7 +57,8 @@ describe( 'Gutenberg Editor - Writing Flow', () => {
 
 			await editorPage.toggleFormatting( 'Strikethrough' );
 
-			// TODO: Invoke assertion or snapshot here?
+			const strikethroughScreenshot = await takeScreenshot();
+			expect( strikethroughScreenshot ).toMatchImageSnapshot();
 
 			await editorPage.removeBlock();
 		} );
--- a/__device-tests__/gutenberg-editor-writing-flow.test.js
+++ b/__device-tests__/gutenberg-editor-writing-flow.test.js
@@ -1,7 +1,7 @@
 /**
  * Internal dependencies
  */
-import { tapSelectAllAboveElement } from './helpers/utils';
+const { tapSelectAllAboveElement } = e2eUtils;
 import { takeScreenshot } from './utils';
 const { blockNames } = editorPage;
 
--- a/__device-tests__/gutenberg-editor-writing-flow.test.js
+++ b/__device-tests__/gutenberg-editor-writing-flow.test.js
@@ -49,7 +49,6 @@ describe( 'Gutenberg Editor - Writing Flow', () => {
 			await editorPage.toggleFormatting( 'Italic' );
 			await editorPage.toggleFormatting( 'Bold' );
 
-
 			await editorPage.typeTextToTextBlock( paragraphBlockElement, '' );
 			await editorPage.typeTextToTextBlock(
 				paragraphBlockElement,
--- a/__device-tests__/gutenberg-editor-writing-flow.test.js
+++ b/__device-tests__/gutenberg-editor-writing-flow.test.js
@@ -1,19 +1,14 @@
 /**
  * Internal dependencies
  */
-const { tapSelectAllAboveElement } = e2eUtils;
+const { selectTextFromElement, tapSelectAllAboveElement } = e2eUtils;
 import { takeScreenshot } from './utils';
 const { blockNames } = editorPage;
 
 describe( 'Gutenberg Editor - Writing Flow', () => {
 	describe( 'TC007 - Test format detection under the cursor', () => {
 		it( 'checks that the proper format buttons are selected when the cursor is under', async () => {
-			const paragraphText = [
-				'Text to type',
-				'into the block',
-				'to test various',
-				'formatting options',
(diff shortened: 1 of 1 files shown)


derekblank:
Note for reviewers: this PR provides an example for how we could structure Rich Text Formatting Writing Flow tests. There are todo items remaining, including considerations for adding visual testing. Through this PR I am seeking to evaluate if this structure would provide adequate E2E test coverage for these cases, or if another strategy should be used (like moving these cases to the host apps, for example). Any feedback is welcome and appreciated. 🙇

__device-tests__/gutenberg-editor-writing-flow.test.js
```diff
@@ -0,0 +1,104 @@
...
+
+			// On a rich-text based component, add bold, italic, strikethrough and link formatted text
+			await editorPage.addNewBlock( blockNames.paragraph );
+			const paragraphBlockElement = await editorPage.getTextBlockAtPosition(
+				blockNames.paragraph
+			);
+
+			await editorPage.typeTextToTextBlock(
+				paragraphBlockElement,
+				paragraphHTML
```
dcalhoun: Rather than typing raw block markup, we might should either (A) initialize the post with the raw block markup via [`setHtmlContent`](https://github.com/WordPress/gutenberg/blob/dfdfaa3135b32af9018a41b832c2538c49adf747/packages/react-native-editor/__device-tests__/pages/editor-page.js#L271-L284) or (B) type plain text while also interacting with the formatting buttons, i.e. tapping the bold button and then writing text. The former would likely be a faster test, but markup could become outdated. 

All that said, use the second approach likely makes sense, even if it is a slower test, for the sake of stability.

__device-tests__/gutenberg-editor-writing-flow.test.js
```diff
@@ -0,0 +1,104 @@
...
+			);
+
+			// Select bold text
+			// TODO: verify if this is the best method to select text formatting
+			await editorPage.elementsByAccessibilityId( 'bold' );
```
dcalhoun: If we want to verify the resulting markup is accurate, we might should rely upon [`getHtmlContent`](https://github.com/WordPress/gutenberg/blob/dfdfaa3135b32af9018a41b832c2538c49adf747/packages/react-native-editor/__device-tests__/pages/editor-page.js#L261-L269) and assert with a [Jest snapshot](https://jestjs.io/docs/snapshot-testing). 

That said, a question worth asking is whether this test case should focus on markup or merely the status of formatting buttons. We may not need to assert that the markup is correct within this test. WDYT?

__device-tests__/gutenberg-editor-writing-flow.test.js
```diff
@@ -0,0 +1,104 @@
...
+
+			// Check that the proper format buttons get selected
+			// TODO: Identify best method to identify format buttons
+
+			// TODO: Invoke assertion or snapshot here?
```
dcalhoun: We could either assert the presence of specific styles on the buttons, i.e. find the button with a query and verify the styles are applied, or rely upon a visual snapshot. My opinion is that this is a perfect candidate for a visual snapshot. 

We would need to type formatted text, move the cursor to a specific position within the formatted text, and assert with a visual snapshot. 

WDYT?

__device-tests__/gutenberg-editor-writing-flow.test.js
```diff
@@ -0,0 +1,104 @@
...
+			);
+
+			// Type a sentence with a word misspelled that will be autocorrected by the editor
+			await editorPage.typeTextToTextBlock(
+				headingBlockElement,
+				headingText
+			);
+
+			// Highlight the word that was autocorrected
+			// Simulated double tap. TODO: Verify if this is the best method to highlight text
```
dcalhoun: There are several existing test helpers. One of which is [`tapSelectAllAboveElement`](https://github.com/WordPress/gutenberg/blob/dfdfaa3135b32af9018a41b832c2538c49adf747/packages/react-native-editor/__device-tests__/helpers/utils.js#L336-L344). So, there may be opportunity to re-use existing helpers. 

The quantity of test helpers is overwhelming to me, personally. There may be opportunity to refactor or remove some. I imagine the `tapSelectAllAboveElement` has value and is worth keeping/using, but I wanted to share my perspective on the helpers regardless for context.

__device-tests__/gutenberg-editor-writing-flow.test.js
```diff
@@ -0,0 +1,104 @@
...
+			await editorPage.tapElement( headingBlockElement );
+			await editorPage.tapElement( headingBlockElement );
+
+			// Check that the word has no formatting applied
+			const selectedText = await editorPage.getTextForParagraphBlockAtPosition(
+				1
+			);
+			// TODO: verify if a snapshot should be used in place of an assertion, or both
+			expect( selectedText ).toMatch( headingText );
```
dcalhoun: We _could_ use a snapshot. However, I do not think the current assertion is incorrect. It is a simple assertion and easy to maintain. 

IMO, snapshots are incredibly helpful for asserting against expectations that are non-trivial to write (writing lots of markup) or change frequently (markup could change). 

If we did use a snapshot here, I might select an inline snapshot given the output is so small.

__device-tests__/gutenberg-editor-writing-flow.test.js
```diff
@@ -0,0 +1,104 @@
+/**
+ * Internal dependencies
+ */
+const { blockNames } = editorPage;
+
+describe( 'Gutenberg Editor - Writing Flow', () => {
+	describe( 'TC007 - Test format detection under the cursor', () => {
```
dcalhoun, line 9: Do we believe there is benefit in retaining the `TC###` labels? My thought is that were useful when test cases were manual and needed to be referenced frequently by humans. However, for automated tests, I'm unsure they provide value. WDYT?

dcalhoun (commented):
Well done! This is a good start covering important interactions. 👏🏻 Thank you!

You may have already done so, but I encourage reviewing helpers in [`editor-page.js`](https://github.com/WordPress/gutenberg/blob/dfdfaa3135b32af9018a41b832c2538c49adf747/packages/react-native-editor/__device-tests__/pages/editor-page.js) and [`utils`](https://github.com/WordPress/gutenberg/blob/dfdfaa3135b32af9018a41b832c2538c49adf747/packages/react-native-editor/__device-tests__/helpers/utils.js#L336-L344). There are a lot, so it easy to miss helpers or struggle to understand their "correct" application. I struggle myself to make sure I use the "correct" ones frequently. 

> Through this PR I am seeking to evaluate if this structure would provide adequate E2E test coverage for these cases, or if another strategy should be used (like moving these cases to the host apps, for example).

I do believe the tests will provide adequate coverage. We could tweak them to have more coverage (use visual snapshots, interact with buttons rather than typing markup) or less coverage (initialize post with markup). Either approach is valid. It is a balance of coverage, speed, and stability. 

In regards to host apps, my thought is that we should put as many tests into the Gutenberg-related repositories as possible. This will further set them up as standalone, tested modules and allow us to catch regression earlier in the development cycle. 

There are some test cases that I believe will have to be tested in the host apps (e.g. media uploads) because the code for that feature is currently very spread across the codebases. Long term, it be ideal if even features such as media uploads were covered in tests within the Gutenberg repositories. That will likely require refactoring the current code and establishing an API (for host apps) that is sound, documented, and well tested. 

Let me know what you think!

derekblank:
Thanks for the feedback, that all sounds good. I'll review the inline notes, and respond from there. 👍

__device-tests__/gutenberg-editor-writing-flow.test.js
derekblank: > If we want to verify the resulting markup is accurate, we might should rely upon [getHtmlContent](https://github.com/WordPress/gutenberg/blob/dfdfaa3135b32af9018a41b832c2538c49adf747/packages/react-native-editor/__device-tests__/pages/editor-page.js#L261-L269) and assert with a [Jest snapshot](https://jestjs.io/docs/snapshot-testing).

In the code comment when I referenced "select text formatting", I was referring to the actual act of pressing the text formatting buttons themselves. In that context, the "select" verb referred to getting some sort of testID or accessibility label for the bold, italic, strikethrough buttons (etc) in the bottom button bar in order to press it. However, after reading your [comment](https://github.com/wordpress-mobile/gutenberg-mobile/pull/5610#pullrequestreview-1365244824) about reviewing the multitude of available test helpers to gain more insight, I became aware of the [`toggleFormatting`](https://github.com/WordPress/gutenberg/blob/dfdfaa3135b32af9018a41b832c2538c49adf747/packages/react-native-editor/__device-tests__/pages/editor-page.js#L585-L593) helper, which I believe will solve what my initial TODO comment was trying to verify. I've updated the test to type plain text and toggle the formatting buttons:

```
await editorPage.typeTextToTextBlock(
	paragraphBlockElement,
	'plaintext string'
);

// Toggle various formatting options, then proceed with typing more text
await editorPage.toggleFormatting( 'Bold' );
```

However, you raise a good points about whether or not we should verify the markup is accurate with an assertion:

> That said, a question worth asking is whether this test case should focus on markup or merely the status of formatting buttons. We may not need to assert that the markup is correct within this test. WDYT?

I definitely think checking the HTML and asserting the value would be helpful here. When I perform this test manually, I always switch to HTML mode to verify it, with the notion that the markup we're "sending" to the web is accurate and valid. I think replicating this step in the automated test would be valuable, so I'll update that as well, and thanks for the call-out. 👍

__device-tests__/gutenberg-editor-writing-flow.test.js
derekblank: I see both sides here, but feel that leaving the labels in does still provide some value while we are transitioning between manual and automated test cases. While the [Rich Text Formatting test cases](https://github.com/wordpress-mobile/test-cases/blob/trunk/test-cases/gutenberg/writing-flow/rich-text-formatting.md) are still extant, I feel that keeping the `TC###` numbers helps as a shorthand to refer and differentiate between the remaining seven test cases. Eventually, if we were to remove that doc altogether and only use automated tests, I think it would make sense to drop the `TC###` values from the test descriptions.

__device-tests__/gutenberg-editor-writing-flow.test.js
derekblank: > We would need to type formatted text, move the cursor to a specific position within the formatted text, and assert with a visual snapshot.

I follow and agree with using a visual snapshot here. 👍 

Could you describe more about the need to move the cursor to a specific position within the text, and what the expected result might be? I recognize that we could move the cursor via the various [`element.getLocation`](https://github.com/WordPress/gutenberg/blob/dfdfaa3135b32af9018a41b832c2538c49adf747/packages/react-native-editor/__device-tests__/helpers/utils.js#L336-L344)-based helpers, but don't quite follow the assertion we're looking for in doing so.

__device-tests__/gutenberg-editor-writing-flow.test.js
derekblank: Thanks for outlining this. I've [updated the code](https://github.com/wordpress-mobile/gutenberg-mobile/blob/7e7576602fc06c4baa3a49140550b158c2311fbd/__device-tests__/gutenberg-editor-writing-flow.test.js#L10-L51) for option (B), to type plaintext and interact with the formatting buttons. 

With this approach, do you feel like it would make sense to take a visual snapshot at each step? E.g.:

```
await editorPage.typeTextToTextBlock(
	paragraphBlockElement,
        'text'
);

// Toggle various formatting options, then proceed with typing more text
await editorPage.toggleFormatting( 'Bold' );

const screenshot = await takeScreenshot();
expect( screenshot ).toMatchImageSnapshot();

// ...repeat with taking a screenshot and an assertion
// for each of the remaining formatting options [?]
```

__device-tests__/gutenberg-editor-writing-flow.test.js
```diff
@@ -0,0 +1,127 @@
...
+			await tapSelectAllAboveElement(
+				editorPage.driver,
+				headingBlockElement
+			);
+
+			// Check that the word has no formatting applied
+			const selectedText = await editorPage.getTextForParagraphBlockAtPosition(
+				1
+			);
+			expect( selectedText ).toMatch( headingText );
```
dcalhoun: I _think_ it does, but wanted to ask to be sure: does `getTextForParagraphBlockAtPosition` return HTML for the input or the plain text value? If the it is the latter, then it wouldn't assert the lack of formatting. 

It may make most sense to assert the output of [`getHtmlContent`](https://github.com/WordPress/gutenberg/blob/dfdfaa3135b32af9018a41b832c2538c49adf747/packages/react-native-editor/__device-tests__/pages/editor-page.js#L261-L269) instead.

__device-tests__/gutenberg-editor-writing-flow.test.js
dcalhoun: > With this approach, do you feel like it would make sense to take a visual snapshot at each step?

Yes, I think visual snapshots is the most appropriate way to assert the visual styles of an active button.

__device-tests__/gutenberg-editor-writing-flow.test.js
dcalhoun: > I definitely think checking the HTML and asserting the value would be helpful here. When I perform this test manually, I always switch to HTML mode to verify it, with the notion that the markup we're "sending" to the web is accurate and valid. I think replicating this step in the automated test would be valuable, so I'll update that as well, and thanks for the call-out. 👍

My only objection to asserting the markup in this test is that we could likely assert the markup output in an integration test, or we may already do so now. That said, doing so here in this (required visually-driven) e2e test likely will not add much runtime to the test.

__device-tests__/gutenberg-editor-writing-flow.test.js
dcalhoun: > Could you describe more about the need to move the cursor to a specific position within the text, and what the expected result might be? I recognize that we could move the cursor via the various [`element.getLocation`](https://github.com/WordPress/gutenberg/blob/dfdfaa3135b32af9018a41b832c2538c49adf747/packages/react-native-editor/__device-tests__/helpers/utils.js#L336-L344)-based helpers, but don't quite follow the assertion we're looking for in doing so.

The formatting buttons toggle their active state based upon the text beneath the cursor position. Generally, when manually performing this test case, we type text with various formatting and then move the cursor around the text to verify the formatting buttons toggle their state correctly based upon the text beneath the cursor position. 

Ideally we assert the formatting buttons display correcting after moving the cursor around text with various formats applied.

__device-tests__/gutenberg-editor-writing-flow.test.js
```diff
@@ -0,0 +1,127 @@
...
+			await tapSelectAllAboveElement(
+				editorPage.driver,
+				headingBlockElement
+			);
+
+			// Check that the word has formatting applied
+			const selectedText = await editorPage.getTextForParagraphBlockAtPosition(
+				1
+			);
+			expect( selectedText ).toMatch( headingText );
```
dcalhoun, line 133: Similar to my other comment, it may make most sense to assert the output of [`getHtmlContent`](https://github.com/WordPress/gutenberg/blob/dfdfaa3135b32af9018a41b832c2538c49adf747/packages/react-native-editor/__device-tests__/pages/editor-page.js#L261-L269) instead. 

This currently asserts that `selectedText` matches `headingText` (`'Mispelled'`). The latter is a plain text string, but the test description states we expected formatting applied.

__device-tests__/gutenberg-editor-writing-flow.test.js
derekblank: Thanks for clarifying. I've updated the test to use `selectTextFromElement`, copying the technique from [gutenberg-editor-sanity-test-4](https://github.com/wordpress-mobile/gutenberg-mobile/blob/9df217c30ebf208a32f7b61f2d7104ba0a876306/__device-tests__/gutenberg-editor-sanity-test-4-visual.test.js#L104-L107). Selecting the text and applying the formatting toggles the active state of the formatting buttons.

1. Do you feel this approach (using the sequence of `typeTextToTextBlock` + `selectTextFromElement` + `toggleFormatting` + `takeScreenshot`) is adequate to cover the test case of checking that "the proper format buttons are selected when the cursor is under"? It appears to be adequate to me, but I could be misinterpreting the result.
2. There seems to be an issue with toggling the formatting of strikethrough. [This code](https://github.com/wordpress-mobile/gutenberg-mobile/blob/2f6a456c7a970c21cd4df24bb4379bda11088294/__device-tests__/gutenberg-editor-writing-flow.test.js#L52-L67) to generate the strikethrough formatting text appears to never run, and I have been unable to determine why. Does anything look obvious to you? I've included a video of the behavior when running the test, but it may also help to run the test locally if you have a moment.

https://user-images.githubusercontent.com/643285/231393331-2f99439d-6d57-414f-bc67-770305dee31b.mov

__device-tests__/gutenberg-editor-writing-flow.test.js
derekblank: > I think it does, but wanted to ask to be sure: does getTextForParagraphBlockAtPosition return HTML for the input or the plain text value? 

This seems like it should be trivial, but I could not figure out a way to log out the value for `getTextForParagraphBlockAtPosition` to the console but in order to answer this. What I was attempting to do was something like this:

```
const selectedText = await editorPage.getTextForParagraphBlockAtPosition(
	1
);

console.log( selectedText );
```

...and I would have expected that `selectedText` value would log to the terminal console in order to verify if it is sent as plain text or with HTML. Is this approach something that is not common (or possible) when developing with Appium? I even tried using [Appium Inspector](https://github.com/appium/appium-inspector), and also could not find anything in the documentation regarding logging to the console, so I feel like I'm missing something that should be more obvious.

Do you use another approach when wanting to log out values to the console? Or, how would you approach the question of whether or not `getTextForParagraphBlockAtPosition` returns HTML or a plain text value?

Regardless of that, I do agree that asserting the output of `getHtmlContent` would make more sense here, so I will update.

__device-tests__/gutenberg-editor-writing-flow.test.js
derekblank: > My only objection to asserting the markup in this test is that we could likely assert the markup output in an integration test, or we may already do so now.

That makes sense. I'll verify if this is already covered in an integration test, and assert here if not, or otherwise utilize the snapshots instead.

__device-tests__/gutenberg-editor-writing-flow.test.js
dcalhoun: > Do you feel this approach (using the sequence of typeTextToTextBlock + selectTextFromElement + toggleFormatting + takeScreenshot) is adequate to cover the test case of checking that "the proper format buttons are selected when the cursor is under"? 

Yes, I do believe it is adequate. It is not exactly how I personally execute the test manually, but there will always be various ways to accomplish a given task. 

> There seems to be an issue with toggling the formatting of strikethrough. [This code](https://github.com/wordpress-mobile/gutenberg-mobile/blob/2f6a456c7a970c21cd4df24bb4379bda11088294/__device-tests__/gutenberg-editor-writing-flow.test.js#L52-L67) to generate the strikethrough formatting text appears to never run, and I have been unable to determine why. Does anything look obvious to you? I've included a video of the behavior when running the test, but it may also help to run the test locally if you have a moment.

The following error outputted when running the test states the strikethrough button element is unfound. 

```
  ● Gutenberg Editor - Writing Flow › TC007 - Test format detection under the cursor › checks that the proper format buttons are selected when the cursor is under

    [waitForElementByXPath("//XCUIElementTypeButton[@name=\"Strikethrough\"]",{},2000)] Element condition wasn't satisfied!
```

My best guess for that occurring is that the strikethrough button is hidden because it is outside the visible bounds of the horizontal scroll view. We need to manually scroll the view and then execute the `toggleFormatting` helper or update the `toggleFormatting` helper to loop between querying and scrolling until it finds the target (similar to the [`findBlockButton`](https://github.com/WordPress/gutenberg/blob/8039d57822457b2a639142375beb6dac9e6bf468/packages/react-native-editor/__device-tests__/pages/editor-page.js#L428-L496) helper).

__device-tests__/gutenberg-editor-writing-flow.test.js
dcalhoun: If you wait long enough, the test ultimately outputs the following warning: 

```
"(//XCUIElementTypeButton[contains(@name, "Paragraph Block. Row 1")])" is still not visible after 25 retries!
```

This is a result of the test subject being a Heading block and attempting to utilize `getTextForParagraphBlockAtPosition`, which appears to only target Paragraph blocks. This is why you do not see the expected console log. 

After temporarily changing the test subject to a Paragraph, I logged `getTextForParagraphBlockAtPosition` and found it outputs a plain text value excluding HTML. So, asserting with this helper will not allow us to achieve the test's goal. We should use `getHtmlContent` as planned.

__device-tests__/gutenberg-editor-writing-flow.test.js
dcalhoun: Noting that we should apply [the planned use of `getHtmlContent`](https://github.com/wordpress-mobile/gutenberg-mobile/pull/5610#discussion_r1164201444) here as well.

PR #5679, opened 2023-04-18
Add VideoPress embed url to disabled cache list

jhnstn:
Partial fix for: https://github.com/wordpress-mobile/gutenberg-mobile/issues/5603

Depends on https://github.com/WordPress/gutenberg/pull/49901

To test:

- Apply this diff 
```diff
diff --git a/packages/react-native-editor/src/api-fetch-setup.js b/packages/react-native-editor/src/api-fetch-setup.js
index 9280903b65..f8ab8dd5bd 100644
--- a/packages/react-native-editor/src/api-fetch-setup.js
+++ b/packages/react-native-editor/src/api-fetch-setup.js
@@ -87,6 +87,8 @@ export const shouldEnableCaching = ( path ) => {
 		DISABLED_CACHING_ENDPOINTS
 	);
 
+	console.log( 'disabled caching paths', disabledCachingEndpoints );
+
 	const isDisabled = disabledCachingEndpoints.some( ( pattern ) =>
 		pattern.test( path )
 	);
```
2. Open the app with the local Metro server.
3. Verify the VideoPress embed url is present in the log message 

PR submission checklist:

### DIFF
--- a/src/setup.js
+++ b/src/setup.js
@@ -59,6 +59,15 @@ const setupHooks = () => {
 			};
 		}
 	);
+
+	// Hook to disable caching for specific endpoints.
+	addFilter(
+		'native.disable_caching_endpoints',
+		'gutenberg-mobile',
+		( endpoints ) => {
+			return [ ...endpoints, '/oembed.*videopress.com/i' ];
+		}
+	);
 };
 
 export default () => {


src/setup.js
```diff
@@ -59,6 +59,15 @@ const setupHooks = () => {
 			};
 		}
 	);
+
+	// Hook to disable caching for specific endpoints.
+	addFilter(
+		'native.disable_caching_endpoints',
+		'gutenberg-mobile',
+		( endpoints ) => {
+			return [ ...endpoints, '/oembed.*videopress.com/i' ];
```
fluiddot, line 68: I'm wondering if we could just simply disable the caching for all requests to `/oembed` endpoint (i.e. regex pattern. `/oembed\/1\.0\/proxy\?.*/i`). We usually use this endpoint via the `getEmbedPreview` selector that is already caching the response, so probably we don't need to have a second caching layer, WDYT?

jhnstn: Sure, I just didn't want to introduce too many changes in this PR and keep it focused on the immediate need. I'm fine with opening that up to all embeds if you think it's safe.

fluiddot: If we open it to all embeds, we could do it directly in Gutenberg so we won't need the WP hook and make any changes in GB-mobile.

jhnstn: Actually thinking if we want to avoid caching all of the embed routes we can do this in Gutenberg directly.

PR #5692, opened 2023-04-21
Enable VideoPress block on iOS

fluiddot:
**Related PRs:**
* https://github.com/wordpress-mobile/WordPress-iOS/pull/20580

This PR enables the VideoPress block on iOS as it will released first on that platform. In the next few days, we plan to ship the Android version.

**NOTE:** ⚠️ The VideoPress block only works on Simple WPCOM sites.

### iOS
1. Open the app and connect it with the local Metro server.
2. Open/create a post.
3. Tap ➕ button to open the block inserter.
4. Observe that the VideoPress block is displayed.
5. Tap on the VideoPress block to add it.
6. Add a video.
7. Observe that the video is displayed and can be played.

**Example:**

## Android

1. Follow the steps from the iOS test case.
2. Observe that the VideoPress block is displayed while in dev mode.
3. Tap on the three dots button -> Debug Menu -> Settings.
4. Disable JS Dev Mode.
5. Reload the editor by tapping on the three dots button -> Reload.
6. Observe that the VideoPress block is displayed as unsupported.
7. Tap ➕ button to open the block inserter.
8. Observe that the VideoPress block is NOT displayed.

**NOTE:** Repeat steps 3 and 4 (enable JS Dev Mode) to go back to dev mode.

**Example:**

PR submission checklist:

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -1,6 +1,6 @@
 Unreleased
 ---
-* [***] Enable VideoPress block on iOS [#5692]
+* [***] Enable VideoPress block (only on iOS and Simple WPCOM sites) [#5692]
 * [*] Allow content that is pasted into the post title to populate the post body correctly [#35733]
 
 1.93.1
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -3,6 +3,10 @@ Unreleased
 * [***] Enable VideoPress block on iOS [#5692]
 * [*] Allow content that is pasted into the post title to populate the post body correctly [#35733]
 
+1.93.1
+---
+* [**] Fix regression with the Color hook and ColorPanel [https://github.com/wordpress-mobile/gutenberg-mobile/pull/5683]
+
 1.93.0
 ---
 * [***] [iOS] Fixed iOS scroll jumping issue by refactoring KeyboardAwareFlatList improving writing flow and caret focus handling [https://github.com/wordpress-mobile/gutenberg-mobile/pull/5533]
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -1,5 +1,6 @@
 Unreleased
 ---
+* [***] Enable VideoPress block on iOS [#5692]
 * [*] Allow content that is pasted into the post title to populate the post body correctly [#35733]
 
 1.93.0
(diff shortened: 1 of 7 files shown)


jetpack
fluiddot, line 1: I bumped the Jetpack ref to incorporate the changes from https://github.com/Automattic/jetpack/pull/30071. Note that it's not pointing to the latest commit from `trunk`.

src/jetpack-editor-setup.js
```diff
@@ -31,7 +32,8 @@ const supportedJetpackBlocks = {
 		available: __DEV__,
 	},
 	'videopress/video': {
-		available: __DEV__,
+		// VideoPress block will be released first on iOS.
+		available: Platform.isIOS,
```
SiobhyB: Should we also keep this available on `__DEV__` to make it easier to work on Android locally?

fluiddot: Good point. I'll update it to use `__DEV__` on Android 👍 .

fluiddot: This has been fixed in https://github.com/wordpress-mobile/gutenberg-mobile/pull/5692/commits/ceee1a8398b989f74c3663f538d25eea577f77c3.

fluiddot: Note that I've reverted the update in the unit tests (https://github.com/wordpress-mobile/gutenberg-mobile/pull/5692/commits/19e0db1399f7115ea183d546409b6ef9e9281996). It's no longer needed as the block will be available on both platforms in the test environment.

RELEASE-NOTES.txt
```diff
@@ -1,5 +1,6 @@
 Unreleased
 ---
+* [***] Enable VideoPress block on iOS [#5692]
```
SiobhyB: I'm wondering if we should specify that this is only for simple WordPress.com sites? I know this limitation tripped a couple of people up when testing internally, so it could be worth clarifying for any users who might check out the release notes.

SiobhyB (approved):
LGTM! I left one small question/comment about the release notes, but nothing at all blocking, going ahead to approve :D 😎

RELEASE-NOTES.txt
fluiddot: Good point, I'll expand the entry specifying that it's enabled on iOS and only for simple WP sites 👍 . Similarly, I'll also update the release notes in https://github.com/wordpress-mobile/WordPress-iOS/pull/20580.

fluiddot: I've updated the release notes in https://github.com/wordpress-mobile/gutenberg-mobile/pull/5692/commits/30af943480f8bdd461356ea926f37a2c0973d1fd.

PR #5715, opened 2023-05-03
Add initial VideoPress test for Android

SiobhyB:
This PR introduces two key changes:

1. **Initial test for VideoPress block:** A new test has been added to ensure the successful insertion of the VideoPress block into the editor. This test checks the visibility of the VideoPress block and verifies the expected HTML output. For now, this test only runs on Android, iOS support will be introduced in a separate PR.

2. **Update to testMatch pattern:** The testMatch pattern in the project's Jest configuration has been adjusted to include subdirectories under the 'test' directory, thus enabling us to create a nested directory for VideoPress. 

To test: 

* Run `npm run test src/test/videopress/edit.js` from the project's directory.
* Ensure all tests pass as expected on this PR.

<hr />

PR submission checklist:


jest.config.js
```diff
@@ -27,7 +27,7 @@ module.exports = {
 	preset: './gutenberg/node_modules/react-native/jest-preset.js',
 	setupFiles: [ '<rootDir>/' + configPath + '/setup.js', './jest.setup.js' ],
 	setupFilesAfterEnv: [ '<rootDir>/' + configPath + '/setup-after-env.js' ],
-	testMatch: [ '<rootDir>/src/**/test/*.[jt]s?(x)' ],
+	testMatch: [ '<rootDir>/src/**/test/**/*.[jt]s?(x)' ],
```
fluiddot, line 30: In case want to add files with test helpers alongside the tests, this pattern would conflict. Not a blocker but we might adjust it in the future if we have this need.

src/test/videopress/edit.js
```diff
@@ -0,0 +1,66 @@
...
+	setupJetpackEditor( { blogId: 1, isJetpackActive: true } );
+	registerJetpackBlocks( defaultProps );
+} );
+
+afterAll( () => {
+	// Clean up registered blocks
+	getBlockTypes().forEach( ( block ) => {
+		unregisterBlockType( block.name );
+	} );
+} );
```
fluiddot: This part is not really needed when using `setupCoreBlocks` as it already unregisters all blocks ([reference](https://github.com/WordPress/gutenberg/blob/cd25a95a972c75ec999f06e816e8bdc28035e936/test/native/integration-test-helpers/setup-core-blocks.js#L36-L41)).

fluiddot (approved):
LGTM 🎊 !

src/test/videopress/edit.js
SiobhyB: Would we still need it to unregister the Jetpack blocks?

SiobhyB: Apologies for the silly question, slow moment from me! I looked at the linked reference and can see that all the blocks will be unregistered as part of that helper. Gone ahead to remove this cleanup in https://github.com/wordpress-mobile/gutenberg-mobile/pull/5715/commits/635d4bcc1b096e316d6015a9b6a8ac6822ecd191.

PR #5739, opened 2023-05-08
iOS XCFramework distribution – Part 2: Fetch React via CocoaPods

mokagio:
Sets up the React dependency integration via CocoaPods. Doing so result in a noisy diff, and also required additional tooling support, which is why I isolated it in a dedicated PR.

## Testing

If CI is green, we are good. The code doesn't actually use the dependencies anywhere, so the test is not that accurate. But, I already have the next step ready, #5740, and I don't think it's worth spending more time here.

PR submission checklist:


ios-xcframework/Podfile
```diff
@@ -0,0 +1,97 @@
...
+def gutenberg_dependencies
+  podspec_prefix = '..'
+
+  # FBReactNativeSpec needs special treatment because of react-native-codegen code generation
+  pod 'FBReactNativeSpec',
+      podspec: "#{podspec_prefix}/third-party-podspecs/FBReactNativeSpec/FBReactNativeSpec.podspec.json"
+
+  DEPENDENCIES.each do |pod_name|
+    pod pod_name, podspec: "#{podspec_prefix}/third-party-podspecs/#{pod_name}.podspec.json"
+  end
```
crazytonyli, line 65: Question: it appears these `DEPENDENCIES` podspec files are generated from [generate-podspecs.sh](https://github.com/wordpress-mobile/gutenberg-mobile/blob/1c64b735fc15ea47b64d3be447397e0ebc55d230/bin/generate-podspecs.sh#L76). Should we run that script before running `pod install` against this Podfile?

mokagio: Good question 🤔 

The outcome of running that script is tracked in the repo, and I think it's meant to run only when the React Native version changes, which is infrequently. So I'd say it's not really necessary. 

Having said that, it shouldn't hurt to run the script anyway, just in case?

Keen to hear what the folks more familiar with the Gutenberg setup think.

mokagio: One thing to remember if we go down that route is that at the moment the script is interactive:

https://github.com/wordpress-mobile/gutenberg-mobile/blob/1c64b735fc15ea47b64d3be447397e0ebc55d230/bin/generate-podspecs.sh#L29

We'll want to add an option to bypass that prompt.

mokagio: Actually, thinking about it, with the current setup running the script in CI would not work because of the lack of node modules 😞

crazytonyli: I'm not familiar with this repo at all, not to mention the release process. Please feel free to ignore this comment. But if I were to guess, all the podspecs in `third-party-podspecs` are dependecies of the `Gutenberg` pod, correct? If that's the case, we could just find all the podspec in the folder and use them add them as `pod ...`, instead of having a hard-coded list of dependencies?

crazytonyli (approved):

ios-xcframework/Podfile
mokagio: > all the podspecs in `third-party-podspecs` are dependecies of the `Gutenberg` pod, correct? If that's the case, we could just find all the podspec in the folder and use them add them as `pod ...,` instead of having a hard-coded list of dependencies?

I think that would work. I'll try it as a followup PR. It'd be really cool to remove that long list! 🙇‍♂️

fluiddot (commented):
Sorry for sharing this comment after the PR merge. I'd like to note that we already use Cocoapods in Gutenberg for the demo project ([reference](https://github.com/WordPress/gutenberg/blob/trunk/packages/react-native-editor/ios/Gemfile)), so I'm wondering if we could somehow unify this so we don't need to manage different Ruby and gems versions on both repositories (i.e. GB-mobile and Gutenberg). WDYT?

fluiddot:
~~@mokagio I'm also noticing that after this PR got merged into `trunk`, the job `Build iOS RN XCFramework` is failing on new commits ([example](https://buildkite.com/automattic/gutenberg-mobile/builds/5721#01882575-10c6-40fd-b216-e5b1f7ff114e)). Any idea what's producing the failure? Thanks 🙇 !~~

**UPDATE:** I realized this will be fixed with https://github.com/wordpress-mobile/gutenberg-mobile/pull/5779, so forget my above comment.

Issue #5734, opened 2023-05-05
Integration tests


fluiddot:
## Description

Add integration tests to cover the following three main areas of VideoPress block:
- [x] Upload video (https://github.com/wordpress-mobile/gutenberg-mobile/pull/5735)
- [x] Replace video (https://github.com/wordpress-mobile/gutenberg-mobile/pull/5825)
- [x] Block settings (https://github.com/wordpress-mobile/gutenberg-mobile/pull/5751, https://github.com/wordpress-mobile/gutenberg-mobile/pull/5754, https://github.com/wordpress-mobile/gutenberg-mobile/pull/5778)
- [ ] ~~Setting synchronization~~
- [x] Player + caption (https://github.com/wordpress-mobile/gutenberg-mobile/pull/5826)

fluiddot:
Heads up that integration tests on iOS might be affected by this issue: https://github.com/wordpress-mobile/gutenberg-mobile/issues/5742

fluiddot:
`Setting synchronization` item has been moved to a separate issue: https://github.com/wordpress-mobile/gutenberg-mobile/issues/5951

Issue #5742, opened 2023-05-08
Infinite recursion on integration tests when running them on iOS platform


fluiddot:
**Describe the bug**
While working on https://github.com/wordpress-mobile/gutenberg-mobile/issues/5734, I spotted that some of the tests I'm writing fail without a specific reason based on the test logic. The error I'm encountering is the following one, which is triggered when initializing the editor:

```
Ran 100000 timers, and there are still more! Assuming we've hit an infinite recursion and bailing out...

      26 |
      27 | 		// Advance all timers allowing store resolvers to resolve.
    > 28 | 		act( () => jest.runAllTimers() );
         | 		                ^
      29 |
      30 | 		// The store resolvers perform several API fetches during editor
      31 | 		// initialization. The most straightforward approach to ensure all of them

      at FakeTimers.runAllTimers (node_modules/@jest/fake-timers/build/legacyFakeTimers.js:178:13)
      at runAllTimers (gutenberg/test/native/integration-test-helpers/wait-for-store-resolvers.js:28:19)
      at callback (gutenberg/node_modules/@testing-library/react-native/src/act.ts:35:24)
      at act (gutenberg/node_modules/react/cjs/react.development.js:2512:16)
      at actImplementation (gutenberg/node_modules/@testing-library/react-native/src/act.ts:34:25)
      at gutenberg/test/native/integration-test-helpers/wait-for-store-resolvers.js:28:3
      at asyncGeneratorStep (gutenberg/node_modules/@babel/runtime/helpers/asyncToGenerator.js:3:24)
      at _next (gutenberg/node_modules/@babel/runtime/helpers/asyncToGenerator.js:25:9)
      at gutenberg/node_modules/@babel/runtime/helpers/asyncToGenerator.js:32:7
      at tryCallTwo (gutenberg/node_modules/promise/lib/core.js:45:5)
      at doResolve (gutenberg/node_modules/promise/lib/core.js:200:13)
      at new Promise (gutenberg/node_modules/promise/lib/core.js:66:3)
      at gutenberg/node_modules/@babel/runtime/helpers/asyncToGenerator.js:21:12
      at fn (gutenberg/test/native/integration-test-helpers/with-fake-timers.js:21:23)
      at asyncGeneratorStep (gutenberg/node_modules/@babel/runtime/helpers/asyncToGenerator.js:3:24)
      at _next (gutenberg/node_modules/@babel/runtime/helpers/asyncToGenerator.js:25:9)
[...]

PR #5751, opened 2023-05-10
VideoPress block: Initial setting tests

SiobhyB:
Part of the coverage efforts outlined in https://github.com/wordpress-mobile/gutenberg-mobile/issues/5734

## Description

The changes in this PR introduce tests that ensure the VideoPress block's title, description, and Playback settings can be successfully updated, as well as some of the settings in the Playback Bar Color and Privacy and Rating panels.

_Note, the following are not covered by this PR, and will be worked on in separate PRs:_

* Rating and Privacy picker selection.
* Various color selections for Playback Bar Color.
* Setting synchronization between the app and the web. 
 
## To test:

* Verify the HTML in [the snapshot file](https://github.com/wordpress-mobile/gutenberg-mobile/blob/f8736c8dd5a706deb525308e8a5e637e6804a331/src/test/videopress/__snapshots__/edit.js.snap) matches with the reality of what the block's HTML would look after making the relevant changes.
* Run the `TEST_RN_PLATFORM=ios npm run test src/test/videopress/edit.js` and `TEST_RN_PLATFORM=android npm run test src/test/videopress/edit.js` commands from the terminal to verify the tests pass as expected. Also, verify the tests on this PR pass.

<hr />

PR submission checklist:


src/test/videopress/edit.js
```diff
@@ -44,5 +55,116 @@ describe( 'VideoPress block', () => {
 
 		const expectedHtml = `<!-- wp:videopress/video /-->`;
 		expect( getEditorHtml() ).toBe( expectedHtml );
+
+		// Reset editor to avoid side effects from built-in timers in addBlock() function on iOS
+		await initializeEditor();
```
SiobhyB: Without this reset, the first test in the second test suite (and only the first test) was consistently failing (no matter what that test was). I believe this may be the result of the extra delay added to the `addBlock` helper in https://github.com/WordPress/gutenberg/pull/50381.

src/test/videopress/constants.js
```diff
@@ -0,0 +1,27 @@
...
+export const PLAYBACK_SETTINGS = [
+	'Autoplay',
+	'Loop',
+	'Muted',
+	'Show Controls',
+	'Play Inline',
+	'Preload Metadata',
+];
+
+export const PLAYBACK_BAR_COLOR_SETTINGS = [ 'Dynamic color' ];
```
SiobhyB: I plan to work on adding the color selections in a future PR.

src/test/videopress/local-helpers/constants.js
```diff
@@ -0,0 +1,23 @@
...
+export const PLAYBACK_SETTINGS = [
+	'Autoplay',
+	'Loop',
+	'Muted',
+	'Show Controls',
+	'Play Inline',
+	'Preload Metadata',
+];
+
+export const PLAYBACK_BAR_COLOR_SETTINGS = [ 'Dynamic color' ];
```
SiobhyB, line 18: This will be added to in an upcoming PR.

src/test/videopress/local-helpers/constants.js
fluiddot, line 1: Ah interesting, this will be also helpful for the upload test cases.

src/test/videopress/local-helpers/utils.js
```diff
@@ -0,0 +1,50 @@
...
+ */
+import { VIDEOPRESS_BLOCK_HTML } from './constants';
+
+export const initializeBlockWithHTML = async (
+	initialHtml = VIDEOPRESS_BLOCK_HTML
+) => {
+	const screen = await initializeEditor( { initialHtml } );
+
+	return { ...screen };
+};
```
fluiddot: Not sure if this function is reducing the test boilerplate 🤔 . It would be the same as doing:
```
const screen = await initializeEditor( { initialHtml: VIDEOPRESS_BLOCK_HTML  } );
```
vs
```
const screen = await initializeBlockWithHTML( VIDEOPRESS_BLOCK_HTML );
```
Do we expect to add more functionality in the future?

src/test/videopress/edit.js
fluiddot: @SiobhyB what error are you getting? Asking in case it's related to this issue https://github.com/wordpress-mobile/gutenberg-mobile/issues/5742.

In theory, we wouldn't need to initialize the editor after the test finishes 🤔 .

src/test/videopress/edit.js
```diff
@@ -44,5 +51,84 @@ describe( 'VideoPress block', () => {
...
+
+		changeTextOfTextInput( input, "The video's new description!" );
+	} );
+
+	/*
+	 * PLAYBACK SETTINGS
+	 * Loop through each of the playback settings and toggle on/off
+	 */
+	PLAYBACK_SETTINGS.forEach( ( setting ) => {
+		it( `should update Playback Settings section's ${ setting } setting`, async () => {
```
fluiddot, line 111: Jest provides [a function](https://jestjs.io/docs/api#testeachtablename-fn-timeout) to run the same tests with different data, sharing here in case you want to use it.

src/test/videopress/edit.js
```diff
@@ -44,5 +51,84 @@ describe( 'VideoPress block', () => {
...
+
+describe( "Update VideoPress block's settings", () => {
+	let screen;
+
+	beforeEach( async () => {
+		// Arrange
+		screen = await initializeBlockWithHTML();
+
+		await selectAndOpenBlockSettings( screen );
+	} );
```
fluiddot: Nitpick: Usually in integration tests, we have all the arrange, act, and expectation parts contained within the `it` function. This introduces a bit of redundancy but in general, helps readability as you can see in the same place what the test case is doing.

fluiddot (commented):
When running the tests I got several warnings with the following message
```
Warning: An update to Player inside a test was not wrapped in act(...).

    When testing, code that causes React state updates should be wrapped into act(...):

    act(() => {
      /* fire events that update state */
    });
```

Seems the player component is being updated after executing some of the actions. It would be great to incorporate a check for this in the test cases to avoid the warning.

src/test/videopress/edit.js
SiobhyB: @fluiddot, if you comment out this line and run `TEST_RN_PLATFORM=ios npm run test src/test/videopress/edit.js`, the first test in the second test suite (no matter what that test is) fails with the following:

```
    TypeError: Cannot read property 'presentPicker' of null

      195 | 			if ( autoOpen && isIOS ) {
      196 | 				this.pickerTimeout = setTimeout(
    > 197 | 					() => this.picker.presentPicker(),
          | 					                  ^
      198 | 					PICKER_OPENING_DELAY
      199 | 				);
      200 | 			} else {
```

If you navigate to that code directly and change the conditional so that `this.picker.presentPicker` runs without a delay, the test passes. Similarly, the test will also pass after commenting out the `addBlock` call, which makes me think that the failure is related to timing/delays, leading to this hack.

The PR related to the issue you linked to, https://github.com/WordPress/gutenberg/commit/023631c27cfecd9684d85a4bf92570158660edbf, doesn't resolve the issue unfortunately.

fluiddot: Ah ok, seems we'd need to run the timers after adding the block via the inserter menu to avoid this. I tried adding the following change and seemed to work.

```patch
diff --git forkSrcPrefix/src/test/videopress/edit.js forkDstPrefix/src/test/videopress/edit.js
index 7dec3ce42eff2e6b1af50e7cffe8eda9a32acf9a..94c975a65299c23d146f2109c77ae71247a38a8d 100644
--- forkSrcPrefix/src/test/videopress/edit.js
+++ forkDstPrefix/src/test/videopress/edit.js
@@ -45,6 +45,8 @@ describe( 'VideoPress block', () => {
 		// Add block
 		await addBlock( screen, 'VideoPress' );
 
+		jest.runOnlyPendingTimers();
+
 		// Get block
 		const videoPressBlock = await getBlock( screen, 'VideoPress' );
 		expect( videoPressBlock ).toBeVisible();
```

I was curious why I didn't experience this in the upload test cases. However, in that case, since the media picker is opened during the editor initialization, the timers are automatically run.

src/test/videopress/local-helpers/utils.js
SiobhyB: Ah, you're right, thanks for pointing this out. I don't expect to add further functionality here in the future, I believe it ended up this way due to some local explorations and changes. I've gone ahead to remove this as a helper in https://github.com/wordpress-mobile/gutenberg-mobile/pull/5751/commits/5293b8b128b97f0df329446a693c9c8c48ae4dea.

src/test/videopress/edit.js
SiobhyB: Happy to go with the standard here, updated in https://github.com/wordpress-mobile/gutenberg-mobile/pull/5751/commits/77aa2c9e8a0dd73ac4dca1bb0c7b7731c96eec44.

src/test/videopress/edit.js
SiobhyB: Ah, wow, thank you Carlos, such a seemingly simple solution to something I was racking my brain over. 🤦‍♀️ I've gone ahead to apply your fix in https://github.com/wordpress-mobile/gutenberg-mobile/pull/5751/commits/d56e303a6ab9570e5acd897aec9f06c8725af4c7.

SiobhyB:
@fluiddot thank you for the review 🙇‍♀️ 

> Seems the player component is being updated after executing some of the actions. It would be great to incorporate a check for this in the test cases to avoid the warning.

I've gone ahead to address this in https://github.com/wordpress-mobile/gutenberg-mobile/pull/5751/commits/881272b3af89171b610adaa9d32d3117bee6a86e, and believe I've addressed your other feedback too. Let me know if there's anything else that needs improving here.

src/test/videopress/edit.js
```diff
@@ -38,6 +45,8 @@ describe( 'VideoPress block', () => {
 		// Add block
 		await addBlock( screen, 'VideoPress' );
 
+		jest.runOnlyPendingTimers();
+
```
fluiddot: When running the command `TEST_RN_PLATFORM=ios npm run test src/test/videopress/edit.js` I'm getting the following warning:
```
 A function to advance timers was called but the timers APIs are not mocked with fake timers. Call `jest.useFakeTimers({legacyFakeTimers: true})` in this test file or enable fake timers for all tests by setting {'enableGlobally': true, 'legacyFakeTimers': true} in Jest configuration file.
    Stack Trace:

          46 | 		await addBlock( screen, 'VideoPress' );
          47 |
        > 48 | 		jest.runOnlyPendingTimers();
             | 		     ^
          49 |
          50 | 		// Get block
          51 | 		const videoPressBlock = await getBlock( screen, 'VideoPress' );

          Error:
          at FakeTimers._checkFakeTimers (node_modules/@jest/fake-timers/build/legacyFakeTimers.js:394:13)
          at FakeTimers.runOnlyPendingTimers (node_modules/@jest/fake-timers/build/legacyFakeTimers.js:189:10)
          at Object.runOnlyPendingTimers (src/test/videopress/edit.js:48:8)
          at asyncGeneratorStep (node_modules/@babel/runtime/helpers/asyncToGenerator.js:3:24)
          at _next (node_modules/@babel/runtime/helpers/asyncToGenerator.js:25:9)
          at tryCallOne (gutenberg/node_modules/promise/lib/core.js:37:12)
          at gutenberg/node_modules/promise/lib/core.js:123:15
          at flush (gutenberg/node_modules/asap/raw.js:50:29)
```

Probably we could use the `withFakeTimers` helper to address the issue:
```
// When the block is inserted, it automatically opens the media picker.
// On iOS, this picker is displayed using a timer, so we need to run it
// to allow any DOM update.
await withFakeTimers( () => jest.runOnlyPendingTimers() );
```

fluiddot (commented):
I noticed several logs with the following message when running the tests. Seems the block is trying to fetch the metadata via `fetchVideoItem` function but fails. If we don't need that data for these tests, maybe we could mock that function, WDYT?

```
videopress:lib:fetch-video-item updating retry from 0 to 1
```

src/test/videopress/edit.js
SiobhyB: Thank you, I've gone ahead to implement that in https://github.com/wordpress-mobile/gutenberg-mobile/pull/5751/commits/b0be14a9e46dcbff0aa1bfd02d910a67f7e8d615.

fluiddot (approved):
LGTM 🎊 !

I confirmed that all tests passed ✅ . Not a blocker, but we might consider adding the mock I suggested [here](https://github.com/wordpress-mobile/gutenberg-mobile/pull/5751#pullrequestreview-1422037209) to avoid the fetch failures when the block tries to sync the metadata.

SiobhyB:
Thanks @fluiddot! I'll go ahead to merge as-is and address mocking `fetchVideoItem` as part of a separate PR. 🙇‍♀️

PR #5754, opened 2023-05-10
VideoPress block: Add tests to verify Privacy and Rating settings

SiobhyB:
Part of the coverage efforts outlined in https://github.com/wordpress-mobile/gutenberg-mobile/issues/5734

## Description

* The changes in this PR iterate on the work done in https://github.com/wordpress-mobile/gutenberg-mobile/pull/5751 to add test coverage for the Privacy and Rating settings. 
* A new helper, [`pressSettingInPicker`](https://github.com/wordpress-mobile/gutenberg-mobile/blob/e46c8346c1670aed9190141987b6b4fd8296957c/src/test/videopress/local-helpers/utils.js#L62-L87), has been introduced to loop through and select each of the available options in the Privacy and Rating pickers.

## To test:

* Verify the HTML the changes to [the snapshot file](https://github.com/wordpress-mobile/gutenberg-mobile/pull/5754/files#diff-e223c3972444b923239b4ee32c702e552bee64ddf0b437a84ed292a5416097b6) matches with the reality of what the block's HTML would look after making the relevant changes.
* Run the `TEST_RN_PLATFORM=ios npm run test src/test/videopress/edit.js` and `TEST_RN_PLATFORM=android npm run test src/test/videopress/edit.js` commands from the terminal to verify the tests pass as expected. Also, verify the tests on this PR pass.

<hr />

PR submission checklist:


src/test/videopress/edit.js
```diff
@@ -117,6 +120,44 @@ describe( "Update VideoPress block's settings", () => {
 		} );
 	} );
 
+	/*
+	 * RATING SETTINGS
+	 * Loop through each of the possible ratings and select each one
+	 */
+	RATING_OPTIONS.forEach( ( option, index ) => {
+		// Skip the default setting, as it is already selected
+		if ( index === 0 ) return;
```
SiobhyB, line 138: Another reason behind skipping the default setting is that, out-of-the-box, it leads to a `Found multiple elements with text: G` error on Android. This is due to the fact the setting is already selected and used as a label for the picker component. I don't feel it's necessary to add test coverage for this case (selecting an already selected setting) so have opted for the most straightforward approach of skipping it. I'd be curious if others have different opinions on this.

src/test/videopress/local-helpers/utils.js
```diff
@@ -48,3 +49,39 @@ export const pressSettingInPanel = async ( screen, panel, setting ) => {
...
+	};
+
+	// Setup the picker and select option
+	const { selectOption } = setupPicker(
+		screen,
+		options,
+		pickerButtonPressed
+	);
+
+	await selectOption( option );
```
fluiddot: In relation to WordPress/gutenberg#50493, I've made the following change and I didn't experience any issue on iOS with the picker. @SiobhyB not sure if I'm overlooking anything, so I'm wondering if you could share the example that drove you to create WordPress/gutenberg#50493 to have a better context 🤔.

```suggestion
	// Setup the picker and select option
	const { selectOption } = setupPicker( screen, options );

	fireEvent.press( getByText( setting ) );

        // Use `act` here as render updates might happen after selecting an option.
	await act( () => selectOption( option ) );
```

src/test/videopress/local-helpers/constants.js
```diff
@@ -17,6 +17,10 @@ export const PLAYBACK_SETTINGS = [
 
 export const PLAYBACK_BAR_COLOR_SETTINGS = [ 'Dynamic color' ];
 
+export const RATING_OPTIONS = [ 'G', 'PG-13', 'R' ];
+
+export const PRIVACY_OPTIONS = [ 'Site default (Public)', 'Public', 'Private' ];
```
fluiddot, line 22: The first privacy option (`Site default (Public)`) can also be `Site default (Private)` if the site is private. I'm wondering if it would be worth adding a test case for this, WDYT? Note that it would require mocking the hook or the fetch request that returns the site privacy value.

src/test/videopress/edit.js
fluiddot: > Another reason behind skipping the default setting is that, out-of-the-box, it leads to a Found multiple elements with text: G error on Android. This is due to the fact the setting is already selected and used as a label for the picker component.

Maybe for this case, we could query the text using the [`within` helper](https://callstack.github.io/react-native-testing-library/docs/api#within-getqueriesforelement) to limit the scope to the `Privacy and Rating` screen.

> I don't feel it's necessary to add test coverage for this case (selecting an already selected setting) so have opted for the most straightforward approach of skipping it. I'd be curious if others have different opinions on this.

I agree it's not necessary to test selecting the already selected setting. The only part that could be tested is that after a selection is made, the option is displayed on the bottom sheet. Although, I think this logic should have been already tested by the unit tests of the `SelectControl` component.

SiobhyB: > Maybe for this case, we could query the text using the [within helper](https://callstack.github.io/react-native-testing-library/docs/api#within-getqueriesforelement) to limit the scope to the Privacy and Rating screen.

I considered this when putting together the PR, but made edits in other files to get this to work (as "Privacy and Rating" also appears multiple times within the component). That's what led me to believe this approach was the most straightforward option. Am I missing another way to achieve this perhaps?

src/test/videopress/local-helpers/utils.js
SiobhyB: Oh, thanks for this Carlos, this stems from me misunderstanding how the `setupPicker` should function. I had been calling the `fireEvent.press( getByText( setting ) )` prior to setting it up, with the believe it was supposed to be set up upon being called. Thank you again for correcting me understanding, I've addressed this in https://github.com/wordpress-mobile/gutenberg-mobile/pull/5754/commits/37692881b062e5329c8c84042e78733460bb203a.

SiobhyB:
~Closing, as discussed in https://github.com/wordpress-mobile/gutenberg-mobile/pull/5754#discussion_r1194962570.~ Edit: Wrong PR! 🤦‍♀️

gutenberg
fluiddot, line 1: Since this PR doesn't need recent changes from Gutenberg, we could consider not updating the Gutenberg ref, WDYT?

jetpack
fluiddot, line 1: I'm wondering if we need to bump the Jetpack reference, is there anything from recent changes we'd like to have in this PR?

src/test/videopress/edit.js
fluiddot: This approach is great to me, I just wanted to note that using the [`within` helper](https://callstack.github.io/react-native-testing-library/docs/api#within-getqueriesforelement) might help in case we'd like to query elements that display value is contained multiple times within the content.

fluiddot (approved):
Awesome work @SiobhyB 🏅 ! LGTM 🎊 !

jetpack
SiobhyB: Thanks for catching! I've reverted in https://github.com/wordpress-mobile/gutenberg-mobile/pull/5754/commits/7adeb1cb178098dae9013209980f9dba2040dfad.

gutenberg
SiobhyB: I've gone ahead to revert the change to the submodule in https://github.com/wordpress-mobile/gutenberg-mobile/pull/5754/commits/7adeb1cb178098dae9013209980f9dba2040dfad, thank you for catching this!

src/test/videopress/edit.js
SiobhyB: Thanks Carlos! If I'm missing a way to use it here, let me know, otherwise I'll bear this in mind for the future 🙇‍♀️

src/test/videopress/local-helpers/constants.js
SiobhyB: I think that's a great idea! I looked a bit into what was needed and see that it would involve mocking the `useSyncMedia` hook in order to get the `privateEnabledForSite` value, like [here](https://github.com/Automattic/jetpack/blob/3897da7aa9a4ac01cd04772ae9519903d46d60b4/projects/packages/videopress/src/client/block-editor/blocks/video/edit.native.js#L92-L93). I think that would be easier to get working when that hook has been mocked for testing the syncing, so will make a note to follow up with a different PR after that. :)

PR #5769, opened 2023-05-12
iOS XCFramework distribution – Part 4: Upload to S3

mokagio:
Builds on top of #5738 , #5739 , and #5740 to upload a ZIP with the Gutenberg XCFrameworks on S3.

To test: Verify CI runs successfully, then download the tar.gz from `https://d2twmm2nzpx3bg.cloudfront.net/Gutenberg-<commit sha1>.tar.gz`

I had to introduce a number of other changes to actually get the XCFrameoworks to work in WordPress, so this PR ended up a bit bloated. Most noteworthy:

- Changing the folder structure of the archive
- Adding `bundle/ios/App.js` to the XCFramework resources

You can verify the integration in [this test PR in WordPress iOS](https://github.com/wordpress-mobile/WordPress-iOS/pull/20792). Admittedly, in that build there are UI tests issues. They _could_ be related to the XCFramework packaging, but not with the distribution via S3 itself.

---

PR submission checklist:


ios-xcframework/fastlane/Fastfile
```diff
@@ -0,0 +1,33 @@
...
+  end
+
+  upload_to_s3(
+    bucket: 'a8c-apps-public-artifacts',
+    key: name,
+    file: xcframework_zip_path,
+    auto_prefix: false,
+    skip_if_exists: false
+  )
+end
```
mokagio, line 39: We could use Fastlane to generate the `xcframework`s at this point. But I consider that an incremental refinement rather than a precondition for the S3 upload.

.buildkite/pipeline.yml
```diff
@@ -23,42 +23,63 @@ steps:
...
     plugins:
-      - automattic/a8c-ci-toolkit#2.15.1
+      - automattic/a8c-ci-toolkit#2.16.0
       - peakon/git-shallow-clone#v0.0.1:
           depth: 1
     agents:
       queue: mac
     env:
-      IMAGE_ID: xcode-14.2
+      IMAGE_ID: xcode-14.3
```
mokagio: These updates were not required by the changes in the PR. I did them just to stay up to date.

.buildkite/publish-react-native-ios-artifacts.sh
```diff
@@ -1,19 +1,28 @@
...
-cd ./ios-xcframework
+# Make Homebrew run a bit faster
+export HOMEBREW_NO_AUTO_UPDATE=1
+export HOMEBREW_NO_INSTALL_CLEANUP=1
+
+echo "--- :arrow_down: Download iOS JS bundle"
+buildkite-agent artifact download bundle/ios/App.js .
 
 echo "--- :rubygems: Setting up Gems"
+cd ./ios-xcframework
```
mokagio: I could have left the `cd` at the start of the script and downloaded the artifact in `..`, but it seemed convoluted.

ios-xcframework/Config/Gutenberg-Shared.xcconfig
```diff
@@ -3,6 +3,12 @@ CLANG_ANALYZER_NONNULL = YES
 CLANG_ANALYZER_NUMBER_OBJECT_CONVERSION = YES_AGGRESSIVE
 CLANG_CXX_LANGUAGE_STANDARD = gnu++20
 CLANG_ENABLE_MODULES = YES
+// If CLANG_ENABLE_MODULE_DEBUGGING is YES (or not specified, which defaults to YES), then CocoaPods says:
+//
+// Linking a static library that was built with -gmodules, but the module cache was not found.
+// Redistributable static libraries should never be built with module debugging enabled.
+// The debug experience will be degraded due to incomplete debug information.
+CLANG_ENABLE_MODULE_DEBUGGING = NO;
```
mokagio, line 11: This change was [part of the previous PR](https://github.com/wordpress-mobile/gutenberg-mobile/pull/5740). For it to be here, I must have messed up a `git rebase` resulting in that commit not landing in `trunk`.

ios-xcframework/XCFrameworkScaffold.xcodeproj/project.pbxproj
```diff
@@ -26,6 +26,7 @@
 		3F12364D29F6B21300AF54A4 /* editor-behavior-overrides.js in Resources */ = {isa = PBXBuildFile; fileRef = 3F12364429F6B21300AF54A4 /* editor-behavior-overrides.js */; };
 		3F12364E29F6B21300AF54A4 /* local-storage-overrides.json in Resources */ = {isa = PBXBuildFile; fileRef = 3F12364529F6B21300AF54A4 /* local-storage-overrides.json */; };
 		3F12365029F6B23300AF54A4 /* supported-blocks.json in Resources */ = {isa = PBXBuildFile; fileRef = 3F12364F29F6B23300AF54A4 /* supported-blocks.json */; };
+		3FA0C5812A1C8C9700600A9A /* App.js in Resources */ = {isa = PBXBuildFile; fileRef = 3FA0C5802A1C8C9700600A9A /* App.js */; };
```
mokagio, line 29: Gutenberg needs the "app bundle" at runtime in the client apps.

ios-xcframework/build.sh
```diff
@@ -97,15 +97,52 @@ do
...
+# For CocoaPods to find the XCFrameworks in the archive, it needs to have a certain folder structure.
+#
+# See for example what Firebase does,
+# https://github.com/CocoaPods/Specs/blob/master/Specs/e/2/1/FirebaseAnalytics/9.6.0/FirebaseAnalytics.podspec.json
+#
+# Archive/
+#   - Frameworks/
+#     - A.xcframework
+#     - B.xcframework
+#   - dummy.txt
```
mokagio, line 124: Note that even though I put all this in place, I'm still unable to get CocoaPods to find the xcframeworks in WordPress iOS 😭

I decided to leave the setup here in the hope to figure out how to make it work in the future

ios-xcframework/podspec.erb
```diff
@@ -0,0 +1,27 @@
...
+  }
+
+  s.vendored_frameworks = [
+    'Frameworks/Aztec.xcframework',
+    'Frameworks/Gutenberg.xcframework',
+    'Frameworks/React.xcframework',
+    'Frameworks/RNTAztecView.xcframework',
+    'Frameworks/yoga.xcframework'
+  ]
+end
```
mokagio, line 27: Once we'll have proven XCFrameworks are stable to use, we shall update the release automation to use this template in updating the main `Gutenberg.podspec`.

.buildkite/pipeline.yml
```diff
@@ -66,9 +66,9 @@ steps:
     depends_on: js-bundles
     command: .buildkite/publish-react-native-ios-artifacts.sh
     artifact_paths:
-      - ios-xcframework/build/xcframeworks/*.zip
+      - ios-xcframework/build/xcframeworks/*.tar.gz
```
jkmassel, line 69: Just a heads up that if we want to distribute this using SwiftPM we'll need to publish a ZIP version as well

ios-xcframework/fastlane/Fastfile
```diff
@@ -0,0 +1,84 @@
...
+
+# rubocop:disable Metrics/AbcSize
+# rubocop:disable Metrics/MethodLength
+def upload_podspec_to_s3(id:)
+  ios_version = '13.0'
+  swift_version = '5.0'
+  gutenberg_package = JSON.parse(File.read(File.join(XCFRAMEWORK_PROJECT_ROOT_FOLDER, '..', 'package.json')))
```
jkmassel:
```suggestion
  gutenberg_package = JSON.parse(File.read(File.join(File.dirname(XCFRAMEWORK_PROJECT_ROOT_FOLDER), 'package.json')))
```

This is a nit, but is more portable?

ios-xcframework/fastlane/Fastfile
```diff
@@ -0,0 +1,84 @@
...
+
+  Dir.mktmpdir do |tmp_dir|
+    podspec_path = File.join(tmp_dir, 'podspec')
+    File.open(podspec_path, 'w') do |file|
+      file.write(podspec_erb.result(context))
+    end
+
+    upload_to_s3(
+      bucket: S3_BUCKET,
+      key: "Gutenberg-#{id}.podspec",
```
jkmassel:
```suggestion
      key: "gutenberg-mobile/Gutenberg-#{id}.podspec",
```

Can we namespace this? Makes browsing the bucket easier

ios-xcframework/podspec.erb
```diff
@@ -0,0 +1,27 @@
...
+
+  s.version = '<%= version %>'
+
+  s.license = { type: '<%= license %>' }
+
+  s.ios.deployment_target = '<%= ios_version %>'
+  s.swift_version = '<%= swift_version %>'
+
+  s.source = {
+    http: "https://d2twmm2nzpx3bg.cloudfront.net/Gutenberg-<%= id %>.tar.gz"
```
jkmassel:
```suggestion
    http: "https://cdn.a8c-ci.services/Gutenberg-<%= id %>.tar.gz"
```

Fun little polish 🙂

ios-xcframework/fastlane/Fastfile
```diff
@@ -0,0 +1,84 @@
...
+
+  unless File.exist? xcframework_archive_path
+    UI.user_error! "Could not find XCFramework archive to upload at #{xcframework_archive_path}"
+  end
+
+  name = options.fetch(:name, nil)
+
+  if name.nil?
+    name = "Gutenberg-#{last_git_commit[:commit_hash]}.#{archive_type}"
+    UI.message "No name provide, will default to #{name}"
```
jkmassel:
```suggestion
    UI.message "No name provided, will default to #{name}"
```

typo

ios-xcframework/fastlane/Fastfile
```diff
@@ -0,0 +1,84 @@
...
+  upload_xcframework_to_s3(
+    path: xcframework_archive_path,
+    name_on_s3: "Gutenberg-#{tag}.#{archive_type}"
+  )
+end
+
+def upload_xcframework_to_s3(path:, name_on_s3:)
+  upload_to_s3(
+    bucket: S3_BUCKET,
+    key: name_on_s3,
```
jkmassel: Can we namespace this? Makes browsing the bucket easier?
```suggestion
    key: "gutenberg-mobile#{name_on_s3}",
```

jkmassel (dismissed):

ios-xcframework/fastlane/Fastfile
mokagio: @jkmassel the function make the name customizabile, but it's practically already namespaced because the only usages we have ATM are "Gutenberg-..." 

Do you think that's enough?

I could make the code less flexible and bake the prefix at this level, with the caller only able to pass the SHA1 / tag value to concatenate.

ios-xcframework/podspec.erb
mokagio: Excellent! I wasn't aware of that URL. Much neater 😄

.buildkite/pipeline.yml
mokagio: It should be trivial to add when the time comes. 

I haven't looked into that, but I assume we'll also need a different package structure (e.g. without `dummy.txt`). That, too, will be a matter of calling `zip` with the appropriate params.

ios-xcframework/fastlane/Fastfile
mokagio: Appreciated 👍

ios-xcframework/fastlane/Fastfile
jkmassel: Ah oops – I forgot to include the `/` 🤦 

The "directory" part of the path helps organize the list as seen here:

Screenshot 2023-06-06 at 9 08 36 AM

.buildkite/pipeline.yml
crazytonyli: It seems a bit wasteful that same files are duplicated in two places, one in Buildkite's artifactory and the other in our public S3 bucket. It'd be nice if we can upload a "redirect link" as a buildkite artifact which takes use to the files in the public S3 bucket.

.buildkite/pipeline.yml
```diff
@@ -66,9 +66,9 @@ steps:
     depends_on: js-bundles
```
crazytonyli, line 66: Add "& Publish to S3" to match the Android steps?

Screenshot 2023-06-08 at 11 55 02 AM

ios-xcframework/BuildPhases/check_app_js.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash -eu
+
+APP_JS_PATH="../bundle/ios/App.js"
+
+if [[ ! -f "$APP_JS_PATH" ]]; then
+  cat <<- ERROR
+error: Could not find required $APP_JS_PATH.
+
+Please run the following command to generate the file before running this script:
```
crazytonyli: I presume "this script" means the `ios-xcframework/build.sh` script? I wonder if we should be explicit or say "before building this project" instead?

ios-xcframework/build.sh
```diff
@@ -97,32 +97,49 @@ do
...
-# that version is already available.
-log 'compression' 'Zipping React XCFrameworks'
-ZIP_PATH=$XCFRAMEWORKS_DIR/ReactNative.zip
-zip -rq "$ZIP_PATH" $XCFRAMEWORKS_DIR -x \
-  $XCFRAMEWORKS_DIR/Aztec.xcframework \
-  $XCFRAMEWORKS_DIR/Gutenberg.xcframework \
-  $XCFRAMEWORKS_DIR/RNTAztecView.xcframework \
-  $XCFRAMEWORKS_DIR/Pods_Gutenberg.xcframework
-echo "React Native XCFrameworks ZIP generated at $ZIP_PATH"
+# TODO: There has to be a RegEx that can do what this sequence does in one go. Something along the lines of s/(^|\s|\[)Gutenberg\.(?!self)/\1/g but that doesn't throw errors
```
crazytonyli, line 107: Here is an answer I got from ChatGPT: https://regex101.com/r/W38ky9/1. But I'm not sure if it'd cover all your cases...

An alternative could be using https://github.com/apple/swift-syntax to parse and update the `.swiftinterface` file.

ios-xcframework/fastlane/Fastfile
```diff
@@ -0,0 +1,73 @@
...
+# rubocop:disable Metrics/AbcSize
+# rubocop:disable Metrics/MethodLength
+def upload_podspec_to_s3(id:)
+  ios_version = '13.0'
+  swift_version = '5.0'
+  gutenberg_package = JSON.parse(File.read(File.join(File.dirname(XCFRAMEWORK_PROJECT_ROOT_FOLDER), 'package.json')))
+  version = gutenberg_package['version']
+  # This ought to be read from package.json, but there isn't one at the moment.
+  # license = GUTENBERG_PACKAGE['license']
+  license = 'GPL'
```
crazytonyli: Have you considered moving the hard-coded iOS version, Swift version, and license to the ERB template? I feel like it's probably more easy and intuitive to update those fields in `podpsec.erb` than this script.

ios-xcframework/fastlane/Fastfile
```diff
@@ -0,0 +1,73 @@
...
+  context = binding
+
+  podspec_erb = ERB.new(File.read(File.join(XCFRAMEWORK_PROJECT_ROOT_FOLDER, 'podspec.erb')))
+
+  Dir.mktmpdir do |tmp_dir|
+    podspec_path = File.join(tmp_dir, 'podspec')
+    File.open(podspec_path, 'w') do |file|
+      file.write(podspec_erb.result(context))
+    end
```
crazytonyli: Nitpick: We could create the podspec file in one line: `File.write(podspec_path, podspec_erb.result(context))`

ios-xcframework/fastlane/Fastfile
```diff
@@ -0,0 +1,73 @@
...
+  upload_to_a8c_s3(file: xcframework_archive_path, name_on_s3: "Gutenberg-#{tag}.#{archive_type}")
+end
+
+# rubocop:disable Metrics/AbcSize
+# rubocop:disable Metrics/MethodLength
+def upload_podspec_to_s3(id:)
+  ios_version = '13.0'
+  swift_version = '5.0'
+  gutenberg_package = JSON.parse(File.read(File.join(File.dirname(XCFRAMEWORK_PROJECT_ROOT_FOLDER), 'package.json')))
+  version = gutenberg_package['version']
```
crazytonyli, line 53: I noticed the version in xcframework's Info.plist was 1.0. It might be good to keep versions in these different files consistent.

ios-xcframework/build.sh
mokagio: Thanks! I'll keep that for a followup PR I that's okay.

mokagio: Tracked at internal ref paaHJt-4WG-p2#comment-7616

.buildkite/pipeline.yml
mokagio: Good point. Addressed in 83deadc9aed73a6b693f86e471ec66718442af3a

In the future, maybe we could split the build from the upload, to use less macOS agent time. Then again, uploading doesn't take that long and it might be silly to upload to Buildkite, download from Buildkite, and finally upload to S3 🤔

mokagio: > Then again, uploading doesn't take that long and it might be silly to upload to Buildkite, download from Buildkite, and finally upload to S3 🤔

Ahah, @crazytonyli I just read your [comment below](https://github.com/wordpress-mobile/gutenberg-mobile/pull/5769/files/57f9dd1c4233913696578d89194b7527399be189#r1222286309). 😄

.buildkite/pipeline.yml
mokagio: @crazytonyli 👍 sounds like a good followup

ios-xcframework/BuildPhases/check_app_js.sh
mokagio: Good point. Addressed in 2c4bd9326044c285ca81d1d6c2b8592da2c8a695

ios-xcframework/fastlane/Fastfile
mokagio: Sounds good. I don't remember why I decided to inject them from here.

mokagio: See 9962787ce3019a319eb4949ed3372dd91bf338d9

ios-xcframework/fastlane/Fastfile
mokagio: Neat. Addressed in 44a114d47da905713c875b31ffc9a122bd4ffade

ios-xcframework/fastlane/Fastfile
mokagio: Good point. 

🤔 Maybe we could move the `CFBundleVersion...` from hardcoded to a build setting, then update that value at build time via a build phase that reads it from the `package.json`.

Alternatively, we could edit it with `plutil` in the XCFramework build script

Tracked in paaHJt-4WG-p2#comment-7617

mokagio:
@crazytonyli @jkmassel I should have either addressed or tracked all your concerns and suggestions.

At the time of writing, the `tar.gz` for the latest commit hasn't been published yet, but here's a [link to a recent CI build](https://buildkite.com/automattic/wordpress-ios/builds/14584) using one of the archives published from here.

The UI tests are failing, but I doubt that has anything to do with how the package is uploaded to S3, which is the focus of this PR.

.buildkite/pipeline.yml
crazytonyli: Yep, not a blocker at all. I have had a quick research and don't think Buildkite supports uploading a "redirect link". But it does support [saving artifact to a S3 bucket](https://buildkite.com/docs/agent/v3/cli-artifact#using-your-private-aws-s3-bucket). Many we can configure the pipeline to tell Buildkite to save the tar.gz and podspec file to the correct location in our public artifacts S3 bucket.

ios-xcframework/podspec.erb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+
+Pod::Spec.new do |s|
+  s.name = 'Gutenberg'
+  s.summary = 'Printing since 1440'
+  s.homepage = 'https://github.com/wordpress-mobile/gutenberg-mobile'
+  s.authors = 'Automattic'
+
+  s.version = '<%= JSON.parse(File.read(File.join(File.dirname(File.dirname(File.expand_path(__dir__))), 'package.json')))['version'] %>'
```
crazytonyli: I find this file path a bit confusing, because the `__dir__` here is actually the directory of the `Fastfile`, not directory of this erb file. We can tell the ERB library the template file path, which would make this part of code easier to understand.

```diff
diff --git a/ios-xcframework/fastlane/Fastfile b/ios-xcframework/fastlane/Fastfile
index 913612dc..1fbd8a73 100644
--- a/ios-xcframework/fastlane/Fastfile
+++ b/ios-xcframework/fastlane/Fastfile
@@ -39,7 +39,9 @@ end
 def upload_podspec_to_s3(id:)
   context = binding

-  podspec_erb = ERB.new(File.read(File.join(XCFRAMEWORK_PROJECT_ROOT_FOLDER, 'podspec.erb')))
+  filepath = File.join(XCFRAMEWORK_PROJECT_ROOT_FOLDER, 'podspec.erb')
+  podspec_erb = ERB.new(File.read(filepath))
+  podspec_erb.filename = filepath

   Dir.mktmpdir do |tmp_dir|
     podspec_path = File.join(tmp_dir, 'podspec')
diff --git a/ios-xcframework/podspec.erb b/ios-xcframework/podspec.erb
index feecc3d4..525baa49 100644
--- a/ios-xcframework/podspec.erb
+++ b/ios-xcframework/podspec.erb
@@ -6,7 +6,7 @@ Pod::Spec.new do |s|
   s.homepage = 'https://github.com/wordpress-mobile/gutenberg-mobile'
   s.authors = 'Automattic'

-  s.version = '<%= JSON.parse(File.read(File.join(File.dirname(File.dirname(File.expand_path(__dir__))), 'package.json')))['version'] %>'
+  s.version = '<%= JSON.parse(File.read(File.expand_path('../package.json', __dir__)))['version'] %>'

   <%# This ought to be read from package.json, but there isn't one at the moment. %>
   s.license = { type: 'GPL2' }
```

mokagio: Great catch. I didn't realize the inconsistency. I kept tweaking the path till it worked 😅 

What do you think of this approach?

```diff
--- a/ios-xcframework/fastlane/Fastfile
+++ b/ios-xcframework/fastlane/Fastfile

@@ -39,12 +39,18 @@ end
 def upload_podspec_to_s3(id:)
   context = binding
 
-  podspec_erb = ERB.new(File.read(File.join(XCFRAMEWORK_PROJECT_ROOT_FOLDER, 'podspec.erb')))
-
-  Dir.mktmpdir do |tmp_dir|
-    podspec_path = File.join(tmp_dir, 'podspec')
-    File.write(podspec_path, podspec_erb.result(context))
-    upload_to_a8c_s3(file: podspec_path, name_on_s3: "Gutenberg-#{id}.podspec")
+  # chdir into the XCFramework project's root so that when the ERB access its
+  # path, it's the correct one relative to its location, not the one altered by
+  # the Fastlane execution.
+  Dir.chdir(XCFRAMEWORK_PROJECT_ROOT_FOLDER) do
+    podspec_erb = ERB.new(File.read(File.join(XCFRAMEWORK_PROJECT_ROOT_FOLDER, 'podspec.erb')))
+
+    Dir.mktmpdir do |tmp_dir|
+      podspec_path = File.join(tmp_dir, 'podspec')
+      File.write(podspec_path, podspec_erb.result(context))
+      upload_to_a8c_s3(file: podspec_path, name_on_s3: "Gutenberg-#{id}.podspec")
+    end
   end
 end
 
--- a/ios-xcframework/podspec.erb
+++ b/ios-xcframework/podspec.erb
@@ -6,7 +6,7 @@ Pod::Spec.new do |s|
   s.homepage = 'https://github.com/wordpress-mobile/gutenberg-mobile'
   s.authors = 'Automattic'
 
-  s.version = '<%= JSON.parse(File.read(File.join(File.dirname(File.dirname(File.expand_path(__dir__))), 'package.json')))['version'] %>'
+  s.version = '<%= JSON.parse(File.read(File.join(File.dirname(File.dirname(File.expand_path(__FILE__))), 'package.json')))['version'] %>'
 ```

Alternatively, I could take a step back on making the ERB smart, and make it expect the version value to be injected. Doing so would be self-explanatory, whereas both solutions we brought forward so far need some additional explanation/understating.

Comment (crazytonyli) [reply ios-xcframework/podspec.erb]: I think reverting the ERB file to expect the version value makes sense to me. The template would become slightly simpler and all the variable setting are done in the Fastfile. 👍

Comment (mokagio) [reply ios-xcframework/podspec.erb]: Addressed in a4570b03, plus realized that the iOS and Swift versions were already isolated, so read them from `xcconfig` in 70bf0607.

I tested the result with this diff

```diff
--- a/ios-xcframework/fastlane/Fastfile
+++ b/ios-xcframework/fastlane/Fastfile
@@ -26,7 +26,7 @@ lane :upload_xcframework_to_s3 do |options|
     UI.message "No name provided, will default to #{name}"
   end
 
-  upload_to_a8c_s3(file: xcframework_archive_path, name_on_s3: name)
+  # upload_to_a8c_s3(file: xcframework_archive_path, name_on_s3: name)
   upload_podspec_to_s3(id: last_git_commit[:commit_hash])
 
   next unless is_ci
@@ -62,7 +62,8 @@ def upload_podspec_to_s3(id:)
   Dir.mktmpdir do |tmp_dir|
     podspec_path = File.join(tmp_dir, 'podspec')
     File.write(podspec_path, podspec_erb.result(context))
-    upload_to_a8c_s3(file: podspec_path, name_on_s3: "Gutenberg-#{id}.podspec")
+    puts File.read(podspec_path)
+    # upload_to_a8c_s3(file: podspec_path, name_on_s3: "Gutenberg-#{id}.podspec")
   end
 end
 # rubocop:enable Metrics/AbcSize
```

crazytonyli (approved):
Thanks for addressing my feedbacks!

Considering we save build artifacts for every commit (maybe not every single one of them, but most of them), and we can't delete the saved artifacts, the S3 bucket is going to keep increasing. I wonder if we should configure the S3 bucket to move old artifacts to [Glacier storage class](https://aws.amazon.com/s3/storage-classes/).

mokagio:
@crazytonyli 

> Thanks for addressing my feedbacks!

Thank you for all the help!

> Considering we save build artifacts for every commit (maybe not every single one of them, but most of them), and we can't delete the saved artifacts, the S3 bucket is going to keep increasing. I wonder if we should configure the S3 bucket to move old artifacts to [Glacier storage class](https://aws.amazon.com/s3/storage-classes/).

@jkmassel @oguzkocer I don't think the bucket we're using here is the same as the one for the Android counterpart, but maybe this is something you have thought about already?

oguzkocer:
> @jkmassel @oguzkocer I don't think the bucket we're using here is the same as the one for the Android counterpart, but maybe this is something you have thought about already?

Yeap, it's a different bucket.

PR #5772, opened 2023-05-12
Update VideoPress block support from dev-only to iOS-only

fluiddot:
Now that the VideoPress block has been released on iOS, we should update the Supported block list.

**To test:**
N/A

PR submission checklist:


SiobhyB (approved):
Nice catch, thanks for updating!

PR #5806, opened 2023-05-26
Build both Android and iOS bundles in Buildkite CI

mokagio:
Building the iOS JS bundle achieves two goals:

- Allows the step that builds the XCFramework to add the bundle as a resource without spending time setting up all the node stack. This is efficient because we make use of the work that has already run for the Android bundle
- Brings us one step closer to being able to run all our CI tasks in Buildkite over Circle CI

_These were originally part of #5769 but I keep running into last mile issues to the get generated and downloaded XCFramework to work, so I thought I'd track these orthogonal changes in a dedicated PR._

To test: Verify build was successful in Buildkite

image

PR submission checklist:

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -74,4 +74,4 @@ steps:
     agents:
       queue: mac
     env:
-      IMAGE_ID: xcode-14.2
+      IMAGE_ID: xcode-14.3
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -13,8 +13,8 @@ steps:
   - block: "Request trigger Android bundle and builds"
     branches: "dependabot/submodules/*"
 
-  - label: "Bundle Android"
-    key: "bundle-android"
+  - label: "Build JS Bundles"
+    key: "js-bundles"
     plugins:
       - docker#v3.8.0:
           image: "public.ecr.aws/automattic/gb-mobile-image:latest"
@@ -34,6 +34,8 @@ steps:
 
         buildkite-agent artifact upload bundle/android/App.js
 
+        npm run bundle:ios
+        buildkite-agent artifact upload bundle/ios/App.js
   - label: "Build Android RN Aztec & Publish to S3"
     key: "publish-react-native-aztec-android"
     plugins:
@@ -43,7 +45,7 @@ steps:
 
   - label: "Build Android RN Bridge & Publish to S3"
     depends_on:
-      - "bundle-android"
+      - "js-bundles"
       - "publish-react-native-aztec-android"
     plugins:
       - *publish-android-artifacts-docker-container
@@ -51,6 +53,7 @@ steps:
         .buildkite/publish-react-native-bridge-android-artifacts.sh
 
   - label: Build iOS RN XCFramework
+    depends_on: js-bundles
     command: .buildkite/publish-react-native-ios-artifacts.sh
     artifact_paths:
       - ios-xcframework/build/xcframeworks/*.zip
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -26,15 +26,25 @@ steps:
     command: |
         source /root/.bashrc
 
+        echo "--- :node: Setup Node environment"
         nvm install && nvm use
+
+        echo "--- :npm: Install Node dependencies"
         npm ci --unsafe-perm --prefer-offline --no-audit --no-progress
 
+        echo "--- :package: Run bundle prep work"
         npm run prebundle:js
+
+        echo "--- :android: Build Android bundle"
         npm run bundle:android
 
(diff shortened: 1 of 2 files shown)


.buildkite/pipeline.yml
```diff
@@ -23,34 +23,54 @@ steps:
...
         source /root/.bashrc
 
+        echo "--- :node: Setup Node environment"
         nvm install && nvm use
+
+        echo "--- :npm: Install Node dependencies"
         npm ci --unsafe-perm --prefer-offline --no-audit --no-progress
 
+        echo "--- :package: Run bundle prep work"
         npm run prebundle:js
```
mokagio, line 36: For reference and future optimization avenues, this is the command that takes up the bulk of the execution time.

.buildkite/publish-react-native-ios-artifacts.sh
```diff
@@ -1,16 +1,22 @@
 #!/bin/bash -eu
 
-cd ./ios-xcframework
+# Make Homebrew run a bit faster
+export HOMEBREW_NO_AUTO_UPDATE=1
+export HOMEBREW_NO_INSTALL_CLEANUP=1
```
mokagio: Having these exports here rather than inline is mostly cosmetic. I did it because I was `brew install`ing another tool at some point in development, but that's no longer the case.

I considered using the `env` node in the step definition in `pipeline.yml` but I wasn't sure whether we propagate those values to the scripts we run. Besides, given the code that uses these var is in this file, it make sense to me to leave them here.

mokagio: Also notice it's been discussed (internal ref p1682611494399739-slack-C020Q0Z579V ) adding these to our VM setup upstream. When that will happen, we'll be able to remove these from here.

.buildkite/pipeline.yml
```diff
@@ -23,34 +23,54 @@ steps:
             # Allow WP-CLI to be run as root, otherwise it throws an exception.
             # Reference: https://git.io/J9q2S
             - "WP_CLI_ALLOW_ROOT=true"
+      - peakon/git-shallow-clone#v0.0.1:
+          depth: 1
```
mokagio: This plugin makes checking out the repo ~ 2 minutes faster

**Before ([build](https://buildkite.com/automattic/gutenberg-mobile/builds/5869#018855c3-bc85-45cb-93be-41386965a05d))**

image


**After ([build](https://buildkite.com/automattic/gutenberg-mobile/builds/5870#0188567c-9da5-4726-85bd-a2b5d132a314))**

image

mokagio: 🤔 I guess I could DRY this by defining it in `x-common-params`.

I was hoping there was a way to add a plugin to every step, but [the docs don't mention it](https://buildkite.com/docs/plugins/using#using-yaml-anchors-with-plugins).

.buildkite/publish-react-native-ios-artifacts.sh
crazytonyli: `HOMEBREW_NO_AUTO_UPDATE` is already included. `HOMEBREW_NO_INSTALL_CLEANUP` is not. But I don't think we need the second one? I don't imagine "cleaning up old formula" would have an big impact on the performance.

> HOMEBREW_NO_INSTALL_CLEANUP
>
> If set, brew install, brew upgrade and brew reinstall will never automatically cleanup installed/upgraded/reinstalled formulae or all formulae every HOMEBREW_CLEANUP_PERIODIC_FULL_DAYS days. Alternatively, HOMEBREW_NO_CLEANUP_FORMULAE allows specifying specific formulae to not clean up.

.buildkite/pipeline.yml
```diff
@@ -23,42 +26,58 @@ steps:
...
     plugins:
       - automattic/a8c-ci-toolkit#2.15.1
-      - peakon/git-shallow-clone#v0.0.1:
-          depth: 1
+      - *git-shallow-clone
     agents:
       queue: mac
     env:
-      IMAGE_ID: xcode-14.2
+      IMAGE_ID: xcode-14.3
```
crazytonyli, line 77: Considering we are updating Xcode, may as well update it to the latest one `14.3.1-rc.1` which is available on CI. 😸

mokagio: During development, I experienced an issue which was resolved, I think, by putting the builder and consumer projects on the same Xcode version 😳 

For the moment, I'd like to keep the two in sync.

mokagio:
@crazytonyli thank you so much for the review on this. I should have addressed all your comments. I did that via a rebase to keep the history neat but it came at the downside of having to re-review the entire diff.

oguzkocer (approved):
Looks good to me :shipit:

crazytonyli (approved):

PR #5856, opened 2023-06-13
Upload a dedicated XCFramework podspec for tags

mokagio:
When building #5769 , I focused only on commits and didn't notice I hand't updated the automation to generate and upload a `podspec` for the tag version. This PR does that.

[CI build from the commit](https://buildkite.com/automattic/gutenberg-mobile/builds/6126#0188b381-0274-4c6c-b95a-5cf2b3040c3b):

image

[CI build from the tag](https://buildkite.com/automattic/gutenberg-mobile/builds/6128):

image

## Testing

See it in action at https://github.com/wordpress-mobile/WordPress-iOS/pull/20845/commits/3ed5255584704b3b3eb030446523f06a381e5aff

---

PR submission checklist:

### DIFF
--- a/ios-xcframework/fastlane/Fastfile
+++ b/ios-xcframework/fastlane/Fastfile
@@ -35,13 +35,15 @@ lane :upload_xcframework_to_s3 do |options|
 
   next if tag.nil? || tag.empty?
 
-  # Upload a dedicated podspec identified by the tag but pointing to the archive made for the commit, to avoid storing a duplicate archive for the tag.
-  upload_podspec_to_s3(archive_id: last_git_commit[:commit_hash], url_id: tag)
+  # Tradeoff: We are "wasting" space on S3 by uploading the same archive under different name.
+  # What we get in return is a 1-1 correspondency between archive and podspec, which keeps the setup simpler.
+  upload_to_a8c_s3(file: xcframework_archive_path, name_on_s3: tag)
+  upload_podspec_to_s3(archive_id: tag)
 end
 
 # rubocop:disable Metrics/AbcSize
 # rubocop:disable Metrics/MethodLength
-def upload_podspec_to_s3(archive_id:, url_id: nil)
+def upload_podspec_to_s3(archive_id:)
   xcconfig = Xcodeproj::Config.new(
     File.join(XCFRAMEWORK_PROJECT_ROOT_FOLDER, 'Config', 'Gutenberg-Shared.xcconfig')
   ).to_hash
@@ -60,19 +62,18 @@ def upload_podspec_to_s3(archive_id:, url_id: nil)
 
   podspec_erb = ERB.new(File.read(File.join(XCFRAMEWORK_PROJECT_ROOT_FOLDER, 'podspec.erb')))
 
-  s3_id = url_id.nil? ? archive_id : url_id
-
   Dir.mktmpdir do |tmp_dir|
     podspec_path = File.join(tmp_dir, 'podspec')
     File.write(podspec_path, podspec_erb.result(context))
-    upload_to_a8c_s3(file: podspec_path, name_on_s3: "Gutenberg-#{s3_id}.podspec")
+    upload_to_a8c_s3(file: podspec_path, name_on_s3: "Gutenberg-#{archive_id}.podspec")
   end
 end
 # rubocop:enable Metrics/AbcSize
 # rubocop:enable Metrics/MethodLength
 
 # Notice skip_if_exists is true by default.
-# CI might run on the same commit twice, once for the commit and once for the tag, and we don't want to fail the build on the second run.
+# CI might run on the same commit twice, once for the commit and once for the tag.
+# We don't want to fail on the second run.
 def upload_to_a8c_s3(file:, name_on_s3:, prefix: 'gutenberg-mobile', skip_if_exists: true)
   upload_to_s3(
     bucket: 'a8c-apps-public-artifacts',
--- a/ios-xcframework/fastlane/Fastfile
+++ b/ios-xcframework/fastlane/Fastfile
@@ -12,22 +12,14 @@ before_all do
   check_for_toolkit_updates unless is_ci || ENV['FASTLANE_SKIP_TOOLKIT_UPDATE_CHECK']
 end
 
-lane :upload_xcframework_to_s3 do |options|
+lane :upload_xcframework_to_s3 do
   xcframework_archive_path = File.join(XCFRAMEWORK_PROJECT_ROOT_FOLDER, "build/xcframeworks/Gutenberg.#{XCFRAMEWORK_ARCHIVE_TYPE}")
 
   unless File.exist? xcframework_archive_path
     UI.user_error! "Could not find XCFramework archive to upload at #{xcframework_archive_path}"
   end
 
-  name = options.fetch(:name, nil)
-
-  if name.nil?
-    name = archive_name(id: last_git_commit[:commit_hash])
-    UI.message "No name provided, will default to #{name}"
-  end
-
-  upload_to_a8c_s3(file: xcframework_archive_path, name_on_s3: name)
-  upload_podspec_to_s3(archive_id: last_git_commit[:commit_hash])
+  upload_xcframework_files_to_s3(archive_path: xcframework_archive_path, id: last_git_commit[:commit_hash])
 
   next unless is_ci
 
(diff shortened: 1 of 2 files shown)


ios-xcframework/fastlane/Fastfile
```diff
@@ -59,16 +60,20 @@ def upload_podspec_to_s3(id:)
...
-    upload_to_a8c_s3(file: podspec_path, name_on_s3: "Gutenberg-#{id}.podspec")
+    upload_to_a8c_s3(file: podspec_path, name_on_s3: "Gutenberg-#{s3_id}.podspec")
   end
 end
 # rubocop:enable Metrics/AbcSize
 # rubocop:enable Metrics/MethodLength
 
-def upload_to_a8c_s3(file:, name_on_s3:, prefix: 'gutenberg-mobile', skip_if_exists: false)
+# Notice skip_if_exists is true by default.
+# CI might run on the same commit twice, once for the commit and once for the tag, and we don't want to fail the build on the second run.
```
jkmassel: Edge case here, but what happens if a tag is re-issued? 

IINM, we'd wind up with the old code being forever associated with the tag, right?

Would it be preferable to default to replace-if-exists?

ios-xcframework/fastlane/Fastfile
```diff
@@ -27,20 +27,21 @@ lane :upload_xcframework_to_s3 do |options|
...
 
-  upload_to_a8c_s3(file: xcframework_archive_path, name_on_s3: "Gutenberg-#{tag}.#{archive_type}")
+  # Upload a dedicated podspec identified by the tag but pointing to the archive made for the commit, to avoid storing a duplicate archive for the tag.
+  upload_podspec_to_s3(archive_id: last_git_commit[:commit_hash], url_id: tag)
 end
 
 # rubocop:disable Metrics/AbcSize
 # rubocop:disable Metrics/MethodLength
-def upload_podspec_to_s3(id:)
+def upload_podspec_to_s3(archive_id:, url_id: nil)
```
jkmassel: I might be missing something, but I'm not sure why we need both parameters here? 

AFAICT it looks like we just use whichever is provided as the S3 key, so could we just accept a `key:` param?

I _think_ what you're going for here is that we could avoid re-uploading the binary (and just have two podspecs pointing to the same commit hash?), but I wonder if it's worth (deliberately) wasting the S3 space to have tag-specific binaries, not just commit hashes?

WDYT?

jkmassel (dismissed):
One potential bug, one suggested improvement

ios-xcframework/fastlane/Fastfile
mokagio: Ah, good point 🤔 

I like the override approach. It's what I usually use in these cases. Unfortunately, [our Fastlane action](https://github.com/wordpress-mobile/release-toolkit/blob/d6b1f424a308ca0d4e23f1705f51b07f72031684/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/upload_to_s3.rb) doesn't support that yet:

```
+--------------------------------------------------------------------------------+
|                                  upload_to_s3                                  |
+--------------------------------------------------------------------------------+
| Uploads a given file to S3                                                     |
|                                                                                |
| Uploads a file to S3, and makes a pre-signed URL available in the lane context |
|                                                                                |
| Created by Automattic                                                          |
+--------------------------------------------------------------------------------+

+----------------+----------------------------+------------+---------+
|                        upload_to_s3 Options                        |
+----------------+----------------------------+------------+---------+
| Key            | Description                | Env Var(s) | Default |
+----------------+----------------------------+------------+---------+
| bucket         | The bucket that will       |            |         |
|                | store the file             |            |         |
| key            | The path to the file       |            |         |
|                | within the bucket. If      |            |         |
|                | `nil`, will default to     |            |         |
|                | the `file's basename       |            |         |
| file           | The path to the local      |            |         |
|                | file on disk               |            |         |
| auto_prefix    | Generate a derived prefix  |            | true    |
|                | based on the filename      |            |         |
|                | that makes it harder to    |            |         |
|                | guess the URL of the       |            |         |
|                | uploaded object            |            |         |
| skip_if_exists | If the file already        |            | false   |
|                | exists in the S3 bucket,   |            |         |
|                | skip the upload (and       |            |         |
|                | report it in the logs),    |            |         |
|                | instead of failing with    |            |         |
|                | `user_error!`              |            |         |
+----------------+----------------------------+------------+---------+
```

I could implement it support for that and followup, while keeping fingers crossed that we don't run into a reused tag in the meantime? 🤞

ios-xcframework/fastlane/Fastfile
mokagio: @jkmassel 

> why we need both parameters here?

We generated the `podspec` on the fly, in a tmp folder. The setup we have at the moment use `Gutenberg-<sha1>.podspec` to download the `<sha1>.tar.gz` archive. For the tag version, `Gutenberg-<tag>.podspec`, we need to name the `podspec` with the `tag`  but still need the `sha1` to identify the `tar.gz` to download.

> I think what you're going for here is that we could avoid re-uploading the binary (and just have two podspecs pointing to the same commit hash?), but I wonder if it's worth (deliberately) wasting the S3 space to have tag-specific binaries, not just commit hashes?

That's exactly what I was going for 😄 

I guess at this time storage is cheaper than maintaining convoluted code. I'll update.

ios-xcframework/fastlane/Fastfile
```diff
@@ -27,20 +27,23 @@ lane :upload_xcframework_to_s3 do |options|
...
   next unless is_ci
 
   tag = ENV.fetch('BUILDKITE_TAG', nil)
 
   next if tag.nil? || tag.empty?
 
-  upload_to_a8c_s3(file: xcframework_archive_path, name_on_s3: "Gutenberg-#{tag}.#{archive_type}")
+  # Tradeoff: We are "wasting" space on S3 by uploading the same archive under different name.
+  # What we get in return is a 1-1 correspondency between archive and podspec, which keeps the setup simpler.
+  upload_to_a8c_s3(file: xcframework_archive_path, name_on_s3: tag)
```
crazytonyli: Is this change intentional? Does using `tag` as the S3 object name mean the framework URL would be something like `https://cdn.a8c.service/gutenberg-mobile/1.2.3`? I don't think that URL matches the one specified in the podspec file.

mokagio: Good catch @crazytonyli. Funny enough, I noticed that about the same time as you did and was working on a fix: 2e84f32

crazytonyli: Never mind. I see this part has been updated almost immediately after my comment 😄

mokagio:
Just a followup with the behavior in CI of the updated code:

image

ios-xcframework/fastlane/Fastfile
```diff
@@ -77,3 +79,7 @@ def upload_to_a8c_s3(file:, name_on_s3:, prefix: 'gutenberg-mobile', skip_if_exi
     skip_if_exists: skip_if_exists
   )
 end
+
+def archive_name(id:)
+  "Gutenberg-#{id}.#{XCFRAMEWORK_ARCHIVE_TYPE}"
+end
```
crazytonyli: This function is called only in one place, what do you think moving it to its caller?

ios-xcframework/fastlane/Fastfile
crazytonyli: We can potentially refactor the action to accept `if_exists=skip|replace|error` to cover different scenarios. Meanwhile, do you think we should add a TODO or FIXME here? What worries me is, if this edge case does happen, we wouldn't even notice it unless something terribly wrong happens.

crazytonyli (approved):
Suggested adding a TODO/FIXME to track the re-issuing tag edge case. Other than that, all looks good to me.

ios-xcframework/fastlane/Fastfile
mokagio: Good point. Addressed in 1ca7eaa6

ios-xcframework/fastlane/Fastfile
mokagio: Addressed in [`aa9d7da` (#5856)](https://github.com/wordpress-mobile/gutenberg-mobile/pull/5856/commits/aa9d7dad642da33b9e2a6944cbbba4530256863d) and https://github.com/wordpress-mobile/release-toolkit/pull/495

PR #5863, opened 2023-06-14
WIP – XCFramework GitHub release

mokagio:
Fixes #

To test:

PR submission checklist:

### DIFF
--- a/ios-xcframework/fastlane/Fastfile
+++ b/ios-xcframework/fastlane/Fastfile
@@ -6,6 +6,7 @@ default_platform :ios
 
 XCFRAMEWORK_PROJECT_ROOT_FOLDER = File.dirname(File.expand_path(__dir__))
 XCFRAMEWORK_ARCHIVE_TYPE = 'tar.gz'
+XCFRAMEWORK_ARCHIVE_PATH = File.join(XCFRAMEWORK_PROJECT_ROOT_FOLDER, "build/xcframeworks/Gutenberg.#{XCFRAMEWORK_ARCHIVE_TYPE}")
 
 before_all do
   # Ensure we use the latest version of the toolkit
@@ -13,13 +14,11 @@ before_all do
 end
 
 lane :upload_xcframework_to_s3 do
-  xcframework_archive_path = File.join(XCFRAMEWORK_PROJECT_ROOT_FOLDER, "build/xcframeworks/Gutenberg.#{XCFRAMEWORK_ARCHIVE_TYPE}")
-
   unless File.exist? xcframework_archive_path
-    UI.user_error! "Could not find XCFramework archive to upload at #{xcframework_archive_path}"
+    UI.user_error! "Could not find XCFramework archive to upload at #{XCFRAMEWORK_ARCHIVE_PATH}"
   end
 
-  upload_xcframework_files_to_s3(archive_path: xcframework_archive_path, id: last_git_commit[:commit_hash])
+  upload_xcframework_files_to_s3(archive_path: XCFRAMEWORK_ARCHIVE_PATH, id: last_git_commit[:commit_hash])
 
   next unless is_ci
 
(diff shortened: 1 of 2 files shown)


ios-xcframework/fastlane/Fastfile
```diff
@@ -5,28 +5,27 @@ require 'xcodeproj'
 default_platform :ios
 
 XCFRAMEWORK_PROJECT_ROOT_FOLDER = File.dirname(File.expand_path(__dir__))
+XCFRAMEWORK_ARCHIVE_TYPE = 'tar.gz'
+XCFRAMEWORK_ARCHIVE_PATH = File.join(PROJECT_ROOT_FOLDER, "build/xcframeworks/Gutenberg.#{XCFRAMEWORK_ARCHIVE_TYPE}")
```
mokagio:
```suggestion
XCFRAMEWORK_ARCHIVE_PATH = File.join(XCFRAMEWORK_PROJECT_ROOT_FOLDER, "build/xcframeworks/Gutenberg.#{XCFRAMEWORK_ARCHIVE_TYPE}")
```

ios-xcframework/fastlane/Fastfile
```diff
@@ -29,7 +28,28 @@ lane :upload_xcframework_to_s3 do
...
+
+  create_release(
+    repository: 'wordpress-mobile/gutenberg-mobile',
+    version: "This is a test – id #{Time.now.to_i}",
+    # Passing nil here will result in no release description text on GitHub
+    release_notes_file_path: nil,
+    release_assets: XCFRAMEWORK_ARCHIVE_PATH.to_s,
+    # TODO: read from options, options.fetch(:beta_release, true)
+    prerelease: true
+  )
```
fluiddot, line 52: @mokagio If the tag is created for a regular release, it's likely that the GitHub release is also created when this is executed. In fact, the tag is usually created when publishing the GitHub release.

I know the PR is draft but I'd like to share a couple of questions:
1. Should we also upload the podspec? In the future, it would be great if each GitHub release contains all the needed artifacts to integrate Gutenberg Mobile in the app.
2. Now that the Javascript bundle (i.e. the file `App.js`) is embedded in the XCFramework, I'm wondering if we could also upload the source maps. These files are required to symbolicate the stack traces of crashes. Currently, we are versioning them in the repository ([reference](https://github.com/wordpress-mobile/gutenberg-mobile/blob/trunk/bundle/ios/App.js.map)), but makes more sense to store them somewhere else. WDYT?

Let me know if I can help you with this, I'll be more than happy to contribute 🙂!

mokagio: Thanks for looking into this @fluiddot !

> Should we also upload the podspec? In the future, it would be great if each GitHub release contains all the needed artifacts to integrate Gutenberg Mobile in the app.

We upload the `podspec` as part of the XCFramework generation. Someone downloading the XCFramework from the release artifacts manually would presumably be integrating the framework manually, too. Otherwise, all they need to do is point their `Podfile` to the version they desire.

The idea to upload the artifacts here is so that, once the `podspec` in the root of the repo is up to date with the new setup, we'll be able to use GitHub releases to store the stable binaries, which seems a bit more community / OSS friendly.

> I'm wondering if we could also upload the source maps.

Sure can do

> If the tag is created for a regular release, it's likely that the GitHub release is also created when this is executed. In fact, the tag is usually created when publishing the GitHub release.

Good point 🤔 It's been a while since I wrote this code, but I guess I was assuming a change in workflow where we'd be pushing the tag from the release manager machine and let the automation do the rest on the GitHub side.

Would that work?

I should think about it a bit more and check whether it could fit with the existing release checklist.

fluiddot: > We upload the `podspec` as part of the XCFramework generation. Someone downloading the XCFramework from the release artifacts manually would presumably be integrating the framework manually, too. Otherwise, all they need to do is point their `Podfile` to the version they desire.

True, good point.

> The idea to upload the artifacts here is so that, once the `podspec` in the root of the repo is up to date with the new setup, we'll be able to use GitHub releases to store the stable binaries, which seems a bit more community / OSS friendly.

That's great, thanks for the info 🙇 !

> Good point 🤔 It's been a while since I wrote this code, but I guess I was assuming a change in workflow where we'd be pushing the tag from the release manager machine and let the automation do the rest on the GitHub side.
> 
> Would that work?

This would be one option, but we'd need to review the current release process to adapt this step. Currently, the tag is always created via publishing a GitHub release.

> I should think about it a bit more and check whether it could fit with the existing release checklist.

Good idea! As a side note, I'm going to [propose a change in the release process](https://github.com/wordpress-mobile/release-toolkit-gutenberg-mobile/pull/143) and create the GitHub release (as a draft) at the beginning of the process. So, I'd say we can confirm that when this part is run, we'll always have a release created.

mokagio:
This work is superseded (or will soon be superseded) by `gmb-cli`, https://github.com/wordpress-mobile/release-toolkit-gutenberg-mobile/tree/54d97c7d14172f2099dee998e48553f8e15ae3c6/gbm-cli#readme

PR #5874, opened 2023-06-16
Upgrade React Native `0.71.11`

fluiddot:
## Related PRs
* [Gutenberg](https://github.com/WordPress/gutenberg/pull/51303)
* [WordPress-Android](https://github.com/wordpress-mobile/WordPress-Android/pull/18613)
* [WordPress-iOS](https://github.com/wordpress-mobile/WordPress-iOS/pull/20956)

**To test:**
Follow testing instructions from https://github.com/WordPress/gutenberg/pull/51303.

PR submission checklist:


jest.config.js
```diff
@@ -56,6 +56,9 @@ module.exports = {
 		'test/helpers$': '<rootDir>/' + configPath + '/helpers.js',
 		jetpackConfig:
 			'<rootDir>/jetpack/tools/js-tools/jest/jest-jetpack-config.js',
+		// Workaround for Jest not having ESM support yet
+		// Reference: https://t.ly/9ap_
+		uuid: require.resolve( 'uuid' ),
```
dcalhoun, line 61: From reviewing the URL in this comment, my understanding is that we can remove this workaround when we [upgrade to `uuid@9`](https://github.com/uuidjs/uuid/issues/451#issuecomment-1206284480). Is that accurate? It would likely be beneficial to note this in the inline comment.

src/test/videopress/edit.js
```diff
@@ -35,14 +36,21 @@ import {
...
 	// Register Jetpack blocks
-	setupJetpackEditor( { blogId: 1, isJetpackActive: true } );
+	setupJetpackEditor( {
+		blogId: 1,
+		isJetpackActive: true,
+	} );
 	registerJetpackBlocks( DEFAULT_PROPS );
+
+	// Mock request reponses
+	setupApiFetch( generateFetchMocks() );
```
dcalhoun: Why this now required due to the RN upgrade? Does this relate to https://github.com/facebook/react-native/pull/34659?

jest.config.js
fluiddot: Ah, true. I haven't tested this myself, but based on the comments that seems the case. I'll update the inline comment to reflect this. Thanks 🙇 !

src/test/videopress/edit.js
fluiddot: Yep, it looks related to that change. Without this change, the tests fail with the following error:
```
TypeError: Cannot read property 'then' of undefined

      53 | 			method: 'POST',
      54 | 			body: fetchParams.body,
    > 55 | 		} ).then( response => {
         | 		   ^
      56 | 			const { metadata_token } = response;
      57 | 			if ( ! metadata_token ) {
      58 | 				console.warn( 'Token is not achievable' ); // eslint-disable-line no-console
```

I understand that when `Promise` was polyfilled in Jest tests, we didn't need to explicitly mock API responses (i.e. `api-fetch` library).

jest.config.js
fluiddot: I tried using `uuid` version 9.x but I still had to use this workaround to address the issue. There might be something else producing the issue, especially as it's not happening in Gutenberg unit tests which I assume we use an older version `uuid` 🤔 .

SiobhyB (approved):
Approved via https://github.com/WordPress/gutenberg/pull/51303#pullrequestreview-1537530236.

For posterity, noting that https://github.com/wordpress-mobile/gutenberg-mobile/pull/5973 needs to be merged in for the tests to pass, and to avoid the issue with crashing on iOS.

fluiddot:
> For posterity, noting that #5973 needs to be merged in for the tests to pass, and to avoid the issue with crashing on iOS.

The PR has been already merged in https://github.com/wordpress-mobile/gutenberg-mobile/pull/5874/commits/378e2330385ca3279f60f4f04b2683d405e3c1e5.

dcalhoun (approved):

fluiddot:
As a side note, https://github.com/wordpress-mobile/gutenberg-mobile/pull/5874/commits/b837251232ac5ec1240bd852cda176e872e9b074 was needed because I updated `Podfile` in https://github.com/wordpress-mobile/gutenberg-mobile/pull/5874/commits/23d9fbd0e729d45823cae7a90c8d35625d2a67be 🤦.

PR #5884, opened 2023-06-20
HACK – Port all XCFramework-publishing automation from trunk on 1.97.1

mokagio:
This is just so we can fetch the XCFramework for 1.97.1 in the Jetpack and WordPress iOS apps, even though there is not an official tag build for it.

The practical benefit is that we can merge the integration work without having to wait for a new Gutenberg mobile version to ship, see https://github.com/wordpress-mobile/WordPress-iOS/pull/20717


mokagio:
This has done its job

PR #5894, opened 2023-06-22
test: Fix CI task failures

dcalhoun:
## Description

Fixes failing CI tasks.

### test: Fix broken input query

It is unclear  me why this change is necessary, but it resolves the
following test failures.

Query returned a single input, which led to the second input being
undefined:

```
    TypeError: Cannot read property 'props' of undefined

        11 |  */
        12 | export const changeTextOfTextInput = ( textInput, text ) => {
    > 13 |      fireEvent( textInput, 'focus' );
            |      ^
        14 |      fireEvent.changeText( textInput, text );
        15 | };
        16 |

        at props (gutenberg/node_modules/@testing-library/react-native/src/fireEvent.ts:116:22)
        at getEventHandler (gutenberg/node_modules/@testing-library/react-native/src/fireEvent.ts:103:19)
        at findEventHandler (gutenberg/node_modules/@testing-library/react-native/src/fireEvent.ts:136:19)
        at changeTextOfTextInput (gutenberg/test/native/integration-test-helpers/text-input-change-text.js:13:2)
        at Object.<anonymous> (src/test/videopress/edit.js:132:3)
        at asyncGeneratorStep (node_modules/@babel/runtime/helpers/asyncToGenerator.js:3:24)
        at _next (node_modules/@babel/runtime/helpers/asyncToGenerator.js:25:9)
        at tryCallOne (gutenberg/node_modules/promise/lib/core.js:37:12)
        at gutenberg/node_modules/promise/lib/core.js:123:15
        at flush (gutenberg/node_modules/asap/raw.js:50:29)
```

The broken query led to an incorrect snapshot outcome:

```
    Snapshot name: `Update VideoPress block's settings should update description 1`

    - Snapshot  - 1
    + Received  + 1

    - "<!-- wp:videopress/video {"title":"default-title-is-file-name","description":"The video's new description!","useAverageColor":false,"id":1,"guid":"AbCdEfGh","privacySetting":2,"allowDownload":false,"rating":"G","isPrivate":true,"duration":2803} /-->"
    + "<!-- wp:videopress/video {"title":"default-title-is-file-name","description":"","useAverageColor":false,"id":1,"guid":"AbCdEfGh","privacySetting":2,"allowDownload":false,"rating":"G","isPrivate":true,"duration":2803} /-->"

        226 |     afterEach( async () => {
        227 |             // Assert
    > 228 |             expect( getEditorHtml() ).toMatchSnapshot();
            |                                       ^
        229 |     } );
        230 | } );
        231 |
```

### test: Add to-do note regarding test improvements

In general, we should not wrap `fireEvent` in an `act` call, as it
already leverages `act` itself. Wrapping it is an indicator of a deeper
issue: there are state updates for which we are not awaiting. Ideally,
we add a new query to await the addition/update/removal of UI to satisfy
the act warning.

## Testing Instructions
Verify CI tasks succeed.

## PR submission checklist

### DIFF
--- a/src/test/videopress/edit.js
+++ b/src/test/videopress/edit.js
@@ -122,12 +122,7 @@ describe( "Update VideoPress block's settings", () => {
 
 		fireEvent.press( screen.getByText( 'Description' ) );
 
-		const allDescriptionInputs = screen.getAllByPlaceholderText(
-			'Add description'
-		);
-
-		// The BottomSheetControl's input field is accessed via the component's second placeholder
-		const input = allDescriptionInputs[ 1 ];
+		const input = screen.getByPlaceholderText( 'Add description' );
 
 		changeTextOfTextInput( input, "The video's new description!" );
 	} );
--- a/src/test/videopress/local-helpers/utils.js
+++ b/src/test/videopress/local-helpers/utils.js
@@ -46,6 +46,8 @@ export const pressSettingInPanel = async ( screen, panel, setting ) => {
 	fireEvent.press( getByText( panel ) );
 
 	// Toggle the specified setting
+	// TODO: Determine the cause of state updates and explicitly wait for them,
+	// instead of wrapping firEvent in act.
 	await act( () => fireEvent.press( getByText( setting ) ) );
 };
 


geriux (approved):
LGTM!

dcalhoun:
@SiobhyB requesting your specific review to share these changes and also provide you the opportunity to let me know if I am overlooking anything. Happy to create a follow PR if anything needs to change. Thanks!

SiobhyB:
@dcalhoun, thank you for fixing this! 

There were recent changes to the description placeholder in https://github.com/Automattic/jetpack/pull/31134 that I now see led to these failures. Previously, the "Add description" placeholder always appeared twice on the screen. As we're now using a different initial placeholder for videos that don't belong to a site, "Add description" will only appear once in the test setup. 

Tests hadn't been on my mind when reviewing https://github.com/Automattic/jetpack/pull/31134. 🤦‍♀️  This case is a good reminder of the need to be more diligent in creating companion Gutenberg Mobile PRs when making changes in Jetpack, especially for the VideoPress block.

> In general, we should not wrap fireEvent in an act call, as it already leverages act itself. Wrapping it is an indicator of a deeper issue: there are state updates for which we are not awaiting. Ideally, we add a new query to await the addition/update/removal of UI to satisfy the act warning.

Thank you for the note here, I appreciate the thought in documenting parts of the code that seem off, as well as expanding on why.

I believe the deeper issue is related to React Navigation, which is used for the `Bottomsheet.Subsheet` component being tested. The following issues are referenced in other places in the codebase ([here](https://github.com/WordPress/gutenberg/blob/ee36ebc6ccd6becf57a7e346a083ec67546d973b/packages/block-library/src/image/test/edit.native.js#L258-L260) and [here](https://github.com/WordPress/gutenberg/blob/ee36ebc6ccd6becf57a7e346a083ec67546d973b/packages/block-library/src/paragraph/test/edit.native.js#L241-L243)) where we've implemented similar workarounds:

- https://github.com/WordPress/gutenberg/issues/35685#issuecomment-961919931
- https://github.com/react-navigation/react-navigation/issues/9701

Do you think the similar workaround here make sense, with that context? If so, I can go ahead to create a follow up PR with an explanation in the comment. If you think there is a better way to work around this issue, I'd be keen to learn and can follow up with that, too.

dcalhoun:
@SiobhyB thank you for explaining the unexpected breakage. I was perplexed and did not know where to start to determine the origin. Your explanation makes sense. It is further encouragement to work towards moving our tests to the `Automattic/jetpack` repo. 

Your thoughts on the `act` warnings are accurate, particularly that you point to https://github.com/WordPress/gutenberg/issues/35685#issuecomment-961919931. That comment summarizes the latest understanding: there is not a bug in RNTL or React Navigation, we need to await state updates anytime a navigation event occurs. 

In regards to approaches to resolving `act` warnings, more specifically those related to navigation events...

```js
// Approach 1
await act( () => fireEvent.press( getByText() ) ); 
```

The above (Approach 1) is a bit of a "sledge hammer" approach and is ill-advised. It basically says "some _asynchronous_ change is going to happen" in addition to the _synchronous_ change that `fireEvent` will await due to its own usage of `act`. 

```js
// Approach 2
fireEvent.press( getByText() ); 
await findByText(/Actual text that changed due to asynchronous state update/); 
```

The above (Approach 2) pointedly awaits the _specific asynchronous_ change that occurs. This better communicates testing intent and helps avoid future confusion that may occur by using Approach 1; e.g. new, unexpected `act` warnings arise that the first "sledge hammer" does not resolve. 

Identifying the specific state changing or the visible UI change to await can be challenging. For this specific context, when I debugged this failure, it appeared that [`isRequestingEmbedPreview` changes](https://github.com/Automattic/jetpack/blob/2cdf41cff76706db4e6b8a27c3bdbdcf4336056f/projects/packages/videopress/src/client/block-editor/blocks/video/components/player/index.native.js#L89) when _toggling the setting_, not during the navigation event, causing the `Player` to re-render. I noted this may actually be observable when using the app in-person as the video embed flashes whenever one toggles a setting.[^1] 

My idea to solve this was to await the removal of the "Loading" text or identify some other visible change in the video player to await. However, I noticed in the tests the "Loading" text is always displayed, the test never results in the video "rendering."[^2] At this point, I abandoned resolving the `act` warning this morning and left the inline comment. Ideally, we follow up to identify UI to query against and await updates. 

[^1]: Rather than awaiting the state updates, a better `act` warning solution might be to avoid unnecessary re-renders triggered by these state updates. This might apply to attributes that do not modify the presentation, e.g. `loop`, but would not apply to attributes that do result in visible changes, e.g. `seekbarColor `. We still need to re-render and await things that modify visuals. 
[^2]: Part of me wondered if this always-loading state relates to and may be addressed by a [recent change](https://github.com/wordpress-mobile/gutenberg-mobile/pull/5874/files/2f89a3be5fb7976e50d85cf5f3e1da1d8c0e7793#r1238636459) in the RN Upgrade project.

PR #5846, opened 2023-06-09
Update XCFramework pods after running `bin/generate-podspecs.sh`

mokagio:
@jhnstn recently opened https://github.com/wordpress-mobile/gutenberg-mobile/pull/5843 to address a `bundle exec pod install` failing locally for the XCFramework generation project.

As far as I can tell, the sequence of events that resulted in that failure was:

1. Some RN dependencies were updated in Gutenberg via https://github.com/WordPress/gutenberg/pull/50731
2. Those new versions were eventually fetched in this repo (I'm guessing via `bin/generate-podspec.js`?) in https://github.com/wordpress-mobile/gutenberg-mobile/commit/ed9625d394b117fc9271fdf642ff9167e95d03cd
3. Suddenly, the version in the XCFramework `Podfile.lock` was pointing to one that no longer existed locally

Assuming that `bin/generate-podspec.js` is the single point through which we updated the `third-party-podspecs`, then I think we can avoid this issue in the future by making the script also update the pods used by the XCFramework project, which is what this PR does.

### Testing

I tried to run the script locally, but it failed like so:

```
WARNING! hermes-engine.podspec doesn't have a 'tag' or 'commit' field, or doesn't point to a SHA verified file. Either modify this script to add a patch during the podspec generation or modify the original hermes-engine.podspec in the source repo.
```

<details>
<summary>Full log</summary>

```
➜ ./bin/generate-podspecs.sh
Working directory: /Users/gio/Developer/a8c/gutenberg-mobile
If your node_modules folder isn't up-to-date please run 'npm install' first and re-run the script. Is your node_modules folder up-to-date? [y/N] y
Enter the commit hash of previous commit. If this is the first-time running this script, enter 0, then commit generated files and re-rerun the script and this time use the previous commit hash: ddd8905b097ace9177cb2aa5d5414c923ccb482a
Please delete '/Users/gio/Developer/a8c/gutenberg-mobile/third-party-podspecs' folder manually before continuing. This script will re-generate it. Did you delete it? [y/N] y
Generating podspec for boost
Generating podspec for RCT-Folly
Generating podspec for DoubleConversion
Generating podspec for glog
Generating podspec for react-native-blur
   ==> Patching react-native-blur podspec
Generating podspec for RNCMaskedView
Generating podspec for react-native-slider
Generating podspec for RNCClipboard
Generating podspec for RNGestureHandler
Generating podspec for react-native-get-random-values
Generating podspec for BVLinearGradient
Generating podspec for RNReanimated
[...]

PR #5901, opened 2023-06-23
build: Fix generate-podspecs script errors

dcalhoun:
## Description 

Fix a couple of failures when running `bin/generate-podspecs.sh`. This is a follow up to https://github.com/wordpress-mobile/gutenberg-mobile/pull/5846. 

* build: Fix script directory changes

    Rely upon pushd and popd to simplify the "breadcrumbs" of navigating
    into and out of multiple directories.
* build: Only update xcframework Podfile.lock with valid targets

    We are required to run `bin/generate-podspecs.sh` twice to capture the
    correct target. This change ensures the xcframework is only updated on
    the second run with the correct target, otherwise the following error
    occurs:

    ```
    Installing FBReactNativeSpec 0.69.4

    [!] Error installing FBReactNativeSpec
    [!] /opt/homebrew/bin/git -C /var/folders/c5/mkbb257d7fg4myx88dvsndp80000gn/T/d20230623-6856-98pfen checkout --quiet 0

    error: pathspec '0' did not match any file(s) known to git
    ```

## Testing Instructions

1. Run `./bin/generate-podspecs.sh`
1. Verify the following errors do not occur:
    ```
    ./bin/generate-podspecs.sh: line 201: pushd: ios-xcframework: No such file or directory
    ```

    ```
    Installing FBReactNativeSpec 0.69.4

    [!] Error installing FBReactNativeSpec
    [!] /opt/homebrew/bin/git -C /var/folders/c5/mkbb257d7fg4myx88dvsndp80000gn/T/d20230623-6856-98pfen checkout --quiet 0

    error: pathspec '0' did not match any file(s) known to git
    ```

PR submission checklist:

### DIFF
--- a/bin/generate-podspecs.sh
+++ b/bin/generate-podspecs.sh
@@ -17,8 +17,8 @@ function warn_missing_tag_commit() {
 }
 
 # Change to the expected directory.
-cd "$( dirname "$0" )"
-cd ..
+pushd "$( dirname "$0" )" > /dev/null
+popd > /dev/null
 WD=$(pwd)
 echo "Working directory: $WD"
 
@@ -96,7 +96,7 @@ done
 
 # Generate the React Native podspecs
 # Change to the React Native directory to get relative paths for the RN podspecs
-cd "$NODE_MODULES_DIR/react-native"
+pushd "$NODE_MODULES_DIR/react-native" > /dev/null
 
 RN_DIR="./"
 SCRIPTS_PATH="./scripts/"
@@ -196,8 +196,9 @@ do
         mv "$TMP_FBReactNativeSpec" "$DEST/FBReactNativeSpec/FBReactNativeSpec.podspec.json"
     fi
 done
+popd > /dev/null
 
 echo 'Updating XCFramework Podfile.lock with these changes'
-pushd ios-xcframework
+pushd ios-xcframework > /dev/null
 bundle exec pod update
-popd
+popd > /dev/null
--- a/bin/generate-podspecs.sh
+++ b/bin/generate-podspecs.sh
@@ -198,6 +198,8 @@ do
 done
 popd > /dev/null
 
+# We are required to run this script twice to capture the correct target.
+# 0 is the value set on the first running this script to generate the podspecs.
 if [[ "$COMMIT_HASH" != "0" ]]; then
     echo 'Updating XCFramework Podfile.lock with these changes'
     pushd ios-xcframework > /dev/null
--- a/bin/generate-podspecs.sh
+++ b/bin/generate-podspecs.sh
@@ -198,7 +198,9 @@ do
 done
 popd > /dev/null
 
-echo 'Updating XCFramework Podfile.lock with these changes'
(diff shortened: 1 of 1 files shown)


bin/generate-podspecs.sh
```diff
@@ -17,8 +17,8 @@ function warn_missing_tag_commit() {
 }
 
 # Change to the expected directory.
-cd "$( dirname "$0" )"
-cd ..
+pushd "$( dirname "$0" )" > /dev/null
+popd > /dev/null
```
mokagio, line 21: The way I'm reading this, we `cd`/`pushd` into `dirname $0`, and immediately after `cd`/`popd` out of it.

Is this necessary to get some other automation to kick in? Otherwise, what do you think of removing it?

mokagio, line 21: Love the use of `pushd` and `popd`. The output redirect to `/dev/null` is a bit cumbersome to do on every usage, though.

We _could_ change the shebang to run on ZSH, and use `setopt PUSHDSILENT` at the start of the script. What do you think? 

Not a blocker! Just a suggestion.

bin/generate-podspecs.sh
```diff
@@ -196,8 +196,11 @@ do
         mv "$TMP_FBReactNativeSpec" "$DEST/FBReactNativeSpec/FBReactNativeSpec.podspec.json"
     fi
 done
+popd > /dev/null
 
-echo 'Updating XCFramework Podfile.lock with these changes'
-pushd ios-xcframework
-bundle exec pod update
-popd
+if [[ "$COMMIT_HASH" != "0" ]]; then
```
mokagio, line 203: A someone with little knowledge of this setup, it took some digging to understand where the 0 value came from.


```suggestion
# We are required to run this script twice to capture the correct target.
# 0 is the value set on the first running this script to generate the podspecs.
if [[ "$COMMIT_HASH" != "0" ]]; then
```

mokagio (approved):
Thank you for these updates @dcalhoun 🙇‍♂️

bin/generate-podspecs.sh
dcalhoun: From my testing, this logic ensures the rest of the script is executed from the project root directory, regardless if one invokes this shell script via `./bin/generate-podspecs.sh` or `cd ./bin && ./generate-podspecs.sh`. It is definitely not necessary, but I suppose it is a safeguard to avoid any confusion. 

If you feel this is unnecessary overhead, please feel free to remove it once this branch merges into your branch.

dcalhoun: I would rate my bash knowledge at 3, on a scale from 1 (non-existent) to 10 (expert). 😄 I am not opposed to modifying the approach at all. Please feel free to do so after this merges into your branch. 

My only question would be: would relying upon ZSH work in all relevant environments or would we experience errors on CI from the absence of ZSH?

bin/generate-podspecs.sh
dcalhoun: Adding a note makes total sense. Thank you.

PR #5916, opened 2023-06-29
Fix VideoPress block test cases

fluiddot:
This PR aims to fix the failures in https://github.com/wordpress-mobile/gutenberg-mobile/pull/5874 related to unit tests.

⚠️  It's expected that some of the CI jobs fails due to the RN upgrade. The jobs that this PR aims to fix are:
- `ci/circleci: Test Android`
- `ci/circleci: Test iOS`

We could omit the rest as they will be solved in other PRs.

**To test:**
* Run the command: `npm run test`.
* Observe that all tests pass.

Alternatively, we could rely on the CI job that runs the unit tests.

PR submission checklist:


src/test/videopress/edit.js
```diff
@@ -37,8 +36,17 @@ import {
...
 	generateFetchMocks,
+	generateBlockHTML,
 } from './local-helpers/utils';
 
+const VIDEO_TITLE = 'default-title-is-file-name';
+const FETCH_MOCKS_METADATA = {
+	title: VIDEO_TITLE,
+	description: '',
+	post_id: 1,
+	duration: 2803,
```
fluiddot: These parameters are not really being tested but decided to include them to avoid updating the Jest snapshots. Especially, as updating snapshots usually makes the changes harder to review.

src/test/videopress/local-helpers/utils.js
```diff
@@ -214,6 +251,15 @@ export const generateFetchMocks = ( {
...
 		},
+		{
+			request: {
+				path: `/wpcom/v2/videopress/${ guid }/check-ownership/${ postID }`,
+				method: 'GET',
+			},
+			response: {
+				'video-belong-to-site': belongsToSite,
+			},
+		},
```
fluiddot, line 262: We need to mock this response, otherwise, some of the block settings like the title will be locked.

src/test/videopress/edit.js
dcalhoun: I have no objections to reviewing a snapshot change if it means the stored snapshot is more helpful or realistic. However, if these mocked parameters achieve a more realistic picture, then it makes sense to keep them IMO.

src/test/videopress/local-helpers/utils.js
```diff
@@ -123,47 +123,84 @@ export const expectShowMediaOptions = ( screen, { title, options } ) => {
...
+ * Generates the block settings related to privacy
+ * based on video and site privacy.
  *
  * @param {Object} options
- * @param {string} [options.guid]        VideoPress GUID.
- * @param {string} [options.title]       Title of the video.
- * @param {string} [options.description] Description of the video.
- * @param {string} [options.isPrivate]   True if the video is private.
+ * @param {string} [options.isVideoPrivate] True if the video is private. If not defined, it will use site's privacy.
+ * @param {string} [options.isSitePrivate]  True if the site is private.
```
dcalhoun: These appear to be inaccurate. 
```suggestion
 * @param {boolean} [options.isVideoPrivate] True if the video is private. If not defined, it will use site's privacy.
 * @param {boolean} [options.isSitePrivate]  True if the site is private.
```

src/test/videopress/local-helpers/utils.js
```diff
@@ -123,47 +123,84 @@ export const expectShowMediaOptions = ( screen, { title, options } ) => {
...
+	// 1 = Private
+	// 2 = Site default
+	if ( typeof isVideoPrivate === 'undefined' ) {
+		return {
+			privacySetting: 2,
+			isPrivate: isSitePrivate,
+		};
+	}
+	return {
+		privacySetting: isPrivate ? 1 : 0,
```
dcalhoun: `isPrivate` appears to be undefined. Is this meant to be `isVideoPrivate` or `isSitePrivate` instead? This oversight may be causing (erroneous?) snapshot changes of `isPrivate` in this PR.

src/test/videopress/local-helpers/utils.js
fluiddot: Ah, good catch, thanks for the suggestion.

src/test/videopress/local-helpers/utils.js
fluiddot: Ah, my bad. This is meant to use `isVideoPrivate`. I'm afraid I renamed the parameter and forgot to update this part. I'll push a fix.

fluiddot: I've just fixed this via https://github.com/wordpress-mobile/gutenberg-mobile/pull/5916/commits/2ab22a26e307bc47747083046d8faab230d5d81d.

src/test/videopress/edit.js
fluiddot: To be honest, these parameters don't provide any value in the snapshot. They aren't meant to be updated in the test cases but since the VideoPress block updates settings automatically from the API when rendering the block, they are included. I'm happy with both options, but if including them in the tests gives a wrong impression of their usage, I can revert it.

dcalhoun: I have no strong preference. I suppose I mostly wanted to convey that we should not be afraid of updating snapshots. If a given snapshot is too unwieldily to review, a more appropriate alternative may be forgoing snapshots altogether and asserting against specific outcomes, e.g. attributes within the snapshot.

fluiddot: > I have no strong preference. I suppose I mostly wanted to convey that we should not be afraid of updating snapshots.
Yep, I think in this case we can update the snapshots as we don't have that many to make the review complex 👍.

> If a given snapshot is too unwieldily to review, a more appropriate alternative may be forgoing snapshots altogether and asserting against specific outcomes, e.g. attributes within the snapshot.

Good point. In general, I'd rather lean on snapshots to simplify test logic. But I agree that in some cases there's a good value on asserting specific outcomes. For these test cases, I'm hesitant about which one would fit better. For now, I'd prefer to keep them as-is and only focus on the fixes, if that makes sense.

dcalhoun: > For now, I'd prefer to keep them as-is and only focus on the fixes, if that makes sense.

Absolutely. 🚀

src/test/videopress/__snapshots__/edit.js.snap
```diff
@@ -1,39 +1,39 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`Update VideoPress block's settings should update Playback Bar Color section's Dynamic color setting 1`] = `"<!-- wp:videopress/video {"title":"default-title-is-file-name","description":"","id":1,"guid":"AbCdEfGh","privacySetting":2,"allowDownload":false,"rating":"G","isPrivate":true,"duration":2803} /-->"`;
+exports[`Update VideoPress block's settings should update Playback Bar Color section's Dynamic color setting 1`] = `"<!-- wp:videopress/video {"title":"default-title-is-file-name","description":"","id":34,"guid":"AbCdEfGh","privacySetting":2,"allowDownload":false,"rating":"G","isPrivate":true,"duration":1200} /-->"`;
```
fluiddot, line 3: The `id` and `duration` attributes have been modified due to using the default values specified in the util function (reference). They don't have an impact on the test cases, so we can omit them.

https://github.com/wordpress-mobile/gutenberg-mobile/blob/68a1ccd7f4ab9334b596e042606887bed4162813/src/test/videopress/local-helpers/utils.js#L203-L229

src/test/videopress/edit.js
fluiddot: I've just applied this suggestion via https://github.com/wordpress-mobile/gutenberg-mobile/pull/5916/commits/68a1ccd7f4ab9334b596e042606887bed4162813.

dcalhoun (approved):

src/test/videopress/local-helpers/utils.js
```diff
@@ -123,47 +123,84 @@ export const expectShowMediaOptions = ( screen, { title, options } ) => {
...
+	// 2 = Site default
+	if ( typeof isVideoPrivate === 'undefined' ) {
+		return {
+			privacySetting: 2,
+			isPrivate: isSitePrivate,
+		};
+	}
+	return {
+		privacySetting: isVideoPrivate ? 1 : 0,
+		isPrivate,
```
fluiddot, line 151: Sorry I didn't notice before 🤦 , but this should be:
```
isPrivate: isVideoPrivate,
```

I'll push a fix directly in the base branch (https://github.com/wordpress-mobile/gutenberg-mobile/pull/5874/commits/43492164e045b219b002fd7f66b5e1c74dc4aab1) 💨

PR #5967, opened 2023-07-17
Remove unneeded fix for text font-weight style

fluiddot:
Removes an unneeded workaround on Android related to correcting the `font-weight` style when using numerical values. This is no longer needed as a fix was introduced in React Native 0.65 ([reference - _Fix font weight numeric values_](https://github.com/facebook/react-native/blob/main/CHANGELOG-pre-070.md#android-specific-40)).

As a side note, I debugged the workaround and noticed that it wasn't actually being applied 😅. The `style` prop is being fetched from the wrong object, as the following code is always returning `undefined`:

https://github.com/wordpress-mobile/gutenberg-mobile/blob/0130fbb9fb2592fbb819431774440dd13a666ef7/src/text-font-weight-correct.js#L55

As an example, here the props of `baseText` for an element that has the `font-weight` style attribute set:
```
{"children": <RCTText accessible={true} allowFontScaling={true} ellipsizeMode="tail" isHighlighted={false} selectionColor={null} style={{"color": "#2e4453", "fontSize": 14, "fontWeight": "600", "marginTop": 4, "textAlign": "center"}}>Table</RCTText>, "value": true}
```

The styles should be fetched from the children element not `baseText`. I updated the code to enable the workaround but still the original font-weight issue is no longer happening.

**To test:**
Following the test cases from [the PR that introduced the workaround](https://github.com/wordpress-mobile/gutenberg-mobile/pull/2141):
* Open a post that contains an unsupported block (e.g. Table block).
* Observe that the block's title text is rendered with proper font-weight.

PR submission checklist:


geriux (approved):
LGTM! Thanks for fixing this issue 👏 

So nice to remove an old workaround that wasn't even being used 😅 🚀

PR #5968, opened 2023-07-17
Flatten embedded frameworks when generating XCFramework

fluiddot:
The goal of this PR is to flatten frameworks nested under the `Frameworks` folder of other frameworks. Especially, the Hermes framework which is nested under the Gutenberg framework.

This has been achieved by simply moving the embedded frameworks to the root level (https://github.com/wordpress-mobile/gutenberg-mobile/pull/5968/commits/e03356d4bb3cba27c5e5b7448caa814a11d8bd47), plus including the Hermes framework in the vendor framework list of the Gutenberg `podspec` (https://github.com/wordpress-mobile/gutenberg-mobile/pull/5968/commits/355f4319c3be6e31fc509d42bb41698078385d62).

## To test

**Preparation:**
* Check out the branch `gutenberg/tweak-pod-deps-by-rn-version` from the WP-iOS repository ([PR reference](https://github.com/wordpress-mobile/WordPress-iOS/pull/21021)).

### Test via CI
* Wait until `Build iOS RN XCFramework & Publish to S3` CI job finishes.
* Update the Gutenberg Mobile reference to the commit `64a5fed3fd28e7fa4aa25edc78f39b76617e5f71`.
* Run the command `bundle install && bundle exec pod install`.
  **NOTE:** If any error is encountered, try to run the command `rm -rf build && rm -rf Pods` before installing Pods.
* Build and run the WP-iOS app using a physical device.
* Observe that the app doesn't crash and that the editor works as expected.

### Test via local installation

⚠️ Note that this testing approach is mainly to test XCFramework integration without having to publish the framework.

* Apply the following patch:
```
diff --git forkSrcPrefix/ios-xcframework/Gutenberg.podspec forkDstPrefix/ios-xcframework/Gutenberg.podspec
new file mode 100644
index 0000000000000000000000000000000000000000..ccba32ed4e4053b51f638d3238e4213324f34ca9
--- /dev/null
+++ forkDstPrefix/ios-xcframework/Gutenberg.podspec
@@ -0,0 +1,28 @@
+
+Pod::Spec.new do |s|
+  s.name = 'Gutenberg'
+  s.summary = 'Printing since 1440'
+  s.homepage = 'https://github.com/wordpress-mobile/gutenberg-mobile'
+  s.authors = 'Automattic'
+
+  s.version = '1.98.1'
+
+  s.license = { type: 'GPL-2' }
+
+  s.ios.deployment_target = '15.0'
+  s.swift_version = '5.0'
+
+  s.source = {
+    http: 'file://' + __dir__ + '/build/xcframeworks/Gutenberg.tar.gz'
+  }
+
+  s.vendored_frameworks = [
+    'Frameworks/Aztec.xcframework',
+    'Frameworks/Gutenberg.xcframework',
+    'Frameworks/React.xcframework',
+    'Frameworks/RNTAztecView.xcframework',
+    'Frameworks/yoga.xcframework',
+    'Frameworks/hermes.xcframework'
+  ]
+end
```
* In the WP-iOS project, change [the Gutenberg Mobile `podspec` reference](https://github.com/wordpress-mobile/WordPress-iOS/blob/636d972c239d15818166fd06b7ee417b55d99fe2/Gutenberg/cocoapods_helpers.rb#L45) to: `../gutenberg-mobile/ios-xcframework/Gutenberg.podspec`
* Run the command `bundle install && bundle exec pod install`.
  **NOTE:** If any error is encountered, try to run the command `rm -rf build && rm -rf Pods` before installing Pods.
* Build and run the WP-iOS app using a physical device.
* Observe that the app doesn't crash and that the editor works as expected.

PR submission checklist:


ios-xcframework/build.sh
```diff
@@ -77,25 +100,21 @@ PLATFORM_SIMULATOR=iphonesimulator
...
+# 2. Flatten embedded frameworks for every framework in the archives
+flattenFrameworks $PLATFORM_IOS
+flattenFrameworks $PLATFORM_SIMULATOR
+
+# 3. Create XCFrameworks for every framework in the archives
 #
 # Notice how we loop on $PLATFORM_IOS as a way to get the frameworks for both platforms.
 # We could use either platform to achieve the same result.
-for FRAMEWORK in $(find "$ARCHIVES_ROOT/$PLATFORM_IOS.xcarchive/Products/Library/Frameworks" -type d -name "*.framework");
+for FRAMEWORK in "$ARCHIVES_ROOT/$PLATFORM_IOS.xcarchive/Products/Library/Frameworks"/*;
```
fluiddot, line 111: This change implies that we'll only loop over the frameworks at the root level. Embedded frameworks (i.e. nested in framework folders) will be skipped as it should have been processed in the previous step.

ios-xcframework/build.sh
```diff
@@ -56,6 +56,29 @@ function archive {
...
+      do
+        CURRENT_EMBEDDED_FRAMEWORK_NAME=$(basename "$EMBEDDED_FRAMEWORK" .framework)
+
+        log 'package' "Flattening $CURRENT_EMBEDDED_FRAMEWORK_NAME"
+
+        mv "$EMBEDDED_FRAMEWORK" "$ARCHIVES_ROOT/$PLATFORM.xcarchive/Products/Library/Frameworks"
+      done
+      rm -rf "$FRAMEWORK/Frameworks"
+    fi
+  done
```
fluiddot, line 79: This function simply moves embedded frameworks (i.e. nested under the `Frameworks` folder within another framework) to the root level.

.buildkite/pipeline.yml
```diff
@@ -32,9 +32,6 @@ steps:
         echo "--- :npm: Install Node dependencies"
         npm ci --unsafe-perm --prefer-offline --no-audit --no-progress
 
-        echo "--- :package: Run bundle prep work"
-        npm run prebundle:js
-
```
fluiddot, line 37: I skipped this step to speed up the bundle generation, it should be reverted before merging the PR.

ios-xcframework/build.sh
AliSoftware: I'm curious if this would be enough to make it work, i.e. I wonder if we don't need to update some `@loadpath` (via some `install_name_tool` dance) so that the dynamic loader (dyld) would be able to properly load the (ex-embded) framework from the parent framework (i.e. load Hermes from Gutenberg) correctly (especially if the load path as printed by `otool -L` would show Hermes being referenced as a relative path from Gutenberg) 🤔 

I _think_ that if you run `otool -L DerivedData/Build/Products/…/WordPress.app/WordPress` on the compiled and linked WordPress executable after it has integrated the `Gutenberg.xcframework` via CocoaPods, you should be able to check that both Gutenberg and Hermes are referenced as `@rpath/[Name].framework/[Name]`? Though I'd be curious to see if `otool -L DerivedData/Build/Products/…/WordPress.app/Frameworks/Gutenberg.framework/Gutenberg` would reference Hermes itself (and if so, how) or not at that point (I'm not familiar with the setup that Gio put in place and how Gutenberg and Hermes interact / might have link-time dependencies on each other)

AliSoftware: In practice, if we can find out how to do it, I think it would be better to have the archive action generate Hermes and Gutenberg as separate flattened frameworks to begin with at compile time, rather than have Hermes be generated as an embed framework by Xcode in the `xcarchive` it generates, and then have our scripts try to de-embed it manually after the fact… 🤔

fluiddot: Thank you so much @AliSoftware for the feedback about this approach 🙇 !

> I'm curious if this would be enough to make it work, i.e. I wonder if we don't need to update some `@loadpath` (via some `install_name_tool` dance) so that the dynamic loader (dyld) would be able to properly load the (ex-embded) framework from the parent framework (i.e. load Hermes from Gutenberg) correctly (especially if the load path as printed by `otool -L` would show Hermes being referenced as a relative path from Gutenberg) 🤔

I'm testing the changes in combination with https://github.com/wordpress-mobile/WordPress-iOS/pull/21021, and so far building the app locally for both the simulator and physical device seems to work. I can take a quick look at the commands you share to validate that references are not broken, although I understand I'd probably experience a build failure or runtime crash in that case 🤔.

> In practice, if we can find out how to do it, I think it would be better to have the archive action generate Hermes and Gutenberg as separate flattened frameworks to begin with at compile time, rather than have Hermes be generated as an embed framework by Xcode in the `xcarchive` it generates, and then have our scripts try to de-embed it manually after the fact… 🤔

Good idea. I assume this would be done by adding a phase script in the Xcode project to flatten the embedded frameworks, is this accurate?

AliSoftware: > I assume this would be done by adding a phase script in the Xcode project to flatten the embedded frameworks, is this accurate?

I was more thinking about selecting "Do Not Embed" in the "Frameworks and Libraries" settings of the `XCFrameworkScaffold.xcodeproj`:
image

_Though to be honest, I haven't looked at how the Gutenberg.framework compilation works (how the scaffold project is configured in details and how this interacts with the build scripts like `build.sh` and the declarations in the `podspec` and the potential custom scripts / `post_install` hooks in client app and all that), so I'd prefer to let @mokagio chime in on the details there than trying to grasp at straws without knowing the details about the bigger picture._

fluiddot: > I _think_ that if you run `otool -L DerivedData/Build/Products/…/WordPress.app/WordPress` on the compiled and linked WordPress executable after it has integrated the `Gutenberg.xcframework` via CocoaPods, you should be able to check that both Gutenberg and Hermes are referenced as `@rpath/[Name].framework/[Name]`? Though I'd be curious to see if `otool -L DerivedData/Build/Products/…/WordPress.app/Frameworks/Gutenberg.framework/Gutenberg` would reference Hermes itself (and if so, how) or not at that point (I'm not familiar with the setup that Gio put in place and how Gutenberg and Hermes interact / might have link-time dependencies on each other)

I've run the commands and seems the reference is correct. For both the app's executable and Gutenberg framwork, they point to Hermes (`@rpath/hermes.framework`).


Result of `otool -L DerivedData/.../Build/Products/Debug-iphoneos/Jetpack.app/Jetpack`:

```
[...]
@rpath/Gutenberg.framework/Gutenberg (compatibility version 1.0.0, current version 1.0.0)
@rpath/hermes.framework/hermes (compatibility version 0.12.0, current version 0.12.0)
[...]
```

Result of `otool -L DerivedData/.../Build/Products/Debug-iphoneos/Jetpack.app/Frameworks/Gutenberg.framework/Gutenberg`:

```
[...]
@rpath/Gutenberg.framework/Gutenberg (compatibility version 1.0.0, current version 1.0.0)
@rpath/hermes.framework/hermes (compatibility version 0.12.0, current version 0.12.0)
[...]
```

.buildkite/pipeline.yml
mokagio: I'm surprised the build worked without it 🤔 Maybe it's no longer necessary and we should not add it back?

fluiddot: The `prebundle` is needed to include the localizations, but skipping it doesn't break the editor. If it's not run, the editor would be only in English.

fluiddot: When installing the dependencies, we download all localizations for all plugins via the command `i18n:check-cache`. This includes tons of strings that are not even used in the editor, leading to a bigger bundle with an increase of around an extra 20 MB. The command `i18n:update` analyzes the strings referenced within the bundles generated by React Native, and removes the unneeded ones. As well as it generates the localization strings files ([example](https://github.com/wordpress-mobile/gutenberg-mobile/blob/trunk/bundle/ios/GutenbergNativeTranslations.swift)) to incorporate into the host apps.

https://github.com/wordpress-mobile/gutenberg-mobile/blob/6f1d7e7d7cc2d40f8ede8ad06430262def575414/package.json#L61

Hence, the `prebundle` command is not strictly needed when generating the bundle, but strongly recommended to optimize the bundle size. In this regard, now that we'll generate both Android and iOS bundles in CI, I'm thinking of reconsidering this step for the sake of reducing the build time. In the end, we'd only need the optimal bundle when releasing a new version of Gutenberg Mobile, so we might add a condition to the job to run `prebundle` in that situation.

https://github.com/wordpress-mobile/gutenberg-mobile/blob/a842544636680b1690adf5aa11f24e0cbf00ed63/package.json#L73

mokagio: Thank you for the explanation @fluiddot.

> In the end, we'd only need the optimal bundle when releasing a new version of Gutenberg Mobile, so we might add a condition to the job to run `prebundle` in that situation.

I see... We could do this by checking if the CI build is running on a tag

ios-xcframework/build.sh
mokagio: > I was more thinking about selecting "Do Not Embed" in the "Frameworks and Libraries" settings of the XCFrameworkScaffold.xcodeproj:

FWIW, #5973 uses that approach.

fluiddot:
Closing this PR in favor of https://github.com/wordpress-mobile/gutenberg-mobile/pull/5973.

PR #5976, opened 2023-07-18
Expand analytics for block insertion/replacement

geriux:
**Related PRs:**
- https://github.com/WordPress/gutenberg/pull/52735

This PR improves the analytics we have related to block insertion, it brings some code [from Calypso](https://github.com/Automattic/wp-calypso/blob/80bf6344bb4917633b90096716062701d45adae8/apps/wpcom-block-editor/src/wpcom/features/tracking.js) to take into account block replacement and getting the source of where a block was inserted or replaced.

To test:

**Precondition:** 
- Run the editor using the host apps with metro running.

## Inserter

- Add blocks using the Inserter
- Check the logs (Xcode/Logcat)
- **Expect** to see the event `editor_block_inserted` with the `insert_method` property with `header-inserter`

## Quick inserter

- Add blocks using the quick Inserter (Available when the title is selected or when no block is selected) and the quick blocks buttons in the Header Toolbar.
- Check the logs (Xcode/Logcat)
- **Expect** to see event `editor_block_inserted` with the `insert_method` property with `quick-inserter`

## Slash inserter

- Add blocks using slash inserter.
- Check the logs (Xcode/Logcat)
- **Expect** to see event `editor_block_inserted` with the `insert_method` property with `slash-inserter`

PR submission checklist:


geriux:
Hey @fluiddot 👋 I have added you as a reviewer for this PR, let me know if you could take a look whenever you can. Thank you!

fluiddot:
> Hey @fluiddot 👋 I have added you as a reviewer for this PR, let me know if you could take a look whenever you can. Thank you!

Sure, I can take a look 👀 .

src/analytics/redux/tracked_events.js
```diff
@@ -19,6 +69,41 @@ function getBlockObject( block ) {
...
+ */
+const trackBlockReplacement = ( originalBlockIds, blocks, ...args ) => {
+	const metaData = args?.[ 3 ] ?? {};
+	const insert_method = getBlockInserterUsed( originalBlockIds, metaData );
+
+	trackBlocksHandler( blocks, 'editor_block_inserted', ( { name } ) => ( {
+		block_name: name,
+		insert_method,
+	} ) );
+};
```
fluiddot: Not sure if we should track all block replacements as insertions. For instance, transforming a block triggers the `editor_block_inserted` event when it's not an actual insertion. You can check this by adding a block and transforming it into another one.

I understand we need to listen for block replacement to identify the insertions via the slash inserter, however, maybe we should condition generating the `editor_block_inserted` only for that case. WDYT?

src/analytics/redux/tracked_events.js
```diff
@@ -111,20 +196,10 @@ function handleBlockMovedByPosition( clientIds, toIndex ) {
...
-			);
-		},
-		insertBlocks( blocks ) {
-			trackBlocksHandler(
-				blocks,
-				'editor_block_inserted',
-				( { name } ) => ( { block_name: name } )
-			);
-		},
+		insertBlock: trackBlockInsertion,
```
fluiddot: I noticed that the `insertBlock` Redux action is actually invoking `insertBlocks`. Could we simply track events for the latter?

https://github.com/WordPress/gutenberg/blob/d27c43fda419995a80b1cbdeafe2b24d9a0e2164/packages/block-editor/src/store/actions.js#L541-L548

src/analytics/redux/tracked_events.js
```diff
@@ -19,6 +69,41 @@ function getBlockObject( block ) {
...
+
+/**
+ * Track block insertion.
+ *
+ * @param {Object | Array} blocks block instance object or an array of such objects
+ * @param {Array}          args   additional insertBlocks data e.g. metadata containing inserter method.
+ * @return {void}
+ */
+function trackBlockInsertion( blocks, ...args ) {
+	const metaData = args?.[ 3 ] ?? {};
```
fluiddot: I'm wondering if, instead of accessing the arguments by index, we could directly fetch the `meta` argument defined in the arguments of the `insertBlock` Redux action:

https://github.com/WordPress/gutenberg/blob/d27c43fda419995a80b1cbdeafe2b24d9a0e2164/packages/block-editor/src/store/actions.js#L534-L540

I.e. apply the following change:
```
trackBlockInsertion( blocks, ...args ) {
```
=>
```
function trackBlockInsertion(
	blocks,
	index,
	rootClientId,
	updateSelection,
	initialPosition,
	meta
) {
```

src/analytics/redux/tracked_events.js
```diff
@@ -111,20 +196,10 @@ function handleBlockMovedByPosition( clientIds, toIndex ) {
...
-		insertBlocks( blocks ) {
-			trackBlocksHandler(
-				blocks,
-				'editor_block_inserted',
-				( { name } ) => ( { block_name: name } )
-			);
-		},
+		insertBlock: trackBlockInsertion,
+		insertBlocks: trackBlockInsertion,
+		replaceBlock: trackBlockReplacement,
```
fluiddot, line 222: Same as the above comment, since `replaceBlock` is invoking `replaceBlocks`, could we simply track events for the latter?

https://github.com/WordPress/gutenberg/blob/d27c43fda419995a80b1cbdeafe2b24d9a0e2164/packages/block-editor/src/store/actions.js#L416-L418

fluiddot (commented):
I checked the events generated and confirmed it's adding the proper inserter source 🎊. However, I noticed that we are generating the `editor_block_inserted` event in more cases than before ([reference](https://github.com/wordpress-mobile/gutenberg-mobile/pull/5976#discussion_r1267771581)).

src/analytics/redux/tracked_events.js
```diff
@@ -19,6 +69,41 @@ function getBlockObject( block ) {
...
+/**
+ * Track block replacement.
+ *
+ * @param {Array}          originalBlockIds ids or blocks that are being replaced
+ * @param {Object | Array} blocks           block instance object or an array of such objects
+ * @param {Array}          args             additional insertBlocks data e.g. metadata containing inserter method.
+ * @return {void}
+ */
+const trackBlockReplacement = ( originalBlockIds, blocks, ...args ) => {
+	const metaData = args?.[ 3 ] ?? {};
```
fluiddot: Similar to [this comment](https://github.com/wordpress-mobile/gutenberg-mobile/pull/5976#discussion_r1267781815), we could consider defining the arguments coming from `replaceBlock`/`replaceBlocks`.

src/analytics/redux/tracked_events.js
geriux: Nice catch! I agree, we should avoid tracking all replacements. 

>I understand we need to listen for block replacement to identify the insertions via the slash inserter, however, maybe we should condition generating the editor_block_inserted only for that case. WDYT?

This is a good idea, I've applied it in https://github.com/wordpress-mobile/gutenberg-mobile/pull/5976/commits/61c29a21bd1a29482680086789c2ab7028dbe2f2

Let me know what you think, thanks!

src/analytics/redux/tracked_events.js
geriux: Thanks for the suggestion! I'm currently checking why `insertBlocks` doesn't work but `insertBlock` does 🙃

src/analytics/redux/tracked_events.js
fluiddot: Yeah, thanks! Definitely, that workaround will limit triggering the event.

src/analytics/redux/tracked_events.js
geriux: So I had to keep both, it only tracks the first dispatched action, so if `insertBlock` is dispatched, that's the action that will be tracked and not the `insertBlocks` that's called within.

I updated here https://github.com/wordpress-mobile/gutenberg-mobile/pull/5976/commits/effe376cb59ea829723d0d8538e4bfc8731441db let me know what you think 😄

src/analytics/redux/tracked_events.js
geriux: Updated it in https://github.com/wordpress-mobile/gutenberg-mobile/pull/5976/commits/7277128f7526d0557303f58a89bca48120c3f88e and https://github.com/wordpress-mobile/gutenberg-mobile/pull/5976/commits/effe376cb59ea829723d0d8538e4bfc8731441db#diff-b264ddd07fecd2f9655557832fb8f89baa62f7f89c387e80df664c26ceb36084

src/analytics/redux/tracked_events.js
geriux: More context of this approach in https://github.com/wordpress-mobile/gutenberg-mobile/pull/5976#discussion_r1268144468 but after some testing I realized that only the action that was dispatched initially is the one tracked

src/analytics/redux/tracked_events.js
geriux: Good idea! Updated in https://github.com/wordpress-mobile/gutenberg-mobile/pull/5976/commits/7277128f7526d0557303f58a89bca48120c3f88e

src/analytics/redux/tracked_events.js
fluiddot: Ah, ok. If only the first dispatched action is processed let's keep both.

src/analytics/redux/tracked_events.js
```diff
@@ -19,6 +69,66 @@ function getBlockObject( block ) {
...
+ * @param {?Object}        meta             Optional Meta values to be passed to the action object.
+ *
+ * @return {void}
+ */
+const trackBlockReplacement = (
+	originalBlockIds,
+	blocks,
+	indexToSelect,
+	// eslint-disable-next-line no-unused-vars
+	initialPosition = 0,
```
fluiddot: I think we don't need to disable the ESLint rule if we don't set a default value for `initialPosition`. We are not using this argument, so I'd say it's safe to remove the default value.

src/analytics/redux/tracked_events.js
```diff
@@ -19,6 +69,66 @@ function getBlockObject( block ) {
...
+ *
+ * @return {void}
+ */
+function trackBlockInsertion(
+	blocks,
+	index,
+	rootClientId,
+	updateSelection,
+	// eslint-disable-next-line no-unused-vars
+	initialPosition = 0,
```
fluiddot: I think we don't need to disable the ESLint rule if we don't set a default value for `initialPosition`. We are not using this argument, so I'd say it's safe to remove the default value.

src/analytics/redux/tracked_events.js
```diff
@@ -111,20 +221,35 @@ function handleBlockMovedByPosition( clientIds, toIndex ) {
...
+		insertBlock( blocks, index, rootClientId, updateSelection, meta ) {
+			trackBlockInsertion(
 				blocks,
-				'editor_block_inserted',
-				( { name } ) => ( { block_name: name } )
+				index,
+				rootClientId,
+				updateSelection,
+				undefined,
+				meta
```
fluiddot: When I suggested [this](https://github.com/wordpress-mobile/gutenberg-mobile/pull/5976#discussion_r1267771581), we had the following code:

https://github.com/wordpress-mobile/gutenberg-mobile/blob/c6163ef83ba8ca7c3a30b03126ef93faefe5c004/src/analytics/redux/tracked_events.js#L197-L198

With this change, seems we can remove all the unused arguments, i.e.:
```
insertBlock( blocks, index, rootClientId, updateSelection, meta ) {
	trackBlockInsertion( blocks, meta );
},
```

This way we could also remove:

https://github.com/wordpress-mobile/gutenberg-mobile/blob/effe376cb59ea829723d0d8538e4bfc8731441db/src/analytics/redux/tracked_events.js#L117-L121

WDYT?

src/analytics/redux/tracked_events.js
```diff
@@ -111,20 +221,35 @@ function handleBlockMovedByPosition( clientIds, toIndex ) {
...
+		) {
+			trackBlockInsertion(
 				blocks,
-				'editor_block_inserted',
-				( { name } ) => ( { block_name: name } )
+				index,
+				rootClientId,
+				updateSelection,
+				initialPosition,
+				meta
```
fluiddot: Similar to [this comment](https://github.com/wordpress-mobile/gutenberg-mobile/pull/5976#discussion_r1268329007), we could remove the unused arguments.

src/analytics/redux/tracked_events.js
geriux: 😅 thanks for catching that! I've updated in https://github.com/wordpress-mobile/gutenberg-mobile/pull/5976/commits/94df0273bad2f0dc1a9422b4a590e5ecb7b3beb5#diff-b264ddd07fecd2f9655557832fb8f89baa62f7f89c387e80df664c26ceb36084L87

src/analytics/redux/tracked_events.js
geriux: Thanks for the suggestion! I ended up removing this one in https://github.com/wordpress-mobile/gutenberg-mobile/pull/5976/commits/94df0273bad2f0dc1a9422b4a590e5ecb7b3beb5#diff-b264ddd07fecd2f9655557832fb8f89baa62f7f89c387e80df664c26ceb36084L121 as your other suggestion about unneeded params, thanks!

src/analytics/redux/tracked_events.js
geriux: Good idea! It'll simplify it a bit, thanks! Updated in https://github.com/wordpress-mobile/gutenberg-mobile/pull/5976/commits/94df0273bad2f0dc1a9422b4a590e5ecb7b3beb5

src/analytics/redux/tracked_events.js
geriux: Updated it in https://github.com/wordpress-mobile/gutenberg-mobile/pull/5976/commits/94df0273bad2f0dc1a9422b4a590e5ecb7b3beb5 thanks!

src/analytics/redux/tracked_events.js
```diff
@@ -19,6 +69,53 @@ function getBlockObject( block ) {
...
+/**
+ * Track block replacement.
+ *
+ * @param {Array}          originalBlockIds ID(s) or blocks that are being replaced
+ * @param {Object | Array} blocks           Block instance object or an array of such objects
+ * @param {number}         indexToSelect    Index of replacement block to select.
+ * @param {0|-1|null}      initialPosition  Index of caret after in the selected block after the operation.
+ * @param {?Object}        meta             Optional Meta values to be passed to the action object.
+ *
+ * @return {void}
```
fluiddot: Nitpick: if the function doesn't return a value probably we don't need to document the `return` statement.

src/analytics/redux/tracked_events.js
```diff
@@ -5,6 +5,56 @@
...
+const INSERTERS = {
+	HEADER_INSERTER: 'header-inserter',
+	SLASH_INSERTER: 'slash-inserter',
+	QUICK_INSERTER: 'quick-inserter',
+};
+
+/**
+ * Guess which inserter was used to insert/replace blocks.
+ *
+ * @param {string[]|string} originalBlockIds ids or blocks that are being replaced
```
fluiddot: Nitpick: Since we can pass a single value, we could reflect this in the argument description.
```suggestion
 * @param {string[]|string} originalBlockIds ID(s) of block(s) that are being inserted/replaced
```

src/analytics/redux/tracked_events.js
```diff
@@ -19,6 +69,53 @@ function getBlockObject( block ) {
 	return select( 'core/block-editor' ).getBlock( block ) || {};
 }
 
+/**
+ * Track block replacement.
+ *
+ * @param {Array}          originalBlockIds ID(s) or blocks that are being replaced
```
fluiddot: Nitpick: Based on the argument's description, I'm not sure if it could be confusing that we mention "ID(s) **or** blocks". I understand we only allow ID(s) **of** blocks. What do you think about updating this? I.e.:
```suggestion
 * @param {Array}          originalBlockIds ID(s) of block(s) that are being replaced
```

fluiddot (approved):
LGTM 🎊 ! I've added some minor comments related to function documentation, but feel free to skip them.

src/analytics/redux/tracked_events.js
geriux: Makes sense! Updated in https://github.com/wordpress-mobile/gutenberg-mobile/pull/5976/commits/b2e4ee34b0b97710deb3dbf3aca02d306da753f8

src/analytics/redux/tracked_events.js
geriux: Updated in https://github.com/wordpress-mobile/gutenberg-mobile/pull/5976/commits/b2e4ee34b0b97710deb3dbf3aca02d306da753f8

src/analytics/redux/tracked_events.js
geriux: That's correct, we only allow ID(s) of blocks so I updated it as suggested in https://github.com/wordpress-mobile/gutenberg-mobile/pull/5976/commits/b2e4ee34b0b97710deb3dbf3aca02d306da753f8

PR #5982, opened 2023-07-20
Fix crash when registering non-core blocks

fluiddot:
## Issue

With https://github.com/WordPress/gutenberg/pull/52417, now blocks are memoized in `useEntityBlockEditor` hook. This hook [is used in the `EditorProvider` component](https://github.com/WordPress/gutenberg/blob/6d22c92b664927d0894cec816adfb07769af8688/packages/editor/src/components/provider/index.js#L58), so the first call is made right after the editor is initialized. When the initial content is parsed [here](https://github.com/WordPress/gutenberg/blob/eac3f039b0bd094af35291987390ea1c5cc473d9/packages/core-data/src/entity-provider.js#L180), it processes the blocks based on the block types registered at that time.

The Jetpack blocks [are registered on the root component's `componentDidMount` handler](https://github.com/WordPress/gutenberg/blob/1e04b10d77c28bb785a2ee7ea70eb7c713682f6e/packages/react-native-editor/src/index.js#L63-L66), which is executed after the `useEntityBlockEditor` hook is invoked the first time. Therefore, when parsing the initial content Jetpack blocks are marked as missing blocks, as they haven't been registered yet by that time.

https://github.com/wordpress-mobile/gutenberg-mobile/blob/25c8620b7186d8fc56e548ec0b78f08688fc2b88/src/jetpack-editor-setup.js#L149-L152

**Example content parsed:**

```
[
  {
    "attributes": { "content": "VideoPress block:", "dropCap": false },
    "clientId": "48445782-74e3-44a7-8d42-a4f75cc783f5",
    "innerBlocks": [],
    "isValid": true,
    "name": "core/paragraph",
    "originalContent": "<p>VideoPress block:</p>",
    "validationIssues": []
  },
  {
    "attributes": {
      "originalContent": "<!-- wp:videopress/video {\"title\":\"file_example_mp4_480_1_5mg\",\"description\":\"\",\"id\":6037,\"guid\":\"6gn6ukfN\",\"privacySetting\":2,\"allowDownload\":false,\"rating\":\"G\",\"isPrivate\":true,\"duration\":30466} /-->",
      "originalName": "videopress/video",
      "originalUndelimitedContent": ""
    },
    "clientId": "5c9a01f4-35e2-42ad-94e4-bc20025523dc",
    "innerBlocks": [],
    "isValid": true,
    "name": "core/missing",
    "originalContent": "<!-- wp:videopress/video {\"title\":\"file_example_mp4_480_1_5mg\",\"description\":\"\",\"id\":6037,\"guid\":\"6gn6ukfN\",\"privacySetting\":2,\"allowDownload\":false,\"rating\":\"G\",\"isPrivate\":true,\"duration\":30466} /-->",
    "validationIssues": []
  }
]
```

When the `Missing` blocks are rendered, [this line](https://github.com/WordPress/gutenberg/blob/eac3f039b0bd094af35291987390ea1c5cc473d9/packages/block-library/src/missing/index.js#L29), which calculates the label for the block, produces the crash. Seems that by that time, the proper Jetpack block is fetched for the `originalBlockType` variable. However, those don't have the `settings` attribute defined.

WordPress/gutenberg#52417 actually uncovered an issue regarding how we register Jetpack blocks, as we were registering them probably too late in the initialization cycle.

## Solution

The workaround for this is to [add a new WP hook that allows Gutenberg Mobile to register non-core blocks](https://github.com/WordPress/gutenberg/pull/52791), like Jetpack blocks, right after the core blocks are registered. This way we ensure that all blocks are registered sequentially (first core blocks, then non-core blocks) before any other calculations are made within the editor.

## To test

1. Create a post.
2. Add Jetpack blocks (e.g. VideoPress block).
3. Save the post and close the editor.
4. Re-open the post.
5. Observe that no crash is produced.
6. Observe that none of the blocks are rendered as missing.

PR submission checklist:

### DIFF
--- a/src/block-experiments-setup.js
+++ b/src/block-experiments-setup.js
@@ -16,7 +16,7 @@ const blockExperiments = [ 'jetpack/layout-grid' ];
 const setupHooks = () => {
 	// Hook triggered after the editor is rendered
 	addAction(
-		'native.render',
+		'native.post-register-core-blocks',
 		'gutenberg-mobile-block-experiments',
 		( props ) => {
 			const capabilities = props.capabilities ?? {};
--- a/src/jetpack-editor-setup.js
+++ b/src/jetpack-editor-setup.js
@@ -147,9 +147,13 @@ const setupHooks = () => {
 	} );
 
 	// Hook triggered after the editor is rendered
-	addAction( 'native.render', 'gutenberg-mobile-jetpack', ( props ) => {
-		registerJetpackBlocks( props );
-	} );
+	addAction(
+		'native.post-register-core-blocks',
+		'gutenberg-mobile-jetpack',
+		( props ) => {
+			registerJetpackBlocks( props );
+		}
+	);
 };
 
 const setupStringsOverrides = () => {


geriux (approved):
LGTM! Nice approach to fix this regression 🚀 The Gutenberg side of changes were approved in https://github.com/WordPress/gutenberg/pull/52791#pullrequestreview-1539579291

I tested different blocks and it's working as expected on both iOS and Android using the host apps.

fluiddot:
For some reason, all jobs are shown here as "waiting", however, they are running in CircleCI and Buildkite 🤔.

**UPDATE:** I ended up pushing an empty commit just to trigger CI jobs.

Issue #6023, opened 2023-08-03
Hermes stack traces can't be symbolicated


fluiddot:
# Problem

Currently, we can't symbolicate stack traces of Android crashes related to the editor. This is caused by:
1. We don't store source maps of the bundle generated for Android. They can be manually generated by checking out the code for a specific version, although as far as I checked, they don't provide an accurate result when symbolicating.
2. The symbolicating steps outlined in [the React Native documentation](https://reactnative.dev/docs/0.71/symbolication) don't work for Hermes. Seems we need an ad-hoc solution to symbolicate them.

Starting in the app's version `23.0`, we'll have the same problem on iOS as we recently enabled Hermes.

# Solution

## Store source maps in GitHub releases

- [x] https://github.com/wordpress-mobile/gutenberg-mobile/pull/6032
- [x] https://github.com/wordpress-mobile/release-toolkit-gutenberg-mobile/pull/143

Source maps are only needed when creating a release, as it's the version that is released to users. Hence, we can attach the source maps generated during the release process to the GitHub release.

**UPDATE:** It's important to use the same source maps produced when generating the bundle. Otherwise, we might encounter wrong symbolicated stack traces.

This will be a manual step for now, but we could automate it in CI. In this regard, there's an [ongoing effort to upload the iOS XCFramework to GitHub releases](https://github.com/wordpress-mobile/gutenberg-mobile/pull/5863). Probably we could follow the same approach and upload other artifacts like the source maps ([reference](https://github.com/wordpress-mobile/gutenberg-mobile/pull/5863/files#r1282985264)).

### 1. ~~Preliminary exploration~~

**⚠️ This approach has been discarded with more simpler script using the command `metro-symbolicate` (see section 2).**

I explored in the branch [`add/node-symbolicate-stacktrace-script`](https://github.com/wordpress-mobile/gutenberg-mobile/compare/add/add/node-symbolicate-stacktrace-script) a script that is likely to address this problem. It uses the Metro symbolication library, similar to the [`metro-symbolicate` npm command](https://reactnative.dev/docs/0.71/symbolication), but in a two-steps process:
1. Get the position using the source map generated for the byte-coded bundle, which points to the source bundle.
2. Using the above, get the position using the source map generated for the original source.

As an example, here you can see the values obtained from the source maps for a single stack trace entry:
```
{
  stacktraceEntry: 'Gn@1:172289',
  bytecodePosition: {
    line: 58,
    column: 25161,
[...]

PR #6032, opened 2023-08-04
Upload composed source maps to Buildkite

fluiddot:
**Related to https://github.com/wordpress-mobile/gutenberg-mobile/issues/6023.**

This PR adds steps to the build workflow to upload the source maps to Buildkite. This way we'll store the most accurate source maps of bundle produced in CI. When symbolicating a stack trace, it's important to use the same source maps produced for the bundle used in the app. Otherwise, the stack trace will point to the wrong source code position.

As a side note, I compared a symbolicated stack trace between a source map generated locally and the one from the CI job and confirmed they don't match.

Currently, two source map files are produced when generating the bundle:
- `App.js.map`: Source map of the bytecode source.
- `App.text.js.map`: Source map of the javascript code.

They can be composed using a script provided by React Native ([`react-native/scripts/compose-source-maps.js`](https://github.com/facebook/react-native/blob/v0.71.11/scripts/compose-source-maps.js)) to simplify the process of symbolicating a stack trace. Further information about source maps and Hermes can be found in [Sentry documentation](https://docs.sentry.io/platforms/react-native/manual-setup/hermes/).

**To test:**
* Observe that all CI succeed.
* Check the `buildkite/gutenberg-mobile/build-js-bundles` job and observe in the Artifacts tab that it generated the following files:
Screenshot 2023-08-04 at 15 23 36

PR submission checklist:

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -44,15 +44,19 @@ steps:
 
         echo "--- :android: Build Android bundle"
         npm run bundle:android
+        node gutenberg/node_modules/react-native/scripts/compose-source-maps.js bundle/android/App.text.js.map bundle/android/App.js.map -o bundle/android/App.composed.js.map
 
         echo "--- :arrow_up: Upload Android bundle artifact"
         buildkite-agent artifact upload bundle/android/App.js
+        buildkite-agent artifact upload bundle/android/App.composed.js.map
 
         echo "--- :ios: Build iOS bundle"
         npm run bundle:ios
+        node gutenberg/node_modules/react-native/scripts/compose-source-maps.js bundle/ios/App.text.js.map bundle/ios/App.js.map -o bundle/ios/App.composed.js.map
 
         echo "--- :arrow_up: Upload iOS bundle artifact"
         buildkite-agent artifact upload bundle/ios/App.js
+        buildkite-agent artifact upload bundle/ios/App.composed.js.map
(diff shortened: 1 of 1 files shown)


SiobhyB (approved):
@fluiddot, LGTM, thank you for all the work to improve our processes 🙏

.buildkite/pipeline.yml
```diff
@@ -44,15 +44,19 @@ steps:
 
         echo "--- :android: Build Android bundle"
         npm run bundle:android
+        node gutenberg/node_modules/react-native/scripts/compose-source-maps.js bundle/android/App.text.js.map bundle/android/App.js.map -o bundle/android/App.composed.js.map
```
fluiddot: @SiobhyB I realized that source maps will only be needed for releases, so I'm going to limit the upload when building from a Git tag. Sorry for the last-minute change.

fluiddot: The change can be checked in the following Buildkite jobs:
- https://buildkite.com/automattic/gutenberg-mobile/builds/6753
- https://buildkite.com/automattic/gutenberg-mobile/builds/6754

SiobhyB: Good catch! Makes total sense to limit the uploads to tags.

SiobhyB (approved):
@fluiddot, thanks for catching that this could be limited to tags. 🙇‍♀️

PR #6048, opened 2023-08-09
Add `cache-builder.yml` pipeline

mokagio:
This pipeline will be used by a scheduled job on Buildkite without further changes require in this repo.

To test: See this manually scheduled build: https://buildkite.com/automattic/gutenberg-mobile/builds/6790

image

_The warning is expected, given this was the first cache ever 😄_

---

PR submission checklist:

### DIFF
--- a/.buildkite/cache-builder.yml
+++ b/.buildkite/cache-builder.yml
@@ -5,7 +5,7 @@
 # It's meant to rebuild various CI caches on a periodic async basis, so as
 # not to waste time on every CI job updating the cache.
 
-common_params:
+x-common-params:
   - &common_plugins
     - automattic/a8c-ci-toolkit#2.18.2
     - automattic/git-s3-cache#1.1.4:
--- a/.buildkite/cache-builder.yml
+++ b/.buildkite/cache-builder.yml
@@ -1,0 +1,23 @@
+# This pipeline runs via Buildkite's scheduled jobs feature.
+#
+# It's meant to rebuild various CI caches on a periodic async basis, so as
+# not to waste time on every CI job updating the cache.
+
+common_params:
+  - &common_plugins
+    - automattic/a8c-ci-toolkit#2.18.2
+    - automattic/git-s3-cache#1.1.4:
+        bucket: a8c-repo-mirrors
+        repo: automattic/gutenber-mobile/
+
+steps:
+  # Build the Git Repo cache
+  #
+  # Because this repo is so large, we periodically create a Git Mirror and copy it to S3,
+  # from where it can be fetched by agents more quickly than from GitHub, and so that
+  # agents then have less new commits to `git pull` by using that cache as starting point
+  - label: ':git: Rebuild git cache'
+    command: cache_repo a8c-repo-mirrors
+    plugins: *common_plugins
+    agents:
+      queue: default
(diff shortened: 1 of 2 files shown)


.buildkite/cache-builder.yml
```diff
@@ -0,0 +1,23 @@
...
+  # Build the Git Repo cache
+  #
+  # Because this repo is so large, we periodically create a Git Mirror and copy it to S3,
+  # from where it can be fetched by agents more quickly than from GitHub, and so that
+  # agents then have less new commits to `git pull` by using that cache as starting point
+  - label: ':git: Rebuild git cache'
+    command: cache_repo a8c-repo-mirrors
+    plugins: *common_plugins
+    agents:
+      queue: default
```
mokagio, line 25: Other repos cache the pods as well. I skipped this here because the command used for the caching expects `Podfile` to be in the root.

We'll get a good speed bump from caching the Git repo. We can consider caching the pods later on.

.buildkite/cache-builder.yml
```diff
@@ -0,0 +1,23 @@
+# This pipeline runs via Buildkite's scheduled jobs feature.
```
AliSoftware, line 3: 💄  So that you get syntax highlighting and autocomplete in IDEs like Visual Studio:
```suggestion
# yaml-language-server: $schema=https://raw.githubusercontent.com/buildkite/pipeline-schema/main/schema.json
---
# This pipeline runs via Buildkite's scheduled jobs feature.
```

.buildkite/cache-builder.yml
```diff
@@ -0,0 +1,23 @@
...
+#
+# It's meant to rebuild various CI caches on a periodic async basis, so as
+# not to waste time on every CI job updating the cache.
+
+common_params:
```
AliSoftware: 💄  Nitpick: doesn't really matter, but since the `pipeline.yml` file uses `x-common-params` as the dummy key used to contain all our anchors we want to use later, it can make sense to make both keys consistent (either by using `x-common-params` in this new `cache-builder.yml`, or by using `common_params` key in `pipeline.yml`, it doesn't really matter which as long as we're consistent) 🙃

AliSoftware (approved):
Looking good. 🚢 

ℹ️  Side note that I noticed that you manually updated the pipeline steps in Buildkite's UI to account for `$${PIPELINE:-pipeline.yml}`, but **your manual change won't work** because of that `$$` escape sequence which should have been a `$` when used in the Buildkite UI (while it's correct—and necessary—to use `$$` in the `.tf` file though, to escape it and prevent terraform from doing the env var evaluation itself, but instead let terraform unescape that `$$` from the `.tf` YAML into a verbatim `$` in the Buildkite settings, so that Buildkite will be the one doing the env var evaluation on each pipeline runs)

So if you were surprised that your manual testing of the setup—that you might have tried while waiting on the `buildkite-ci` PR to land—didn't work as expected, that would be why.

.buildkite/cache-builder.yml
mokagio: Oh, good point. I hadn't noticed that. Consistency FTW.

```suggestion
x-common-params:
```

mokagio:
@AliSoftware thank you for pointing out the `$$` misuse.

> So if you were surprised that your manual testing of the setup—that you might have tried while waiting on the buildkite-ci PR to land—didn't work as expected, that would be why.

For the record, my test succeeded. That is, I wanted the a build with the cache builder to run [and it did](https://buildkite.com/automattic/gutenberg-mobile/builds/6790). I guess I was lucky in that the parser must have used the [`$` escaping](https://buildkite.com/docs/agent/v3/cli-pipeline#environment-variable-substitution-escaping-the-dollars-character) in that instance, too.

image

PR #6051, opened 2023-08-09
test: Add jest-watch-typeahead

dcalhoun:
Simplify scoping test runs in watch mode. This addition mirrors an
already completed addition in Gutenberg proper. Relates to https://github.com/WordPress/gutenberg/pull/51869.

To test: 

1. Enable Jest watch mode: `npm run test -- --watch`
2. Press <kbd>p</kbd> to enable test file name filtering. 
3. Type to begin filtering, select a test file. 
4. Verify only the selected file runs. 
5. Press <kbd>t</kbd> to enable test name filtering. 
6. Type to begin filtering, select a test. 
7. Verify only the selected test runs. 

PR submission checklist:


jest.config.js
```diff
@@ -74,4 +74,8 @@ module.exports = {
 		'node_modules/(?!(simple-html-tokenizer|@react-native-community|(jest-)?react-native|@react-native|react-clone-referenced-element|is-plain-obj))',
 	],
 	reporters: [ 'default', 'jest-junit' ],
+	watchPlugins: [
+		'jest-watch-typeahead/filename',
+		'jest-watch-typeahead/testname',
```
fluiddot, line 79: I'm wondering if instead of adding the packages, we could simply point to the module in Gutenberg:

```suggestion
		'./gutenberg/node_modules/jest-watch-typeahead/filename.js',
		'./gutenberg/node_modules/jest-watch-typeahead/testname.js',
```

WDYT?

fluiddot (commented):
I noticed this change only works for unit/integration tests. When running the command `TEST_RN_PLATFORM=ios npm run device-tests:local -- --watch`, although it's using Watch it doesn't find any test file 🤔. I wouldn't block this PR due to this but eventually, it would be great to also have this feature available when running E2E tests.

jest.config.js
dcalhoun: My understanding is the general recommendation/convention is to install development dependencies  at the root of a monorepo. This repository isn't a monorepo in the traditional Node.js ecosystem format with a `packages` directory, but I still consider it a monorepo.

We could treat this repository as a "passthrough" of sorts, minimizing the configuration and always referencing configuration from child packages, but I feel we will likely hit a roadblock at some point with that approach, e.g. Metro requiring explicit one-off configuration in this repository (that may already be true). 

So, my thought is that we keep the configuration here as-is and structure this project in a manner that directly communicates its development dependencies.

dcalhoun (commented):
> I noticed this change only works for unit/integration tests. When running the command `TEST_RN_PLATFORM=ios npm run device-tests:local -- --watch`, although it's using Watch it doesn't find any test file 🤔. I wouldn't block this PR due to this but eventually, it would be great to also have this feature available when running E2E tests.

This should not longer be the case now that the sibling https://github.com/WordPress/gutenberg/pull/53479 is incorporated into this branch. Apologies for the confusion.

jest.config.js
fluiddot: > We could treat this repository as a "passthrough" of sorts, minimizing the configuration and always referencing configuration from child packages, but I feel we will likely hit a roadblock at some point with that approach, e.g. Metro requiring explicit one-off configuration in this repository (that may already be true).

Good point. My main concern about the "monorepo" approach is regarding the maintenance of packages that are referenced in both repositories, especially for matching versions. Probably for this package won't be a problem, but for others using different versions might lead to incompatibilities. Also, I feel the Gutenberg Mobile is coupled enough to Gutenberg to consider the "passthrough" as a valid option.

That said, I'm not opposed to the approach but it would be great if we follow this approach in other areas like configurations. For instance, the [Metro configuration is inherited from Gutenberg](https://github.com/wordpress-mobile/gutenberg-mobile/blob/b1891908b3eda693350e585f75db30440a66eabb/metro.config.js#L5), and something similar happens with the [Jest configuration for UI tests](https://github.com/wordpress-mobile/gutenberg-mobile/blob/b1891908b3eda693350e585f75db30440a66eabb/jest_ui.config.js#L1).

> So, my thought is that we keep the configuration here as-is and structure this project in a manner that directly communicates its development dependencies.

Sound good to me 😊 .

dcalhoun: > My main concern about the "monorepo" approach is regarding the maintenance of packages that are referenced in both repositories, especially for matching versions. Probably for this package won't be a problem, but for others using different versions might lead to incompatibilities.

That is a fair concern. However, I imagine this concern applies to many — if not all — of the existing [development dependencies](https://github.com/wordpress-mobile/gutenberg-mobile/blob/b1891908b3eda693350e585f75db30440a66eabb/package.json#L13-L59), correct? 

> That said, I'm not opposed to the approach but it would be great if we follow this approach in other areas like configurations. For instance, the [Metro configuration is inherited from Gutenberg](https://github.com/wordpress-mobile/gutenberg-mobile/blob/b1891908b3eda693350e585f75db30440a66eabb/metro.config.js#L5), and something similar happens with the [Jest configuration for UI tests](https://github.com/wordpress-mobile/gutenberg-mobile/blob/b1891908b3eda693350e585f75db30440a66eabb/jest_ui.config.js#L1).

I did note that the integration configuration differed from the UI configuration in this manner. I suppose that may oversight.

fluiddot (approved):
LGTM 🎊 ! I've just tested running unit and E2E tests and worked like a charm. Thanks @dcalhoun for introducing this useful improvement 🙇 !

jest.config.js
fluiddot: > That is a fair concern. However, I imagine this concern applies to many — if not all — of the existing [development dependencies](https://github.com/wordpress-mobile/gutenberg-mobile/blob/b1891908b3eda693350e585f75db30440a66eabb/package.json#L13-L59), correct?

Absolutely, this concern is not specific to this package. I think in the end this concern originated from the topic of where to position the Gutenberg Mobile repository, and how dependent or independent should be with other repositories like the host apps, Jetpack and Gutenberg itself. I know we've surfaced a few times the idea of merging Gutenberg Mobile into the Gutenberg repository, probably that's why I lean a bit toward the "passthrough" approach 😅 .

PR #6076, opened 2023-08-17
feat: Register the Paywall block

dcalhoun:
Register the Paywall block for the mobile editor. This enables rendering the Paywall block for posts that contain it, but the Paywall block is not available for insertion with the block inserter.

## Related PRs

* https://github.com/WordPress/gutenberg/pull/53883
* https://github.com/Automattic/jetpack/pull/32553
* https://github.com/wordpress-mobile/WordPress-iOS/pull/21359
* https://github.com/wordpress-mobile/WordPress-Android/pull/18998

### Install a _Jetpack_ app prototype build ([Jetpack iOS](https://github.com/wordpress-mobile/WordPress-iOS/pull/21359#issuecomment-1682607689), [Jetpack Android](https://github.com/wordpress-mobile/WordPress-Android/pull/18998#issuecomment-1682552711)), and then...

**Paywall block render**
1. Create and save a post on the web containing a Paywall block: `<!-- wp:jetpack/paywall /-->`
2. Open the same post in the Jetpack mobile on iOS or Android. 
3. Verify the Paywall block renders and displays a "Exclusive content below this line" label. 

**Paywall block insertion disabled**
1. Open the mobile editor. 
2. Within a Paragraph block, type `/paywall`
3. Verify the Paywall block is not available. 
4. Tap the <kbd>+</kbd> button above the keyboard to open the mobile editor block inserter. 
5. Verify the Paywall block is not available. 

### Install a _WordPress_ app prototype build ([WPiOS](https://github.com/wordpress-mobile/WordPress-iOS/pull/21359#issuecomment-1682614391), [WPAndroid](https://github.com/wordpress-mobile/WordPress-Android/pull/18998#issuecomment-1682550810)), and then...

**Paywall block remains unsupported**
1. Create and save a post on the web containing a Paywall block: `<!-- wp:jetpack/paywall /-->`
2. Open the same post in the WordPress mobile editor on iOS or Android. 
3. Verify the Paywall block renders as "unsupported." 

PR submission checklist:

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -1,6 +1,7 @@
 Unreleased
 ---
 * [*] Search Control - Prevent calling TextInput's methods when undefined [https://github.com/wordpress-mobile/gutenberg-mobile/pull/6090]
+* [**] Add basic support to view, relocate, and remove the Jetpack Paywall block. [https://github.com/wordpress-mobile/gutenberg-mobile/pull/6076]
 
 1.102.0
 ---
--- a/package-lock.json
+++ b/package-lock.json
@@ -63,7 +63,7 @@
 		},
 		"gutenberg/packages/babel-plugin-import-jsx-pragma": {
 			"name": "@wordpress/babel-plugin-import-jsx-pragma",
-			"version": "4.22.0",
+			"version": "4.23.0",
 			"dev": true,
 			"license": "GPL-2.0-or-later",
 			"engines": {
@@ -75,7 +75,7 @@
 		},
 		"gutenberg/packages/babel-preset-default": {
 			"name": "@wordpress/babel-preset-default",
-			"version": "7.23.0",
+			"version": "7.24.0",
 			"dev": true,
 			"license": "GPL-2.0-or-later",
 			"dependencies": {
@@ -98,7 +98,7 @@
 		},
 		"gutenberg/packages/browserslist-config": {
 			"name": "@wordpress/browserslist-config",
-			"version": "5.22.0",
+			"version": "5.23.0",
 			"dev": true,
 			"license": "GPL-2.0-or-later",
 			"engines": {
@@ -107,7 +107,7 @@
 		},
 		"gutenberg/packages/element": {
 			"name": "@wordpress/element",
-			"version": "5.16.0",
+			"version": "5.17.0",
 			"dev": true,
 			"license": "GPL-2.0-or-later",
 			"dependencies": {
@@ -160,7 +160,7 @@
 		},
 		"gutenberg/packages/escape-html": {
 			"name": "@wordpress/escape-html",
-			"version": "2.39.0",
+			"version": "2.40.0",
 			"dev": true,
 			"license": "GPL-2.0-or-later",
 			"dependencies": {
@@ -172,7 +172,7 @@
 		},
 		"gutenberg/packages/eslint-plugin": {
 			"name": "@wordpress/eslint-plugin",
-			"version": "14.12.0",
+			"version": "15.0.0",
 			"dev": true,
 			"license": "GPL-2.0-or-later",
 			"dependencies": {
@@ -422,7 +422,7 @@
 		},
 		"gutenberg/packages/prettier-config": {
 			"name": "@wordpress/prettier-config",
-			"version": "2.22.0",
+			"version": "2.23.0",
 			"dev": true,
 			"license": "GPL-2.0-or-later",
 			"engines": {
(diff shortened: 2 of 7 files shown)


dcalhoun:
@derekblank while the Paywall block renders as expected when running the Metro server locally, it does not appear for either iOS or Android when running the prototype builds or a local debug build without the Metro server running. I am unsure why that is the case currently.

dcalhoun:
> @derekblank while the Paywall block renders as expected when running the Metro server locally, it does not appear for either iOS or Android when running the prototype builds or a local debug build without the Metro server running. I am unsure why that is the case currently.

I identified the root issue and addressed it in 5f7cb78dad419f893a1b1404a757c1e982cb2fb9. The Android prototype build worked as expected. The iOS prototype build is still building currently.

derekblank (approved):
I was able to follow the testing instructions successfully for both platforms, on both apps. LGTM! 🚀

lmischner:
I was able to take a look on both Android and iOS platforms (but not tablet).
I manipulated UI in posts with this block (different font sizes, bold/non-bold, landscape vs portrait), used VoiceOver+Talkback
I also tried interacting with the block itself, moving it various ways, e.g.

Questions/potential issues:
* Is it intentional that the Paywall Block icon during drag & drop is green rather than black/white like other blocks?
* Is the enforcement of one paywall block per post enforced with the new reduced scope? (I can get a second paywall block if I duplicate a grouped block that includes the paywall block)
* I can't ungroup blocks if one of them is the paywall block. (The option is available, but when I tap it the group remains a group). 
* Are we able to force wrapping earlier with large font sizes? With large font, the lines around "EXCLUSIVE CONTENT BELOW HERE" are practically dots


Let me know if you need device info or more explicit repro steps! I saw all of the above on both iOS and Android.

dcalhoun:
Thank you, @lmischner! This is great feedback. 

> Is it intentional that the Paywall Block icon during drag & drop is green rather than black/white like other blocks?

No, I do not believe so. The icons for other Jetpack blocks — e.g. Contact Info — are presented as black/white. I created https://github.com/wordpress-mobile/gutenberg-mobile/issues/6099 to address this. 

> Is the enforcement of one paywall block per post enforced with the new reduced scope? (I can get a second paywall block if I duplicate a grouped block that includes the paywall block)

Since we are not enabling the ability to _insert_ Paywall blocks, I considered the single block per post enforcement unrelated. However, duplicating via a Group is an interesting approach. Thanks for sharing that. The More block — which also should be limited to a single block — appears to [suffer from the same behavior](https://github.com/WordPress/gutenberg/issues/22718) on both web and mobile. It could be that recursive checks for the enforcement may be avoided for performance reasons. 

<details><summary>Side note: Web Duplicate Block Experience</summary>

Interestingly, Gutenberg core for web does include an experience that identifies erroneous, duplicative blocks whenever the block is ungrouped. The native editor does not have this. I consider adding this enhancement out of scope and more global. 

<video src="https://github.com/wordpress-mobile/gutenberg-mobile/assets/438664/a1c05c04-50eb-4d97-a8f0-dc80c80b69d5" />

</details>

From my testing, the presence of multiple Paywall blocks does not cause major issues, e.g. a crash. The presence might result in an unexpectedly early "cut off" point for publicly available content, i.e. if an author is unaware a superfluous Paywall block is inserted earlier in the content than the desired location of the second Paywall block. Recovery from this feels relatively manageable in the form of realizing the oddity, locating the duplicative Paywall block, and removing it. 

Given all this, I feel the risk of not addressing this unexpected before shipping seems low. 

> I can't ungroup blocks if one of them is the paywall block. (The option is available, but when I tap it the group remains a group).

This is unexpected. I created https://github.com/wordpress-mobile/gutenberg-mobile/issues/6098 to address it. 

> Are we able to force wrapping earlier with large font sizes? With large font, the lines around "EXCLUSIVE CONTENT BELOW HERE" are practically dots

I agree the bookending lines could benefit from adjustments. It should be easy. From playing with it for a few minutes, it is (frustratingly) proving to _not_ be easy. I'll see what I can do. I will note this effects the existing More block as well.

dcalhoun:
@lmischner after researching the inability to ungroup the Paywall block, I believe we should consider this as "won't fix." I described more details as to why in https://github.com/wordpress-mobile/gutenberg-mobile/issues/6098#issuecomment-1690233530. 

With that, I believe this work is ready to merge. We addressed the horizontal line width in https://github.com/WordPress/gutenberg/pull/53883 and the icon color (https://github.com/wordpress-mobile/gutenberg-mobile/issues/6099).

lmischner:
Thanks, @dcalhoun, for the quick turnaround ! I looked into your investigation on the ungrouping. While I don't love it from a user experience perspective (if we shouldn't be able to ungroup, we shouldn't have the option in the menu / banner saying the blocks have been ungrouped),  it looks like that behavior is not new with this project, so out of scope. All to say, I agree with you that "won't fix" is appropriate here.

Thanks for the two fixes on the others! I'm comfortable with merging this based on your investigation and fixes 🎉

PR #6086, opened 2023-08-18
Read Podfile dependencies from third-party-podspec

mokagio:
The previous approach had them hardcoded. This is better because we now have a single source of truth.

Hat tip @crazytonyli, see https://github.com/wordpress-mobile/gutenberg-mobile/pull/5739#discussion_r1193227183

Notice that the only `Podfile.lock` change is in the checksum, meaning that the new logic results in the same resolved dependencies as before.

**To test**: Ensure CI is green.

**PR submission checklist**:


ios-xcframework/Podfile
```diff
@@ -9,57 +9,40 @@ HERMES_ENABLED = ENV.fetch('HERMES_ENABLED', true)
...
-
-  computed_dependencies.each do |pod_name|
-    pod pod_name, podspec: "#{podspec_prefix}/third-party-podspecs/#{pod_name}.podspec.json"
-  end
-
+def setup_rnreanimated_pod_from_fork
   # This is required to workaround an issue with RNReanimated after upgrading to version 2.17.0
   rn_node_modules = File.join(Dir.pwd, '..', 'gutenberg', 'node_modules')
   raise "Could not find node modules at given path #{rn_node_modules}" unless File.exist? rn_node_modules
+
```
mokagio, line 18: RuboCop suggests leaving an empty line after `raise` for readability 😉

ios-xcframework/Podfile.lock
```diff
@@ -674,6 +674,6 @@ SPEC CHECKSUMS:
   WordPress-Aztec-iOS: 7d11d598f14c82c727c08b56bd35fbeb7dafb504
   Yoga: f7decafdc5e8c125e6fa0da38a687e35238420fa
 
-PODFILE CHECKSUM: 1caec69a3de9c6e37113c8eb5dbdb2d2b52c5108
+PODFILE CHECKSUM: 8f2b84cf5105963b4ef56ad18d6d64ab24334789
```
mokagio: Notice the only change is in the checksum. This mean that the resolved dependencies with this new approach are exactly the same as before 🎉 

The different checksum is due to the text content of `Podfile` having changed.

ios-xcframework/Podfile
```diff
@@ -9,57 +9,40 @@ HERMES_ENABLED = ENV.fetch('HERMES_ENABLED', true)
...
 
   # Use a custom RNReanimated version while we coordinate a fix upstream
   pod 'RNReanimated', git: 'https://github.com/wordpress-mobile/react-native-reanimated', branch: 'wp-fork-2.17.0'
 end
 
+def gutenberg_dependencies
+  podspec_extension = '.podspec.json'
+
+  Dir["../third-party-podspecs/*#{podspec_extension}"].each do |podspec_path|
+    pod_name = File.basename(podspec_path, podspec_extension)
```
crazytonyli: It might be more robust to use the podspec json's `name` property as the `pod_name`, instead of the filename? We probably won't use a different name for the podspec.json file intentionally, but there might be a typo, omitting a dash (i.e. RCTFolly.podspec.json for the RCT-Folly pod), or something else, which would cause a mismatch between the filename and the pod name.

crazytonyli (approved):

ios-xcframework/Podfile
mokagio: Done via f865eb5

mokagio:
@dcalhoun @geriux @SiobhyB @fluiddot 

Is there anything I'm missing with this implementation that uses the contents of `third-party-specs` as the source of truth for which dependencies to include?

Thanks!

geriux:
> Is there anything I'm missing with this implementation that uses the contents of `third-party-specs` as the source of truth for which dependencies to include?

Nope! We are good at using `third-party-specs` as the source of truth 🚀

geriux:
I see React-related podspecs deleted, is this because of the XCFramework that they are not needed? We might need to update the [bin/generate-podspecs.sh](https://github.com/wordpress-mobile/gutenberg-mobile/blob/trunk/bin/generate-podspecs.sh) script to avoid adding them back whenever we update a dependency 🤔

mokagio:
@geriux thank you for mentioning that. I run `bin/generate-podspec.sh` and it did in fact recreate the specs I removed. Clearly there's more work to do to centralize the source of truth.

geriux:
> @geriux thank you for mentioning that. I run `bin/generate-podspec.sh` and it did in fact recreate the specs I removed. Clearly there's more work to do to centralize the source of truth.

No problem! I think if we just remove the `# Generate the React Native podspecs` part within the `bin/generate-podspecs.sh` script we'd be set.

fluiddot:
> Is there anything I'm missing with this implementation that uses the contents of `third-party-specs` as the source of truth for which dependencies to include?

The approach looks good to me Gio, thanks for working on this 🙇 ! The only caveat, as @geriux shared in https://github.com/wordpress-mobile/gutenberg-mobile/pull/6086#issuecomment-1692182236, we'd need to prevent re-generating the React-related pods in the `generate-podspecs.sh` script.

mokagio:
@geriux @fluiddot I updated the script to generates the `third-party-podspecs` to only generated the ones we currently need. The approach could do with some more work and optimization, but in my tests locally it did not generate additional or removed existing `.podspecs`.

If one of you can confirm, I'll proceed with merging.

mokagio:
image

I'll take those emoji reactions as confirmation. I just merged `trunk` into this branch to 1) update CI to match the new requirements (Lint and Unit Tests from Buildkite) and 2) verify locally that the script didn't change the `third-party-podspecs` content.

Enabling auto-merge

ios-xcframework/Podfile
```diff
@@ -9,57 +10,42 @@ HERMES_ENABLED = ENV.fetch('HERMES_ENABLED', true)
...
+    # while working on an upstream fix
+    #
+    # https://github.com/software-mansion/react-native-reanimated/pull/4684
+    if pod_name == 'RNReanimated'
+      setup_rnreanimated_pod_from_fork
+      next
+    end
+
+    # Skip React-jsc unless using Hermes
+    next if pod_name == 'React-jsc' && HERMES_ENABLED == false
```
fluiddot: When Hermes is enabled we'd only need to include `React-hermes`, so in theory, we could filter out `React-jsc`. I know the condition was set this way but I'm wondering if it was originally wrong 🤔.

fluiddot: I've generated the XCFramework without the `React-jsc` pod and the build succeeded locally. I've also tested using a locally generated XCFramework in the WordPress-iOS project and the app runs successfully too. That said, I think we could safely remove it when Hermes is enabled.

```suggestion
    next if pod_name == 'React-jsc' && HERMES_ENABLED == true
```

bin/generate-podspecs.sh
```diff
@@ -128,82 +137,11 @@ do
...
-        echo "
-        def min_ios_version_supported
-        '12.4'
-        end
-        " >> "$REACT_NATIVE_CODEGEN_UTILS_MODIFIED_PATH"
-        # Modify the get_react_codegen_spec method to return the result using print and JSON.pretty
-        sed -i '' -e "s/:git => ''/:git => 'https:\/\/github.com\/facebook\/react-native.git', :tag => 'v$RN_VERSION'/" "$REACT_NATIVE_CODEGEN_UTILS_MODIFIED_PATH"
-        sed -i '' -e 's/return spec/print JSON.pretty_generate(spec)/' "$REACT_NATIVE_CODEGEN_UTILS_MODIFIED_PATH"
-        # Run get_react_codegen_spec and generate React-Codegen.podspec.json
-        ruby -r "./scripts/cocoapods/codegen_utils_modified.rb" -e "CodegenUtils.new.get_react_codegen_spec('$PACKAGE_JSON_PATH', hermes_enabled:$HERMES_ENABLED)" > "$DEST/React-Codegen.podspec.json"
```
fluiddot, line 184: Nitpick: After removing this line, we can also remove the `HERMES_ENABLED` variable definition in line 110.

bin/generate-podspecs.sh
```diff
@@ -115,7 +117,14 @@ do
     pod=$(basename "$podspec" .podspec)
     path=$(dirname "$podspec")
 
+    # The only one we need is React-jsc, all the others can be skipped.
+    if [[ $pod != 'React-jsc' ]]; then
+      echo "Skipping copying React Native $pod podspec..."
+      continue
+    fi
```
fluiddot, line 123: I was hesitant about whether to skip the `React-jsc` pod if Hermes is enabled, however, I think this decision could be made by the XCFramework project. Hence, this script will generate all podspecs independently of the engine used (i.e. JSC vs Hermes).

fluiddot (commented):
Thanks @mokagio for the improvements 🙇 !

I've only added a minor comment, that shouldn't block the PR but would be interesting to include to avoid adding an unnecessary pod.

Probably something for another PR, but I noticed we still have the logic that prompts us to run the `generate-podspecs. ` twice:

https://github.com/wordpress-mobile/gutenberg-mobile/blob/1261f7b51ec1b8b2190bb14bc14ef782359c327e/bin/generate-podspecs.sh#L39-L43

https://github.com/wordpress-mobile/gutenberg-mobile/blob/1261f7b51ec1b8b2190bb14bc14ef782359c327e/bin/generate-podspecs.sh#L148-L155

I understand that this is no longer needed as the `FBReactNativeSpec` pod is now generated in the XCFramework project, is this accurate?

bin/generate-podspecs.sh
mokagio: +1.

I think it would be handy to skip generating the jsc podspec if we were running the `generate-podspecs.sh` and `bundle exec pod install` as part of the same operation, so that both would read the same "Hermes enabled" config. But given that's not the case, for the moment at least, then "this script will generate all podspecs independently of the engine used (i.e. JSC vs Hermes)" is the way to go.

bin/generate-podspecs.sh
mokagio: Good catch. Thanks!

Address in 5039d5c9935c8c625fbd030d2281822e28cbd81e

ios-xcframework/Podfile
mokagio: Good point @fluiddot. 

I must have gotten mixed up when changing the syntax and didn't make the JSC = JavaScriptCore connection.

I guess we never noticed because we've been only testing with Hermes enabled for a while. 

I decided not to use your suggestion because I wanted to add more details, see ad3c8f9098b90d8ff4677f33efb3fd1c6abcfbb7

mokagio:
> I understand that this [running `pod install` twice] is no longer needed as the FBReactNativeSpec pod is now generated in the XCFramework project, is this accurate?

I'll need to double check this and follow up if necessary.

mokagio:
@fluiddot @geriux I addressed your new suggestions, thanks again. I'd appreciate a final look over, just in case.

mokagio:
image

Thanks for the `trunk` mereg @fluiddot. CI is green. I'll take your reaction as a 👍 to merge this.

fluiddot:
> image
> Thanks for the `trunk` mereg @fluiddot. CI is green. I'll take your reaction as a 👍 to merge this.

@mokagio To be honest, I couldn't fully test it last week when I shared those reactions. From the code perspective, the PR looks good to me but I was planning to generate an installable build as a double-check. Since CI is green let's keep this as-is and we'll test next week an installable build when we have more availability. Thanks 🙇 !

fluiddot:
> > image
> > Thanks for the `trunk` mereg @fluiddot. CI is green. I'll take your reaction as a 👍 to merge this.
> 
> @mokagio To be honest, I couldn't fully test it last week when I shared those reactions. From the code perspective, the PR looks good to me but I was planning to generate an installable build as a double-check. Since CI is green let's keep this as-is 
and we'll test next week an installable build when we have more availability. Thanks 🙇 !

I tested using the XCFramework and a local Gutenberg installation and haven't found any build and runtime errors 🎊.

PR #6101, opened 2023-08-23
Use the same version as `package.json` in XCFramework

mokagio:
See @crazytonyli suggestion here: https://github.com/wordpress-mobile/gutenberg-mobile/pull/5769#discussion_r1222309995

**To test**: Check the value in the generated `Info.plist`, either by running `npm run xcframework:build` locally or by downloading and inspecting the CI artifact:

Screenshot 2023-08-23 at 8 49 33 pm

---

PR submission checklist:


crazytonyli:
Thanks for taking my suggestion! What do you think about passing the version configurations to [the xcodebuild command here](https://github.com/wordpress-mobile/gutenberg-mobile/blob/da001ebc1e641992ca79ff6201f2fb4fc11717d7/ios-xcframework/build.sh#L56-L56)?

dcalhoun (commented):
The test plan succeeded for me. 🎉 Thanks!

mokagio:
> What do you think about passing the version configurations to [the xcodebuild command here](https://github.com/wordpress-mobile/gutenberg-mobile/blob/da001ebc1e641992ca79ff6201f2fb4fc11717d7/ios-xcframework/build.sh#L56-L56)?

Thanks @crazytonyli !

I think that would work and it looks better that using `PlistBuddy` after the file has been created. At the same time, having the setting in the build script alone means we _might_ lose it in case we change the automation (eg by moving to Fastlane). If it's as a build phase in the target, it will always run as long as we build the target.

Am I missing something?

crazytonyli:
The current solution looks good to me. But I'm not sure what responsibilities should the "scaffold" project holds. We now have build related tasks in both the `build.sh` shell script and the Xcode project's build phases. I just thought maybe we should consolidate the build related tasks in one place. The `build.sh` seems to be a good place for them. Also, we get to keep the scaffold Xcode project as lean as possible, which could be beneficial too.

BTW, I believe you can also pass build settings to fastlane actions too. Fastlane is usually backwards compatible with `xcodebuild`.

mokagio:
> Also, we get to keep the scaffold Xcode project as lean as possible, which could be beneficial too.

Okay. I can stand behind that 👍 

> BTW, I believe you can also pass build settings to fastlane actions too. Fastlane is usually backwards compatible with xcodebuild.

Sure. I'm pretty sure that would be possible out of the box. My point was that with the settings in the xcodeproj they would automatically apply to a build call made from Fastlane.

mokagio:
@crazytonyli addressed your suggestion in https://github.com/wordpress-mobile/gutenberg-mobile/pull/6101/commits/e277e3858eba84175f6f2dc06d3fcde01b1c4d1d

Thanks! 🙇‍♂️

ios-xcframework/BuildPhases/set_current_project_version.sh
```diff
@@ -0,0 +1,13 @@
...
+dsym_plist="$DWARF_DSYM_FOLDER_PATH/$DWARF_DSYM_FILE_NAME/Contents/Info.plist"
+
+version=$("$(dirname "$0")"/get_gutenberg_version)
+
+for plist in "$target_plist" "$dsym_plist"; do
+  if [ -f "$plist" ]; then
+    /usr/libexec/PlistBuddy -c "Set :CFBundleVersion $version" "$plist"
+    /usr/libexec/PlistBuddy -c "Set :CFBundleShortVersionString $version" "$plist"
+  fi
+done
```
crazytonyli: This script is not used anymore, I think?

ios-xcframework/build.sh
```diff
@@ -63,6 +66,9 @@ if [ ! -d $HERMES_XCFRAMEWORK ]; then
   exit 1
 fi
 
+# Fail early if cannot fetch version
+VERSION=$(./bin/get_gutenberg_version)
```
crazytonyli: I think we can use `jq` to replace this custom JSON parsing script: `jq -e -r '.version' ./package.json`?

ios-xcframework/bin/get_gutenberg_version
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env ruby
+
+# frozen_string_literal: true
+
+require 'json'
+
+# get the path to the parent folder using Dir methods
+package_name = 'package.json'
+path_to_package_json = File.join(File.dirname(File.dirname(File.dirname(File.absolute_path(__FILE__)))), package_name)
```
crazytonyli: What do think using [`File.expand_path`](https://www.rubydoc.info/stdlib/core/File.expand_path), instead of multiple `dirname` calls?

crazytonyli (changes requested):
Thanks for taking my suggestions! I have left a couple of comments which may help further simplify the solution.

ios-xcframework/BuildPhases/set_current_project_version.sh
mokagio: Thanks! I had forgotten about that one.

mokagio: Done via 6d5a033d89e51b5c84b29816743b74258aa92fbd

ios-xcframework/build.sh
mokagio: I considered it, but decided against it to avoid having to ensure the tool is available locally. 

Having said that... I guess it's safe to assume that to be the case and it doesn't take much to fail the script early if that's not the case 🤔

ios-xcframework/bin/get_gutenberg_version
mokagio: Can you elaborate on this suggestion?

The code uses `File.dirname` to have a platform-independent `..`. That is go up three parent folders.

There's no need for this to be platform-independent to be fair, but I find it a good habit to get into 🤷‍♂️

ios-xcframework/build.sh
mokagio: Addressed in 8779e2e88cf17fef54b45d8a14ecbe95b1f3a167

ios-xcframework/bin/get_gutenberg_version
mokagio: BTW, addressing this is no longer necessary since the automation uses `jq` as of 8779e2e88cf17fef54b45d8a14ecbe95b1f3a167

crazytonyli (approved):
:shipit:

PR #6134, opened 2023-08-30
Duplicated "iOS Build" step from CircleCI to Buildkite

mokagio:
_WIP_

---

To test:

PR submission checklist:


mokagio:
Getting some

```
HTTP Status: 500
{
	"message": "An unknown server-side error occurred while processing the command.
       Original error: Could not extract Info.plist from 'GutenbergDemo.app':
          Plist file doesn't exist:
            '/var/folders/3n/z0_2p7gn03q5x5mt40c4lj1m0000kr/T/202386-3093-1uz43f4.nw69/GutenbergDemo.app/Info.plist'"
}
```

mokagio:
Got to the bottom of it, I think. 

If you look at the [Buildkite logs](https://buildkite.com/automattic/gutenberg-mobile/builds/7164#018a695f-f47a-4488-9baf-6e554109722c/890) when zipping the `GutenbergDemo.app` package, there is indeed no `Info.plist`. Compare them to the CircleCI ones, and you'll see that [they do have the file](https://app.circleci.com/pipelines/github/wordpress-mobile/gutenberg-mobile/21744/workflows/1c3349a0-e20c-404a-8bc4-37b24107eabc/jobs/127562/steps).

I believe this is happening because the previous step, `npm run core test:e2e:build-app:ios` failed silently. See [logs here](https://buildkite.com/automattic/gutenberg-mobile/builds/7164#018a695f-f47a-4488-9baf-6e554109722c/837-852) because of the provisioning profile issue already noticed in https://github.com/WordPress/gutenberg/pull/54079

mokagio:
Okay, so the reason was a dumb forgetting that piping output will lose exit code [unless explicitly specified](https://github.com/WordPress/gutenberg/compare/f3a2c1c6e727b4ac94e5c123748ba27acbd904e8...0af56e560eac189d4a23a19e854db406bfe8107e#diff-189a63d46ad00127c49d550987f8ed257a6fd739ef1642c307985e203315a990).

That is why the build step didn't fail despite the script failing, therefore resulting in the tests running but without access to a fully built `.app`.

As for the reason why the script, that is the `xcodebuild` call, failed, by looking at the Buildkite logs once more we can see that there is no iPhone 13 Simulator available so we go back to the code signing process looking using the Mac machine as the target, which fails. I'll fix that next, once established that the `set -o pipefail` does its job.

mokagio:
Happy to say that the test annotation and Slack reporting worked fine :) 

Screenshot 2023-09-07 at 1 05 52 pm

Screenshot 2023-09-07 at 1 06 01 pm

mokagio:
Closing in favor of https://github.com/wordpress-mobile/gutenberg-mobile/pull/6179, which took these changes minus all the Git noise.

PR #6135, opened 2023-08-30
Update Buttons block dark mode screenshot to reduce failures in E2E tests

fluiddot:
This PR aims to reduce the failures we have had recently in Android E2E tests, specifically for the test case: "Buttons block - should render dark mode colors" ([example](https://app.circleci.com/pipelines/github/wordpress-mobile/gutenberg-mobile/21667/workflows/c115fdc9-5ec0-4f42-923d-a67cc7c03893/jobs/126846/tests#failed-test-0)). An extra time has been added on Android to ensure that the screenshot is taken after the bottom sheet's scrollbars fade out after opening.

**To test:**
* Observe that all CI jobs pass.

PR submission checklist:


__device-tests__/gutenberg-editor-sanity-test-4-visual.test.js
```diff
@@ -216,7 +216,9 @@ describe( 'Gutenberg Editor - Test Suite 4', () => {
 
 			// Open and wait for block settings
 			await editorPage.openBlockSettings();
-			await editorPage.driver.sleep( 500 );
+			// On Android, wait extra time before taking the screenshot
+			// to let the bottom sheet's scrollbar fade out.
+			await editorPage.driver.sleep( isAndroid() ? 2000 : 500 );
```
dcalhoun: Thinking out loud — an alternative is to focus the screenshot on a specific element or crop the screenshot by a few pixels. Genuinely uncertain if that approach is appropriate for the spirit of this test, merely offering it for discussion. WDYT?

dcalhoun (approved):
Thank you!

__device-tests__/gutenberg-editor-sanity-test-4-visual.test.js
fluiddot: Not sure if we can focus on a specific element as the test case aims to check how the block and settings look in dark mode. However, I agree that alternatively to increase the waiting time we could crop the screenshot by a few pixels from the right to exclude the scrollbar. I don't think we'll miss any element by cropping it a little.

fluiddot: I'll update the changes with the above proposal.

fluiddot: @dcalhoun We don't have an easy way to crop the screenshot from the right and bottom edges, so I went ahead and added a new parameter to `takeScreenshot` util (https://github.com/wordpress-mobile/gutenberg-mobile/pull/6135/commits/193d6836afcaa2403e361ba9bf784aca56d87f3e).

Additionally, the screenshot and test case have been updated in https://github.com/wordpress-mobile/gutenberg-mobile/pull/6135/commits/961ea91a390194abfa59a749f450571dad233e27 and https://github.com/wordpress-mobile/gutenberg-mobile/pull/6135/commits/eb584b3b16ff1b8307165ae67611142e649922f8.

Let me know if you could take another look at the PR changes, thanks 🙇 !

__device-tests__/utils.js
```diff
@@ -51,17 +51,29 @@ export async function takeScreenshotByElement( element, { padding } = {} ) {
...
  *
- * @param    {Object}         options                   Options
- * @param    {boolean}        [options.withoutKeyboard] Prevents showing the keyboard in the screenshot.
- * @param    {CropScreenshot} [options.crop]            Specify values to crop the screenshot.
+ * @typedef {Object} CropOffsetScreenshot
+ *
+ * @property {number}               left                      Left offset to apply to the crop values, in pixels.
+ * @property {number}               top                       Top offset to apply to the crop values, in pixels.
+ * @property {number}               right                     Right offset to apply to the crop values, in pixels.
+ * @property {number}               bottom                    Bottom offset to apply to the crop values, in pixels.
```
dcalhoun: Maybe a crazy idea: is there a parameter design that allows us to avoid adding a new `cropOffset` parameter?

I'm wonder how we might use negative numbers to achieve the outcome we desire without expanding the API surface area. 

```js
takeScreenshot({  right: -20  });
```

The issue with the current design — which is not directly used anywhere outside of `takeScreenshotByElement` — is that it does not have a `right` value, but uses a combination of `x` and `width`. Would it be worthwhile to redesign these arguments to enable negative values? Would we be able to retain the simplicity in use of, say, `takeScreenshot({  withoutKeyboard: true  })`? 

Off the top of my head, I am picturing something like the following. 

```js
/**
 * @typedef {Object} CropScreenshot
 *
 * @property {number} top    Top offset, in pixels.
 * @property {number} right  Right offset, in pixels.
 * @property {number} bottom Bottom offset, in pixels.
 * @property {number} left   Left offset, in pixels.
 * @property {number} width  Width, in pixels.
 * @property {number} height Height, in pixels.
 */
```

Is this worthwhile or a waste of time?

geriux:
Hey @fluiddot 👋 . We will cut the `1.103.0` release today. I'm bumping this PR to the next milestone, but please let me know if you’d rather us work to include this PR in `1.103.0`. Thanks!

fluiddot:
> Hey @fluiddot 👋 . We will cut the `1.103.0` release today. I'm bumping this PR to the next milestone, but please let me know if you’d rather us work to include this PR in `1.103.0`. Thanks!

Yeah, no problem. This PR only affects E2E tests so no need to include it in `1.103.0`. Thanks for the heads up 🙇 !

__device-tests__/utils.js
fluiddot: I thought about adding the `right` and `bottom` offset parameters to the `crop` object, but I felt that it could be confusing to have two different coordinate formats in the same object (i.e. `x, y` format vs `top, left` format). My rationale for creating a second parameter was mainly driven by that fact.

> I'm wonder how we might use negative numbers to achieve the outcome we desire without expanding the API surface area.
> 
> ```js
> takeScreenshot({  right: -20  });
> ```

Ah, interesting. After checking your comment, I realized that using the word `right` here might be confusing. The way I see the offset values to transform the surface area is more like the [`padding` CSS property](https://developer.mozilla.org/en-US/docs/Web/CSS/padding). So if you set a right value, it would push the content to the left. This makes me think that probably the word `offset` can be misleading for this parameter. We could change it to `padding` to better match what we try to achieve by cutting some pixels from the borders. WDYT?

> Off the top of my head, I am picturing something like the following.
> 
> ```js
> /**
>  * @typedef {Object} CropScreenshot
>  *
>  * @property {number} top    Top offset, in pixels.
>  * @property {number} right  Right offset, in pixels.
>  * @property {number} bottom Bottom offset, in pixels.
>  * @property {number} left   Left offset, in pixels.
>  * @property {number} width  Width, in pixels.
>  * @property {number} height Height, in pixels.
>  */
> ```
> 
> Is this worthwhile or a waste of time?

Yeah, this approach is interesting. My only concern is how the width and height affect the other parameters. Could you share some examples? It would help me to better understand how each parameter interacts with the cropping surface area. Thanks!

Alternatively, and following my comment regarding the padding, we could have the following parameters:

```js
/**
 * @typedef {Object} CropSurface
 *
 * @property {number}         x                         X offset, in pixels.
 * @property {number}         y                         Y offset, in pixels.
 * @property {number}         width                     Width, in pixels.
 * @property {number}         height                    Height, in pixels.
 *
 * @typedef {Object} CropPadding
 *
 * @property {number}         left                      Padding to apply to the left side, in pixels.
 * @property {number}         top                       Padding to apply to the top side, in pixels.
 * @property {number}         right                     Padding to apply to the right side, in pixels.
 * @property {number}         bottom                    Padding to apply to the bottom side, in pixels.
 *
 * @typedef {Object} CropScreenshot
 *
 * @property {CropSurface}    surface                   Surface dimensions to crop the screenshot. The default values will try to fit the screenshot size.
 * @property {CropPadding}    padding                   Padding to apply to the cropping surface. By default, no padding is applied.
 */
```

WDYT?

dcalhoun: As I reflect on my thoughts further, I believe they may miss my original goal of simplifying the utility. 

I do not believe using the name "padding" would clarify things either, as (1) `padding` is already used in `takeScreenshotByElement` for the inverse purpose — adding space — and (2) negative padding isn't a concept in CSS. 

My recommendation would be to leave the implementation as-is or remove it and use a hard-coded `width` value instead, e.g. `takeScreenshot({ width: WIDTH_SANS_SCROLLBAR })`. The latter could make sense as the device properties are static and this is (currently) a one-off usage. 

WDYT?

fluiddot: > My recommendation would be to leave the implementation as-is or remove it and use a hard-coded `width` value instead, e.g. `takeScreenshot({ width: WIDTH_SANS_SCROLLBAR })`. The latter could make sense as the device properties are static and this is (currently) a one-off usage.

I'm hesitant to use a hard-coded value for the sake of keeping the tests as device-agnostic as possible. I agree we currently use a single device, but if in the future we use a different one or even multiple, this approach would need to be revisited.

One option I'm thinking of is to use `takeScreenshotByElement` and target the block list element. This way we'll capture the entire content area and using negative padding should crop the screenshots as we need. WDYT?

dcalhoun: Fair enough regarding device-agnostic tests. 

Using the approach `takeScreenshotByElement` and negative padding sounds like it could work well. Do you think we'd overload the `padding` argument to accept both a number (`-20`) and object (`{ right: -20 }`) or require an object with top/right/bottom/left? I think I lean towards overloading so that we retain the easy of use for universal padding with a number.

fluiddot: > Using the approach `takeScreenshotByElement` and negative padding sounds like it could work well. Do you think we'd overload the `padding` argument to accept both a number (`-20`) and object (`{ right: -20 }`) or require an object with top/right/bottom/left? I think I lean towards overloading so that we retain the easy of use for universal padding with a number.

Yep, I agree with overloading the argument and accepting both number and object 👍.

__device-tests__/gutenberg-editor-sanity-test-4-visual.test.js
```diff
@@ -219,7 +219,19 @@ describe( 'Gutenberg Editor - Test Suite 4', () => {
...
-			const screenshot = await takeScreenshot();
+			let screenshot;
+			if ( isAndroid() ) {
+				// On Android, we take the screenshot using the block list element
+				// in order to crop some pixels using the padding argument.
+				const editorScreen = await editorPage.waitForElementToBeDisplayedByXPath(
+					'//android.widget.FrameLayout[@resource-id="android:id/content"]'
+				);
+				screenshot = await takeScreenshotByElement( editorScreen, {
+					padding: { right: -10 },
```
fluiddot, line 230: Negative values in padding can be used to shrink the cropping surface area.

__device-tests__/gutenberg-editor-sanity-test-4-visual.test.js
```diff
@@ -219,7 +219,19 @@ describe( 'Gutenberg Editor - Test Suite 4', () => {
...
 
 			// Visual test check
-			const screenshot = await takeScreenshot();
+			let screenshot;
+			if ( isAndroid() ) {
+				// On Android, we take the screenshot using the block list element
+				// in order to crop some pixels using the padding argument.
+				const editorScreen = await editorPage.waitForElementToBeDisplayedByXPath(
+					'//android.widget.FrameLayout[@resource-id="android:id/content"]'
+				);
```
fluiddot, line 228: This element covers the editor's content area. It might be interesting to use it as the default element when taking a screenshot, especially as it excludes device UI like the status bar. Hence, we won't need to manually remove it as we do [here](https://github.com/wordpress-mobile/gutenberg-mobile/blob/9a12cfac8e7f25ffab99ef0d8c4f2d787160fbe2/__device-tests__/utils.js#L121-L126).

fluiddot (commented):
@dcalhoun I've updated the PR based on https://github.com/wordpress-mobile/gutenberg-mobile/pull/6135#pullrequestreview-1603143099. Let me know if you could take a look, thanks 🙇 !

__device-tests__/utils.js
fluiddot: > Yep, I agree with overloading the argument and accepting both number and object 👍.

Done https://github.com/wordpress-mobile/gutenberg-mobile/pull/6135/commits/9bd3294236378186ac4e428361da874cfaacd514 and used https://github.com/wordpress-mobile/gutenberg-mobile/pull/6135/commits/9a12cfac8e7f25ffab99ef0d8c4f2d787160fbe2.

__device-tests__/gutenberg-editor-sanity-test-4-visual.test.js
dcalhoun: > This element covers the editor's content area. It might be interesting to use it as the default element when taking a screenshot, especially as it excludes device UI like the status bar. Hence, we won't need to manually remove it as we do [here](https://github.com/wordpress-mobile/gutenberg-mobile/blob/9a12cfac8e7f25ffab99ef0d8c4f2d787160fbe2/__device-tests__/utils.js#L121-L126).

This seems like a sensible default. 💯

dcalhoun (approved):
I believe this a very solid, robust implementation. Thank you for your patience as we navigated finding this design together. 🚀 👏🏻

__device-tests__/gutenberg-editor-sanity-test-4-visual.test.js
fluiddot: I'll work on incorporating this in a separate PR, as well as explore if the same can be achieved on iOS 👍.

PR #6153, opened 2023-09-04
Switch to Buildkite for lint and unit tests checks

mokagio:
What it says in the title.

To test, notice how:

**1. The checks have gone from CircleCI:**

image

I unblocked the optional UI tests steps to make sure the logic to do so still worked.

**2. There is a new "Lint and Unit Tests" check in the GitHub checks**

image

**After this is approved and before merging it** I'll update the GitHub configuration to replace the CircleCI checks with the new unified Buildkite checks

image

Notice that in the close future we'll split the Buildkite checks in three, which will require a further checks-settings switcheroo. That's an accepted tradeoff because we value moving away from CircleCI ASAP.

---

PR submission checklist:

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -46,6 +46,9 @@ steps:
     artifact_paths:
       - ./logs/*.log
       - ./reports/test-results/*.xml
+    notify:
+      - github_commit_status:
+          context: Lint and Unit Tests
 
   - label: "Build JS Bundles"
     depends_on: "unit-tests"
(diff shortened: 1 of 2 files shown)


mokagio:
I will admin merge this because:

1. The changes are self-evident, no-logic ones. They are merely removing steps from CircleCI and setting a label in Buildkite. A label name that will change in the near future, too.
2. This will allow me to make the GitHub config changes in a period with still relatively few people online, therefore minimizing dev-workflow disruption.

PR #6180, opened 2023-09-11
Update .ruby-version to 3.2.2

SiobhyB:
Ruby was updated from `2.7.4` to `3.2.2` in the iOS codebase in https://github.com/wordpress-mobile/WordPress-iOS/pull/21498. This PR brings Gutenberg Mobile in alignment with that change.

Related:

* https://github.com/WordPress/gutenberg/pull/54897

To test:

* Verify CI tests pass.

<hr />

PR submission checklist:


derekblank:
@mokagio Also updating the Ruby version here. I noted that the CI was failing with:

```
An error occurred while installing bigdecimal (1.4.4), and Bundler cannot continue.

In Gemfile:
  fastlane-plugin-wpmreleasetoolkit was resolved to 8.1.0, which depends on
    bigdecimal
```

I then noted that the [release-toolkit CHANGELOG](https://github.com/wordpress-mobile/release-toolkit/blob/99a42e2f5256b71f796241805ac53bfec399d044/CHANGELOG.md?plain=1#L59-L60) drops the dependency on bigdecimal and supports Ruby 3 in version 9.0+, so I updated `fastlane-plugin-wpmreleasetoolkit` to 9.1.0 here. Let me know if there is a reason not to do so, or if there is anything else we should update within the Gutenberg XCFramework here.

geriux:
> @mokagio Also updating the Ruby version here. I noted that the CI was failing with:
> 
> ```
> An error occurred while installing bigdecimal (1.4.4), and Bundler cannot continue.
> 
> In Gemfile:
>   fastlane-plugin-wpmreleasetoolkit was resolved to 8.1.0, which depends on
>     bigdecimal
> ```
> 
> I then noted that the [release-toolkit CHANGELOG](https://github.com/wordpress-mobile/release-toolkit/blob/99a42e2f5256b71f796241805ac53bfec399d044/CHANGELOG.md?plain=1#L59-L60) drops the dependency on bigdecimal and supports Ruby 3 in version 9.0+, so I updated `fastlane-plugin-wpmreleasetoolkit` to 9.1.0 here. Let me know if there is a reason not to do so, or if there is anything else we should update within the Gutenberg XCFramework here.

It looks like [WordPress iOS uses](https://github.com/wordpress-mobile/WordPress-iOS/blob/bd143e2aa4664db419779de92162d6b8514c4028/Gemfile#L20) version `9.0` not sure if they need to be consistent or not 🤔

ios-xcframework/Podfile.lock
```diff
@@ -676,4 +676,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: 1caec69a3de9c6e37113c8eb5dbdb2d2b52c5108
 
-COCOAPODS: 1.12.1
+COCOAPODS: 1.13.0
```
geriux, line 670: Was this change needed? I see we use `1.12.1` in [WordPress iOS](https://github.com/wordpress-mobile/WordPress-iOS/blob/bd143e2aa4664db419779de92162d6b8514c4028/Podfile.lock#L246C12-L246C18)

derekblank: This change was the output of running `bundle exec pod install` (which was a command prompted by CI failures). 1.13.0 refers to the latest [Cocoapods release](https://github.com/CocoaPods/CocoaPods/releases), and 1.12.1 was the previous version. I'm not certain if they need to be consistent with WordPress iOS either.

@mokagio Any issues with using fastlane-plugin-wpmreleasetoolkit 9.1.0 or Cocoapods 1.13.0 here? Or should we use 9.0 and 1.12.1?

mokagio: It's okay to use 1.13.0 here. We don't need to keep the versions in sync between projects because:

1. Fastlane is used for project automation, there is no information exchanged between the two setup.
2. CocoaPods is used to put together dependencies and generate a binary framework. It's highly unlikely that a difference in the library version will result in issues.

Anyways, we'll soon update CocoaPods to version 1.13.0 in WordPress iOS, too, given it has [a fix for Xcode 15](https://github.com/CocoaPods/CocoaPods/releases/tag/1.13.0).

mokagio: Also wanted to say that updating the release-toolkit to 9.x was the "correct" thing to do in this case because that's the version officially compatible with Ruby 3 👍 Thanks @derekblank

mokagio (approved):

mokagio:
Screenshot 2023-10-04 at 12 28 19 pm

@SiobhyB the branch was a bit out of date which is why it was blocked by a CI task not having reported. I merged the latest `trunk` which should sort that out.

geriux:
Hey @derekblank 👋 I think it'd be a good idea to bring the changes merged in https://github.com/WordPress/gutenberg/pull/54897 to Gutenberg mobile so both repos are synced.

derekblank:
> Hey @derekblank 👋 I think it'd be a good idea to bring the changes merged in [WordPress/gutenberg#54897](https://github.com/WordPress/gutenberg/pull/54897) to Gutenberg mobile so both repos are synced.

@geriux Thanks! Updated here:
* https://github.com/wordpress-mobile/gutenberg-mobile/pull/6261

PR #6187, opened 2023-09-12
Add block themes visual tests support

geriux:
**Related PRs:**
- https://github.com/WordPress/gutenberg/pull/54376

This PR adds the functionality to be able to test themes using our visual E2E testing setup.

It adds a new util `fetchTheme` that handles downloading the themes from their GitHub repos. The idea of dynamically downloading them is to have the latest version released to be aware of any breaking changes that might affect the mobile editor.

A new visual test is added for block-based themes:
- Initial empty state of the editor with theme colors
- Having a few standard blocks to check their font size, line height, and styles.

To test CI checks should pass (Full tests)

PR submission checklist:


fluiddot:
Hey @geriux 👋 . We will cut the `1.104.0` release tomorrow (September 14th). I plan to circle back and bump this PR to the next milestone then, but please let me know if you’d rather us work to include this PR in `1.104.0`. Thanks 🙇 !

__device-tests__/utils.js
```diff
@@ -157,3 +157,98 @@ export async function takeScreenshot( { withoutKeyboard, crop } = {} ) {
 	const resizedImage = await image.resize( widthSize, jimp.AUTO );
 	return resizedImage.getBufferAsync( jimp.MIME_PNG );
 }
+
+/**
+ * Generates the appropriate URL to fetch the theme data based on the provided theme details.
+ *
+ * @param {Object}  theme                  Object containing details about the theme.
+ * @param {string}  theme.name             The name of the theme.
+ * @param {boolean} theme.isWordPressTheme Flag indicating whether the theme is a WordPress theme. Defaults to false.
```
fluiddot: Nitpick: If this parameter is optional we can wrap it with brackets.
```suggestion
 * @param {boolean} [theme.isWordPressTheme] Flag indicating whether the theme is a WordPress theme. Defaults to false.
```

__device-tests__/gutenberg-editor-functionality-test-themes-visual.test.js
```diff
@@ -0,0 +1,97 @@
...
+ */
+const { blockNames } = editorPage;
+import { takeScreenshot, fetchTheme } from './utils';
+
+const THEMES = [
+	{ name: 'didone' },
+	{ name: 'organizer' },
+	{ name: 'montagna' },
+	{ name: 'twentytwentythree', isWordPressTheme: true },
+];
```
fluiddot, line 12: I'm curious about the rationale for picking these themes, Were they chosen because they are top-used or for another reason?

__device-tests__/utils.js
```diff
@@ -157,3 +157,98 @@ export async function takeScreenshot( { withoutKeyboard, crop } = {} ) {
...
+ * @return {Promise<Object>} A promise that resolves to a stringified JSON object containing the theme data.
+ * @throws Will throw an error if the fetch operation fails.
+ */
+export async function fetchTheme( theme ) {
+	const themeJSONLink = getThemeLink( theme );
+
+	return await fetch( themeJSONLink )
```
fluiddot: I think we don't need to use `await` here as we are using the `then` Promise functions.

__device-tests__/utils.js
```diff
@@ -157,3 +157,98 @@ export async function takeScreenshot( { withoutKeyboard, crop } = {} ) {
...
+			const supportedElements = [
+				'h1',
+				'h2',
+				'h3',
+				'h4',
+				'h5',
+				'h6',
+				'link',
+				'button',
+			];
```
fluiddot: If these are constants, we might consider moving it outside the function. WDY?T

__device-tests__/utils.js
```diff
@@ -157,3 +157,98 @@ export async function takeScreenshot( { withoutKeyboard, crop } = {} ) {
...
+		.then( ( data ) => {
+			const rawFeatures = data?.settings;
+			const rawStyles = data?.styles;
+			const supportedBlocks = [
+				'core/paragraph',
+				'core/button',
+				'core/quote',
+				'core/pullquote',
+				'core/quote',
+				'core/search',
```
fluiddot: I'm wondering if we could create this list dynamically based on the supported blocks listed here:

https://github.com/wordpress-mobile/gutenberg-mobile/blob/3969d7efb282abd521e7a2c8ee2ed5e5a244137d/src/block-support/supported-blocks.json#L2-L38

__device-tests__/utils.js
```diff
@@ -157,3 +157,98 @@ export async function takeScreenshot( { withoutKeyboard, crop } = {} ) {
...
+			}
+
+			// We filter the supported features because of limitations in sending this data
+			// during app launch on Android.
+			const filteredRawFeatures = {
+				color: rawFeatures?.color,
+				typography: {
+					fontSizes: rawFeatures?.typography?.fontSizes,
+				},
+			};
```
fluiddot: I'm curious about the limitations that force us to filter the supported features. Do you foresee any impact due to this reason in future tests?

__device-tests__/utils.js
```diff
@@ -157,3 +157,98 @@ export async function takeScreenshot( { withoutKeyboard, crop } = {} ) {
...
+					if ( rawStyles?.elements?.[ key ] ) {
+						acc[ key ] = rawStyles.elements[ key ];
+					}
+					return acc;
+				}, {} ),
+			};
+
+			return {
+				rawFeatures: JSON.stringify( filteredRawFeatures ),
+				rawStyles: JSON.stringify( filteredRawStyles ),
```
fluiddot, line 255: These values will be stringified when being passed when launching the demo app ([reference](https://github.com/WordPress/gutenberg/blob/7652de9ed159d4e9698e432c06b83e56777160c8/packages/react-native-editor/__device-tests__/helpers/utils.js#L758-L780)). I'm wondering if we could simply return the original values and let the data transformation happen when initializing the editor. WDYT?

fluiddot (commented):
@geriux It's awesome that we can now test themes 🤩 , thank you so much for incorporating these new features into our testing framework 🙇 !

I added some minor comments that shouldn't block the PR, plus some questions.

__device-tests__/utils.js
geriux: The issue is that if we allow passing more global styles for other blocks that we do support but don't support rendering their custom styles, it increases the length of the JSON that we send through the launch app params. On Android, if it's too long it gets cut off 😅

__device-tests__/utils.js
geriux: Good idea, I'll move it outside 👍

__device-tests__/utils.js
geriux: I think it'll depend on how often we'll work on expanding Global styles support in the app. We haven't added any new features in a while so it's hard to predict when we will encounter this limitation as a blocker. If in the future we need to add extra features, we would probably need to rethink how we pass this data to the demo app.

__device-tests__/gutenberg-editor-functionality-test-themes-visual.test.js
geriux: I used internal data of different areas where themes are selected while also keeping the `twentytwentythree` theme that ships with core WordPress.

__device-tests__/utils.js
fluiddot: > On Android, if it's too long it gets cut off 😅

Yep, this might be a limitation in the future. I'm wondering if we could create a temporal JSON file with all the data and pass it in the launch arguments instead 🤔.

__device-tests__/utils.js
fluiddot: Ok, I see. So the features filtered here depend on the Global styles supported in the native version.

fluiddot:
@geriux I noticed that one of the E2E tests related to themes failed ([Gutenberg Editor Visual test for Block Themes › `twentytwentythree` renders different blocks correctly](https://app.circleci.com/pipelines/github/wordpress-mobile/gutenberg-mobile/21893/workflows/b135a174-550c-4af8-a342-7bb57d58ee98/jobs/128556?invite=true#step-110-3795_104)). Per the screenshot, seems the post content is scrolled some pixels above when comparing to the baseline.

image

fluiddot:
Hey @geriux 👋, we are cutting the `1.105.0` release today, so I'll bump this PR to the next milestone.

.buildkite/commands/test-ios.sh
```diff
@@ -62,6 +62,6 @@ else
     echo "npm run $TESTS_CMD failed."
     echo "For more details about the failed tests, check the Buildkite annotation, the logs under the '$SECTION' section and the tests results in the artifacts tab."
 
-    annotate_test_failures "$JEST_JUNIT_OUTPUT_FILE" --slack "build-and-ship"
+    # annotate_test_failures "$JEST_JUNIT_OUTPUT_FILE" --slack "build-and-ship"
```
geriux: This will be removed before merging

__device-tests__/utils.js
geriux: Updated in https://github.com/wordpress-mobile/gutenberg-mobile/pull/6187/commits/976e642d8648890d35522140348080460a5dd706 thanks for the suggestion!

__device-tests__/utils.js
geriux: Updated in https://github.com/wordpress-mobile/gutenberg-mobile/pull/6187/commits/d22db6b2ab10e7283a55d91dc993a027bf194c71

__device-tests__/utils.js
geriux: Updated in https://github.com/wordpress-mobile/gutenberg-mobile/pull/6187/commits/87b79c47d61d40fc9c8a4cc6c170968e99bfb80b

__device-tests__/utils.js
geriux: > Yep, this might be a limitation in the future. I'm wondering if we could create a temporal JSON file with all the data and pass it in the launch arguments instead 🤔.

What do you mean by passing the data in the launch arguments? I thought this is what we are doing [in this code](https://github.com/WordPress/gutenberg/blob/trunk/packages/react-native-editor/__device-tests__/helpers/utils.js#L758-L768) I haven't seen another way to pass data through the launch functionality 😅

geriux:
> @geriux I noticed that one of the E2E tests related to themes failed ([Gutenberg Editor Visual test for Block Themes › `twentytwentythree` renders different blocks correctly](https://app.circleci.com/pipelines/github/wordpress-mobile/gutenberg-mobile/21893/workflows/b135a174-550c-4af8-a342-7bb57d58ee98/jobs/128556?invite=true#step-110-3795_104)). Per the screenshot, seems the post content is scrolled some pixels above when comparing to the baseline.

Yeah this took me a bit to figure out how to fix, I ended up changing the logic we use to scroll to the top on iOS, we usually use the `getTitleElement` helper with the `autoscroll` option set to `true`, this uses `swipeUp` internally to reach to the top but it was causing different values because sometimes it wouldn't scroll up to the very top.

So I added a new `tapStatusBariOS` util that will be used within `getTitleElement` this way we make sure it scrolls to the top every time.

I also saw that typing with a large font would cause flakiness due to the auto-scroll functionality to keep the keyboard below the caret, so I updated the logic of `typeStringIos` with a sleep timer to wait for that auto-scroll animation to finish.

__device-tests__/utils.js
fluiddot: The change I was thinking of is that instead of passing the raw data in the launch arguments, as we currently do, create a temporal JSON file with all the data and pass its file path in the launch arguments. This way the app would fetch all the data from a file, which I assume we won't have limitations in terms of the content length. The tricky part is that this JSON file should be accessible by the app in SauceLabs 🤔.

In any case, I think this approach would need further exploration. Since we don't have yet the need to pass complex structures, let's omit it for now.

__device-tests__/utils.js
geriux: The thing is that the editor expects these data to be string values, so if we remove this and just keep the `JSON.stringify` in `launchApp` the native side will get `rawFeatures` and `rawStyles` as JSON objects, then we will need to convert them into strings again. So I thought I'd simplify that whole process 😅

__device-tests__/utils.js
geriux: Ohh I see! Yeah it'd be a bit more complicated as we would need to bundle these files in the builds but I think it's an interesting idea to explore, as you mentioned for future work.

Maybe something to consider if we work on expanding Global styles support in the app.

__device-tests__/utils.js
fluiddot: Ah, ok I see. I didn't realize that the editor expects this data as strings 👍.

fluiddot (approved):
LGTM 🎊 ! Awesome work @geriux 🏅 !

⚠️ Before merging, let's check https://github.com/wordpress-mobile/gutenberg-mobile/pull/6187/files#r1341512059 and revert the change.

__device-tests__/utils.js
```diff
@@ -5,6 +5,25 @@ import jimp from 'jimp';
...
+const SUPPORTED_GLOBAL_STYLES_ELEMENTS = [
+	'h1',
+	'h2',
+	'h3',
+	'h4',
+	'h5',
+	'h6',
+	'link',
+	'button',
+];
```
dcalhoun, line 25: Is there a way for us to have a single source of truth for this in the `WordPress/gutenberg` repository? I am fearful this will (possibly silently) become out of date and difficult to maintain. 

I briefly searched for a simple solution, but was unable to find one. 😞

dcalhoun (commented):
Great work! 

I left one comment as I'm curious of your thoughts on the subject. 🙇🏻

jhnstn:
@geriux I'm seeing the same scrolling issue 
image
https://buildkite.com/automattic/gutenberg-mobile/builds/7529#018b2620-7722-4fd8-a450-88eee5973a6d

__device-tests__/utils.js
geriux: Sorry for the late reply!

> Is there a way for us to have a single source of truth for this in the WordPress/gutenberg repository? I am fearful this will (possibly silently) become out of date and difficult to maintain.

I think we could have it with the rest of the configuration related to global styles for mobile so it's all in the same place.

I can draft a PR with a possible solution this week 👍

geriux:
> @geriux I'm seeing the same scrolling issue

Thanks for the report @jhnstn ! I will take a look when we start migrating this test to Appium 2.

PR #6200, opened 2023-09-14
Read `RNTAztecView.podspec` from its source in Gutenberg

mokagio:
As far as I can see, so far we duplicated `RNTAztecView.podspec` in this repo based on the version in Gutenberg.

This meant doing things such as synchronizing updating the `WordPress-Aztec-iOS` dependency in Gutenberg and then in this repo.

By reading the podspec from Gutenberg instead of duplicating it we fulfill the dependency requirement for the XCFramework without needing to update a file twice. To bump the Aztec dependency, we only need to update Gutenberg itself.

I suspect `RNTAztecView.podspec` was defined locally because in the past we had something that read it, but that is no longer the case, even in the legacy `Gutenberg.podspec` that's still tracked in the repo but no longer used in our main client, WordPress iOS. Notice how the dependency is referred in the standard form here:
https://github.com/wordpress-mobile/gutenberg-mobile/blob/c70afb8da680196206b6f7f50b7d76c9b3cb51ba/Gutenberg.podspec#L27 That notation means that CocoaPods would look for it among its sources. If we were reading it from the repo, we'd need to specify it with `path:`. That's all to say, I consider this change safe to make.

To test: Verify the XCFramework published successfully.

PR submission checklist:


fluiddot (approved):
Now that we publish GBM via the XCFramework, I understand that the podspecs `RNTAztecView` and `Gutenberg` are no longer needed. I checked that the XCFramework was published successfully and E2E tests passed, so this PR LGTM 🎊 !

PR #6204, opened 2023-09-14
Release 1.104.0

fluiddot:
Release for Gutenberg Mobile 1.104.0

## Related PRs

- Gutenberg: https://github.com/WordPress/gutenberg/pull/54458
- WPAndroid: https://github.com/wordpress-mobile/WordPress-Android/pull/19190
- WPiOS: https://github.com/wordpress-mobile/WordPress-iOS/pull/21576

## Extra PRs that Landed After the Release Was Cut

No extra PRs yet. 🎉

### RichText - Update logic for the placeholder text color
- **PR:** https://github.com/WordPress/gutenberg/pull/54259
- **Issue:** https://github.com/wordpress-mobile/gutenberg-mobile/issues/6069

### feat: Insert from URL leverages a bottom sheet
- **PR:** https://github.com/WordPress/gutenberg/pull/54096
- **Issue:** https://github.com/wordpress-mobile/gutenberg-mobile/issues/4936

## Test plan

Once the installable builds of the main apps are ready, perform a quick smoke test of the editor on both iOS and Android to verify it launches without crashing. We will perform additional testing after the main apps cut their releases.

## Release Submission Checklist

- [x] Verify Items from test plan have been completed
- [x] Check if `RELEASE-NOTES.txt` is updated with all the changes that made it to the release. Replace `Unreleased` section with the release version and create a new `Unreleased` section.
- [x] Check if `gutenberg/packages/react-native-editor/CHANGELOG.md` is updated with all the changes that made it to the release. Replace `## Unreleased` with the release version and create a new `## Unreleased`.
- [x] Bundle package of the release is updated.

PR #6219, opened 2023-09-21
Bump ESLint and Prettier packages to match versions in Gutenberg

fluiddot:
This PR updates ESLint, some ESLint plugins, and Prettier packages following the changes recently made in Gutenberg (https://github.com/WordPress/gutenberg/pull/54539).

These updates are needed to avoid failures when running the command `npm run lint`, as it was identified when trying to merge the release `1.104.0` into `trunk` (https://github.com/wordpress-mobile/gutenberg-mobile/pull/6204). Here is the sequence of errors found when trying to address the failures:

### Error 1

`eslint-plugin-prettier` package was recently introduced ([reference](https://github.com/WordPress/gutenberg/pull/54539/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R198)) with the version `5.0.0`. This version uses a newer version of ESLint than the one currently installed in GBM _(version installed in GBM `6.8.0` vs. [version referenced in the ESLint plugin `8.0.0`](https://github.com/prettier/eslint-plugin-prettier/blob/b2b93c30c5679d38a094589eb608d0f5d7615aaf/package.json#L40C6-L40C12))_. This explains the error of not loading the rule `context.getPhysicalFilename`, as it was introduced in ESLint version `7.28.0` ([reference](https://eslint.org/blog/2021/06/eslint-v7.28.0-released/#context.getphysicalfilename())).

This problem was solved by upgrading the ESLint version in GBM to match [the one used in Gutenberg (`^8.3.0`)](https://github.com/WordPress/gutenberg/blob/191fd0f84223a543af0a2d0893567f558a9e547d/packages/scripts/package.json#L61).

**NOTE 1:** In Gutenberg, the ESLint version installed is the one set by the `@wordpress/scripts` package. Since we don't use it in GBM, we have to set our own ESLint version.

**NOTE 2:** Alternatively, we could add the `scripts` package as a dev dependency in GBM. However, if we reference it via a local path, ESLint won't be automatically installed. This is mentioned in NPM documentation:

> note: Packages linked by local path will not have their own dependencies installed when npm install is ran in this case. You must run npm install from inside the local path itself.

### Error 2

The `@typescript-eslint/eslint-plugin` plugin can't be loaded due to the following error: 

```
TypeError: Failed to load plugin '@typescript-eslint' declared in '.eslintrc.js » plugin:@wordpress/recommended': Class extends value undefined is not a constructor or null
Referenced from: /Users/fluiddot/workspace/gutenberg-mobile/gutenberg/packages/eslint-plugin/index.js
```

This was solved by upgrading to `6.4.1`, [as referenced in `@wordpress/eslint-plugin` package](https://github.com/WordPress/gutenberg/blob/191fd0f84223a543af0a2d0893567f558a9e547d/packages/eslint-plugin/package.json#L35).

Alternatively, we could point to the npm version. However, this seems to not work because peer dependencies are not installed at the root level and ESLint plugins are only imported from that location.

### Error 3

The `eslint-plugin-import` plugin can't be loaded:

```
Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: Failed to load plugin 'import' declared in '.eslintrc.js » plugin:@wordpress/recommended » /Users/fluiddot/workspace/gutenberg-mobile/gutenberg/packages/eslint-plugin/configs/recommended-with-formatting.js': Package subpath './lib/util/glob-util' is not defined by "exports" in /Users/fluiddot/workspace/gutenberg-mobile/node_modules/eslint/package.json
```

Similar to the above solutions, we need to upgrade it to `2.25.2` [as referenced in the Gutenberg package](https://github.com/WordPress/gutenberg/blob/191fd0f84223a543af0a2d0893567f558a9e547d/packages/eslint-plugin/package.json#L41).

### Error 4

The `eslint-plugin-prettier` plugin fails when using Prettier:

```
TypeError: prettier.resolveConfig is not a function
    at /Users/fluiddot/workspace/gutenberg-mobile/node_modules/eslint-plugin-prettier/worker.js:38:24
```

This was solved by upgrading Prettier to version `3.0.3-beta-3` as applied in [the PR](https://github.com/WordPress/gutenberg/pull/54539).

After solving the above errors, several files were formatted after running `npm run lint:fix`.

**To test:**
* Check that all CI jobs pass.

PR submission checklist:


geriux (approved):
LGTM! Thanks for addressing this 🚀 

FYI I've restarted some CI jobs since they failed due to a network error, hopefully, they'll pass this time.

PR #6269, opened 2023-10-06
Use `UnsupportedBlockDetails` component in Missing block

fluiddot:
**Related PRs:**
* https://github.com/WordPress/gutenberg/pull/55133

**To test:**
Follow testing instructions from https://github.com/WordPress/gutenberg/pull/55133

PR submission checklist:

### DIFF
--- a/src/jetpack-editor-setup.js
+++ b/src/jetpack-editor-setup.js
@@ -177,8 +177,7 @@ const setupStringsOverrides = () => {
 
 			if (
 				onlyCoreBlocks &&
-				jetpackBlockNames.includes( blockName ) &&
-				videoPressBlock
+				( jetpackBlockNames.includes( blockName ) || videoPressBlock )
 			) {
 				return null;
 			}
--- a/src/jetpack-editor-setup.js
+++ b/src/jetpack-editor-setup.js
@@ -169,13 +169,16 @@ const setupStringsOverrides = () => {
 			const { capabilities } = select( blockEditorStore ).getSettings();
 			const onlyCoreBlocks = capabilities?.onlyCoreBlocks === true;
 
-			const jetpackBlockNames = Object.keys( supportedJetpackBlocks ).map(
-				( name ) =>
-					// Add `jetpack` prefix if the block name doesn't contain a prefix.
-					/^(\w|-)+$/.test( name ) ? `jetpack/${ name }` : name
+			const namespacedBlockNames = Object.keys(
+				supportedJetpackBlocks
+			).map( ( name ) =>
+				/^(\w|-)+$/.test( name ) ? `jetpack/${ name }` : name
 			);
 
-			if ( onlyCoreBlocks && jetpackBlockNames.includes( blockName ) ) {
+			if (
+				onlyCoreBlocks &&
+				namespacedBlockNames.includes( blockName )
+			) {
 				return null;
 			}
 			return defaultValue;
--- a/src/jetpack-editor-setup.js
+++ b/src/jetpack-editor-setup.js
@@ -170,15 +170,12 @@ const setupStringsOverrides = () => {
 			const onlyCoreBlocks = capabilities?.onlyCoreBlocks === true;
 
 			const jetpackBlockNames = Object.keys( supportedJetpackBlocks ).map(
-				( name ) => `jetpack/${ name }`
+				( name ) =>
+					// Add `jetpack` prefix if the block name doesn't contain a prefix.
+					/^(\w|-)+$/.test( name ) ? `jetpack/${ name }` : name
 			);
 
-			const videoPressBlock = blockName === 'videopress/video';
-
-			if (
-				onlyCoreBlocks &&
-				( jetpackBlockNames.includes( blockName ) || videoPressBlock )
-			) {
+			if ( onlyCoreBlocks && jetpackBlockNames.includes( blockName ) ) {
 				return null;
 			}
 			return defaultValue;
--- a/src/strings-overrides.js
+++ b/src/strings-overrides.js
@@ -36,26 +36,6 @@ addFilter(
 	}
 );
 
-addFilter(
-	'native.missing_block_action_button',
-	'native/missing_block',
-	( defaultValue ) => {
-		const { capabilities } = select( blockEditorStore ).getSettings();
-		const isUnsupportedBlockEditorSupported =
-			capabilities?.unsupportedBlockEditor === true;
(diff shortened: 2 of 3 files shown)


src/jetpack-editor-setup.js
```diff
@@ -177,8 +177,7 @@ const setupStringsOverrides = () => {
 
 			if (
 				onlyCoreBlocks &&
-				jetpackBlockNames.includes( blockName ) &&
-				videoPressBlock
+				( jetpackBlockNames.includes( blockName ) || videoPressBlock )
```
fluiddot: This condition checks if the block passed is a Jetpack block, so it needed to be fixed to meet the expectation. Seems we overlooked this bug when introducing these changes (https://github.com/wordpress-mobile/gutenberg-mobile/pull/5459/files#diff-b38d335848f5255ead813ac6b271ead0d6afc53cca251ade416d02903c367b05) 🤦 .

src/strings-overrides.js
```diff
@@ -21,41 +21,16 @@ addFilter(
 		);
 
 		if (
-			isUnsupportedBlockEditorSupported === false &&
+			isUnsupportedBlockEditorSupported ||
 			canEnableUnsupportedBlockEditor
```
fluiddot, line 25: The condition to check here is that either UBE is available (`isUnsupportedBlockEditorSupported` = true) or if unavailable, it can be enabled (`canEnableUnsupportedBlockEditor`= true`).

src/jetpack-editor-setup.js
dcalhoun: This logic is a bit convoluted IMO. The VideoPress block is within `jetpackBlocknames`.

https://github.com/wordpress-mobile/gutenberg-mobile/blob/8cd476d1d8d151cf058b6f8067ef7cc28594367a/src/jetpack-editor-setup.js#L36-L38

Which subsequently receives an erroneous `jetpack/` prefix.

https://github.com/wordpress-mobile/gutenberg-mobile/blob/8cd476d1d8d151cf058b6f8067ef7cc28594367a/src/jetpack-editor-setup.js#L172-L174

Is there a way for us to better structure this logic to reduce the need for this one-off check? 

My understanding is that we cannot include the `jetpack/` prefix in `supportedJetpackBlocks` because [`getJetpackExtensionAvailability`](https://github.com/Automattic/jetpack/blob/58cc1a05fd388c7ddc3cf3936d1db1eaf9079c73/projects/js-packages/shared-extension-utils/src/get-jetpack-extension-availability.js#L6) requires its absence. 

**Idea A:** We might conditionally add the `jetpack/` prefix when mapping if the block name fails a test for a prefix.

```js
const jetpackBlockNames = Object.keys(supportedJetpackBlocks).map((name) =>
  /(\w|-)+\//.test(blockName) ? name : `jetpack/${name}`
);
```

**Idea B:** We go the opposite direction and strip all block prefixes when setting `jetpackEditorInitialState`. 

```js
const supportedJetpackBlocks = {
  "jetpack/contact-info": {
    available: true,
  },
  "jetpack/paywall": {
    available: true,
  },
  "jetpack/story": {
    available: true,
  },
  "jetpack/tiled-gallery": {
    available: __DEV__,
  },
  "videopress/video": {
    available: true,
  },
};

{
  available_blocks: supportedJetpackBlocks.map((block) => block.match(/\/(.*)/)[1]),
}
```

WDYT?

dcalhoun (approved):
Approved via https://github.com/WordPress/gutenberg/pull/55133#pullrequestreview-1664524787. 

I left one comment for consideration, but we do not need to address it in this PR. We can follow up on it later if desired.

src/jetpack-editor-setup.js
fluiddot: > Is there a way for us to better structure this logic to reduce the need for this one-off check?

Good point. We can definitely improve this logic and simplify it.

> My understanding is that we cannot include the `jetpack/` prefix in `supportedJetpackBlocks` because [`getJetpackExtensionAvailability`](https://github.com/Automattic/jetpack/blob/58cc1a05fd388c7ddc3cf3936d1db1eaf9079c73/projects/js-packages/shared-extension-utils/src/get-jetpack-extension-availability.js#L6) requires its absence.

Yes, that's correct. We'd need to maintain the format of block names in `supportedJetpackBlocks` as-is. These values are used as the `available_blocks` parameter of Jetpack's initial state, which is being used in helpers like [`getJetpackExtensionAvailability`](https://github.com/Automattic/jetpack/blob/58cc1a05fd388c7ddc3cf3936d1db1eaf9079c73/projects/js-packages/shared-extension-utils/src/get-jetpack-extension-availability.js#L6):

https://github.com/wordpress-mobile/gutenberg-mobile/blob/8cd476d1d8d151cf058b6f8067ef7cc28594367a/src/jetpack-editor-setup.js#L47-L49

> **Idea A:** We might conditionally add the `jetpack/` prefix when mapping if the block name fails a test for a prefix.

That's a good idea. Following the code you suggested, we could check the block's name format and include the `jetpack` extension only for those that don't contain the `/` character. The below suggestion is based on the one shared https://github.com/wordpress-mobile/gutenberg-mobile/pull/6269#discussion_r1350277777 but the regular expression only tests for no-slash block names:

```js
const jetpackBlockNames = Object.keys( supportedJetpackBlocks ).map(
( name ) =>
	// Add `jetpack` prefix if the block name doesn't contain a prefix.
	/^(\w|-)+$/.test( name ) ? `jetpack/${ name }` : name
);
```

> **Idea B:** We go the opposite direction and strip all block prefixes when setting `jetpackEditorInitialState`.

This would be another option. I don't have a strong preference but I'd lean toward **Idea A** to keep `supportedJetpackBlocks` as-is.

fluiddot: I applied the above suggestion in https://github.com/wordpress-mobile/gutenberg-mobile/pull/6269/commits/4b17ffbe1cdc6503dd7611b84741911abd4f1d50.

dcalhoun: Your alternative regex for **Idea A** looks good to me. 👍🏻 

Maybe rename `jetpackBlockNames` to `namespacedBlockNames` to communicate the mapping result? That might even mitigate the need for an inline comment.

fluiddot: Good idea, I've just applied this suggestion in https://github.com/wordpress-mobile/gutenberg-mobile/pull/6269/commits/530876a09d46be9c356cca9be4c27764863bd5e9.

PR #6281, opened 2023-10-12
test: Update Unsupported Block tests for Appium 2

dcalhoun:
## Related PRs
- https://github.com/WordPress/gutenberg/pull/55322

## What?

Relates to #6276. Update the Unsupported Block tests for Appium 2 and
WebdriverIO.

## Why?

Improve the stability and quality of the automated tests.

## How?

- Set static device pixel ratio
 WebdriverIO (the replacement for the deprecated WD driver library) does
 not have a method for retrieving a device's pixel ratio. Therefore, this
 sets an explicit value for each of our test devices.
- test: Replace `sleep` with `pause` utility

## Testing Instructions

Local tests should pass for both platforms when running the following commands:

```
TEST_RN_PLATFORM=ios npm run device-tests:local -- gutenberg-editor-unsupported-block-visual
```

```
TEST_RN_PLATFORM=android npm run device-tests:local -- gutenberg-editor-unsupported-block-visual
```

PR submission checklist:


__device-tests__/utils.js
```diff
@@ -40,8 +40,7 @@ const SUPPORTED_GLOBAL_STYLES_ELEMENTS = [
  * @return {Buffer} Sreenshot image.
  */
 export async function takeScreenshotByElement( element, { padding } = {} ) {
-	const sessionCapabilities = await editorPage.driver.sessionCapabilities();
```
dcalhoun, line 43: From my research, WebdriverIO provides a [`getDisplayDensity`](https://webdriver.io/docs/api/appium/#getdisplaydensity) method for Android, but no equivalent for iOS, which is unfortunately the platform on which we use/need pixel ratio.

This implementation instead relies upon a static capability configuration set in the [sibling PR](https://github.com/WordPress/gutenberg/pull/55322/files#diff-14b5427ac3a466e50e4f22aa9b8e218a07cbdc55662effc4bf57b59366b91e40R14).

__device-tests__/image-snapshots/gutenberg-editor-unsupported-block-visual-test-js-gutenberg-editor-visual-test-for-unsupported-block-should-be-able-to-open-the-unsupported-block-web-view-editor-1-android.png
dcalhoun, line 1: @geriux I noticed you captured this change for iOS in https://github.com/wordpress-mobile/gutenberg-mobile/pull/6154. Do you have any ideas why this changed occurred and why it just now shows up for Android here?

geriux: This is kind of a mystery 😅 I thought it was related to the migration to Hermes on iOS and how the capabilities were being sent 🤔 

I see there was a change recently in https://github.com/wordpress-mobile/gutenberg-mobile/pull/6269 but the full tests passed in `trunk`. Were you getting an error when running the test and then you updated the snapshot? Or did you directly run it with the update snapshot option?

dcalhoun: I updated the snapshot after noticing the test failure. 

For me, there are three questions: 

1. What is the origin for the change? 
2. Is the change correct or catching a legitimate error? 
3. Why wasn't this caught on earlier PRs via CI runs? Do we not run these snapshots tests in CI? 

@fluiddot do you happen to know if the updated snapshots discussed here are correct? This seems related to  #6269.

__device-tests__/utils.js
geriux: I've just stumbled on an issue on Android where it is expecting the pixel ratio [see these lines](https://github.com/wordpress-mobile/gutenberg-mobile/pull/6281/files#diff-17df8429cf2b4cca490b47704148d667ae52f41958836ab5a82fb243e69db41eR67-R71). If we update this to:

`const pixelRatio = editorPage.driver.capabilities.pixelRatio` it will work for both cases (if we update the iOS capabilities in the sibling PR to use `pixelRatio` as the name of the fixed values)

dcalhoun: Good catch. I interpret this to mean implementing your [prior feedback](https://github.com/WordPress/gutenberg/pull/55322#discussion_r1360399049) will address this issue. I am implementing that now. 👍🏻

geriux: Thank you!

__device-tests__/image-snapshots/gutenberg-editor-unsupported-block-visual-test-js-gutenberg-editor-visual-test-for-unsupported-block-should-be-able-to-open-the-unsupported-block-web-view-editor-1-android.png
dcalhoun: Strangely enough, running this test again locally now is producing the inverse result — i.e. this snapshot change appears invalid. I'm quite confused as to what is occurring. 😕

fluiddot: @dcalhoun I see that the difference between screenshots is the description message within the bottom sheet. This message is controlled via [the `strings-overrides.js` file](https://github.com/wordpress-mobile/gutenberg-mobile/blob/trunk/src/strings-overrides.js) that is loaded when initializing the editor from Gutenberg Mobile. I noticed that in the screenshot modified in this PR, the text displayed is the default one. I.e. the string override hasn't been applied. My presumption is that the E2E tests have been run using the bundle generated from Gutenberg possibly when running Metro from Gutenberg instead of Gutenberg Mobile. Would be this the case?

dcalhoun: No, I do not have a Metro server running. Your understanding and theory makes sense, though. 

Running the following two commands locally now results in different outcomes for iOS and Android respectively. 

```
npm run test:e2e:ios:local -- -- gutenberg-editor-unsupported-block-visual
```

```
npm run test:e2e:android:local -- -- gutenberg-editor-unsupported-block-visual
```

| iOS | Android |
| - | - |
| gutenberg-editor-unsupported-block-visual-test-js-gutenberg-editor-visual-test-for-unsupported-block-should-be-able-to-open-the-unsupported-block-web-view-editor-1-ios | gutenberg-editor-unsupported-block-visual-test-js-gutenberg-editor-visual-test-for-unsupported-block-should-be-able-to-open-the-unsupported-block-web-view-editor-1-android |

fluiddot: > No, I do not have a Metro server running. Your understanding and theory makes sense, though.
> Running the following two commands locally now results in different outcomes for iOS and Android respectively.

Oh, that's interesting. I wonder if the bundle being used is different in each platform 🤔. However, per the screenshots you shared, seems that now the iOS one is the one that is using the default message, which is unexpected if the strings are being overridden.

geriux: Ah I see, when the Hermes setup [was introduced](https://github.com/wordpress-mobile/gutenberg-mobile/pull/6154) for the iOS demo app, the bundle script for E2E testing is now being done in Gutenberg, so that's why the screenshot is different. I'll work on changing that so it generates the bundle from Gutenberg mobile.

geriux: Thanks for the insights @fluiddot !

fluiddot: > Ah I see, when the Hermes setup https://github.com/wordpress-mobile/gutenberg-mobile/pull/6154 for the iOS demo app, the bundle script for E2E testing is now being done in Gutenberg, so that's why the screenshot is different. I'll work on changing that so it generates the bundle from Gutenberg mobile.

Ah, true. I've just seen that the screenshots on iOS changed (https://github.com/wordpress-mobile/gutenberg-mobile/pull/6154/files#diff-5dfb5280d5562994f2cf3666656b21f830a8aacc80843ed2887a4a21cd3f08ae).

dcalhoun: Thank you for helping debug this oddity. It now makes sense. I reverted the updated Android snapshot. I suppose that may have come from a one-off scenario where I had either manually bundled from _within Gutenberg_ or mistakenly had a Metro server running during a test run.

geriux (approved):
LGTM! Tests succeeded as expected 🚀

PR #6296, opened 2023-10-18
Run linting and unit tests tasks in parallel in CI

mokagio:
Before, we had a 18-21 mins wait time on the lint + testing information:

image

image

By running them in parallel on Docker we only wait ~10 minutes—_assuming all three steps start more or less at the same time_.

image

**Note:** If this is approved, we'll need to update the GitHub configuration to add new individual required checks.

### Testing

See CI builds

PR submission checklist:

### DIFF
--- a/.buildkite/commands/build-ios.sh
+++ b/.buildkite/commands/build-ios.sh
@@ -1,7 +1,7 @@
 #!/bin/bash -eu
 
 echo '--- :node: Setup Node depenendencies'
-npm ci
+npm ci --unsafe-perm --prefer-offline --no-audit --no-progress
 
 echo '--- :ios: Set env var for iOS E2E testing'
 set -x
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -175,6 +175,9 @@ steps:
 
   - label: iOS Build and Sauce Labs
     key: ios-build-and-saucelabs
+    depends_on:
+      - lint
+      - ios-unit-tests
     command: .buildkite/commands/build-ios.sh
     plugins:
       - *ci_toolkit_plugin
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -37,8 +37,6 @@ steps:
     key: lint
     plugins:
       - *gb-mobile-docker-container
-      - *ci_toolkit_plugin
-      - *nvm_plugin
       - *git-cache-plugin
     command: |
       echo "--- :docker: Additional Docker image setup"
@@ -57,8 +55,6 @@ steps:
     plugins:
       - *gb-mobile-docker-container
       - *ci_toolkit_plugin
-      - *nvm_plugin
-      - *git-cache-plugin
     command: |
       echo "--- :docker: Additional Docker image setup"
       source /root/.bashrc
@@ -81,8 +77,6 @@ steps:
     plugins:
       - *gb-mobile-docker-container
       - *ci_toolkit_plugin
-      - *nvm_plugin
-      - *git-cache-plugin
     command: |
       echo "--- :docker: Additional Docker image setup"
       source /root/.bashrc
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -56,8 +56,8 @@ steps:
       - github_commit_status:
           context: Lint
 
-  - label: Unit Tests
-    key: unit-tests
+  - label: Android Unit Tests
+    key: android-unit-tests
     plugins:
       - *gb-mobile-docker-container
       - *ci_toolkit_plugin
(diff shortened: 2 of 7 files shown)


.buildkite/pipeline.yml
```diff
@@ -33,32 +33,76 @@ x-common-params:
...
+      - *ci_toolkit_plugin
+      - *nvm_plugin
+      - *git-cache-plugin
     command: |
+      echo "--- :docker: Additional Docker image setup"
       source /root/.bashrc
 
-      echo "--- :node: Setup Node environment"
+      echo "--- :docker::node: Setup Node environment"
       nvm install && nvm use
```
mokagio, line 46: I tried relying on the [nvm plugin](https://github.com/Automattic/nvm-buildkite-plugin) but, despite successfully installing nvm and the node version form the `.nvmrc` file in the repo, when the commands run in Docker, the Node version was not set as expected:

image

See https://buildkite.com/automattic/gutenberg-mobile/builds/7569#018b40c7-f76a-4ace-ae1c-2b73f44c5ab5

Maybe that has got to do with the Docker container spinning up after nvm did its setup?

.buildkite/commands/build-ios.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash -eu
 
 echo '--- :node: Setup Node depenendencies'
-npm ci
+npm ci --unsafe-perm --prefer-offline --no-audit --no-progress
```
dcalhoun, line 4: What is the intent with the addition of `--unsafe-perm`?

.buildkite/pipeline.yml
dcalhoun: I lack familiarity with the CI infrastructure — specifically `gb-mobile-docker-container` and `nvm_plugin` — but my perception of Docker is that its containers are generally used to create an environment with all the necessary requirement ready to go. 

For this particular context, is the environment provided by `gb-mobile-docker-container` _not_ already containing the correct Node.js version? Is `nvm_plugin` supposed to execute _within_ the Docker container environment to set the Node.js version? 

> Maybe that has got to do with the Docker container spinning up after nvm did its setup?

If the order of events you outline is true, then, yes, I imagine that is why the Docker Node.js environment is incorrect. If `nvm` commands are not run _within_ the Docker environment, then it is not surprising that running the commands had no impact on the Docker environment.

dcalhoun (commented):
Looking good! I left a few comments largely for my own education.

.buildkite/commands/build-ios.sh
mokagio: It was intentional in that other commands running in Docker use it, so I ported it for consistency:

https://github.com/wordpress-mobile/gutenberg-mobile/blob/81c9ed87db77693ab8e7c030906850f925698232/.buildkite/pipeline.yml#L47

https://github.com/wordpress-mobile/gutenberg-mobile/blob/81c9ed87db77693ab8e7c030906850f925698232/.buildkite/pipeline.yml#L73

As of whether it's appropriate, I don't know. I didn't dig into it at the time...

From [the docs](https://docs.npmjs.com/cli/v6/using-npm/config#unsafe-perm):

> **unsafe-perm**
>
> – Default: false if running as root, true otherwise
> - Type: Boolean
>
> Set to true to suppress the UID/GID switching when running package scripts. If set explicitly to false, then installing as a non-root user will fail.

If I'm reading it correctly, it seems like it should be redundant given that our scripts don't run as root and therefore the value is true by default.

.buildkite/pipeline.yml
mokagio: > but my perception of Docker is that its containers are generally used to create an environment with all the necessary requirement ready to go.

That's my understanding of how Docker should be used, too.

The image we have does setup `nvm` and installs the LTS version, but it maybe there's more to do to enable automatically switching to the version from `.nvmrc` when changing the folder?

https://github.com/wordpress-mobile/docker-gb-mobile-image/blob/25f008b93f9fe190f76c68f3a9719b45ad99a6cc/Dockerfile#L16

My approach at the moment is to iterate locally, which is faster. Once an implementation is locked in, we can push the additional setup in the Docker image (if we'll still be using it) to improve performance. What do you think?

.buildkite/commands/build-ios.sh
dcalhoun: Ah, OK. I was unaware of the existing usage and user context. It makes sense to be consistent. Thank you!

.buildkite/pipeline.yml
dcalhoun: > The image we have does setup `nvm` and installs the LTS version, but it maybe there's more to do to enable automatically switching to the version from `.nvmrc` when changing the folder?

Yes, this may be the issue. `nvm` does not automatically switch Node.js versions when changing directories. [Custom scripting](https://github.com/nvm-sh/nvm/tree/v0.39.5#deeper-shell-integration) is required to enable such a feature with `nvm`. 

It appears the nvm plugin does call `nvm use` — which should be enough to ensure the correct Node.js version is set for subsequent commands — but maybe later commands are executed in a different session? 

> My approach at the moment is to iterate locally, which is faster. Once an implementation is locked in, we can push the additional setup in the Docker image (if we'll still be using it) to improve performance. What do you think?

This makes a lot of sense. 👍🏻

mokagio:
@crazytonyli @jkmassel can I have your review to see if we can merge this incremental improvement? Thanks!

crazytonyli (approved):
LGTM.

I don't think it's possible to run other plugins inside the docker container launched by the docker plugin. 🤔

PR #6300, opened 2023-10-19
Update annotate_test_failures to only send Slack notification for trunk merges

jostnes:
### Description
This PR updates the condition for the `annotate_test_failures` plugin, where it would now only send Slack notifications for failing tests on `trunk` merges.

This change should reduce the noise from development branches to the `#build-and-ship` channel.


mokagio (approved):
Thanks @jostnes !

PR #6310, opened 2023-10-23
Test – Appium 2 tests in macOS CI agent without Sauce Labs

mokagio:
Work In Progress. Builds on top https://github.com/wordpress-mobile/gutenberg-mobile/pull/6276


.buildkite/commands/test-ios.sh
```diff
@@ -49,7 +49,8 @@ fi
 
 set +e
 echo "$SECTION"
-npm run "$TESTS_CMD"
+# npm run "$TESTS_CMD"
+npm run test:e2e:ios:local
```
geriux, line 53: I know this was just a test but we would also need to avoid exporting the following variables: `TEST_ENV` and `CIRCLE_BRANCH` so the tests run locally.

PR #6312, opened 2023-10-23
Update second Gutenberg Mobile package reference in `package-lock.json`

SiobhyB:
After running `npm install` against `trunk`, the second reference to Gutenberg Mobile's package version in `package-lock.json` was updated from `1.105.0` to `1.106.0`. 

I believe the structure of `package-lock.json` changed following the latest node/npm upgrades in https://github.com/WordPress/gutenberg/pull/53426, which may be why we don't currently account for this second reference as part of our release process. 

To test:

* With this branch checked out, run `npm install` to confirm no changes are suggested.

<hr />

PR submission checklist:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6,7 +6,7 @@
 	"packages": {
 		"": {
 			"name": "gutenberg-mobile",
-			"version": "1.105.0",
+			"version": "1.106.0",
 			"hasInstallScript": true,
 			"devDependencies": {
 				"@babel/core": "^7.20.0",


jhnstn (approved):
Thanks for catching this @SiobhyB !

derekblank:
@SiobhyB I believe this should be included in the upcoming 1.107.0 milestone that will be cut later this week, but just wanted to double check with you first.

SiobhyB:
👋 @derekblank, yes, we'd need to make sure the value is updated to `1.107.0` for the next release (and to reflect the release version for all future releases, too). I think it'd be necessary to run `npm install` somewhere in the release script to ensure it gets updated.

derekblank:
Thanks for clarifying -- I've added it to the 1.107.0 milestone. And good point about running `npm install` during the release as well. 👍

PR #6324, opened 2023-10-25
[Test] Run iOS E2E tests in macOS CI agent 

geriux:
**Related PRs:**
- https://github.com/WordPress/gutenberg/pull/55813

PR submission checklist:


package.json
```diff
@@ -84,9 +84,9 @@
 		"test": "cross-env NODE_ENV=test jest --config ./jest.config.js",
 		"test:update": "cross-env NODE_ENV=test jest --config ./jest.config.js --updateSnapshot",
 		"test:debug": "cross-env NODE_ENV=test node --inspect-brk node_modules/.bin/jest --runInBand --verbose --config jest.config.js",
-		"device-tests": "cross-env NODE_ENV=test jest --maxWorkers=2 --testPathIgnorePatterns='canary|gutenberg-editor-rendering|ipad' --config jest_ui.config.js",
-		"device-tests-canary": "cross-env NODE_ENV=test jest --maxWorkers=2 --testPathPattern=@canary --config jest_ui.config.js",
-		"device-tests-ipad": "cross-env NODE_ENV=test IPAD=true jest --maxWorkers=2 --testPathPattern=@ipad --config jest_ui.config.js",
+		"device-tests": "APPIUM_HOME=~/.appium NODE_ENV=test jest --maxWorkers=1 --testPathIgnorePatterns='canary|gutenberg-editor-rendering|ipad' --config jest_ui.config.js",
```
mokagio: I've been trying to run the tests locally myself on iOS and Android and I haven't understood the recommended mode to run Appium.

[The setup automation checks for a global Appium then defers to the one installed in the local modules](https://github.com/WordPress/gutenberg/blob/820b13599bf12f5efa472e36a780cd63e86a67b3/packages/react-native-editor/bin/test-e2e-setup.sh#L5-L10). What is best? Is there any advantage in installing it globally when using the version in the local `node_modules` "guarantees" every instance will run the same version?

.buildkite/commands/test-ios.sh
```diff
@@ -19,36 +21,46 @@ while [ "$INPUT" != "" ]; do
...
-export TEST_ENV=sauce
+export TEST_ENV=ci
 export JEST_JUNIT_OUTPUT_FILE="reports/test-results/ios-test-results.xml"
 # Set a destination different from the hardcoded one which only works in the
 # older Xcode-setup used by CircleCI
-export RN_EDITOR_E2E_IOS_DESTINATION='platform=iOS Simulator,name=iPhone 13,OS=16.4'
-# This is a relic of the CircleCI setup.
-# It should be removed once the migration to Buildkite is completed.
-export CIRCLE_BRANCH=${BUILDKITE_BRANCH}
+export RN_EDITOR_E2E_IOS_DESTINATION='platform=iOS Simulator,name=iPhone 14,OS=16.4'
```
mokagio: Granted that this is just a test 😄 

But for future reference, it would be useful to interpolate the device here
```suggestion
export RN_EDITOR_E2E_IOS_DESTINATION="platform=iOS Simulator,name=$DEVICE,OS=16.4"
```

geriux: Thanks for the suggestion! I've updated it a different PR https://github.com/wordpress-mobile/gutenberg-mobile/pull/6336 as for the `RN_EDITOR_E2E_IOS_DESTINATION` variable in `test-ios.sh` it wasn't needed so I've removed it 👍

package.json
geriux: As it is currently set up Appium installs in the local `node_modules` folder of the project but the drivers are installed globally so we export the `APPIUM_HOME` path so the drivers can be found. In terms of what's best, I'd say it's more of a preference if you want to have it installed globally.

My idea was to make it work for the migration but I'm open to suggestions. I'm all for improving the setup and making it as easy as possible 🚀 

We also [updated the E2E docs](https://github.com/WordPress/gutenberg/blob/trunk/packages/react-native-editor/__device-tests__/README.md) with the migration, any feedback with the document would be also appreciated if you find any issues when trying to run the tests locally.

mokagio: > In terms of what's best, I'd say it's more of a preference if you want to have it installed globally.

Oh, my preference would be to install everything locally. I find it's a cleaner setup. In fact, if it wasn't for the fact that some of the code needs to run on macOS, I'd even argue for Dockerizing the project so that we can all develop against the same image with all the same libraries etc.

geriux: > Oh, my preference would be to install everything locally

Awesome!

>  if it wasn't for the fact that some of the code needs to run on macOS, I'd even argue for Dockerizing the project so that we can all develop against the same image with all the same libraries etc.

I agree, that would be great if we could 😄

PR #6325, opened 2023-10-25
Remove the 'version' command from package.json

jhnstn:
The `version` command is running the bundle step which is no longer needed to create a release.
Also, using `npm version <newversion>` is the cleanest way to update the package version. 

To test:
N/A

PR submission checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -104,8 +104,7 @@
 		"lint": "eslint . --ext .js",
 		"lint:fix": "npm run lint -- --fix",
 		"xcframework:setup": "./bin/xcframework-setup",
-		"xcframework:build": "npm run xcframework:setup && pushd ./ios-xcframework && ./build.sh && popd",
-		"version": "npm run bundle && git add -A bundle"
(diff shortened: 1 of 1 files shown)


jhnstn:
Note: by default `npm version` will add a git tag but that can be disabled.

package.json
```diff
@@ -104,8 +104,7 @@
 		"lint": "eslint . --ext .js",
 		"lint:fix": "npm run lint -- --fix",
 		"xcframework:setup": "./bin/xcframework-setup",
-		"xcframework:build": "npm run xcframework:setup && pushd ./ios-xcframework && ./build.sh && popd",
-		"version": "npm run bundle && git add -A bundle"
```
dcalhoun, line 108: I was curious of the origin of this script. It appears to have been added in https://github.com/wordpress-mobile/gutenberg-mobile/pull/255 as a way to ease remembering to bundle prior to committing a version bump. Given that, removing the script makes sense now that we bundle on the CI server and automate version bumps with other scripts.

dcalhoun (approved):

PR #6476, opened 2023-12-18
Add instructions for providing Aztec commit number in `ios-xcframework`

SiobhyB:
### Description

It's sometimes desirable to create an installable build based on a recent commit to the [AztecEditor-iOS](https://github.com/wordpress-mobile/AztecEditor-iOS) repository. To do this, there are two steps:

1. Update [these lines](https://github.com/wordpress-mobile/WordPress-iOS/blob/570ecfd18443080c2a90eab23ae2076614ee4925/Podfile#L35-L38) in the `WordPress-iOS` repo's Podfile.
2. Update the Podfile in `ios-xcframework` here in the `gutenberg-mobile` repo.

For the second step, the exact format for adding the commit is not currently easily discoverable. With this PR, I've attempted to rectify this by adding a comment similar to the one found in the `WordPress-iOS` repo. 

@mokagio, what do you think? Looking forward to hearing your thoughts or whether you see other opportunities here. Thanks! 🙇‍♀️  

<hr />

PR submission checklist:


SiobhyB:
@mokagio, pinging again in case this one got lost in the sea of EOY pings :D

ios-xcframework/Podfile
```diff
@@ -68,6 +68,10 @@ use_react_native! path: REACT_NATIVE_PATH, hermes_enabled: HERMES_ENABLED
 
 target 'Gutenberg' do
   pod 'RNTAztecView', path: File.join(GUTENBERG_PATH, 'packages', 'react-native-aztec')
+  # When using a tagged version, leave the WordPress-Aztec-iOS line commented out below.
+  # When using a commit number (to create an installable build) add the hash to the commit parameter.
+  # After changing, remember to run `bundle exec pod install` in this directory to update the Podfile.lock.
+  # pod 'WordPress-Aztec-iOS', :git => 'https://github.com/wordpress-mobile/WordPress-Aztec-iOS.git', :commit => ''
```
mokagio: Nitpick: there is a more compact Ruby dictionary key syntax.

```suggestion
  # pod 'WordPress-Aztec-iOS', git: 'https://github.com/wordpress-mobile/WordPress-Aztec-iOS.git', commit: ''
```

mokagio (commented):
It took me a while to wrap my head around the setup that this suggestion makes a bit clearer works:

Here's my understanding:

- We may at times want to use a different version of `RNTAztecView` and we might want to generate builds to distribute that use that version
- `RNTAztecView` depends on `WordPress-Aztec-iOS`
- It's simpler to force `WordPress-Aztec-iOS` to a different commit by changing it in the `Podfile` here than by 1) updating the `RNTAztecView` `podspec` file in the Gutenberg repo; 2) pushing the change to Gutenberg; 3) updating the submodule in this repo.

@SiobhyB have I got it right?

If so, I'm afraid this approach is an unreliable hack. I haven't verified it manually, but I think that as soon as the `commit` points WordPress-Aztec-iOS to a version of the code where the `spec.version` is different that `1.19.9`—the version [locked](https://github.com/WordPress/gutenberg/blob/beb10327bc8942d5e0a7572242846d8d49d67aac/packages/react-native-aztec/RNTAztecView.podspec#L23-L25) in the RNTAztecView `podspec`—the dependency resolution will fail.

😞 

That's not to say we shouldn't use this approach when convenient or tracking the information for the formatting in the `Podfile`. But maybe there's a bit more explanation that needs to go into how to use it and how it could fail?

ios-xcframework/Podfile
SiobhyB: Thank you! I've applied the suggestion.

SiobhyB:
@mokagio, I think the understanding you outlined is correct. The use case is creating an installable build for designers or developers to test PR changes. The goal is not to actually merge the commit change to `trunk`.

The background where I had a need to create an installable build, which led to this PR, can be found at p1702908292558979-slack-C0126N6JH7Z. 

> It's simpler to force WordPress-Aztec-iOS to a different commit by changing it in the Podfile here than by 1) updating the RNTAztecView podspec file in the Gutenberg repo; 2) pushing the change to Gutenberg; 3) updating the submodule in this repo.

I'm not familiar with that process, but it does sound perhaps overly involved and time-consuming for our use case, in which we simply want a way to test PR changes via an installable build. If it'd be a preferable approach, though, I'd be happy to try and figure out a place to document it.

> If so, I'm afraid this approach is an unreliable hack. I haven't verified it manually, but I think that as soon as the commit points WordPress-Aztec-iOS to a version of the code where the spec.version is different that 1.19.9—the version [locked](https://github.com/WordPress/gutenberg/blob/beb10327bc8942d5e0a7572242846d8d49d67aac/packages/react-native-aztec/RNTAztecView.podspec#L23-L25) in the RNTAztecView podspec—the dependency resolution will fail.

That's good to know. Though, I suspect that the Aztec version is unlikely to be updated in the process of creating/testing changes for a single PR, judging by the [current frequency of releases](https://github.com/wordpress-mobile/AztecEditor-iOS/releases).

> That's not to say we shouldn't use this approach when convenient or tracking the information for the formatting in the Podfile. But maybe there's a bit more explanation that needs to go into how to use it and how it could fail?

Adding more detail is a good call. Do you think the [WordPress iOS Integration Guide](https://github.com/wordpress-mobile/gutenberg-mobile/blob/trunk/WPiOS%20Integration.md) is a sound place for this? If so, I'll update the PR with some updates to that guide when I'm able to carve the time out.

SiobhyB:
@mokagio, I've added some more in-depth instructions to [WPiOS Integration.md](https://github.com/wordpress-mobile/gutenberg-mobile/blob/296e2030a9a3ee8f4a700a639b0f4efb9f9f2d15/WPiOS%20Integration.md), let me know if that captures the type of extra detail you were imagining! My main hope with this PR is to be able to help save people's time in the future, as it took a non-trivial amount of time for me to figure out exactly what was needed to create an installable build for an Aztec change.

mokagio (approved):
Thanks for iterating @SiobhyB 🙇‍♂️

mokagio:
@SiobhyB CI failed with this error:

image

The reason it happened is that `Podfile` has been updated, even though only with comments. When `bundle exec pod install` runs, it tracks a checksum of the _whole_ `Podfile` in the lockfile. I never looked into the rationale in the docs, but I think it's fair to say it's an added level of security to ensure a `Podfile` and `Podfile.lock` match and that when `pod install` reads `Podfile.lock`, the dependencies it will fetch have been resolved according to what is specified in the current `Podfile`. 

That's all to say, after every edit of `Podfile`, even typo fixes or adding comments, one needs to remember to run `bundle exec pod install`.

6f2d80df10c2fd4007ac034ea13da85d5b0ec406 does that.

PR #6487, opened 2023-12-21
Update Node to version 20

fluiddot:
This PR is a follow-up on the Gutenberg PR https://github.com/WordPress/gutenberg/pull/56331 to update Node to the same version used in Gutenberg.

More information can be found in:
https://make.wordpress.org/core/2023/12/20/updating-wordpress-to-use-more-modern-versions-of-node-js-npm-2/

### Jetpack repository
No updates are needed as it already uses node 20 ([reference](https://github.com/Automattic/jetpack/blob/014b0873e1edbe60241b2d79abe25bdccf69baf2/.nvmrc#L1)).

### Block Experiments repository
It currently uses node 16, so eventually we'd need to update it ([reference](https://github.com/Automattic/block-experiments/blob/0f5a7d7544eab76127f8cd6ad7ceb22bd912e510/.nvmrc#L1)). Similar to what happened in GBM, we need to update the dependency `eslint-plugin-jsdoc`. Since the affected package is not strictly related to editor/block logic, looks safe to continue this PR without updating node in this repository.

In any case, I opened https://github.com/Automattic/block-experiments/pull/340 as a follow-up.

**To test:**
* Smoke test the editor.
* Check that all CI jobs succeed.

PR submission checklist:


package.json
```diff
@@ -34,7 +34,7 @@
 		"eslint-plugin-eslint-comments": "3.1.2",
 		"eslint-plugin-import": "2.25.2",
 		"eslint-plugin-jest": "27.2.1",
-		"eslint-plugin-jsdoc": "^39.6.9",
+		"eslint-plugin-jsdoc": "^46.4.6",
```
fluiddot, line 37: This dependency has been updated to support node 20.

package.json
```diff
@@ -7,8 +7,8 @@
 		"scssfiles": "src/*.scss src/**/*.scss"
 	},
 	"engines": {
-		"node": ">=16.0.0",
-		"npm": ">=8 <9"
+		"node": ">=20.10.0",
+		"npm": ">=10.2.3"
```
fluiddot, line 11: This configuration matches the one used in Gutenberg ([reference](https://github.com/WordPress/gutenberg/blob/1275d2f0d02961a8cc109bcc3f3ab76a8609bdd6/package.json#L17-L20)).

geriux (approved):
LGTM! Thanks for updating these, I've smoke-tested the demo app and all of the CI jobs passed correctly 🚀

PR #6525, opened 2024-01-09
Use https git urls in git submodules

crazytonyli:
## Issue

The git-mirror feature in buildkite-agent clones all git repository into a local directory. But it doesn't resolve a relative submodule url to a real git url (based on the "host" repository's URL) and fails to clone the submodule repositories.

## Fix

Changing the relative url to a https URL fixed the issue. I'll report the issue to buildkite too. If they resolve it in the buildkite-agent, we have the option of reverting the submodule url change if needed.

> [!Warning]
> Once this PR is merged and pulled into your local git checkout, you may need to run `git submodule sync` to update the submodule URLs in your work copy.

PR submission checklist:

### DIFF
--- a/.buildkite/commands/build-ios.sh
+++ b/.buildkite/commands/build-ios.sh
@@ -13,7 +13,7 @@ IOS_DEVICE_NAME=$(jq -r '.ios.local.deviceName' "$CONFIG_FILE")
 IOS_PLATFORM_VERSION=$(jq -r '.ios.buildkite.platformVersion' "$CONFIG_FILE")
 # Set a destination different from the hardcoded one which only works in the
 # older Xcode-setup used by CircleCI
-export RN_EDITOR_E2E_IOS_DESTINATION="platform=iOS Simulator,name=$IOS_DEVICE_NAME,OS=$IOS_PLATFORM_VERSION"
+export RN_EDITOR_E2E_IOS_DESTINATION="platform=iOS Simulator,name=iPhone 15,OS=17.2"
 set +x
 
 echo '--- :react: Build iOS app for E2E testing'
--- a/.buildkite/commands/build-ios.sh
+++ b/.buildkite/commands/build-ios.sh
@@ -7,12 +7,9 @@ echo '--- :ios: Set env var for iOS E2E testing'
 set -x
 export TEST_RN_PLATFORM=ios
 export TEST_ENV=sauce
-CONFIG_FILE="$(pwd)/gutenberg/packages/react-native-editor/__device-tests__/helpers/device-config.json"
-# Uses the local deviceName since SauceLabs uses a different one.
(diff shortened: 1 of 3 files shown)


.buildkite/commands/build-ios.sh
```diff
@@ -7,13 +7,10 @@ echo '--- :ios: Set env var for iOS E2E testing'
...
-# Uses the local deviceName since SauceLabs uses a different one.
-IOS_DEVICE_NAME=$(jq -r '.ios.local.deviceName' "$CONFIG_FILE")
-IOS_PLATFORM_VERSION=$(jq -r '.ios.buildkite.platformVersion' "$CONFIG_FILE")
-# Set a destination different from the hardcoded one which only works in the
-# older Xcode-setup used by CircleCI
-export RN_EDITOR_E2E_IOS_DESTINATION="platform=iOS Simulator,name=$IOS_DEVICE_NAME,OS=$IOS_PLATFORM_VERSION"
+# We must use a simulator that's available on the selected Xcode version
+# otherwsie Xcode fallbacks to "generic destination" which requires provision
+# profiles to built the Demo app.
+export RN_EDITOR_E2E_IOS_DESTINATION="platform=iOS Simulator,name=iPhone 15,OS=17.2"
```
crazytonyli, line 13: @mokagio I didn't change [the `buildkite.platformVersion`](https://github.com/WordPress/gutenberg/blob/e8b0a4c459dd2dd8f6775f312128cc6e7106a6bc/packages/react-native-editor/__device-tests__/helpers/device-config.json#L13), because I think its value depends on the Xcode version used in the pipeline or on our own mac. I'm not sure if we need the dance of updating it in the `wordpress/gutenberg` repo so that pipeline in this repo can pass. I'm not sure if I missed anything though...

mokagio: Thanks @crazytonyli. Yeah I'm not sure either... I think this hardcoded solution is an excellent step to get this over the line.

We can then think about the best type of configuration.

However, I too feel uncomfortable with having to set values for a CI build that runs on this repo via a different repo. Given this handy env var that allows overrides, I think we could use this solution in the long run, too.

mokagio (approved):
The build for the latest commit hasn't finished, but [an earlier one](https://buildkite.com/automattic/gutenberg-mobile/builds/8214#018cf004-81b7-47cd-be59-44833d7193aa) clearly shows CI successfully fetching the submodules 👏

mokagio:
🤔 interesting that CircleCI hasn't reported here. 

My PR, https://github.com/woocommerce/woocommerce-ios/pull/11352, removes CircleCI's `config.yml`, but there are not changes here 🤔 

Given that CircleCI is scheduled to be removed via #11352, I'd say we can merge this as is. Besides, there are no source changes in this PR, only CI configurations.

PR #6558, opened 2024-01-19
Use `android` queue instead of `gb-mobile-...` Docker image

mokagio:
With this switch, we gain support for unit test failure reporting, see [this example build](https://buildkite.com/automattic/gutenberg-mobile/builds/8285#018d15f0-ee5a-460c-813e-88247eb44ca5), and in the order of 1-2 minute faster execution for each of Lint and Unit Tests steps. All that with room for improvement in terms of adding caching layer to make `npm ci` faster.

image

To test: Verify CI is greed.

PR submission checklist:

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -74,6 +74,8 @@ steps:
     agents:
       queue: android
     command: .buildkite/commands/unit-tests-ios.sh
+    env:
+      JEST_JUNIT_OUTPUT_FILE: reports/test-results/ios-test-results.xml
     artifact_paths:
       - ./logs/*.log
       - ./reports/test-results/*.xml
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -40,16 +40,11 @@ steps:
   - label: Lint
     key: lint
     plugins:
-      - *gb-mobile-docker-container
+      - *nvm_plugin
       - *git-cache-plugin
-    command: |
-      echo "--- :docker: Additional Docker image setup"
-      source /root/.bashrc
-
-      echo "--- :docker::node: Set up Node environment"
-      nvm install && nvm use
-
-      .buildkite/commands/lint.sh
+    agents:
+      queue: android
+    command: .buildkite/commands/lint.sh
     notify:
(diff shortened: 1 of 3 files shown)


.buildkite/pipeline.yml
```diff
@@ -120,21 +107,17 @@ steps:
...
+  - label: Build JS Bundles
     depends_on:
       - lint
       - android-unit-tests
       - ios-unit-tests
-    key: "js-bundles"
+    key: js-bundles
     plugins:
-      - *gb-mobile-docker-container
+      - *ci_toolkit_plugin # unused?
```
fluiddot, line 117: @mokagio If we are unsure if it's used, we could try to remove it in a separate PR to determine it. WDYT?

fluiddot (approved):
LGTM 🎊 !

.buildkite/pipeline.yml
```diff
@@ -120,21 +107,17 @@ steps:
...
-  - label: "Build JS Bundles"
+  - label: Build JS Bundles
     depends_on:
       - lint
       - android-unit-tests
       - ios-unit-tests
-    key: "js-bundles"
+    key: js-bundles
     plugins:
-      - *gb-mobile-docker-container
```
fluiddot, line 130: @mokagio I overlooked checking in this PR that the `Build JS Bundles` job skips a build command on PRs. 

https://github.com/wordpress-mobile/gutenberg-mobile/blob/3ce96799f8f99bea17198c609cc885010be82f92/.buildkite/pipeline.yml#L124-L129

Today when creating a tag I bumped into the following error ([reference](https://buildkite.com/automattic/gutenberg-mobile/builds/8388#018d453c-de7f-4266-af9d-174678456092/649-907)):
```
Upgrading WP-CLI
<LOCAL_PATH>/gutenberg-mobile/gutenberg/packages/react-native-editor/bin/generate-pot-files.sh: line 80: php: command not found
```

We need to have PHP installed to use `WP-CLI`. IIRC in the GB-mobile docker image we installed PHP (https://github.com/wordpress-mobile/docker-gb-mobile-image/pull/3), could we do the same for the image used in this job? Thanks 🙇 !

PR #6579, opened 2024-01-25
Increase node max available memory in `Build JS Bundles` job

fluiddot:
Increases the max available memory that `node` can use. This follows the same approach applied in https://github.com/wordpress-mobile/gutenberg-mobile/pull/6503 to avoid CI failures when building Gutenberg, which requires more memory than the default value (2GB).

**To test:**
* Observe that all CI checks pass.
* Observe in https://github.com/wordpress-mobile/gutenberg-mobile/pull/6579/commits/a1bc23ef70bd10097b10a3084b6fc2f09c8e4f1c that the heap size is set to 4G ([reference](https://buildkite.com/automattic/gutenberg-mobile/builds/8380#018d428c-5736-436f-aa24-4b67b3f02c5a/358-397)).

PR submission checklist:

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -110,8 +110,6 @@ steps:
     env:
       NODE_OPTIONS: "--max-old-space-size=4096"
     command: |
-        node -e 'console.log("Heap size limit:",v8.getHeapStatistics().heap_size_limit/(1024*1024))'
-
         echo "--- :npm: Install Node dependencies"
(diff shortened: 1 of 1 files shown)


.buildkite/pipeline.yml
```diff
@@ -1,14 +1,4 @@
 x-common-params:
-  - &gb-mobile-docker-container
```
fluiddot, line 2: I removed this common parameter as it's no longer used in the pipeline configuration.

geriux (approved):
LGTM! Thanks for addressing this 🚀

PR #6596, opened 2024-02-01
Update Gutenberg ref to latest from trunk 

derekblank:
Updates Gutenberg ref to latest from trunk to incorporate E2E test fix from https://github.com/wordpress-mobile/gutenberg-mobile/pull/6588.

### DIFF
--- a/ios-xcframework/Podfile.lock
+++ b/ios-xcframework/Podfile.lock
@@ -3,14 +3,14 @@ PODS:
   - BVLinearGradient (2.7.3):
     - React-Core
   - DoubleConversion (1.1.6)
-  - FBLazyVector (0.71.11)
-  - FBReactNativeSpec (0.71.11):
+  - FBLazyVector (0.71.15)
+  - FBReactNativeSpec (0.71.15):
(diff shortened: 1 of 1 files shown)


mokagio (approved):
@derekblank the PR in the description results closed, but the diff shows it only updated the Gutenberg submodule, to point to https://github.com/WordPress/gutenberg/pull/58408 which is indeed now merged into Gutenberg's `trunk`.

CI is green, and the `Podfile.lock` points the various React Native dependencies to version 0.71.15, which is what we expected (based on my understanding of the context from p1706755410758249-slack-C0126N6JH7Z). So I say we can merge 👍

derekblank:
> The PR in the description results closed, but the diff shows it only updated the Gutenberg submodule, to point to https://github.com/WordPress/gutenberg/pull/58408 which is indeed now merged into Gutenberg's trunk.

You are correct -- I linked to to the closed https://github.com/wordpress-mobile/gutenberg-mobile/pull/6588, but what I meant to say is that we're bringing in the changes from https://github.com/WordPress/gutenberg/pull/58408. Thanks for the review. 🚀

PR #6624, opened 2024-02-09
Update metro config for bundles

geriux:
This PR updates the `bundle` commands to include the updated configuration params to make it work for React Native `0.73.3`.

To test, bundle-related tests should pass.

PR submission checklist:


package.json
```diff
@@ -90,7 +88,7 @@
...
 		"test:e2e:bundle:android": "npm run test:e2e:bundle:android:text && npm run test:e2e:bundle:android:bytecode",
-		"test:e2e:bundle:android:text": "mkdir -p gutenberg/packages/react-native-editor/android/app/src/main/assets && npm run rn-bundle -- --reset-cache --platform android --dev false --minify false --entry-file index.js --bundle-output gutenberg/packages/react-native-editor/android/app/src/main/assets/index.android.text.bundle --assets-dest gutenberg/packages/react-native-editor/android/app/src/main/res",
+		"test:e2e:bundle:android:text": "mkdir -p gutenberg/packages/react-native-editor/android/app/src/main/assets && npm run rn-bundle -- -- -- -- --config ../../../metro.config.js --reset-cache --platform android --dev false --minify false --entry-file index.js --bundle-output ./android/app/src/main/assets/index.android.text.bundle --assets-dest ./android/app/src/main/res",
```
dcalhoun, line 91: It is not clear to me why Android has a dedicated E2E bundle command where as iOS relies upon the default bundle command. It could be for good reason or we might consider aligning the implementation in a separate PR.

dcalhoun (approved):
LGTM. Thanks!

package.json
geriux: > we might consider aligning the implementation in a separate PR

That would be a good idea, maybe even just moving all the scripts to Gutenberg in a way that accepts external params so it can set its custom paths from Gutenberg Mobile

dcalhoun: Good idea. Consolidating scripts would be great. I also think we should follow the `wp-scripts` pattern of moving them out of `package.json` and into dedicated files that are easier to read and test.

geriux: Definitely 💯

PR #6638, opened 2024-02-14
Add `git-cache-plugin` to CI steps that were missing it

mokagio:
@geriux made me notice that not all the CI steps used the `git-s3-cache` plugin, thus missing out on using the cached version of the repo.

Unfortunately, the checkout times seem to be the same with or without the plugin available. I'm a bit rusty on how the plugin works exactly, but I think we ought to look at its internals and see if there's something more to do, or something that's not working out properly, maybe because of the submodules in this repo.

Still, this addition doesn't make things worse. And, assuming we can improve `git-s3-cache` efficacy, makes the whole pipeline ready to benefit from it.

---

PR submission checklist:

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -54,6 +54,7 @@ steps:
     plugins:
       - *nvm_plugin
       - *ci_toolkit_plugin
+      - *git-cache-plugin
     agents:
       queue: android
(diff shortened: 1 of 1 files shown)


geriux (approved):
Thank you Gio! Yeah, it makes sense to include it for the missing jobs, this way once the git cache issues are solved they will benefit from it. 🚀

PR #6719, opened 2024-03-12
Change CI `checkout` approach for faster repo setup

mokagio:
This experimental PR builds on top of a suggestion by @AliSoftware to add a `--filter` to the `git clone` command as explained [in this blog post by GitHub](https://github.blog/2020-12-21-get-up-to-speed-with-partial-clone-and-shallow-clone/).

> **Treeless clones**
> In some repositories, the tree data might be a significant portion of the history. Using `--filter=tree:0`, a treeless clone downloads all reachable commits, then downloads trees and blobs on demand.
> [...]
> Treeless clones are really only helpful for automated builds when you want to quickly clone, compile a project, then throw away the repository. 

At first, I tried setting `--filter=tree:0` via the env var Buildkite offers. That worked well for checking out `gutenberg-mobile` but didn't improve the situation for its submodules. I tinkered with options but didn't get far. As far as I could tell, while one can customize some of the checkout behavior, Buildkite still adds certain parameters to the commands which interfered with the approach.

As such, I tried using [a custom checkout script](https://github.com/Automattic/git-partial-clone-buildkite-plugin/blob/77ac31a82ea20f8f6f5d2734db5dc162c97b39a9/hooks/checkout):

```sh
#!/usr/bin/env bash

# Hooks are sourced, so we need to set flags via `set`, not shebang.
set -eu

echo '[git-partial-clone-plugin] :git: Adding GitHub to known hosts'
ssh-keyscan -t rsa github.com >> "$HOME/.ssh/known_hosts"

# See https://github.blog/2020-12-21-get-up-to-speed-with-partial-clone-and-shallow-clone/
#
# > Treeless clones are really only helpful for automated
# > builds when you want to quickly clone, compile a project,
# > then throw away the repository. In environments like
# > GitHub Actions using public runners, you want to minimize
# > your clone time so you can spend your machine time
# > actually building your software! Treeless clones might be
# > an excellent option for those environments.

echo '[git-partial-clone-plugin] :git: Cloning repo and submodules'
git clone --filter=tree:0 --recurse-submodules --also-filter-submodules "$BUILDKITE_REPO" "$BUILDKITE_BUILD_CHECKOUT_PATH"

cd "$BUILDKITE_BUILD_CHECKOUT_PATH"

git submodule foreach --recursive "git clean -ffxdq"
git clean -ffxdq

echo '[git-partial-clone-plugin] :git: Checking out commit'
git checkout -f "$BUILDKITE_COMMIT"

git submodule sync --recursive
git submodule update --init --recursive
```

The result was that **checking out the repo now takes ~30 seconds** where before it took ~5 minutes.

image

image

The improvement is so good and promising that I'm scared I'm missing something here. Like, why wouldn't Buildkite use this approach out of the box?

Still, I'm putting it out here for your feedback and consideration. Let me know what you think.

_Note:_ The iOS steps running on the `mac` queue don't use the plugin because the Apple Silicon CI optimization for Git @jkmassel built already delivers great speed.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -43,6 +43,8 @@ steps:
       - *ci_toolkit_plugin
       - automattic/git-partial-clone#trunk
       - *nvm_plugin
+    env:
+      BUILDKITE_PLUGINS_ALWAYS_CLONE_FRESH: true
     agents:
       queue: android
     command: .buildkite/commands/lint.sh
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -55,9 +55,9 @@ steps:
   - label: Android Unit Tests
     key: android-unit-tests
     plugins:
+      - automattic/git-partial-clone#mokagio/experiments
       - *nvm_plugin
       - *ci_toolkit_plugin
-      - *git-cache-plugin
     agents:
       queue: android
     command: .buildkite/commands/unit-tests-android.sh
@@ -75,7 +75,7 @@ steps:
     plugins:
       - *nvm_plugin
       - *ci_toolkit_plugin
-      - *git-cache-plugin
+      - automattic/git-partial-clone#mokagio/experiments
     agents:
       queue: android
     command: .buildkite/commands/unit-tests-ios.sh
@@ -97,7 +97,7 @@ steps:
       ./gradlew testDebug
     plugins:
       - *ci_toolkit_plugin
-      - *git-cache-plugin
+      - automattic/git-partial-clone#mokagio/experiments
       - *nvm_plugin
     agents:
       queue: android
@@ -111,7 +111,7 @@ steps:
       ./gradlew test
     plugins:
       - *ci_toolkit_plugin
-      - *git-cache-plugin
+      - automattic/git-partial-clone#mokagio/experiments
       - *nvm_plugin
     agents:
       queue: android
@@ -126,7 +126,7 @@ steps:
       - *gb-mobile-docker-container
       # - *ci_toolkit_plugin # unused?
       # - *nvm_plugin
-      - *git-cache-plugin
+      - automattic/git-partial-clone#mokagio/experiments
     command: |
         source /root/.bashrc
 
@@ -162,7 +162,7 @@ steps:
   - label: "Build Android RN Aztec & Publish to S3"
     key: "publish-react-native-aztec-android"
     plugins:
-      - *git-cache-plugin
+      - automattic/git-partial-clone#mokagio/experiments
       - *publish-android-artifacts-docker-container
     command: .buildkite/commands/publish-react-native-aztec-android-artifacts.sh
 
@@ -171,7 +171,7 @@ steps:
       - "js-bundles"
       - "publish-react-native-aztec-android"
     plugins:
-      - *git-cache-plugin
+      - automattic/git-partial-clone#mokagio/experiments
       - *publish-android-artifacts-docker-container
     command: .buildkite/commands/publish-react-native-bridge-android-artifacts.sh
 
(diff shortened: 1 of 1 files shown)


fluiddot:
@mokagio I noticed in [the article you shared](https://github.blog/2020-12-21-get-up-to-speed-with-partial-clone-and-shallow-clone/#) that for **Treeless clones** there's warning related to submodules:

> ⚠️ Warning: While writing this article, we were putting treeless clones to the test beyond the typical limits. **We noticed that repositories that contain submodules behave very poorly with treeless clones.** Specifically, if you run `git fetch` in a treeless clone, then the logic in Git that looks for changed submodules will trigger a tree request for every new commit! **This behavior can be avoided by running `git config fetch.recurseSubmodules false` in your treeless clones.** We are working on a more robust fix in the Git client.

I wonder if applying the suggestion of using `git config fetch.recurseSubmodules false` might speed up the process. Although, I suspect that using the parameter `--also-filter-submodules` might be an alternative to this, is this accurate?

.buildkite/pipeline.yml
```diff
@@ -14,6 +14,8 @@ x-common-params:
       # Ensure these settings match what's defined in cache-builder.yml
       bucket: a8c-repo-mirrors
       repo: automattic/gutenberg-mobile/
+  - &git-partial-clone-plugin
+    automattic/git-partial-clone#trunk
```
fluiddot: We use fixed versions for other plugins, could we do the same for this?

.buildkite/pipeline.yml
```diff
@@ -40,8 +42,10 @@ steps:
   - label: Lint
     key: lint
     plugins:
+      - *git-partial-clone-plugin
       - *nvm_plugin
-      - *git-cache-plugin
+    env:
+      BUILDKITE_PLUGINS_ALWAYS_CLONE_FRESH: true
```
fluiddot: I might be overlooking this, but I wonder about the reason of setting this environment only in this job. Should we set the same environment variable in other jobs?

fluiddot (approved):
The results are impressive, awesome work @mokagio 🙇 ! All CI jobs are succeeding so seems it works as expected. The only part that we'd need to validate is if the process is checking out the specified commit SHA, but based on the CI logs I assume it's working fine. Unless someone else noticed a blocker, I'd say we could start using it. Thanks 🙇 !

AliSoftware:
> Like, why wouldn't Buildkite use this approach out of the box?

Maybe it's just an oversight on their part, not having tested partial checkouts vs submodules? I think it's worth opening a GitHub issue on Buildkite's repo to suggest that improvement—or alternatively contacting them by email or their support forums.

.buildkite/pipeline.yml
mokagio: Good point. That was a leftover for iterating on the plugin.

Done.

.buildkite/pipeline.yml
mokagio: That was another leftover for iterating on the plugin. The env var makes Buildkite checkout the plugin from scratch instead of using a Buildkite-managed cached version.

Done. Thanks for bringing this up.

mokagio:
@fluiddot 

> I wonder if applying the suggestion of using `git config fetch.recurseSubmodules false` might speed up the process

I thought I tried that, but I went back and look at my commits and the only config I set was `submodule.alternateErrorStrategy info` 😳 

> Although, I suspect that using the parameter `--also-filter-submodules` might be an alternative to this, is this accurate?

I can't say for sure. The docs say

```
       --also-filter-submodules
           Also apply the partial clone filter to any submodules in the repository.
           Requires --filter and --recurse-submodules. This can be turned on by default
           by setting the clone.filterSubmodules config option.
```

As far as I can see, the option was added [in Git 2.36.0](https://git-scm.com/docs/git-clone/2.36.0) as [2.35.0](https://git-scm.com/docs/git-clone/2.35.0)  doesn't have it. That version dates 2022 while the GitHub article dates 2020. So I think Git improved in the meantime and the current approach where we tell it to apply the filter to the submodules recursively is the way to go.

PR #6741, opened 2024-03-15
Update dependabot configuration to check for updates for each submodule at a different time

geriux:
Currently, Dependabot's configuration for submodules is the same for all of them. The issue with that is that it triggers CI at the same time. Usually, it's just for `gutenberg` and `jetpack`, but in some cases, `block-experiments` is more active and it adds one more.

<kbd>Screenshot 2024-03-15 at 14 03 23</kbd>

The main issue is SauceLabs, since we have a limit of 5 jobs to run at a time. So, tests usually fail due to exceeding retry times or getting the connection terminated.

With this proposal, there will be a 2-hour space between each submodule to avoid this problem.

PR submission checklist:

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,11 +1,25 @@
 # To get started with Dependabot version updates, you'll need to specify which
 # package ecosystems to update and where the package manifests are located.
 # Please see the documentation for all configuration options:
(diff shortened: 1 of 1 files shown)


twstokes (approved):
Cool! Makes sense @geriux. 🚀

PR #6745, opened 2024-03-18
Update Dependabot config to use target-branch for one of the configurations.

geriux:
This is a follow-up of https://github.com/wordpress-mobile/gutenberg-mobile/pull/6744

Unfortunately, [those changes failed](https://github.com/wordpress-mobile/gutenberg-mobile/runs/22795776877) with the following error:

```
Update configs must have a unique combination of 'package-ecosystem', 'directory', and 'target-branch'
```

After some investigation, I saw [there's an issue](https://github.com/dependabot/dependabot-core/issues/1778) related to this, where a [workaround is suggested](https://github.com/dependabot/dependabot-core/issues/1778#issuecomment-1988140219) and I'm trying out with this PR.

I've added `target-branch` to the Gutenberg config, which should bypass the error mentioned above.

Since this workaround only works for two configurations, I've merged the Jetpack and Block Experiments since the latter doesn't get too many updates lately so it should only trigger Jetpack updates daily.

PR submission checklist:

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -3,8 +3,11 @@
 # Please see the documentation for all configuration options:
 # https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/
 
-# We are using a configuration for each submodule so they run at a different time
+# We are using two configurations so they run at a different time
 # to avoid multiple PRs running CI and SauceLabs.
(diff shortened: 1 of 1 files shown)


twstokes (approved):
Sounds good @geriux.

PR #6760, opened 2024-03-20
Avoid including source map file in app's binary

fluiddot:
**Related PRs:**
* https://github.com/wordpress-mobile/WordPress-iOS/pull/22874

_Internal ref:_  pcdRpT-5rI-p2#comment-9518

This PR updates the approach for including the React Native source map when publishing the GBM via the XCFramework. The goal is to avoid including it in the final app's binary. Instead of including the file as an asset in the Xcode project, we'll create a new folder to hold both the bundle and source map independent of the iOS frameworks.

The previous approach was introduced in https://github.com/wordpress-mobile/WordPress-iOS/pull/22862.

**To test:**
- Observe that the XCFramework artifact is generated properly.
- Follow testing instructions from WP-iOS PR.

PR submission checklist:

### DIFF
--- a/ios-xcframework/podspec.erb
+++ b/ios-xcframework/podspec.erb
@@ -25,4 +25,6 @@ Pod::Spec.new do |s|
     'Frameworks/yoga.xcframework',
     'Frameworks/hermes.xcframework'
   ]
+
+  s.preserve_paths = 'react-native-bundle-source-map'
 end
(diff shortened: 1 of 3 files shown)


ios-xcframework/podspec.erb
```diff
@@ -25,4 +25,6 @@ Pod::Spec.new do |s|
     'Frameworks/yoga.xcframework',
     'Frameworks/hermes.xcframework'
   ]
+
+  s.preserve_paths = 'react-native-bundle-source-map'
```
fluiddot, line 29: This ensures that the folder will be included in the Pod.

jhnstn (approved):
Works as expected. Ship it!

PR #6861, opened 2024-05-13
Include Block experiments classname package

geriux:
**Related PR:**
- https://github.com/WordPress/gutenberg/pull/61616

This PR adds a new step to install dependencies of the `block-experiments` submodule. Currently, we were relying on Gutenberg to have all needed dependencies for this submodule, but since the recent removal of the `classnames` package, it makes sense this submodule also installs its dependencies as we do it for Jetpack and Gutenberg.

It adds a new bash file to install the `block-experiments` dependencies, which will be called in the `postinstall` script in Gutenberg Mobile's `package.json`

A new item was added to Metro's block list, to not include files from the `@wordpress` package from its dependencies, the reason for this is to solve problems with bringing old references of older versions of the `native` files. This shouldn't affect anything as we are loading these packages from Gutenberg already.

PR submission checklist:

### DIFF
--- a/bin/run-block-experiments-command.sh
+++ b/bin/run-block-experiments-command.sh
@@ -1,6 +1,14 @@
 #!/bin/bash
 set -Eeuo pipefail
 
+local_run_script="${GBM_LOCAL_BLOCK_EXPERIMENTS_RUN_SCRIPT:-./bin/run-block-experiments-command.sh.local}"
+
+if [ -e "$local_run_script" ]
+then
+  source "$local_run_script"
+  exit 0
+fi
+
 # Check if nvm is installed
 [ -z "$NVM_DIR" ] && NVM_DIR="$HOME/.nvm"
 [ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"
--- a/bin/run-block-experiments-command.sh
+++ b/bin/run-block-experiments-command.sh
@@ -13,7 +13,7 @@ command -v nvm >/dev/null 2>&1 || {
 pushd block-experiments
 
 # Set up node requirement for block-experiments
-nvm use
+nvm install
 
 # Check if Yarn is installed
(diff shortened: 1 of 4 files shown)


bin/run-block-experiments-command.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
+set -Eeuo pipefail
+
+# Check if nvm is installed
```
dcalhoun, line 12: If it is not too difficult, would you be willing to add support for managing Node.js versions with something other than `nvm` in the form of a local script? 

An example can be found in the existing, sibling Jetpack command script. 

https://github.com/wordpress-mobile/gutenberg-mobile/blob/13e258a64bb98e6d127bb2e090183d6e6c82e5bb/bin/run-jetpack-command.sh#L4-L10

geriux: Of course! I'll extract this so it can be reused in both files 👍

twstokes (commented):
Will wait for [this item](https://github.com/wordpress-mobile/gutenberg-mobile/pull/6861/files#r1599860208) to be resolved before reviewing. Feel free to re-request after it's been resolved. Thanks!

geriux:
> Will wait for [this item](https://github.com/wordpress-mobile/gutenberg-mobile/pull/6861/files#r1599860208) to be resolved before reviewing. Feel free to re-request after it's been resolved. Thanks!

Done! I decided to not unify both just in case separate configurations are preferred for each submodule 👍

twstokes (approved):
Thank you @geriux. 🚀

geriux:
Thanks @twstokes ! This PR didn't include the merge commit hash from Gutenberg so I'll create another PR to update it shortly.

twstokes:
> Thanks @twstokes ! This PR didn't include the merge commit hash from Gutenberg so I'll create another PR to update it shortly.

Ah! My bad on that. 😬

